Issue #7, opened 2017-01-30
Improvement: Support Terraform Remote Statefiles


jacknagz:
Convert local tfstate to remote tfstate via S3 buckets

PR #24, opened 2017-02-06
[terraform] remote state configuration

jacknagz:

size: small
resolves: #7

* Create new S3 bucket with versioning for the tfstate file
* Add Terraform remote state configuration command


stream_alert_cli/cli.py
```diff
@@ -161,6 +166,27 @@ def _continue_prompt():
...
+            'config',
+            '-backend=s3',
+            '-backend-config="bucket={}"'.format(bucket),
+            '-backend-config="key={}"'.format(s3_key),
+            '-backend-config="region={}"'.format(region),
+            '-backend-config="kms_key_id={}"'.format(kms_key_id),
+            '-backend-config="encrypt=true"'
+        ]
+
+        self.run_command([' '.join(remote_state_opts)], quiet=True, shell=True)
```
austinbyers: Using `shell=True` is a potential security risk. A future downstream application with restricted permissions which only allows users to configure bucket names would then be vulnerable to shell injection. Let's try to avoid `shell=True` if possible.

I think the reason `suprocess.check_call` is not working here is because of the double-quotes in the argument names. Try different alternatives, like single quotes, no quotes, etc. See if there is a Python API that will sanitize/invoke terraform on your behalf.

If all else fails and `shell=True` is the only option, put a comment here about why we have to do it, and then a warning in the global README that tells the user that the config input is assumed to be trusted.

stream_alert_cli/helpers.py
```diff
@@ -25,14 +25,17 @@ def run_command(cls, runner_args, **kwargs):
         default_cwd = 'terraform'
         cwd = kwargs.get('cwd', default_cwd)
 
+        shell = kwargs.get('shell', False)
```
austinbyers: If nothing uses the shell command, let's remove the possibility

PR #38, opened 2017-02-17
[pre_parsers] correct error for s3 objects > 128mb

strcrzy:

size: tiny

fixes an incorrect error message that may confuse users


jacknagz:
👍

Issue #16, opened 2017-02-01
Feature - VPC flow log support


mime-frame:
Feature - VPC flow log support

Likely via http://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/Subscriptions.html?tag=vig-20#LambdaFunctionExample

strcrzy:
this is really 3 things

- [ ] terraform to enable flow logs and pipe to kinesis via cloudwatch subscription (#39)

- [ ] gzip-json record parser (#40)

- [ ] json nested record parser (1784238) 

the gzip and nested json work will need to be merged but after that this should be good to go!

strcrzy:
this should now be technically supported and documented with #39 and #40. 
i will add some integration tests before closing this issue though.

PR #40, opened 2017-02-20
[core] gzip json parser

strcrzy:
size: small
issue: #16

# gzip-json parser
This StreamAlert parser parses deflate-compressed json-encoded strings, which is the format of the payload [provided](http://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html#DestinationKinesisExample) by CloudWatch log subscriptions.

# nested records
the CloudWatch JSON payload contains nested records (the actual logs are at key `logEvents`) so this will only really be useful once 1784238 merges. i will wait until 1784238 has merged and rebase on top of that


jacknagz:
This is great @strcrzy 😄 ! 

Did you experience issues with the KV/CSV parsers? Just trying to get some background on the added `try/except` blocks.

strcrzy:
thanks! 

there's an issue where the `kv` and `csv` parsers were throwing exceptions when they attempted to parse the compressed payloads (they expect string-like payloads and choke on the binary data). 

this will happen when a source declares both (`kv` or `csv`) and `gzip-json` log types, in two situations that i am aware of:
1. `kv`/`csv` log types are declared before `gzip-json` one; these parsers are then attempted first
2. `gzip-json` log type is declared first, but the payload doesn't match the schema; the `kv`/`csv` parsers are then attempted next

in each of these situations it seems appropriate to catch the exception and return `False`

jacknagz:
Awesome, thanks for the background.  Once the nested record parsing support merges, could you also add some tests for this new parser?

strcrzy:
:+1: definitely

stream_alert/parsers.py
```diff
@@ -134,6 +143,22 @@ def _parse_records(self, json_payload):
                         json_records.append(record)
                 elif isinstance(val, dict):
                     return self._parse_records(json_payload[key])
+        # if hints are present, use them to find the nested records
+        elif (hints and len(hints)):
+            records_jsonpath = jsonpath_rw.parse(hints['records'])
+            envelope_schema = hints.get('envelope', {})
+            if len(envelope_schema):
+                self.schema.update({u"envelope": envelope_schema})
```
jacknagz: why make the key `unicode` vs `str`?

stream_alert/classifier.py
```diff
@@ -220,7 +220,6 @@ def _parse(self, payload, data):
         Returns:
             A boolean representing the success of the parse.
         """
-        logger.debug(data)
```
strcrzy: when this line attempts to log a binary payload, the lambda goes 💥

jacknagz: interesting, ok!

stream_alert/parsers.py
```diff
@@ -114,7 +119,9 @@ def _key_check(self, json_records):
                 json_records.remove(json_record)
 
     def _parse_records(self, json_payload):
-        """Iterate over a json_payload to determine if nested records exist
+        """Iterate over a json_payload.
```
strcrzy: 👍

stream_alert/parsers.py
strcrzy: i'm not sure why i put this here. removing it.

stream_alert/parsers.py
```diff
@@ -124,7 +131,9 @@ def _parse_records(self, json_payload):
             A list of dict JSON payloads
         """
         json_records = []
-
+        hints = self.options.get('hints',{})
```
strcrzy: :+1:

strcrzy:
added some docs!

docs/source/conf-schemas.rst
```diff
@@ -146,6 +146,41 @@ This promotes Rule safety, but requires additional time to define the schemas
 
 .. warning:: In Option 2, the schema definition is flexible, but Rule safety is lost because you'll need to use defensive programming when accessing and analyzing fields in `columns`. The use of `req_subkeys` will be advised, see Rules for more details
 
+
+JSON Example: VPC Flow Logs
+---------------------------
+
+VPC flow logs can be delivered via CloudWatch to a kinesis stream. Though they are compressed with deflate, we can use the special ``gzip-json`` parser to decompress them before parsing.CloudWatch logs are delivered as a nested record, so we will need to pass ``hints`` to JSON parser to properly find the nested records::
```
jacknagz: `...them before parsing.Cloudwatch`

pls fix spacing after `parsing.`

strcrzy:
stripped previous nesting implementation, updated docs to reflect the JSONPath approach

stream_alert/parsers.py
```diff
@@ -124,16 +135,23 @@ def _parse_records(self, json_payload):
...
+                self.schema.update({"envelope": envelope_schema})
+                envelope_keys = envelope_schema.keys()
+                envelope_jsonpath = jsonpath_rw.parse("$." + ",".join(envelope_keys))
+                envelope_matches = [match.value for match in envelope_jsonpath.find(json_payload)]
+                envelope = dict(zip(envelope_keys, envelope_matches))
+
+            for match in records_jsonpath.find(json_payload):
+                record = match.value
+                if len(envelope):
+                    record.update({"envelope":envelope})
```
jacknagz: space after "envelope"

PR #41, opened 2017-02-22
[v1.2.0] Nested Record Support, External Alerting, and New Alert Format

jacknagz:
cc @patrobinson 

size: large

* Support external alert handling.  If you pass in the kwarg `return_alerts=True` into the `StreamAlert()` initializer, it'll return an array of alerts.
* Support nested JSON records.
* Code cleanup, add/change unit tests, and more.
* Use a new alert format broken down into three main keys: `record`, `rule_name`, and `metadata`.  Pagerduty example:

screen shot 2017-02-22 at 11 20 27 am


stream_alert/__init__.py
```diff
@@ -1 +1,107 @@
...
+from stream_alert.rules_engine import StreamRules
+from stream_alert.sink import StreamSink
+
+logging.basicConfig()
+logger = logging.getLogger('StreamAlert')
+logger.setLevel(logging.INFO)
+
+__version__ = '1.2.0'
+
+class StreamAlert(object):
```
austinbyers: I don't think we should put a full class in `__init__.py`, because then it can't be imported directly. `__init__.py` should be used for [structuring a package](https://docs.python.org/2.7/tutorial/modules.html#packages), not defining classes.

main.py
```diff
@@ -14,83 +14,11 @@
...
-from stream_alert.config import load_config, load_env
-from stream_alert.classifier import StreamPayload, StreamClassifier
-from stream_alert.pre_parsers import StreamPreParsers
-from stream_alert.rules_engine import StreamRules
-from stream_alert.sink import StreamSink
-
+from stream_alert import StreamAlert
 from rules import (
     sample_rules
 )
```
austinbyers: This can all be on one line now

main.py
```diff
@@ -14,83 +14,11 @@
...
-
-def process_alerts(payload, alerts_to_send):
-    if payload.valid:
-        alerts = StreamRules.process(payload)
-        if alerts:
-            alerts_to_send.extend(alerts)
-    else:
-        logger.debug('Invalid data: %s', payload)
+    """Main Lambda handler function"""
+    StreamAlert().run(event, context)
```
austinbyers: +1 for very simple main file / function!

stream_alert/classifier.py
```diff
@@ -248,11 +257,15 @@ def _parse(self, payload, data):
...
+                typed_data = []
+                for data in parsed_data:
+                    # convert data types per the schema
+                    typed_data.append(self._convert_type(data, schema, options))
+
+                if typed_data:
                     payload.log_source = log_name
                     payload.type = parser_name
-                    payload.record = parsed_and_typed_data
+                    payload.record = typed_data
```
strcrzy: should this be `payload.records` now?

jacknagz: good catch

main.py
jacknagz: going to keep this as is, since the list tends to grow

jacknagz:
@austinbyers @strcrzy PTAL

stream_alert/parsers.py
```diff
@@ -65,24 +65,90 @@ def __init__(self, data, schema, options):
...
+        """
+        json_records = []
+
+        if len(json_payload) == 1:
+            for key, val in json_payload.iteritems():
+                self.nested = True
+                self.nested_keys.append(key)
+                if isinstance(val, list):
+                    for record in val:
+                        json_records.append(record)
```
strcrzy: one thing i was wondering if is it would be worth merging the fields from the enveloping event with the nested events? for cloudwatch log subscriptions, the envelope fields would be `logGroup` `logStream` `owner` and `subscriptionFilters`. as is, it seems like we would lose those fields as we discard the enveloping event

jacknagz: @strcrzy would you mind making this change in your PR? I think it would be more relevant since  you'll be adding tests, etc for this type of event

strcrzy: sure thing!

strcrzy:
:+1:

Issue #57, opened 2017-03-09
Bug: skip file if contents cannot be decoded into valid json


ryandeivert:
In `stream_alert_cli/test.py`, the function for testing kinesis alert rules (`test_kinesis_alert_rules`) should skip each file that cannot be properly decoded into json. Currently, the below error is thrown when a binary file (.DS_Store) is in the directory we're testing against.

**Command**
`stream_alert_cli.py lambda test --func alert --source kinesis`

**Error**
```
INFO:StreamAlertCLI:Error loading .DS_Store, bad JSON
Traceback (most recent call last):
  File "./stream_alert_cli.py", line 128, in <module>
    if __name__ == "__main__": main()
  File "./stream_alert_cli.py", line 125, in main
    cli.run(options)
  File "/Users/<user>/code/streamalert/stream_alert_cli/cli.py", line 59, in run
    self._lambda_runner(options)
  File "/Users/<user>/code/streamalert/stream_alert_cli/cli.py", line 73, in _lambda_runner
    stream_alert_test(options)
  File "/Users/<user>/code/streamalert/stream_alert_cli/test.py", line 216, in stream_alert_test
    test_kinesis_alert_rules()
  File "/Users/<user>/code/streamalert/stream_alert_cli/test.py", line 189, in test_kinesis_alert_rules
    test_records = contents['records']
UnboundLocalError: local variable 'contents' referenced before assignment
```

PR #59, opened 2017-03-09
[cli][testing] skipping files that cannot be dumped to json

ryandeivert:

size: small
resolves: #57 

**changes**
Files that are unable to be dumped to json will now be skipped in the loop.
Also resolving issues in two places where we would try to continue with operations even if the opening of the file failed.

### DIFF
--- a/stream_alert_cli/test.py
+++ b/stream_alert_cli/test.py
@@ -106,7 +106,12 @@ def format_record(test_record):
     elif service == 'kinesis':
         kinesis_path = os.path.join(DIR_TEMPLATES, 'kinesis.json')
         with open(kinesis_path, 'r') as kinesis_template:
-            template = json.load(kinesis_template)
+            try:
+                template = json.load(kinesis_template)
+            except ValueError as err:
+                LOGGER_CLI.error('Error loading kinesis.json: %s', err)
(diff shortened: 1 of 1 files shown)


stream_alert_cli/test.py
```diff
@@ -146,21 +150,22 @@ def test_kinesis_alert_rules():
...
-                except ValueError:
-                    LOGGER_CLI.info('Error loading %s, bad JSON', rule_file)
+                except ValueError as err:
+                    LOGGER_CLI.error('Error loading %s: %s', rule_file, err)
+                    continue
 
-            test_records = contents['records']
-            if len(test_records) == 0:
-                LOGGER_CLI.info('No records to test for %s', rule_name)
+                test_records = contents['records']
```
jacknagz: do we need to indent here if we are using `continue` in the `except` block?

ryandeivert: The except block captures if the file was opened successfully, but the JSON fails to parse.

However, if the file fails to open at all in `with open(..`, having this outside of the `with` block will still cause issues.. yes?

jacknagz: it shouldn't because of the `continue`

PR #65, opened 2017-03-15
[lambda][output] request_helper changes

armtash:

size: small

* Added request_helper optional arg "headers" to pass http headers.
* Added request_helper optional arg "verify" to allow output to ignore
* server certificate checking. Default is verify=True
* Removed request_helper required arg "endpoint" as it's no longer needed.
* request_helper returns its urllib2 response to allow for outputs to
* chain http requests and better validate successful requests.

### DIFF
--- a/stream_alert_output/main.py
+++ b/stream_alert_output/main.py
@@ -21,6 +21,7 @@ import logging
 import os
 import time
 import urllib2
+import ssl
 
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


stream_alert_output/main.py
```diff
@@ -120,7 +121,10 @@ def run(self, alerts):
                 output_func = getattr(self, '_{}'.format(output), None)
                 if output_func:
                     self._setup_output_creds(output)
-                    output_func(rule_name, formatted_alerts)
+                    try:
```
jacknagz: What is the error you're trying to catch here? I'd prefer a strategy where we catch specific errors in each respective `output`.

yeller15: I intended for it to be a general catchall.  An OutputRequestFailure exception is raised when a failure occurs in request_handler, which wasn't caught by anything.  I can modify this block to only catch OutputRequestFailure, or I can revert it to how it was before.

jacknagz: let's keep it how it was initially, otherwise the error raised would never actually fire

jacknagz:
please run `pylint` on this commit, I noticed a couple offenses

jacknagz:
@armtash the changes in here are also present in #66, perhaps we can just close this one out as a duplicate?

armtash:
Sure

Issue #61, opened 2017-03-10
Feature: Support S3 Testing


jacknagz:
### Background
When testing the StreamAlert Lambda functions, it's necessary to create fake `event` records to send to the main handler for analysis.   With AWS Kinesis and SNS this is straightforward, since the `event` contains the data to be analyzed/alerted on.  However with `s3` events, the data must be fetched out of band, which increases the complexity of testing.

### Desired Outcome
The ability to test s3 based rules:

```
{
  "records": [
    {
      "data": "data,in,my,fake,object",
      "description": "Privileged wget commands to non Airbnb assets",
      "trigger": true,
      "source": "my.s3.bucket",
      "key": "my.fake.object",
      "service": "s3"
    }
  ]
}
```

PR #68, opened 2017-03-16
[testing] updating cli testing to include s3 as a service

ryandeivert:

size: small
resolves: #61

## changes ##
Slight refactor to `test.py` to allow for a more generic way of testing various services.
Updating `sources.json` to include various source log types for s3.
`requirements.txt` now includes `moto==0.4.31`, as this is now used for boto3 mocking.
Updated integration rule files to include s3 service examples.

### DIFF
--- a/stream_alert_cli/test.py
+++ b/stream_alert_cli/test.py
@@ -18,9 +18,12 @@ import base64
 import json
 import logging
 import os
+import random
 import re
 import time
 
+import boto3
+from moto import mock_s3
 from stream_alert.handler import StreamAlert
 # import all rules loaded from the main handler
 import main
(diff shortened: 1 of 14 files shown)


ryandeivert:
@jacknagz @mime-frame PTAL

stream_alert_cli/test.py
```diff
@@ -104,36 +107,42 @@ def format_record(test_record):
...
-                LOGGER_CLI.error('Error loading kinesis.json: %s', err)
-                return
-
         template['kinesis']['data'] = base64.b64encode(data)
         template['eventSourceARN'] = 'arn:aws:kinesis:us-east-1:111222333:stream/{}'.format(source)
-        return template
-
     elif service == 'sns':
+        # TODO implement sns testing
         pass
```
jacknagz: didn't you want to raise a `NotImplementedError` here instead of failing silently?

ryandeivert: I wasn't going to raise it there but I don't see why we shouldn't. I'll add it

PR #70, opened 2017-03-16
[docs] updates to the docs to reflect changes to support s3 testing

ryandeivert:

size: small

## changes ##
Updated the Rule Testing section of the docs to reflect changes made to support s3 testing.
Users no longer need to designate a specific service when running tests, and the output format for tests has change significantly.


docs/source/rules.rst
```diff
@@ -154,72 +154,3 @@ Example use of that function within a rule::
         user_whitelist = { 'mike', 'jin', 'jack', 'mary' }
 
         return in_set(user, user_whitelist)
```
jacknagz: add a note which links users to the rules-testing page

ryandeivert: Amended to add note

jacknagz:
cc @mime-frame

Issue #58, opened 2017-03-09
Improvement: investigate PagerDuty incident merging


ryandeivert:
Updating this issue since the de-duping via the `incident_key` was not quite what we were looking for. As an alternative, we should investigate adding support for incident merging through the PD API:
https://v2.developer.pagerduty.com/v2/page/api-reference#!/Incidents/put_incidents_id_merge

Some implications of this that may not be ideal:
- A new incident would have to be created initially, which will probably result in another PD alert that we're trying to avoid.
- After the incident is created, we'd have to query incidents that exist with this incident_key and then merge any open incidents into one (presumably the oldest open incident):
  - https://v2.developer.pagerduty.com/v2/page/api-reference#!/Incidents/get_incidents

Other Notes:
- We could potentially combine something like the `rule_name` with the `host_identifier` or some other unique key to try to help with merging alerts that are from different hosts.
- PD is investigating adding a table like the new Alerts table found on `pagerduty.com/alerts` to the Details section of an incident.
- In the future, there could be a 'Page Once' feature from PD that must be opted-in on a per-incident basis that would aggregate all alerts after the mode is enabled.

**Other Reference**
https://support.pagerduty.com/hc/en-us/articles/230090307-Merging-Incidents

------**Below is the original issue description for this issue**------
PagerDuty supports automated de-duplication by including an incident_key with incidents. We should utilize this to reduce "alert fatigue" and "queue explosion" by setting the 'incident_key' to the StreamAlert rule name ('rule_name'). The incidents should then be merged within PagerDuty.

**Reference**
[https://v2.developer.pagerduty.com/docs/events-api](url)
[https://support.pagerduty.com/hc/en-us/articles/202889630-How-do-I-reduce-the-number-of-notifications-I-am-receiving-for-the-same-incident-event](url)

patrickod:
One thing that'd be great to have for a feature like this is the ability to provide the value by which alerts are grouped, for example grouping by the `.detail.userIdentity.arn` in a CloudWatch would provide a nice summary of rules triggered by a single actor within a given time period.

In our own deployment this would be really helpful as it's often the case that someone will trigger multiple rules at once making a series of changes.

mime-frame:
[...]

PR #71, opened 2017-03-17
[lambda][output] adding `incident_key` for PD output to support de-duping

ryandeivert:

size: small
resolves: #58 

## changes ##
Super tiny change that adds an `incident_key` for pagerduty output to support de-duplication of incidents on the PD side. This `incident_key` will contain the triggering `rule_name`.


mime-frame:
@ryandeivert - can you include testing steps going forward in your PR description?

mime-frame:
@ryandeivert - what happens if there is a rule name that is the same for diff clusters/rules files?

ryandeivert:
@mimeframe great point and question. Working on a way to determine that now.

ryandeivert:
@jacknagz merged in master to resolve conflicts. PTAL and approve.

PR #72, opened 2017-03-20
[terraform][cli][lambda] Staging Environment Deprecation, CLI Config v2, Lambda Func Reorganization

### LINKED ISSUE #55: Improvement: Combine Staging/Production Lambda Aliases
**Background**

Currently there are two AWS Lambda functions which make up StreamAlert:
`alert`: rules processing (rule matching and classification)
`output`: output processing (alerting to desired incident management/comms)

They have separate [AWS Lambda Alias](http://docs.aws.amazon.com/lambda/latest/dg/aliases-intro.html) configurations:
`alert` has both `staging` and `production`
`output` only has `staging`

The difference between the two is that `staging` always points to the `$LATEST` uploaded StreamAlert source code S3 while `production` is a pointer to an immutable published version.  

The difference from within StreamAlert is that `staging` does not send detected alerts to assigned outputs, it instead writes them to Cloudwatch logs.  Staging historically has been used as a smoke test deployment to identify issues early on before it is ran against production data.   However as this project has matured, and continuous integration testing has increased its coverage, we have less of a need for the split.

**Desired Outcome**

To simplify the deployment process, all StreamAlert Lambda functions should have a single alias: `production`.

This provides the following:
* A single, immutable version which can be rolled back as needed.
* No split application logic between staging/production.

jacknagz:
size: large
resolves #55 

## High Level
* This PR is focused around two changes: The removal of the `staging` environment, and renaming of both AWS Lambda Functions.
* Removal of `staging` decreases complexity in the Lambda deployment/event source configuration.
* New Lambda Function names:
  * `alert` => `rule_processor`
  * `output` => `alert_processor`

## Terraform Changes
* Rename both Lambda functions and their associated resources (IAM policies and more).
* Remove the `staging` environment alias from the `rule_processor` Lambda function.
* Create a `production` environment for the `alert_processor`.
* Refactor the `variables.json` for clearer organization and consistent settings between both Lambda functions.
* Update the `cluster_template` to reflect the new `variables.json` layout and Lambda function rename.
* Remove unused `integration_testing` s3 bucket.

## Lambda Changes
* Reorganize both Lambda functions under the same parent `stream_alert` folder.
* Update module imports per rename.
* Update Lambda handlers per rename.
* Gracefully break functions on non-kinesis events.
* Bump version to `1.3.0`

## CLI Changes
* Add a new `CLIConfig` class for reading/writing/modifying `variables.json` with unit tests.
* Update command line options to remove `env` and rename `func` to `processor`.
* Rename `cli.py` to `runner.py` .
* Update Lambda packaging logic for new folder structure.

## Breaking Changes
* Rule files must change their imports from `from stream_alert.rules_engine import StreamRules` to `from stream_alert.rule_processor.rules_engine import StreamRules`.
* The CLI has deprecated `env` and `func` arguments.  Use `processor` instead of `func`.
* A new S3 bucket is created called `prefix_cluster_streamalerts`.  This is the default bucket for alerts designed for s3 output.  It's recommended to s3 sync from the old bucket (stream.alert.output.results) to this new bucket. 

## Testing
* Locally with unit/rule testing
* Integration testing with a demo AWS account to verify end-to-end flow


requirements.txt
```diff
@@ -1,28 +1,40 @@
...
 pyparsing==2.1.10
 python-dateutil==2.6.0
+python-mimeparse==1.6.0
 rsa==3.3
 s3transfer==0.1.10
 six==1.10.0
+testtools==2.2.0
+traceback2==1.4.0
+unittest2==1.1.0
 virtualenv==15.1.0
```
ryandeivert: Have you included all your dev reqs here? It looks like a lot of these aren't actual code dependencies for SA

stream_alert_cli/config.py
```diff
@@ -0,0 +1,180 @@
...
+    def write(self):
+        """Write the current config in memory to disk"""
+        with open(self.filename, 'r+') as varfile:
+            config_out = json.dumps(
+                self.config,
+                indent=4,
+                separators=(',', ': '),
+                sort_keys=True
+            )
+            varfile.seek(0)
```
ryandeivert: Opening with `r+` positions the stream at 0 so this seek is unnecessary.

stream_alert_cli/config.py
```diff
@@ -0,0 +1,180 @@
...
+            varfile.write(config_out)
+            varfile.truncate()
+
+    def detect_version(self):
```
ryandeivert: Seems like this could be converted to a private method `_detect_version`.. unless I'm missing some use-case outside of this class

stream_alert_cli/package.py
```diff
@@ -24,26 +24,27 @@
...
+logging.basicConfig(format='%(name)s [%(levelname)s]: %(message)s')
+LOGGER_CLI = logging.getLogger('StreamAlertCLI')
+LOGGER_CLI.setLevel(logging.INFO)
+
 class LambdaPackage(object):
     """Build and upload a StreamAlert deployment package to S3."""
-    package_folders = ()
-    package_files = ()
+    package_folders = {}
+    package_files = {}
```
austinbyers: Here these are initialized as `dicts`, but the subclasses define them as `sets`. So this should be
```Python
package_folders = set()
package_files = set()
```

However, I think a better way to do this is with [abstract properties](https://docs.python.org/2/library/abc.html#abc.abstractproperty):
```Python
class LambdaPackage(object):
  __metaclass__ = abc.ABCMeta

  @abc.abstractproperty
  def package_folders(self):
    """A set of string folder names..."""
    pass

  @abc.abstractproperty
  def package_files(self):
    """A set of string package file names."""
    pass


class RuleProcessorPackage(LambdaPackage):  
  @property
  def package_folders(self):
    return {'stream_alert/rule_processor', 'rules', 'conf'}
  
  @property
  def package_files(self):
    return {'stream_alert/__init__.py', 'variables.json'}
```

It's annoyingly verbose, but it guarantees that all of the fields are defined by all subclasses and makes the code more resilient to future changes. The properties are accessed just like they are right now (`RuleProcessorPackage().package_folders`)

requirements.txt
jacknagz: I'll remove pyfakefs.  The rest are dependencies from mock

stream_alert_cli/runner.py
```diff
@@ -0,0 +1,344 @@
...
+        LOGGER_CLI.info('Deploying Lambda Functions')
+        # setup remote state
+        refresh_tf_state()
+        # deploy both lambda functions
+        deploy(deploy_opts('all'))
+        # create all remainder infrastructure
+        LOGGER_CLI.info('Building Remainder Infrastructure')
+        tf_runner()
+        # refresh config to get modified variables
+        # refresh_config()
```
austinbyers: Can this be deleted? As a general rule, there shouldn't be commented code without a reason.

terraform/modules/tf_stream_alert/README.md
```diff
@@ -1,13 +1,17 @@
...
-* This Terraform module creates the lambda function to match rules against logs.
+* This Terraform module creates the main AWS Lambda functions to match rules and send alerts.
 
 ## Components
-Creates the following:
-* A S3 bucket for stream_alert lambda source code.
-* The Lambda function with the execution role.
-* A staging and production alias.
-* An SNS topic to send alerts to.
+* A S3 buckets
```
austinbyers: Nit: typo

.travis.yml
```diff
@@ -7,4 +7,4 @@ install:
   - pip install -r requirements.txt
 script: 
   - ./test/scripts/unit_tests.sh
-  - ./test/scripts/integration_test_kinesis.sh
+  - ./test/scripts/rule_test.sh
```
jacknagz: They are also used on the command line for local tests/development

stream_alert_cli.py
```diff
@@ -108,21 +114,20 @@ def build_parser():
...
 
     return parser
 
 def main():
-    logging.basicConfig(level=logging.INFO,
-                        format='%(asctime)s [%(levelname)s] %(message)s',
-                        datefmt='%m/%d/%Y %I:%M:%S%p')
+    # logging.basicConfig(level=logging.INFO,
+    #                     format='%(asctime)s [%(levelname)s] %(message)s',
+    #                     datefmt='%m/%d/%Y %I:%M:%S%p')
```
jacknagz: good catch

stream_alert_cli/config.py
```diff
@@ -0,0 +1,180 @@
...
+        new_config['flow_log_config'] = {
+            'vpcs': [],
+            'subnets': [],
+            'emis': []
+        }
+        new_config['kinesis_streams_config'] = self.config['kinesis_settings']
+        new_config['terraform'] = {
+            'tfstate_s3_key': self.config['tfstate_s3_key'],
+            'tfvars': self.config['tfvars']
+        }
```
jacknagz: yes, see `test/unit`

stream_alert_cli/package.py
```diff
@@ -24,26 +24,27 @@
...
 import pip
 
-from stream_alert_cli.helpers import CLIHelpers
-
 import boto3
 
+# from stream_alert_cli.logger import LOGGER_CLI
+logging.basicConfig(format='%(name)s [%(levelname)s]: %(message)s')
+LOGGER_CLI = logging.getLogger('StreamAlertCLI')
+LOGGER_CLI.setLevel(logging.INFO)
```
jacknagz: accidentally left this in here when testing

stream_alert_cli/config.py
```diff
@@ -0,0 +1,180 @@
...
+        'rule_processor_lambda_config',
+        'terraform'
+    }
+
+    def __init__(self):
+        self.config = self.load()
+        self.version = self.detect_version()
+        if self.version == 1:
+            self.config = self._convert_schema()
+            self.version = self.detect_version()
```
jacknagz: It's calculated again after the conversion to ensure the schema matches

stream_alert_cli/package.py
```diff
@@ -24,26 +24,27 @@
...
-    source_prefix = None
+    config_key = None
 
     def __init__(self, **kwargs):
         self.version = kwargs['version']
         self.config = kwargs['config']
 
     def create_and_upload(self):
-        """Create a Lambda deployment package, checksum it, and upload it to S3.
+        """Create a Lambda deployment package, hash it, and upload it to S3.
```
jacknagz: LOL 👍

stream_alert_cli/config.py
jacknagz: It's a way to ensure our converted schema matches our declared key set

stream_alert_cli/package.py
```diff
@@ -160,66 +163,67 @@ def _sha256sum(package_path):
...
-class AlertPackage(LambdaPackage):
-    """Deployment package class for the AWS Lambda `Alert` function"""
-    package_folders = {'stream_alert', 'rules', 'conf'}
-    package_files = {'main.py', 'variables.json'}
+
+class RuleProcessorPackage(LambdaPackage):
+    """Deployment package class for the StreamAlert Rule Processor function"""
+    package_folders = {'stream_alert/rule_processor', 'rules', 'conf'}
+    package_files = {'stream_alert/__init__.py', 'variables.json'}
+    package_root_dir = '.'
```
jacknagz: They both weren't always `'.'`, and they may change in the future, so I'm going to leave it

stream_alert_cli/logger.py
```diff
@@ -0,0 +1,24 @@
...
+'''
+
+import logging
+
+LOGGER_SA = logging.getLogger('StreamAlert')
+LOGGER_SA.setLevel(logging.INFO)
+
+logging.basicConfig(format='%(name)s [%(levelname)s]: %(message)s')
+LOGGER_CLI = logging.getLogger('StreamAlertCLI')
+LOGGER_CLI.setLevel(logging.INFO)
```
austinbyers: 🎉

stream_alert_cli/runner.py
```diff
@@ -0,0 +1,341 @@
...
+def rollback():
+    """Rollback the current production AWS Lambda version by 1
+
+    Notes:
+        Ignores if the production version is $LATEST
+        Only rollsback if published version is greater than 1
+    """
+    clusters = CONFIG['clusters'].keys()
+    for cluster in clusters:
+        for lambda_function in ('rule_processor', 'alert_processor'):
```
jacknagz: this is a bug actually

ryandeivert: even better 👍

stream_alert_cli/package.py
```diff
@@ -59,23 +56,27 @@ def create_and_upload(self):
...
             # remove generated deployment files
             self._cleanup(package_path, package_sha256_path)
             # set new config values and update
-            self.config[self.source_key] = os.path.join(self.source_prefix,
-                                                        package_name)
-            self.config[self.source_hash_key] = package_sha256
-            CLIHelpers.update_config(self.config)
+            full_package_name = os.path.join(self.package_name, generated_package_name)
+            # make all config changes here
+            print self.config[self.config_key]
```
austinbyers: Is this print statement still supposed to be here?

stream_alert_cli/runner.py
```diff
@@ -0,0 +1,367 @@
...
+    clusters = CONFIG['clusters'].keys()
+    if options.processor == 'all':
+        lambda_functions = {'rule_processor', 'alert_processor'}
+    else:
+        lambda_functions = {'{}_processor'.format(options.processor)}
+
+    for cluster in clusters:
+        for lambda_function in lambda_functions:
+            version_key = '{}_versions'.format(lambda_function)
+            print version_key, cluster
```
austinbyers: Stray print statement?

Issue #73, opened 2017-03-21
Improvement: testing should return an error value on any failure, but only after all tests run


ryandeivert:
When tests for StreamAlert fail, we current break and exit mid-testing. This make for a less than desirable user experience when multiple tests fail that need to be fixed. We should print each failure as it occurs and then exit with an error code (1) at the very end.

PR #74, opened 2017-03-21
[cli][testing] cli testing will now fail gracefully after all tests have run

ryandeivert:

size: small
resolves: #73 

## changes ##
- Refactoring the cli testing code so all tests will run and report failures that occurred during the run. We will no longer bail out early on a test failure.
- Tracking if there are failures and exiting with an error code if any occur throughout the process.

### DIFF
--- a/stream_alert_cli/test.py
+++ b/stream_alert_cli/test.py
@@ -27,7 +27,7 @@ from moto import mock_s3
 from stream_alert.handler import StreamAlert
 # import all rules so they are loaded
 # pylint: disable=unused-import
-from rules import sample_rules, sample_matchers
+import main
 # pylint: enable=unused-import
 
 LOGGER_SA = logging.getLogger('StreamAlert')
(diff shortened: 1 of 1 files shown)


stream_alert_cli/test.py
```diff
@@ -25,8 +25,10 @@
 import boto3
 from moto import mock_s3
 from stream_alert.handler import StreamAlert
-# import all rules loaded from the main handler
-import main
+# import all rules so they are loaded
+# pylint: disable=unused-import
+from rules import sample_rules, sample_matchers
```
jacknagz: we were importing `main` here to avoid having to load rules in multiple places

ryandeivert: IMO importing main here is confusing as to the purpose of doing so. This also avoids the other unnecessary import that occurs by importing main. Happy to change it back if you disagree, though.

PR #76, opened 2017-03-23
Cloudtrail Example Schema, Rule, and Test

jacknagz:
size medium

## Changes
* Add default log types for `cloudtrail` eventVersions `1.04` and `1.05`.
* Add example rule for Cloudtrail
* Fix KeyError bugs in the `JSONParser` and `classifier` when parsing nested S3 data.
* Fix a bug in the `JSONParser. _key_check` where invalid records were not being removed due to the iteration technique.
* Support a new `trigger_count` key in test records to indicate the number of alerts generated by a nested record.

### DIFF
--- a/stream_alert/rule_processor/parsers.py
+++ b/stream_alert/rule_processor/parsers.py
@@ -104,19 +104,26 @@ class JSONParser(ParserBase):
         """
         schema = self._get_schema()
         schema_keys = set(schema.keys())
+        valid_records = []
+        schema_match = False
 
         for json_record in json_records:
             json_keys = set(json_record.keys())
             if json_keys == schema_keys:
+                schema_match = True
                 for key, key_type in schema.iteritems():
                     # If the value is a map of defined key/value pairs
                     if isinstance(key_type, dict) and key_type != {}:
                         # subkey check
                         if set(json_record[key].keys()) != set(schema[key].keys()):
-                            json_records.remove(json_record)
+                            schema_match = False
             else:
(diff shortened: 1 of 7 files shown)


stream_alert/rule_processor/parsers.py
```diff
@@ -104,19 +104,28 @@ def _key_check(self, json_records):
         """
         schema = self._get_schema()
         schema_keys = set(schema.keys())
+        valid_records = []
+        schema_match = False
```
ryandeivert: Since you're appending to a new list now, I think you can simplify this loop and remove the boolean that tracks the validity. Like:

```
        valid_records = []

        for json_record in json_records:
            json_keys = set(json_record.keys())
            if json_keys == schema_keys:
                for key, key_type in schema.iteritems():
                    # If the value is a map of defined key/value pairs
                    if isinstance(key_type, dict) and key_type:
                        # subkey check
                        if set(json_record[key].keys()) == set(schema[key].keys()):
                            valid_records.append(json_record)

            else:
                logger.debug('JSON Key mismatch: %s vs. %s', json_keys, schema_keys)

        return valid_records
```

Also note change for the check for the empty dict (`and key_type` from `and key_type != {}`). This same sort of logic applies to empty lists. len(dict) will also do the same. You can take that change or leave it...

ryandeivert:
LGTM

stream_alert/rule_processor/parsers.py
jacknagz: As discussed offline, this solution does not account for string/integer values in the schema.

PR #80, opened 2017-03-27
[terraform] CloudTrail Support

jacknagz:
size: medium

## Changes
Adds a new Terraform module called `tf_stream_alert_cloudtrail` which accepts a Kinesis Stream as input to write CloudTrail logs to.

The module creates the following:
* AWS CloudTrail with a S3 bucket for long term storage.  If a user has an existing Cloudtrail, they can skip the creation via the `existing_trail` flag in the module (see `README.md` for example).
* CloudWatch event rule to forward all API calls (in all regions) to a designated Kinesis stream.
* IAM policy/role to allow CloudWatch Events to PutRecords to Kinesis.

Additionally, this change comes with two new schema types: `cloudwatch:cloudtrail` and `cloudwatch:ec2_event` along with a rule test/example rules. 

## Testing
All testing was performed in a test AWS account, where end-to-end alerting was verified via S3.

### DIFF
--- a/conf/logs.json
+++ b/conf/logs.json
@@ -85,7 +85,7 @@
       "userAgent": "string",
       "sourceIPAddress": "string",
       "resources": {},
-      "recipientAccountId": "string"
+      "recipientAccountId": "integer"
     },
     "parser": "json",
     "hints": {
@@ -113,7 +113,7 @@
                   "type": "string",
                   "arn": "string",
                   "principalId": "string",
-                  "accountId": "string"
+                  "accountId": "integer"
               },
               "attributes": {
                   "creationDate": "string",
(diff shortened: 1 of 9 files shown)


terraform/modules/tf_stream_alert_cloudtrail/iam.tf
```diff
@@ -0,0 +1,45 @@
+// Role for CloudWatch events
```
strcrzy: can this be extracted into its own module? it would be useful for the flow log module too

terraform/modules/tf_stream_alert_cloudtrail/README.md
```diff
@@ -0,0 +1,85 @@
...
+* Creates an IAM Role/Policy to allow CloudWatch Events to deliver to Kinesis.
+
+## Example
+For users with no existing CloudTrail:
+```
+module "cloudtrail" {
+  source         = "modules/tf_stream_alert_cloudtrail"
+  prefix         = "streamalert"
+  cluster        = "prod"
+  enable_logging = true
```
strcrzy: this defaults to true right? should we elide it to simplify dox?

terraform/modules/tf_stream_alert_cloudtrail/README.md
```diff
@@ -0,0 +1,85 @@
...
+}
+```
+
+To skip the creation of a CloudTrail, set the `existing_trail` option to `true`:
+```
+module "cloudtrail" {
+  source         = "modules/tf_stream_alert_cloudtrail"
+  prefix         = "streamalert"
+  cluster        = "prod"
+  enable_logging = true
```
strcrzy: see above

conf/logs.json
```diff
@@ -136,5 +136,52 @@
     "hints": {
       "records": "Records[*]"
     }
+  },
+  "cloudwatch:cloudtrail": {
```
strcrzy: this diff got me thinking --  i believe that currently we short circuit to accept the first schema match. this works well as long as there are not multiple schemas that match the event. if there are, this could get sketchy.

terraform/modules/tf_stream_alert_cloudtrail/README.md
jacknagz: my thought is to give the user a quick ability to enable/disable as needed

conf/logs.json
jacknagz: @ryandeivert and I did a lot of testing regarding this.  We came to the following conclusion: Cloudtrail logs sent via Cloudwatch Events will always have the schema as defined here.  If a user is filtering on a specific API request (like EC2 for instance), they will need to define a specific schema.

terraform/modules/tf_stream_alert_cloudtrail/iam.tf
jacknagz: How so? This role module uses `events.amazonaws.com` as the Principal while I believe the flow log module delegates to a different service.

PR #82, opened 2017-03-28
[cli] makes sure to write the new config format to disk

jacknagz:
size: small

* When reading a v1 config, make sure to write it to disk.  This ensures that when we invoke Terraform with the new cluster format, all the variables are properly loaded.


ryandeivert:
LGTM

PR #83, opened 2017-03-28
[tf] ensure legacy output processor bucket names are retained

jacknagz:
size: small

Retain the legacy bucket names by replacing all `_` with `.` as it previously was.

### DIFF
--- a/terraform/modules/tf_stream_alert/main.tf
+++ b/terraform/modules/tf_stream_alert/main.tf
@@ -61,7 +61,7 @@ resource "aws_s3_bucket" "streamalerts" {
 
 // Legacy S3 bucket name - All alerts should be copied to the bucket created above.
 resource "aws_s3_bucket" "stream_alert_output" {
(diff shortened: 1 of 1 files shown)


terraform/modules/tf_stream_alert/main.tf
```diff
@@ -61,7 +61,7 @@ resource "aws_s3_bucket" "streamalerts" {
 
 // Legacy S3 bucket name - All alerts should be copied to the bucket created above.
 resource "aws_s3_bucket" "stream_alert_output" {
-  bucket        = "${var.prefix}.${var.cluster}.stream.alert.output.processor.results"
+  bucket        = "${replace("${var.prefix}.${var.cluster}.stream.alert.output.processor.results", "_", ".")}"
```
austinbyers: I always have to look twice at the nested double-quotes, but Terraform seems fine with them

jacknagz: yep, it's fine.

PR #89, opened 2017-04-03
List, Boolean, and Float Schema Types

### LINKED ISSUE #77: Improvement: Support List, Float, and Boolean Schema Types
## Background
When declaring StreamAlert schemas for log types, a user can currently only specify the following values:

```
"key": "string",
"key2": "integer",
"key3": {}
"key4": {
  "nestedKey": "string"
}
```

The schema converts a parsed payload's values into their declared type as shown above.  `{}` is a special type which accepts any Map values, which is representative of a 'loose schema'.  Alternatively, strict schema checking will occur when nested keys are declared within a Map type.

## Desired Outcome
Support `boolean`, `float`, and `[]` options for schema field types.  `[]` should have the same behavior as `{}`.

jacknagz:
size: small
resolves #77 

* Support `[]`, `boolean`, and `float` schema types.
* Update docs.
* Add unit test coverage for these changes.

### DIFF
--- a/docs/source/conf-schemas.rst
+++ b/docs/source/conf-schemas.rst
@@ -28,7 +28,7 @@ Schemas are defined in ``conf/logs.json``
       "recipientAccountId": "string"
     },
     "hints" : {
-      "records": "Records[*]"
+      "records": "Records[*]"               # JSONPath to records
     }
   },
   ....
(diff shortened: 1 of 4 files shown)


docs/source/conf-schemas.rst
```diff
@@ -38,8 +38,9 @@ Here are the basics:
 
 * Keys defined in the ``schema`` must exist
 * Values are weakly/loosely typed and enforced
-* Arrays imply zero or more elements
-* An empty map ({}) implies zero or more elements
+* Supported types: ``string``, ``integer``, ``boolean``, ``float``, ``{}``, and ``[]``
+* An empty array (``[]``) imply zero or more elements of any type
+* An empty map (``{}``) implies any key/value pairs
```
mime-frame: zero or more key/value pairs

stream_alert/rule_processor/classifier.py
```diff
@@ -300,7 +300,21 @@ def _convert_type(self, parsed_data, schema, options):
...
 
+            elif value == 'float':
+                try:
+                    payload[key] = float(payload[key])
+                except ValueError as e:
+                    logger.error('Invalid schema - %s is not a float', key)
+                    return False
+
+            elif value == 'boolean':
+                payload[key] = payload[key].lower() == 'true'
```
mime-frame: use parens so readers don't have to think about operator precedence

jacknagz: There's no precedence happening here

PR #92, opened 2017-04-05
[core] check loaded boolean type before converting

jacknagz:
size: small

* Check the loaded type of the value to convert to `boolean` before casting.  JSON may load a boolean into its correct type while other parsers will need to be converted from a string/unicode.

### DIFF
--- a/stream_alert/rule_processor/classifier.py
+++ b/stream_alert/rule_processor/classifier.py
@@ -308,7 +308,7 @@ class StreamClassifier(object):
                     return False
 
             elif value == 'boolean':
-                payload[key] = payload[key].lower() == 'true'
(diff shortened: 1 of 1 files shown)


stream_alert/rule_processor/classifier.py
```diff
@@ -308,7 +308,13 @@ def _convert_type(self, parsed_data, schema, options):
                     return False
 
             elif value == 'boolean':
-                payload[key] = payload[key].lower() == 'true'
+                if type(payload[key]) in (str, unicode):
```
ryandeivert: this seems overly complex - can we just cast it to a string instead? if it's already a string it doesn't matter, and if it's a boolean it'll cast to `True` or `False`:
`payload[key] = str(payload[key]).lower() == 'true'`

PR #93, opened 2017-04-05
[lambda][alert] fix a bug in getting the s3 bucket name to send alerts

jacknagz:
size: small

Terraform replaces all `_` with `.` to generate the name of the [StreamAlerts S3 bucket](https://github.com/airbnb/streamalert/blob/master/terraform/modules/tf_stream_alert/main.tf#L56-L60).  This change must also occur in the Lambda function to ensure we send alerts to the correct bucket.


jacknagz:
PTAL @ryandeivert

PR #96, opened 2017-04-06
Optional Top Level Keys and Configuration Log Option

### LINKED ISSUE #95: Improvement: Optional Top Level Keys in a Schema and Parser Option Overhaul
## Background

It is becoming increasingly common for certain data types to have inconsistent top level  keys.  This is also a familiar scenario during external API level changes where certain keys may or may not be present in data.  An example of this would be:

type 1:
```{"foo": 1, "bar": 2}```

changing to

type 2:
```{"foo": 1, "bar": 2, "baz": 3}```

To support this behavior, a new option will be added to the `log` declaration in `conf/logs.json`.

## Desired Outcome

Create a new setting in the `log` declaration called `configuration`,  meant for various parser options:

|setting|parser(s)|type|description|
|---|---|---|---|
|optional_top_level_keys|json|Map|Key/value pairs of optional fields with their type.  If an incoming record does not have the key, a default value will be provided based on the declared type.|
|jsonpath|json|String|The JSONPath expression used to identify and parse nested records.|
|envelope_keys|json|Map|When using nested records, there may be top level keys we want to include in the record.  This setting adds key/value pairs to a key called `envelope` in a record.|
|delimiter|csv, kv|String|The delimiter of fields in records.|
|separator|kv|String|The separator between keys and values in records.|

## Example

```
  "carbonblack:ingress.event.procstart": {
    "schema": {
      "cb_server": "string",
      "command_line": "string",
      "computer_name": "string",
      "event_type": "string",
      "expect_followon_w_md5": "boolean",
      "md5": "string",
      "parent_create_time": "integer",
      "parent_md5": "string",
      "parent_path": "string",
      "parent_process_guid": "string",
      "path": "string",
      "pid": "integer",
      "process_guid": "string",
      "sensor_id": "integer",
      "timestamp": "integer",
      "type": "string",
      "username": "string"
    },
    "parser": "json",
    "configuration": {
      "optional_top_level_keys": {
        "parent_md5": "string"
      }
    }
  }
```

### PR DESCRIPTION
**Pull request by jacknagz**
size: medium
resolves #95 

## Library Changes
* Nest all `conf/log` type options under a new key called `configuration`
  * `hints` stays at the top level, since it's not an option, but rather a guard.
  * `delimiter`, `separator` move under `configuration`.
  * `envelope` is now called `envelope_keys` and nested under `configuration`.
  * `records` is now called `json_path` and also nested under `configuration`.
* Create a new `configuration` key called `optional_top_level_keys`.
  * The goal of this option is to allow log schemas to have optional keys with specific types.
  * Sets default values for optional keys not in a specific incoming record.
* Fix various `pylint` offenses.

## Other Changes
* Update unit tests
* Update sample configurations
* Update scripts
  * Add an `autopep`  helper script
  * Update the `unit_test.sh` script to test the `stream_alert_cli` package
  * Enforce 80%+ code coverage for this repo

## Notes
* I will update documentation once the proposed changes are accepted.
* This will break most existing `conf/logs.json` settings if the keys are not updated as described above.

### DIFF
--- a/stream_alert/rule_processor/parsers.py
+++ b/stream_alert/rule_processor/parsers.py
@@ -180,7 +180,7 @@ class JSONParser(ParserBase):
                     json_payload[key_name] = default_optional_values(value_type)
 
         # Handle jsonpath extraction of records
-        if config_options and len(config_options) and records_schema:
+        if config_options and records_schema:
             records_jsonpath = jsonpath_rw.parse(records_schema)
             if len(envelope_schema):
                 self.schema.update({'envelope': envelope_schema})
--- a/stream_alert/rule_processor/parsers.py
+++ b/stream_alert/rule_processor/parsers.py
@@ -144,15 +144,15 @@ class JSONParser(ParserBase):
         json_records = []
         envelope = {}
 
-        hints = self.options.get('hints', {})
-        if hints:
-            records_schema = hints.get('records')
-            envelope_schema = hints.get('envelope', {})
+        config_options = self.options.get('configuration')
+        if config_options:
+            records_schema = config_options.get('json_path')
+            envelope_schema = config_options.get('envelope_keys', {})
 
-        if (hints and len(hints) and records_schema):
+        if (config_options and len(config_options) and records_schema):
             records_jsonpath = jsonpath_rw.parse(records_schema)
             if len(envelope_schema):
-                self.schema.update({"envelope": envelope_schema})
+                self.schema.update({'envelope': envelope_schema})
                 envelope_keys = envelope_schema.keys()
                 envelope_jsonpath = jsonpath_rw.parse("$." + ",".join(envelope_keys))
                 envelope_matches = [match.value for match in envelope_jsonpath.find(json_payload)]
@@ -161,7 +161,7 @@ class JSONParser(ParserBase):
             for match in records_jsonpath.find(json_payload):
                 record = match.value
                 if len(envelope):
-                    record.update({"envelope": envelope})
+                    record.update({'envelope': envelope})
                 json_records.append(record)
         else:
             json_records.append(json_payload)
@@ -202,8 +202,6 @@ class GzipJSONParser(JSONParser):
     def parse(self):
         """Parse a gzipped string into JSON.
 
-        Options:
-            - hints
         Returns:
             - An array of parsed JSON records.
             - False if the data is not Gzipped JSON or the columns do not match.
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: stream_alert/rule_processor/parsers.py [hunk 1]
```diff
@@ -133,26 +135,55 @@ def _parse_records(self, json_payload):
...
+                    return str()
+                elif key == 'integer':
+                    return int()
+                elif key == 'float':
+                    return float()
+                elif key == 'boolean':
+                    return bool()
+                elif key == []:
+                    return list()
+                elif key == OrderedDict():
```
austinbyers: This smells funny. As you point out, lists and dicts cannot be keys in a Python dictionary. So why do you allow them as your own keys? Why isn't it just `key == 'list'` and `key == 'dict'`?

Keep in mind, this will only match if the key is an empty `OrderedDict.` Do you mean to use `type?`

Finally, if `key == OrderedDict(),` you return `dict(),` meaning you've removed the ordering. Why the discrepancy in types?

stream_alert/rule_processor/parsers.py
```diff
@@ -133,26 +135,55 @@ def _parse_records(self, json_payload):
...
+            for key_name, value_type in optional_keys.iteritems():
+                # Update the schema to ensure the record is valid
+                self.schema.update({key_name: value_type})
+                # If the optional key isn't in our parsed json payload
+                if key_name not in json_payload:
+                    # Set default value
+                    json_payload[key_name] = default_optional_values(value_type)
+
+        # Handle jsonpath extraction of records
+        if config_options and len(config_options) and records_schema:
```
austinbyers: The `len` check is redundant: `if config_options` returns `True` only if `config_options` is not None and is non-empty

test/unit/test_classifier.py
```diff
@@ -50,10 +49,10 @@ def make_kinesis_record(**kwargs):
         raw_record = {
             'eventSource': 'aws:kinesis',
             'eventSourceARN': 'arn:aws:kinesis:us-east-1:123456789012:stream/{}'
-                .format(kinesis_stream),
+                              .format(kinesis_stream),
             'kinesis': {
                 'data': base64.b64encode(kinesis_data)
-            }
+                              }
```
austinbyers: This indentation doesn't any sense to me

stream_alert/rule_processor/parsers.py
jacknagz: @austinbyers it's not key == 'list' or 'dict' because in the schema's they are dictated as `[]` or `{}`.  Since we load as an OrderedDict, empty `{}` is converted to `OrderedDict()`.  It's mean't to be empty.

test/unit/test_classifier.py
jacknagz: autopep did it 😢

jacknagz:
Thanks for the review @austinbyers, I just addressed your comments

austinbyers:
LGTM

PR #104, opened 2017-04-09
missing step

mruba:
if you don't have any AWS experience you should guide them through this steep


jacknagz:
@mruba thanks for this PR, could you explain further when you needed to do this step?

jacknagz:
any update on this @mruba ?

PR #108, opened 2017-04-13
[lambda][output] fixing alert dispatching bugs

ryandeivert:
size: small

## Changes ##
* Fixing bug that would attempt to send alerts even if there was an error getting the configured output.
* Updating dispatching so only the record itself gets sent and not the included metadata.
* Changing wording of s3 example output.

### DIFF
--- a/stream_alert/alert_processor/main.py
+++ b/stream_alert/alert_processor/main.py
@@ -95,6 +95,7 @@ def run(loaded_sns_message, context):
         except ValueError:
             LOGGER.error('outputs for rules must be declared with both a service and a '
                          'descriptor for the integration (ie: \'slack:my_channel\')')
+            continue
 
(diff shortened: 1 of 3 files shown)


stream_alert/alert_processor/main.py
```diff
@@ -114,7 +115,7 @@ def run(loaded_sns_message, context):
         output_dispatcher.dispatch(
             descriptor=descriptor,
             rule_name=rule_name,
-            alert=alert
+            alert=alert['record']
```
chunyong-lin: Can you double check if `rule_name` included in the output? From code, looks it is not.
E.g.
```
"rule_name": "monorail_web_untrusted_commands"
```

ryandeivert: fixed it! see outputs.py

PR #114, opened 2017-04-20
[docs] updating documentation on outputs

ryandeivert:
cc @mime-frame 
size: small

* Adding documentation updates to reflect changes made to the alert processor lambda function
  * Demonstrates how to configure a new output for a currently supported service.
  * Walks through the steps that need to be taken to implement support for sending to a new service.
  * Updated to include Phantom as a supported output.
  * Updated to include AWS Lambda as a supported output.

### DIFF
--- a/docs/source/outputs.rst
+++ b/docs/source/outputs.rst
@@ -8,13 +8,110 @@ StreamAlert comes with a flexible alerting framework that can integrate with new
 
 Out of the box, StreamAlert supports:
 
-* **S3**
+* **AWS Lambda**
+* **AWS S3**
 * **PagerDuty**
+* **Phantom**
 * **Slack**
 
-It can also be extended to support any API.  Outputs will be more modular in the near future to better support additional outputs and public contributions.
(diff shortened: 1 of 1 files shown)


docs/source/outputs.rst
```diff
@@ -8,13 +8,110 @@ StreamAlert comes with a flexible alerting framework that can integrate with new
...
 Adhering to the secure by default principle, all API credentials are encrypted and decrypted using AWS Key Management Service (KMS).
+Credentials are stored on AWS S3 and are not packaged with the StreamAlert code. They are downloaded and decrypted on an as-needed basis.
+Credentials are never cached on disk in a decrypted state.
+
+Configuration
+-------------
+
+Adding a new configuration for a currently supported service is handled using ``stream_alert_cli.py``:
+
+ - ``python stream_alert_cli.py output new --service service_name``
```
mime-frame: I suggest using CAPS for placeholder values or something like `<SERVICE_NAME>`

ryandeivert: changed!

PR #117, opened 2017-04-21
[lambda][output] skip sns messages for alarms

ryandeivert:
size: small

## Changes ##
* Skipping sns records that are related to alarms
* Adding a logging.info output that will print the number of records being processed when the alert processor is ran

## Why ##
Alarm messages do not have the same 'default' required key needed to process the message. We don't really care about these messages right now anyway, so we can just skip them without logging an error. If there is an alarm message, it comes in with the format:
```
{
  "AWSAccountId": “1234567890”,
  "AlarmDescription": “This is the alarm description",
  "AlarmName": “This is the alarm name",
  "NewStateReason": "Threshold Crossed: 1 datapoint (0.5) was greater than the threshold (0.0).",
  "NewStateValue": "ALARM",
  "OldStateValue": "OK",
  "Region": "<region>",
  "StateChangeTime": "2017-04-14T23:43:11.196+0000",
  "Trigger": {
    "ComparisonOperator": "GreaterThanThreshold",
    "Dimensions": [
      {
        "name": "Resource",
        "value": “resource:qualifier"
      },
      {
        "name": "FunctionName",
        "value": "<function_name>"
      }
    ],
    "EvaluateLowSampleCountPercentile": "",
    "EvaluationPeriods": 1,
    "MetricName": "Errors",
    "Namespace": "AWS/Lambda",
    "Period": 300,
    "Statistic": "AVERAGE",
    "StatisticType": "Statistic",
    "Threshold": 0,
    "TreatMissingData": "",
    "Unit": ""
  }
}
```

### DIFF
--- a/stream_alert/alert_processor/main.py
+++ b/stream_alert/alert_processor/main.py
@@ -34,7 +34,10 @@ def handler(event, context):
             has been sent from the main StreamAlert Rule processor function
         context [AWSLambdaContext]: basically a namedtuple of properties from AWS
     """
-    for record in event.get('Records', []):
+    records = event.get('Records', [])
+    LOGGER.info('running alert processor for %d records', len(records))
+
+    for record in records:
         sns_payload = record.get('Sns')
         if not sns_payload:
             continue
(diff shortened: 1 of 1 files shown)


stream_alert/alert_processor/main.py
```diff
@@ -44,11 +47,14 @@ def handler(event, context):
...
+            continue
 
-        if not 'default' in loaded_sns_message:
+        if not 'default' in loaded_sns_message and \
+            'AlarmName' in loaded_sns_message:
+            continue    # skip sns messages related to alarms and do not log them
+        elif not 'default' in loaded_sns_message:
             LOGGER.error('Malformed SNS: %s', loaded_sns_message)
-            return
+            continue
```
austinbyers: I've never been a fan of line continuation. In any case, lines 52-57 can be simplified:
```python
if not 'default' in loaded_sns_message:
    if not 'AlarmName' in loaded_sns_message:
        LOGGER.error('...')
    continue
```

ryandeivert: Updated! Good catch

Issue #64, opened 2017-03-14
Feature: Support SNS and SQS as Sources


jacknagz:
## Background

AWS Lambda functions can be invoked from many different sources.  Currently in StreamAlert, only AWS Kinesis and Amazon S3 are supported.  In order to support a SOA based approach, allowing SNS and SQS to invoke StreamAlert's Lambda function will provide additional entry points to StreamAlert.

## Desired Outcome

* Terraform module(s) which configure SNS/SQS permissions and event mappings
* Testing workflows

jacknagz:
Closing this out since SNS is already supported, and SQS is not a push method.  We don't want to introduce latency by having the rule processor poll a stream directly, this is better suited for a StreamAlert App integration.

PR #119, opened 2017-04-25
[lambda][alert] adding support for sns as a source

ryandeivert:
size: small

work in progress for #64 
## changes ##
* Adding sns source support
* Updating tests to include sns sources

### DIFF
--- a/stream_alert/rule_processor/classifier.py
+++ b/stream_alert/rule_processor/classifier.py
@@ -130,11 +130,14 @@ class StreamClassifier(object):
             payload.service = 'kinesis'
         elif 's3' in payload.raw_record:
             payload.service = 's3'
+        elif 'Sns' in payload.raw_record:
+            payload.service = 'sns'
 
         # map the entity name from a record
         entity_mapper = {
             'kinesis': lambda r: r['eventSourceARN'].split('/')[1],
-            's3': lambda r: r['s3']['bucket']['name']
+            's3': lambda r: r['s3']['bucket']['name'],
+            'sns': lambda r: r['EventSubscriptionArn'].split(':')[5]
         }
         # get the entity name
         payload.entity = entity_mapper[payload.service](payload.raw_record)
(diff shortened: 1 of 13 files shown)


jacknagz:
anotha commit
anothaone

stream_alert/rule_processor/config.py
```diff
@@ -67,8 +67,8 @@ def validate_config(config):
 
         # check sources attributes
         elif config_key == 'sources':
-            if not set(settings.keys()).issubset(set(['kinesis', 's3'])):
-                raise ConfigError('Sources missing kinesis or s3 keys')
+            if not set(settings.keys()).issubset(set(['kinesis', 's3', 'sns'])):
```
jacknagz: instead of doing `set([])` let's just do `{}`

stream_alert/rule_processor/classifier.py
```diff
@@ -130,11 +130,14 @@ def map_source(self, payload):
             payload.service = 'kinesis'
         elif 's3' in payload.raw_record:
             payload.service = 's3'
+        elif 'Sns' in payload.raw_record:
+            payload.service = 'sns'
 
         # map the entity name from a record
         entity_mapper = {
```
jacknagz: #simplify (discussed offline)

stream_alert/rule_processor/handler.py
```diff
@@ -86,6 +88,12 @@ def s3_process(self, payload, classifier):
             classifier.classify_record(payload, data)
             self.process_alerts(payload)
 
+    def sns_process(self, payload, classifier):
+        """Process SNS data for alerts"""
+        data = StreamPreParsers.pre_parse_sns(payload.raw_record)
```
jacknagz: this method looks almost the same as the kinesis pre parser, can we simplify?

ryandeivert: I made a small change, but it just mostly consolidates duplicate code into a method. The `pre_parse..` methods pull data from different keys, so this is about as simple as this get (see `pre_parsers.py` for more context).

PR #120, opened 2017-04-25
[lambda][rule] all logs now support log_pattern matching

ryandeivert:
cc @austinbyers, @jacknagz 
size: large

## changes ##
* Moving the `hints` map that was previously used to determine if a log is valid to within the `configuration` in `logs.json` and renaming it `log_patterns`
* Simplified how `options` are passed to parsers and updated the test cases to match.
  * All options are now stored within the `configuration`, so handling this is now much cleaner.
* `ParserBase` class will now handle `log_pattern` checking with the `matched_log_pattern` method. This can be implemented on a per-parser basis if needed, but is currently functioning for all parsers from the base class.
* Parsers now take the data to be parsed as a param instead of storing as a local var
* Fixing bug where `log_source` was not being reset properly for a log payload
* Optimized `log_metadata` parsing to avoid creating a new dictionary with duplicate data. The originally loaded dict will now be pared down to include only valid log schema for the given entities.
* Removed some unused (deprecated?) code in the JSON parser related to nested keys.
* Add support for multiple schema classification.
  * If multiple schemas now match for a log, the processor will check to see if `log_patterns` was configured for the matched schemas and then check to see if any `log_patterns` match for the parsed data.
  * If `matched_log_pattern` does not determine the proper log, the processor will report that this log classified as multiple schema, and then default to the first schema for rule processing.
  * Type conversion for the log only occurs once all parsing and classification has occurred to avoid a ton of extra processing.


stream_alert/rule_processor/classifier.py
```diff
@@ -170,21 +159,19 @@ def log_metadata(self, payload):
...
+        config_logs = self.config['logs']
 
-        all_config_logs = self.config['logs']
-        entity_log_sources = self._payload_logs(payload)
-        for log_source, log_source_attributes in all_config_logs.iteritems():
-            source_pieces = log_source.split(':')
-            category = source_pieces[0]
-            if category in entity_log_sources:
-                metadata[log_source] = log_source_attributes
+        for log_source in config_logs.keys():
```
austinbyers: nit: `.keys()` is unnecessary if `config_logs` is a dictionary:
```python
for log_source in config_logs:
```

stream_alert/rule_processor/classifier.py
```diff
@@ -224,54 +211,82 @@ def _parse(self, payload, data):
...
-                    return True
-        return False
+                passes.append((log_name, parser, parsed_data))
+
+        if not passes:
+            return False
 
-    def _convert_type(self, parsed_data, schema, options):
+        valid_parse = passes[0]
+        if len(passes) > 1:
```
austinbyers: nit: simplify:
```python
if passes:
```

stream_alert/rule_processor/classifier.py
```diff
@@ -224,54 +211,82 @@ def _parse(self, payload, data):
...
+            for i, valid_parses in enumerate(passes):
+                parser = valid_parses[1]
+                for data in valid_parses[2]:
+                    if parser.matched_log_pattern(data, parser.options.get('log_patterns', {})):
+                        matched_parses.append(passes[i])
+                        break
+                    else:
+                        LOGGER.debug('log pattern matching failed for schema: %s', parser.schema)
+
+            if len(matched_parses) != 0:
```
austinbyers:
```python
if matched_parses:
```

stream_alert/rule_processor/classifier.py
```diff
@@ -299,39 +313,44 @@ def _convert_type(self, parsed_data, schema, options):
...
-            elif isinstance(value, list):
-                pass
-
-            elif isinstance(value, (OrderedDict)):
+            elif isinstance(value, OrderedDict):
                 # allow for any value to exist in the map
-                if len(value) == 0:
-                    pass
-                else:
+                if len(value) != 0:
```
austinbyers: Since we know at this point that `value` is an `OrderedDict`, then
```python
if value:
```

mime-frame:
Great work on this @ryandeivert, this is awesome.

stream_alert/rule_processor/classifier.py
```diff
@@ -224,54 +211,82 @@ def _parse(self, payload, data):
...
+        LOGGER.debug('parsed_data: %s', parsed_data)
+
+        typed_data = []
+        for data in parsed_data:
+            # Convert data types per the schema
+            # Use the parser.schema due to updates caused by
+            #   configuration settings such as envelope and optional_keys
+            converted_data = self._convert_type(data, parser.type(), parser.schema, parser.options)
+            if not converted_data:
+                payload.valid = False
```
austinbyers: Was `payload.valid_source` replaced by `payload.valid`? Where is this used?

stream_alert/rule_processor/classifier.py
```diff
@@ -299,39 +313,44 @@ def _convert_type(self, parsed_data, schema, options):
...
-                        parse_csv = get_parser('csv')
-                        parsed_nested_key = parse_csv(payload[key],
-                                                      schema[key],
-                                                      options).parse()
-                        # Call the first element since a list is returned
+                    # skip the 'stream_log_envelope' key that we've added during parsing
+                    if all((key == 'stream_log_envelope', isinstance(payload[key], dict))):
+                        continue
+
+                    if 'log_patterns' in options.keys():
```
austinbyers:
```python
if 'log_patterns' in options:
```

conf/logs.json
```diff
@@ -212,5 +214,89 @@
...
+      }
+    },
+    "parser": "json",
+    "configuration": {
+      "log_patterns": {
+        "type": [
+          "event01"
+        ],
+        "test": {
+          "key01": [
```
mime-frame: any reason to not use keys actually found/used in "the wild"?

stream_alert/rule_processor/classifier.py
```diff
@@ -299,39 +313,44 @@ def _convert_type(self, parsed_data, schema, options):
...
                     # handle nested csv
-                    if isinstance(payload[key], str):
-                        options['hints'] = options['hints'][key]
-                        parse_csv = get_parser('csv')
-                        parsed_nested_key = parse_csv(payload[key],
-                                                      schema[key],
-                                                      options).parse()
-                        # Call the first element since a list is returned
+                    # skip the 'stream_log_envelope' key that we've added during parsing
+                    if all((key == 'stream_log_envelope', isinstance(payload[key], dict))):
```
austinbyers: Simplify:
```python
if key == 'stream_log_envelope' and isinstance(payload[key], dict):
```

This avoids building a tuple object

stream_alert/rule_processor/classifier.py
mime-frame: If possible, escape early vs having long indentation blocks. Example:

```
if len(passes) <= 1)`
    blah
... the rest of the code ...
```

stream_alert/rule_processor/classifier.py
```diff
@@ -142,19 +130,20 @@ def map_source(self, payload):
                 break
 
         # if the payload's entity is found in the config and contains logs
-        if self._payload_logs(payload):
-            payload.valid_source = True
+        self._entity_log_sources = self._payload_logs(payload)
+
+        return self._entity_log_sources and len(self._entity_log_sources) > 0
```
austinbyers: Add return information to the docstring

stream_alert/rule_processor/parsers.py
```diff
@@ -50,77 +50,96 @@ def parser(cls):
...
-        self.payload_type = None
+        self.options = options or {}
 
     @abstractmethod
-    def parse(self):
+    def parse(self, data):
         """Main parser method to be overridden by all Parser classes
 
+        Args:
+            data: Data to be parsed.
```
austinbyers: What's the type? A `dict`?

stream_alert/rule_processor/parsers.py
```diff
@@ -50,77 +50,96 @@ def parser(cls):
...
-    __parserid__ = 'json'
+    def type(self):
+        """Returns the type of parser. Overriden in GzipJSONParser to just return json"""
+        return self.__parserid__
+
+    def matched_log_pattern(self, record, log_patterns):
+        """Return True if all log patterns of this record match"""
+        # Return False immediately if there are no log patterns
+        # or if the data being tested is not a dict
+        if not all((log_patterns, isinstance(record, dict))):
```
austinbyers: Simplify and make more readable:

```python
if not log_patterns or not isinstance(record, dict):
```

stream_alert/rule_processor/parsers.py
```diff
@@ -50,77 +50,96 @@ def parser(cls):
...
+            # the pattern field value in the record
+            try:
+                value = record[field]
+            except (KeyError, TypeError):
+                LOGGER.debug('declared log pattern field [%s] is not a valid field '
+                             'for this record: %s', field, record)
+                continue
+            # append the result of any of the log_patterns being True
+            pattern_result.append(any([fnmatch(value, pattern)
+                                       for pattern in pattern_list]))
```
austinbyers: No need to build the list, use generator syntax instead:
```python
pattern_result.append(
    any(fnmatch(value, pattern) for pattern in pattern_list))
```

stream_alert/rule_processor/parsers.py
```diff
@@ -179,89 +192,94 @@ def default_optional_values(key):
...
             for match in records_jsonpath.find(json_payload):
                 record = match.value
                 if len(envelope):
-                    record.update({'envelope': envelope})
+                    record.update({'stream_log_envelope': envelope})
+
                 json_records.append(record)
-        else:
+
+        if len(json_records) == 0:
```
austinbyers:
```python
if not json_records:
```

stream_alert/rule_processor/parsers.py
```diff
@@ -273,92 +291,64 @@ def _get_reader(self, config_options):
...
-                # if all hint group results are True
-                logger.debug('hint result: %s', hint_result)
-                if all(hint_result):
-                    self.payload_type = 'csv'
-                    for index, key in enumerate(schema):
-                        csv_payload[key] = row[index]
-
-                    csv_payloads.append(csv_payload)
+                # extract the keys from the row via the index
+                csv_payloads.append({key: row[index] for index, key in enumerate(self.schema)})
```
austinbyers: Beautiful! Brings tears to my eyes

conf/logs.json
```diff
@@ -212,5 +214,89 @@
         "owner": "integer"
       }
     }
+  },
+  "carbonblack:event01": {
```
ryandeivert: pls tell me how to link to docs in json? :)

ryandeivert: also this was a constructed log since didn't know if we were concerned about publishing CB formats. I now know that's not the case, so will be updating these will valid CB schemas

stream_alert/rule_processor/classifier.py
austinbyers: The first half isn't necessary either. @ryandeivert pointed out that we can just do:
```python
return bool(self._entity_log_sources)
```

stream_alert/rule_processor/classifier.py
ryandeivert: No - `payload.valid_source` was an unnecessary var. `payload.valid` is different and has been there to store if any log in a payload does not get classified properly.

ryandeivert:
@jacknagz , @mime-frame , @austinbyers made requested changes and some additional updates. PTAL tomorrow. 
face-folding-kirk-googly-eyes-364x196

Issue #25, opened 2017-02-07
Feature: Rule Descriptions via Docstrings


jacknagz:
Idea: Add descriptions/context to rules in function docstrings which are sent along with alerts:

```
@rule('invalid_subnet',
      logs=['osquery'],
      matchers=['logged_in_users'],
      outputs=['pagerduty'])
def invalid_subnet_rule(rec):
    """Catch logins from unauthorized subnets"""
    valid_cidr = IPNetwork('10.2.0.0/24')
    ip = IPAddress(rec['columns']['host'])

    return ip not in valid_cidr
```

Improved alert structure in `rules_engine.py`:
```
alert = {
  'rule_name': rule.rule_name,
  'rule_description': rule.rule_function.__doc__,
  'payload': payload,
  'outputs': rule.outputs
}
```

PR #125, opened 2017-04-28
[lambda][rule] send alerts with rule docstring

ryandeivert:
size: medium
resolve: #25 

## changes ##
* Alerts will now include a `rule_description` key within the `metadata` dictionary that is packaged with the alert.
* Moving the `rule_name` key that is sent with the alert into the `metadata` dictionary as well since it makes more sense and is cleaner.
  * Alerts are now made up of two dictionaries: `metadata` (containing info related to this rule/alert) and `record` (containing the actual payload that triggered this alert).
* Updating tests to conform to the new `metadata` dict.
* Adding the `rule_description` as part of the alert sent to applicable configurations.
  * Slack will now receive a `rule_description` as part of the message text
  * PagerDuty will now receive a `rule_description` as part of the details
  * Phantom will now receive a `rule_description` as part of the container description
  * AWS S3 and Lambda do not receive `rule_description` as part of the alert (as it's not currently necessary and there is no obvious way to handle it).
* Setting a default placeholder for rules that do not provide a docstring to be used for `rule_description`: `'No rule description provided'`

### DIFF
--- a/stream_alert/alert_processor/main.py
+++ b/stream_alert/alert_processor/main.py
@@ -69,9 +69,10 @@ def run(loaded_sns_message, context):
         The alert is another dict with the following structure:
 
         {
-            'rule_name': rule.rule_name,
             'record': record,
             'metadata': {
+                'rule_name': rule.rule_name,
+                'rule_description': rule.rule_function.__doc__,
                 'log': str(payload.log_source),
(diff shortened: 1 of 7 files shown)


mime-frame:
image

PR #126, opened 2017-04-28
[lambda][rule] some s3 optimizations

ryandeivert:
size: small

## changes ##
* Will now yield and parse lines from an s3 file as they are read instead of reading once and then iterating over the lines a second time to parse.

### DIFF
--- a/stream_alert/rule_processor/pre_parsers.py
+++ b/stream_alert/rule_processor/pre_parsers.py
@@ -13,7 +13,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 '''
-
 import base64
 import gzip
 import logging
@@ -65,7 +64,7 @@ class StreamPreParsers(object):
         size = int(raw_record['s3']['object']['size'])
         downloaded_s3_object = cls._download_s3_object(client, bucket, key, size)
 
-        return cls._read_s3_file(downloaded_s3_object)
+        return downloaded_s3_object
 
     @classmethod
     def pre_parse_sns(cls, raw_record):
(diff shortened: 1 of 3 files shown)


stream_alert/rule_processor/pre_parsers.py
```diff
@@ -79,6 +78,32 @@ def pre_parse_sns(cls, raw_record):
...
+
+        if extension == '.gz':
+            for line in gzip.open(downloaded_s3_object, 'r'):
+                yield line.rstrip()
+        else:
+            for line in open(downloaded_s3_object, 'r'):
+                yield line.rstrip()
+
+        # remove the file
+        os.remove(downloaded_s3_object)
```
austinbyers: This doesn't work as expected in Lambda - you have to truncate the file first to avoid using all of the disk space

stream_alert/rule_processor/pre_parsers.py
```diff
@@ -79,6 +78,32 @@ def pre_parse_sns(cls, raw_record):
...
     @classmethod
+    def read_s3_file(cls, downloaded_s3_object):
+        """Parse a downloaded file from S3
+
+        Supports reading both gzipped files and plaintext files.
+
+        Args:
+            downloaded_s3_object (string): A full path to the downloaded file.
+
+        Returns:
```
austinbyers: `Yields:` (not returns)

stream_alert/rule_processor/pre_parsers.py
ryandeivert: Hmm that's strange and doesn't make much sense, but okay. I will add the truncate back.

austinbyers: Yeah, Jack and I both independently observed the phenomenon. `os.remove` is really `os.unlink`, and presumably Lambda's version of `unlinking` is not the same as a traditional OS

jacknagz:
lgtm-wally

PR #128, opened 2017-04-30
[testing] improving rule processor test output by adding some summary info

ryandeivert:
size: small

## changes ##
* Updating the rule processor testing output to show a summary at the end indicating how many tests passed and how many failed out of the total ran.
* The tail of the output will now appear like the following if there are failures:
```
(33/35)	Tests Passed
(2/35)	Tests Failed
	(1/2) test failed for rule: invalid_user [user not in the whitelist (john)]
	(2/2) test failed for rule: sample_cloudwatch_cloudtrail_rule [cloudwatch cloudtrail 'Decrypt' event log]
```
* The `(#/#) Tests Passed` line will also be green, while all lines following it related to failures will be red.
* If all tests pass, only the first line above will be output, but would be:
```
(35/35)	Tests Passed
```
* This should aid in seeing which rules are passing and which are failing as the amount of rules being tested grows.

### update ###
* adding warnings as part of the output. Warnings can occur for three reasons:
  * test file (rule file) is not properly formatted json
  * test file contains a 'records' root key with 0 test records
  * test file contains records that are malformed
* The output for warnings will look like:
```
3 	Warnings
	(1/3) improperly formatted file - ValueError: Extra data: line 1 column 8 - line 1 column 13 (char 7 - 12) [carbonblack_ingress_event_regmod2.json]
	(2/3) no records to test in file [invalid_subnet2.json]
	(3/3) improperly formatted record: {u'test': 20} [sample_cloudwatch_cloudtrail_rule2.json]
```

### DIFF
--- a/stream_alert_cli/test.py
+++ b/stream_alert_cli/test.py
@@ -53,6 +53,33 @@ def report_output(cols, failed):
 
     print '\t{}\ttest ({}): {}'.format(status, *cols)
 
+def report_output_summary(rules_fail_pass):
+    """Helper function to print the summary results of all tests
+    Args:
+        rules_fail_pass [list]: A list containing two lists for failed and passed
+            rule tests. The sublists contain tuples made up of: (rule_name, rule_description)
+    """
+    failed_tests = len(rules_fail_pass[0])
+    passed_tests = len(rules_fail_pass[1])
+    total_tests = failed_tests + passed_tests
+
+    # Print a message indicating how many of the total tests passed
+    print '{}({}/{})\tTests Passed{}'.format(COLOR_GREEN, passed_tests, total_tests, COLOR_RESET)
+
+    # Check if there were failed tests and report on them appropriately
+    if rules_fail_pass[0]:
+        # Print a message indicating how many of the total tests failed
+        print '{}({}/{})\tTests Failed'.format(COLOR_RED, failed_tests, total_tests)
+
+        color = COLOR_RED
+        # Iterate over the rule_name values in the failed list and report on them
+        for i, failure in enumerate(rules_fail_pass[0]):
+            if i == failed_tests-1:
+                # Change the color back so std out is not red
+                color = COLOR_RESET
+            print '\t({}/{}) test failed for rule: {} [{}]{}'.format(i+1, failed_tests, failure[0],
+                                                                     failure[1], color)
+
 def test_rule(rule_name, test_record, formatted_record):
     """Feed formatted records into StreamAlert and check for alerts
     Args:
@@ -215,6 +242,10 @@ def test_alert_rules():
     BOTO_MOCKER.start()
     tests_passed = True
 
+    # Create a list for pass/fails. The first value in the list is a list of tuples for failures,
+    # and the second is list of tuples for passes. Tuple is (rule_name, rule_description)
+    rules_fail_pass = [[], []]
+
     for root, _, rule_files in os.walk(DIR_RULES):
         for rule_file in rule_files:
             rule_name = rule_file.split('.')[0]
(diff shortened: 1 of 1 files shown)


stream_alert_cli/test.py
```diff
@@ -53,6 +53,33 @@ def report_output(cols, failed):
 
     print '\t{}\ttest ({}): {}'.format(status, *cols)
 
+def report_output_summary(rules_fail_pass):
+    """Helper function to print the summary results of all tests
+    Args:
+        rules_fail_pass [list]: A list containing two lists for failed and passed
+            rule tests. The sublists contain tuples made up of: (rule_name, rule_description)
+    """
+    failed_tests = len(rules_fail_pass[0])
```
jacknagz: should we add a key for tests that result in error?

ryandeivert: hm that would be cool but kinda hard to catch... errors are just printed along the way and don't bubble up. any suggestions?

jacknagz: I was thinking catching errors [here](https://github.com/airbnb/streamalert/blob/1fbd2f99a970df31e3a4bb4f3a0bccfd8fc340e6/stream_alert_cli/test.py#L261-L262) and [here](https://github.com/airbnb/streamalert/blob/1fbd2f99a970df31e3a4bb4f3a0bccfd8fc340e6/stream_alert_cli/test.py#L267-L268).  I think it would increase the clarity of the result since tests never actually run in those cases

mime-frame:
> Updating the rule processor testing output to show a summary at the end indicating how many tests passed and how many failed out of the total ran.

Nice, been wanting this :)

stream_alert_cli/test.py
ryandeivert: Oh okay so you mean errors that are local to this method, not errors that might occur throughout processing. Let me think about this a bit since the logic will have to change to accommodate this.

ryandeivert: @jacknagz made some changes to support warnings. PTAL. output will be similar to:
screen shot 2017-05-03 at 8 51 25 pm

stream_alert_cli/test.py
```diff
@@ -53,6 +54,45 @@ def report_output(cols, failed):
...
+    print '{}({}/{})\tTests Passed{}'.format(COLOR_GREEN, passed_tests, total_tests, COLOR_RESET)
+
+    # Check if there were failed tests and report on them appropriately
+    if rules_fail_pass[0]:
+        color = COLOR_RED
+        # Print a message indicating how many of the total tests failed
+        print '{}({}/{})\tTests Failed'.format(color, failed_tests, total_tests)
+
+        # Iterate over the rule_name values in the failed list and report on them
+        for i, failure in enumerate(rules_fail_pass[0]):
```
jacknagz: nit: please use `indx` or equivalent instead of `i` (for here and the block below)

ryandeivert: that's the go in me :(

stream_alert_cli/test.py
```diff
@@ -53,6 +54,47 @@ def report_output(cols, failed):
...
+    Args:
+        rules_fail_pass [list]: A list containing two lists for failed and passed
+            rule tests. The sublists contain tuples made up of: (rule_name, rule_description)
+    """
+    failed_tests = len(rules_fail_pass[0])
+    passed_tests = len(rules_fail_pass[1])
+    total_tests = failed_tests + passed_tests
+
+    # Print a message indicating how many of the total tests passed
+    print '{}({}/{})\tTests Passed{}'.format(COLOR_GREEN, passed_tests, total_tests, COLOR_RESET)
```
jacknagz: Can you add two newlines above the `Tests Passed` string

ryandeivert: Picky much

PR #130, opened 2017-05-02
Remove base64 encoding from SNS input

austinbyers:
Incoming SNS data should not be assumed to be base64-encoded, because it is designed to be a simple notification for email, texting, etc.

For example, the SNS message that the rule processor uses to [invoke the alert processor](https://github.com/airbnb/streamalert/blob/abb--sns-no-base64/stream_alert/rule_processor/sink.py#L112) is not base64-encoded.

@airbnb/streamalert-maintainers

### DIFF
--- a/stream_alert/rule_processor/pre_parsers.py
+++ b/stream_alert/rule_processor/pre_parsers.py
@@ -73,9 +73,9 @@ class StreamPreParsers(object):
         Args:
             raw_record (dict): An SNS message.
 
-        Returns: (string) Base64 decoded data.
+        Returns: (string) SNS message data.
         """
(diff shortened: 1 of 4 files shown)


stream_alert/rule_processor/pre_parsers.py
```diff
@@ -73,9 +73,9 @@ def pre_parse_sns(cls, raw_record):
         Args:
             raw_record (dict): An SNS message.
 
-        Returns: (string) Base64 decoded data.
+        Returns: (string) SNS message data.
         """
-        return base64.b64decode(raw_record['Sns']['Message'])
+        return raw_record['Sns']['Message']
```
austinbyers: Technically, this returns unicode

PR #133, opened 2017-05-04
streamalert - rules.rst - outputs

mime-frame:

Denote that one element must exist in `outputs`


mime-frame:
@jacknagz done

PR #135, opened 2017-05-05
[lambda][output] updating slack output format to support large messages

ryandeivert:
size: medium 

## changes ##
* Slack output will now use attachments and split the message on the Slack recommended size of 4000 characters/bytes.
* Output will also be formatted with tabs and "markdown" styling.

* Messages will look similar to this in Slack (with the proper App name/icon):
image

* Multi-part messages will have a `(Part # of n)` string after `Record`


mime-frame:
Fixed https://github.com/airbnb/streamalert/pull/135

PR #137, opened 2017-05-05
[lambda][rule] processed logs will now be sent immediately to the alert processor

ryandeivert:
size: medium

## changes ##
* Refactoring the rule processor so alerts are sent to the alert processor (via sns) as they are processed.
  * Prior to these changes, we would first iterator over all logs and apply the rule logic against them, caching any that matched a rule along the way. We would finally send them to the alert processor (one at a time) after all logs had been processed.
  * These change will make this all happen sequentially and any logs that match a rule will get sent as we process (instead of caching and sending at the end).
* Will log a message on every 100 records processed from s3 so we can track progress.
  * Adding some logic that will also try to estimate the number of records in the entire file during processing (we use a generator to avoid reading all the lines multiple times, and getting the exact line count would require reading the entire file twice).
* Other indirect gains:
  * boto3 client creation now happens when the `StreamSink` class is instantiated and recycled to avoid continuously re-creating it.
  * the `StreamSink` class does not cache the alerts list and will send on demand
* Removing/updating some old code (ie: things related to `staging`) that is no longer relevant.
* Updating tests where applicable

### DIFF
--- a/stream_alert/rule_processor/sink.py
+++ b/stream_alert/rule_processor/sink.py
@@ -34,7 +34,7 @@ def json_dump(sns_dict, indent_value=None):
     try:
         return json.dumps(sns_dict, indent=indent_value, default=json_dict_serializer)
     except AttributeError as err:
-        logging.error('An error occurred while dumping object to JSON: %s', err)
+        LOGGER.error('An error occurred while dumping object to JSON: %s', err)
         return ""
 
 class SNSMessageSizeError(Exception):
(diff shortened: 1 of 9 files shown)


stream_alert/rule_processor/sink.py
```diff
@@ -117,18 +114,20 @@ def publish_message(self, client, message, topic):
...
-                logging.error('An error occurred while publishing alert: %s', err.response)
-                raise err
-            logger.info('Published %i alert(s) to %s', len(self.alerts), topic)
-            logger.info('SNS MessageID: %s', response['MessageId'])
-        else:
+        if not self._sns_message_size_check(message):
             logging.error('Cannot publish Alerts, message size is too big!')
             raise SNSMessageSizeError('SNS message size is too big! (Max: 256KB)')
+
+        try:
```
jacknagz: can we add SNS tests via moto to verify the messages would send?

Issue #79, opened 2017-03-25
Bug: if pip.main() fails, the lambda package should not be uploaded


ryandeivert:
If the `pip.main(pip_command)` call within `stream_alert_cli/package.py` fails, it does so silently, and the lambda package will continue to be uploaded. We should check for failures and stop the process.

`pip.main()` returns a non-zero error code on failure, so this should be easy to evaluate and handle.

PR #140, opened 2017-05-08
[package] will log error and exit with code 1 if/when pip fails

ryandeivert:
size: small
resolves: #79 

* Checking the `pip.main()` return code and logging error/exiting if anything other than 0 is encountered.

### DIFF
--- a/stream_alert_cli/package.py
+++ b/stream_alert_cli/package.py
@@ -53,7 +53,10 @@ class LambdaPackage(object):
         temp_package_path = self._get_tmpdir()
         self._copy_files(temp_package_path)
         # download third-party libs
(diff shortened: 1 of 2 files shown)


stream_alert_cli/package.py
```diff
@@ -53,7 +53,10 @@ def create_and_upload(self):
         temp_package_path = self._get_tmpdir()
         self._copy_files(temp_package_path)
         # download third-party libs
-        self._resolve_third_party(temp_package_path)
+        if self._resolve_third_party(temp_package_path) != 0:
```
jacknagz: you can remove the `!= 0`, since 0 is falsey anyway

PR #145, opened 2017-05-09
[tf] upgrade statefile to 0.9.0

jacknagz:
size: small

Terraform's most recent minor version (0.9.0) introduced a change in remote state file management.   This PR removes usage of `terraform remote config` commands as it is deprecated in the most recent version, and removes all references to `refresh_state()`.

**IMPORTANT: All users should follow these steps for upgrading:**

```
With the older Terraform version (version 0.8.x), run terraform remote pull.
This will cache the latest legacy remote state data locally. 
We'll use this for a backup in case things go wrong.

Backup your terraform/.terraform/terraform.tfstate file. 
This contains the cache we just pulled. 
Please copy this file to a location outside of your Terraform module.

Run the terraform init command. 
This is an interactive process that will guide you through migrating your existing
remote state to the new backend system. During this step, Terraform may ask 
if you want to copy your old remote state into the newly configured backend.
If you configured the identical backend location, you may say no since it should already be there.

Verify your state looks good by running stream_alert_cli.py terraform build and 
seeing if it detects your infrastructure. Advanced users may run terraform state 
pull which will output the raw contents of your state file to your console. You can 
compare this with the file you saved. There may be slight differences in the serial 
number and version data, but the raw data should be almost identical.
```


jacknagz:
Working on an update to this PR, it seems there are some breaking changes in `0.9.0` to how StreamAlert runs initialization

jacknagz:
Update: I've been in contact with some engineers at Hashicorp about the breaking of the initialization process.  Should have an updated commit soon.

PR #153, opened 2017-05-16
[parsers] fixing envelope_key value bug

ryandeivert:
to @austinbyers 

* Fixing bug where an empty value for a key that is an `envelope_key` would cause `envelope_key` validation to fail in the json parser. Making this change so the code will now look if the _key_ exists, and not worry about what _value_ said key has.


austinbyers:
Great find! A subtle bug

PR #155, opened 2017-05-17
[lambda][outputs] adding a check for credentials

ryandeivert:
to @jacknagz 

### changes ###
* Fixing a bug where the `_load_creds` method could return `None` if there was an error pulling/decrypting credentials. The result was not being validated and we would attempt to load info from a `None` object.
* Also fixing an logging call that was not using our `LOGGER` instance.
* Moving Phantom endpoint urls suffixes to constants on the `PhantomOutput` class.
  * Migrating `container_url` to within the `_setup_container` method since it's not used outside of that scope.

### DIFF
--- a/stream_alert/alert_processor/outputs.py
+++ b/stream_alert/alert_processor/outputs.py
@@ -100,6 +100,9 @@ class PagerDutyOutput(StreamOutputBase):
                 alert [dict]: Alert relevant to the triggered rule
         """
         creds = self._load_creds(kwargs['descriptor'])
+        if not creds:
+            self._log_status(False)
+            return
 
         message = 'StreamAlert Rule Triggered - {}'.format(kwargs['rule_name'])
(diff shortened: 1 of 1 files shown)


stream_alert/alert_processor/outputs.py
```diff
@@ -181,7 +187,7 @@ def _setup_container(self, rule_name, rule_description, container_url, headers):
         try:
             resp_dict = json.loads(resp.read())
         except ValueError as err:
-            logging.error('An error occurred while decoding message to JSON: %s', err)
+            LOGGER.error('an error occurred while decoding phantom response to JSON: %s', err)
```
austinbyers:
```python
LOGGER.exception('An error occurred while decoding phantom response to JSON')
```

austinbyers:
LGTM after nit

PR #156, opened 2017-05-23
[cli][outputs] the cli will handle instances of the secrets bucket not existing better

ryandeivert:
to @jacknagz 
size: small

## changes ##
* The cli will handle instances of the secrets bucket not existing better.
* Also updating docs to include a note about terraforming on first run to get the `streamalert.secrets` bucket.


docs/source/outputs.rst
```diff
@@ -34,6 +34,10 @@ Adding a new configuration for a currently supported service is handled using ``
 For example:
  - ``python stream_alert_cli.py output new --service slack``
 
+.. Note:: If this is the first time you have configured new outputs via the cli, you may see this error for certain services:
+ `An error occurred while sending credentials to S3 for key [<SERVICE>/<KEY>]: The specified bucket does not exist [<PREFIX>.streamalert.secrets]`
+ If you encounter this error, you must first run `python stream_alert_cli.py terraform build`. This ensures the S3 bucket used for storing encrypted secrets is created and only needs to occur once.
```
mime-frame: and only needs to be run once.

ryandeivert: funny because I was tossing these two verbiages back and forth and couldn't settle on one

ryandeivert:
@mime-frame PTAL - tweaked a bit. also removed old `secrets.rst` file

docs/source/secrets.rst
```diff
@@ -1,33 +0,0 @@
-Secrets
```
mime-frame: @ryandeivert why are we deleting the entire `secrets` doc and the link to it?

ryandeivert: This page described the old way of storing secret info, and should have been removed when the outputs refactor happened but was overlooked.

We no longer require the user to use `aws kms` from the command line and handle all the overhead for them.

mime-frame: @ryandeivert i might be derp'ing, but `docs/source/outputs.rst` doesn't appear to cover the new way of doing secrets, where is it covered?

ryandeivert: @mime-frame Just pushed a commit to also remove the link from the TOC in `index.rst`

ryandeivert: @mime-frame secrets are transparently sent kms/s3 behind the scenes and the process outlined `outputs.html#configuration` walks the user through adding the output.

As simple as:
`python stream_alert_cli.py output new --service slack`

Which then prompts for things like channel name and integration key (the latter of which is sent to kms/s3 for encryption/storage).

PR #160, opened 2017-06-01
[cli] changing back a bucket name that was used to test a recent feature

ryandeivert:
to @jacknagz 

* Renaming secrets bucket back to the right name. This was changed to test catching a bucket not found error and properly report it.


jacknagz:
👍

PR #161, opened 2017-06-01
[lambda][rule] fixing up some overlooked code from a previous optimization change

ryandeivert:
to @jacknagz 

* Changing to using the cached SNS client instead of instantiating one inside of a loop every time we publish to sns. This should have been changed previously but was overlooked or overwritten.
* Also caching the topic upon instantiation instead of re-creating it every time.

### DIFF
--- a/stream_alert/rule_processor/sink.py
+++ b/stream_alert/rule_processor/sink.py
@@ -19,7 +19,7 @@ import logging
 import sys
 
 import boto3
-import botocore
+from botocore.exceptions import ClientError
 
 _SNS_MAX_SIZE = (256*1024)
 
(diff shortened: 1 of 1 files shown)


stream_alert/rule_processor/sink.py
```diff
@@ -116,20 +115,19 @@ def publish_message(self, client, message, topic):
...
                 Subject='StreamAlert Rules Triggered'
             )
-        except botocore.exceptions.ClientError as err:
+        except ClientError as err:
             LOGGER.error('An error occurred while publishing alert: %s', err.response)
             raise err
 
         if response['ResponseMetadata']['HTTPStatusCode'] != 200:
-            LOGGER.error('Failed to publish message to sns topic: %s', topic)
+            LOGGER.error('Failed to publish message to sns topic: %s', self.topic)
```
jacknagz: can we just log the topic name (the last element in the ARN)

ryandeivert:
@jacknagz updated PTAL

mime-frame:
image

PR #162, opened 2017-06-05
[lambda][output] moving some variable instantiation to outside a loop…

ryandeivert:
to @jacknagz 

## Changes ##
* Moving some variable instantiation to outside a loop
  * Previously, the same variables would be re-instantiated within a loop with the same value. These are now passed in as a param.
* Fixing bug when loading output config where a `None` object could be returned if the config failed to load.

### DIFF
--- a/stream_alert/alert_processor/main.py
+++ b/stream_alert/alert_processor/main.py
@@ -37,6 +37,14 @@ def handler(event, context):
     records = event.get('Records', [])
     LOGGER.info('Running alert processor for %d records', len(records))
 
+    # A failure to load the config will log the error in load_output_config and return here
+    config = load_output_config()
+    if not config:
+        return
+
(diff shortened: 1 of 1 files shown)


stream_alert/alert_processor/main.py
```diff
@@ -54,9 +62,9 @@ def handler(event, context):
...
             continue
 
-        run(loaded_sns_message, context)
+        run(loaded_sns_message, region, function_name, config)
 
-def run(loaded_sns_message, context):
+def run(loaded_sns_message, region, function_name, config):
     """Send an Alert to its described outputs.
 
     Args:
```
jacknagz: pls update the args

ryandeivert: Thanks - amended previous commit

Issue #44, opened 2017-02-28
Improvement: JSON Cluster Templates


jacknagz:
**Background**

The `$ ./stream_alert_cli.py terraform generate` command is used to dynamically create Terraform files which represent a StreamAlert cluster.  This works by using a [cluster_template](https://github.com/airbnb/streamalert/blob/master/terraform/templates/cluster_template), and passing in variables to render a desired cluster plan.  Since Terraform plans can either be HCL (Hashicorp Config Language) or JSON, a simplified version of this process could occur via Python dictionaries with JSON marshaling.  

Additionally, adding new Terraform modules should be easier to plugin to this dynamic cluster generation process.  It would be ideal if entire clusters could be described in `variables.json` verses one-off variables like `s3_event_buckets`.

**Desired Outcome**

- [ ] Express the entire StreamAlert infrastructure in `variables.json` in the form of Terraform module names.
- [x] Generate all cluster files as JSON instead of HCL (Terraform can understand either).

PR #164, opened 2017-06-06
Terraform 0.9.0, Generative TF Files, and more

jacknagz:
to @austinbyers @strcrzy 

## Background
When HashiCorp upgraded Terraform to 0.9.0, they deprecated `remote_states` in favor of `backend` ([info here](https://www.terraform.io/upgrade-guides/0-9.html)).  This PR makes StreamAlert future proof by moving to `backend`, and includes boilerplate code to dynamically generate entire cluster files (more details below).

To get `backend` to work, we first Terraform several resources in the init phase with local state.  After, the `s3` backend is enabled and then the remainder infrastructure is built.  The same steps happen in reverse during destruction.

## Changes
* Remove `remote_state_refresh()`: This is a deprecated command in Terraform > 0.9.0.
* Remove Jinja templating in favor of generating JSON.
* Better reliability of Terraform commands: Break execution if preceding steps fail.
* Cleanup cluster files on destroy, along with leftover state files.
* Remove default example SNS topic: This breaks Terraform (since it's not a valid ARN)
* Suppress `pip` logger messages

## Upgrade Notes
For users upgrading from 0.8.0 to 0.9.0 state files, run the following commands:

```
With the older Terraform version (version 0.8.x), run terraform remote pull.
This will cache the latest legacy remote state data locally. 
We'll use this for a backup in case things go wrong.

Backup your terraform/.terraform/terraform.tfstate file. 
This contains the cache we just pulled. 
Please copy this file to a location outside of your Terraform module.

# Note: this will wipe any custom Terraform added (at this time)
$ ./stream_alert_cli.py terraform generate
$ ./stream_alert_cli.py terraform init-backend
# Type 'no' at the prompt since your statefile isn't moving

Verify your state looks good by running stream_alert_cli.py terraform build and 
seeing if it detects your infrastructure.
```

For any other issues with upgrading, see the link above in `Background`


stream_alert_cli/runner.py
```diff
@@ -14,16 +14,20 @@
 limitations under the License.
 '''
 
+import os
+import sys
+import shutil
```
austinbyers: Nit: switch `sys` and `shutil` for alphabetical import order

stream_alert_cli/terraform_generate.py
```diff
@@ -0,0 +1,193 @@
...
+        tf_file.write(main_json)
+
+    # Break out early during the init process, clusters aren't needed yet
+    if init:
+        return True
+
+    # Setup clusters
+    for cluster in config['clusters'].keys():
+        if cluster == 'main':
+            raise InvalidClusterName('Rename cluster "main" to something else!')
```
austinbyers: It doesn't look like `InvalidClusterName` is defined in this file? It's defined in `runner.py`

jacknagz: good catch, #copypasta

jacknagz:
Related to #44

PR #165, opened 2017-06-07
S3 Versioning, Cleanup, and Tagging

jacknagz:
to @ryandeivert 

* Enable S3 versioning
* Cleanup tagging on Lambda/Kinesis resources for easier tracking
* Begin to use Terraform policy data resources over JSON strings (provides more reliability when running Terraform)


jacknagz:
Verified this change via running `python stream_alert_cli.py terraform init` in an AWS account

Issue #166, opened 2017-06-07
optional_top_level_keys - unexpected outcome when coupled with json_path & envelope_keys


mime-frame:
@ryandeivert see internal branch for details and repro

mime-frame:
You know things are working with integrations tests fully pass

PR #167, opened 2017-06-08
[parser] nested json records will now have optional keys applied

ryandeivert:
to @jacknagz , @mime-frame 
resolves: #166
size: small

## changes ##
* Nested json records that get extracted via `json_path` will now have `optional_top_level_keys` applied.
* Removing the need to call `update` on a schema to add the default value, and instead logging an error if the key does not exist in the schema. This informs the user that the `optional_top_level_keys` should also be declared as part of the `schema` in logs.json.

## todo ##
* A future PR is to come with a note added to the docs about the need for `optional_top_level_keys` to be declared in the schema.

### DIFF
--- a/conf/logs.json
+++ b/conf/logs.json
@@ -246,6 +246,8 @@
       "cb_server": "string",
       "computer_name": "string",
       "event_type": "string",
+      "filetype": "integer",
+      "filetype_name": "string",
       "link_process": "string",
       "link_sensor": "string",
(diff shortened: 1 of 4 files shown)


jacknagz:
Change looks good - please add a unit test case which uses optional_top_level_keys as well as json_path to verify

ryandeivert:
@jacknagz  amended with new unit test. PTAL

Issue #109, opened 2017-04-13
RFC: Improvement: rules and how they interplay with clusters


mime-frame:
Make it easy to define what clusters a rule should run in

Desired properties (very rough draft, open for comment):

1. "Write once, run anywhere" - meaning, write the rule once, have the option of it executing within any number of clusters vs. copy-pasta if you happen to break your rules out by cluster
2. Each rule shouldn't have to check the cluster it's operating within if you are already logically grouping rules into cluster specific files. Maybe have the option of defining the supported clusters atop the rules file?
3. Promote through process, or code, the idea of logically separating alert logic from the post-processing false-positive filtering logic. Why: this improves readability, it makes it easier to support (1) and (2), and it will make sharing rules with the public easier (e.g. no company, including ours, will be comfortable sharing the filtering logic since it's very environment specific)

PR #174, opened 2017-06-14
JSON Cluster Templates and S3 Logging

### LINKED ISSUE #44: Improvement: JSON Cluster Templates

### LINKED ISSUE #84: [CLI] Pin Terraform Version
Pin the version of Terraform required by the StreamAlert CLI.

### LINKED ISSUE #91: Issue: Not all regions support firehose
Hey, Oblivious this is an AWS issue. The required kinesis firehouse resource's, are available in three regions: EU (Ireland), US East (N. Virginia) and US West (Oregon)

Advice, tips or workarounds welcome!

Cheers,
Andrew

### LINKED ISSUE #123: Bug: think about implications of allowing lambda functions to write to any s3 bucket
As it stands currently, the rule processor could technically write to any of the S3 outputs that are configured (and not just the one designated for its cluster). We should debate whether is this a major concern and resolve by making lambda functions cluster aware.

### LINKED ISSUE #129: Bug: Cluster aware SNS inputs
## Background

#124 introduced the ability to provide SNS topic ARNs as inputs to the StreamAlert rule processor.  This is not cluster aware, and will apply this to all defined clusters.  

## Desired Outcome

Express the input configuration in a way which only applies input SNS topics to a specific cluster.  This requires changes to `inputs.json`, Terraform, and the `rule_processor`.

jacknagz:
to @chunyong-lin @ryandeivert 
resolves #129, resolves #123, preliminary to #109, resolves #91, resolves #84, resolves #44

## Background
StreamAlert's infrastructure is a collection of many Terraform modules.  Each  module represents a piece of the setup, each with a different task (create Lambda functions, Kinesis Streams, Flow Logs, etc).  Currently, a bare minimum setup is configurable out of the box, which limits users without Terraform knowledge from utilizing the newer modules added (flow logs, cloudtrail, etc).  This change allows clusters to be expressed fully as JSON in order to take advantage of these settings.

## Changes
* Split `variables.json` out into two primary files: `lambda.json` and `global.json`.  The former contains lambda settings fed into each function, while global serves for higher level configuration settings.
* Create a default `prod` prototype cluster, here is an [example](https://gist.github.com/jacknagz/505325753fe9c3a2faf4c517bcd938a4) with all possible options (this will be explained in the docs)
* Add S3 access logging bucket, and enable all buckets created by StreamAlert to write to it.  This bucket also comes with a 30 day Glacier policy to keep costs low.  This is a security best practice for auditing.
* Lambda inputs and outputs are now cluster aware.  This means that adding AWS outputs will be a two step process (for now - more details in action items)
* Terraform Variables: Modified to pass in simpler values (strings or ints) instead of maps/lists.  The TF cluster files are now generated via JSON configurations, so we can express with much more granularity our configurations.  Each time you run any relevant CLI command, TF files are automatically regenerated. 
* Update sample rules and schemas

## Testing
Using a test account, I created two clusters (one with all possible modules enabled) and verified alerts were indeed sending.

## Action Items
- [ ] Update documentation
- [ ] Support custom CloudTrail patterns (most likely via a second PR)
- [ ] Enable the CLI to create AWS outputs that are cluster aware, and automatically place into cluster files
- [ ] Investigate whether or not `variables.tf` is still necessary (I don't think it is)


terraform/modules/tf_stream_alert/variables.tf
```diff
@@ -23,44 +23,36 @@ variable "rule_processor_config" {
...
-  default = {}
-}
+variable "alert_processor_memory" {}
+
+variable "alert_processor_timeout" {}
 
 variable "output_lambda_functions" {
-  type    = "map"
-  default = {}
+  type    = "list"
```
chunyong-lin: It is not necessary to specify `type` here, based on [terraform document](https://www.terraform.io/docs/configuration/variables.html) `If this field is omitted, the variable type will be inferred based on the default. If no default is provided, the type is assumed to be string.`

terraform/modules/tf_stream_alert_flow_logs/variables.tf
```diff
@@ -1,19 +1,29 @@
-variable "destination_stream_arn" {}
+variable "destination_stream_arn" {
+  type = "string"
```
jacknagz: @chunyong-lin I have this in almost every module for clarity purposes, i'm going to leave it

terraform/modules/tf_stream_alert/iam.tf
```diff
@@ -150,7 +150,7 @@ resource "aws_iam_role_policy" "streamalert_alert_processor_lambda" {
         "lambda:InvokeFunction"
       ],
       "Effect": "Allow",
-      "Resource": "arn:aws:lambda:${var.region}:${var.account_id}:function:${lookup(var.output_lambda_functions, element(keys(var.output_lambda_functions), count.index))}"
+      "Resource": "arn:aws:lambda:${var.region}:${var.account_id}:function:${element(var.output_lambda_functions, count.index)}"
```
jacknagz: I left it because of the count interpolation

chunyong-lin:
Thanks for the clarification, terraform code LGTM.

stream_alert_cli/terraform_generate.py
```diff
@@ -96,99 +127,160 @@ def generate_main(**kwargs):
...
     main_json = json.dumps(
         generate_main(init=init, config=config),
-        indent=4,
+        indent=2,
         sort_keys=True
     )
     with open('terraform/main.tf', 'w') as tf_file:
         tf_file.write(main_json)
-
+
```
ryandeivert: what's this rogue tab about???

stream_alert_cli/config.py
```diff
@@ -81,105 +39,67 @@ def __getitem__(self, key):
...
-    def _detect_version(self):
-        """Detect the config version
-
-        Returns:
-            [int] detected config version
-        """
-        config_keys = set(self.config.keys())
+        def _config_loader(key, filepath, cluster_file):
+            if not os.path.isfile(filepath):
+                LOGGER_CLI.error('%s not found!', filepath)
```
ryandeivert: Make this error more descriptive - ie, mention the operation (config loading). also quit using so many exclamation points. this is not exciting.

stream_alert_cli/config.py
```diff
@@ -81,105 +39,67 @@ def __getitem__(self, key):
...
-                config_keys
-            ))
+            with open(filepath) as data:
+                try:
+                    if cluster_file:
+                        config['clusters'][key] = json.load(data)
+                    else:
+                        config[key] = json.load(data)
+                except ValueError:
+                    LOGGER_CLI.error('%s is not valid JSON!', filepath)
```
ryandeivert: again, more expressive error message with less excitement.

stream_alert_cli/config.py
```diff
@@ -81,105 +39,67 @@ def __getitem__(self, key):
...
-            raise ConfigError('StreamAlert variables.json file not found!')
 
-        with open(self.filename) as data:
-            try:
-                config = json.load(data)
-            except ValueError:
-                raise ConfigError('StreamAlert variables.json file is not valid JSON!')
-            return config
+        config = {}
+        config['clusters'] = {}
```
ryandeivert: `config = {'clusters': {}}`

stream_alert_cli/config.py
jacknagz: @ryandeivert it is exciting, put a smile on your face sometime :trollface:

jacknagz:
Doc updates are coming

PR #177, opened 2017-06-14
[streamalert] Remove samples, update schemas and add first CloudTrail rule

mime-frame:
to: @jacknagz 

**What**
- Add new and updated log schemas
- Remove sample logs and rules
- Add first CloudTrail rule
- Remove tabs from `conf/sources.json`

**Why**
- Address bugs in the old CarbonBlack schema
- Allow folks to write CloudTrail rules
- In it's current form, samples make it harder for new users to deploy streamalert
- Pave the way for committing more production rules (value out of the box)

**Testing**

`./test/scripts/rule_test.sh`


conf/logs.json
```diff
@@ -182,100 +463,200 @@
...
+        "unknown4": "string",
+        "actionflags": "string",
+        "srcloc": "string",
+        "dstloc": "string",
+        "pkts_sent": "string",
+        "pkts_received": "string"
       }
-    }
+    },
+    "hints": {
```
jacknagz: This should be `"configuration": {"log_patterns": {"type": ["TRAFFIC"]}}`

conf/sources.json
```diff
@@ -1,38 +1,20 @@
...
-				"json_log",
-				"syslog_log",
-				"kv_log",
-				"csv_log",
-				"osquery",
-				"cloudtrail"
-			]
-		}
-	},
-	"sns": {
```
jacknagz: why remove the example SNS topic?

rules/cloudtrail/cloudtrail_critical_api.py
```diff
@@ -0,0 +1,41 @@
+from stream_alert.rule_processor.rules_engine import StreamRules
+from rules.helpers.base import in_set
+
+rule = StreamRules.rule
+disable = StreamRules.disable()
+
+
+@rule(logs=['cloudtrail:api_events'],
+      matchers=[],
+      outputs=['aws-s3:sample.bucket'])
```
jacknagz: it should be `aws-s3:sample_bucket`

stream_alert/rule_processor/main.py
```diff
@@ -18,9 +18,6 @@
 import os
 
 from stream_alert.rule_processor.handler import StreamAlert
-from rules import (
```
jacknagz: if we are getting rid of this, we need an action item to standardize loading matchers from within the loop below

conf/logs.json
mime-frame: I guess internal is broken as well, noted

conf/sources.json
mime-frame: goal is to remove all samples

rules/cloudtrail/cloudtrail_critical_api.py
mime-frame: i guess existing rules were broken too?

ryandeivert: I have a bunch of these I'm fixing now in other rules - please adhere to what jack is suggesting with `aws-s3:sample_bucket`

conf/sources.json
jacknagz: then why retain the s3 input?

conf/logs.json
jacknagz: it's unused

mime-frame: but if used, broken, noted

conf/sources.json
mime-frame: its required for rule integration testing?

stream_alert/rule_processor/main.py
mime-frame: i'll just re-add it back, better suited for another PR

mime-frame:
@jacknagz suggestions addressed except for your s3 comment, see my reply. Thanks! :)

PR #179, opened 2017-06-15
[tests] alert processor testing

ryandeivert:
to @jacknagz 

## changes  ##
* Adding option to `cli` to run rule test on a comma separated list of rules.
  * New options is used like so: `python stream_alert_cli.py lambda test --processor rule --rule cloudtrail_put_bucket_acl,cloudtrail_put_object_acl`
* Updating the alert processor so it can yield back status results to the calling handler. Would like to make this more granular in the future.
* Removing an unused file (`context`)
* Migrating some helper functions from the unit testing framework to the cli package so they can be utilized across tests.
* The cli now supports running tests for the `alert` processor via: `python stream_alert_cli.py lambda test --processor alert`
  * End-to-end testing is also available via: `python stream_alert_cli.py lambda test --processor all`
  * Either of the above commands can be used in conjunction with the new `--rule/-r` flag to specify rules to test against.

Testing output will now look similar to:
```
$ python stream_alert_cli.py lambda test --processor all --rule cloudtrail_critical_api

cloudtrail_critical_api
	[Pass]   [trigger=1]	rule	(kinesis): CloudTrail - Critical API - DeleteSubnet
	[Pass]              	alert	(aws-s3): sending alert to 'sample_bucket'
	[Pass]   [trigger=1]	rule	(kinesis): CloudTrail - Critical API - DeleteVpc
	[Pass]              	alert	(aws-s3): sending alert to 'sample_bucket'
	[Pass]   [trigger=1]	rule	(kinesis): CloudTrail - Critical API - UpdateTrail
	[Pass]              	alert	(aws-s3): sending alert to 'sample_bucket'
	[Pass]   [trigger=1]	rule	(kinesis): CloudTrail - Critical API - StopLogging
	[Pass]              	alert	(aws-s3): sending alert to 'sample_bucket'
	[Pass]   [trigger=1]	rule	(kinesis): CloudTrail - Critical API - DeleteDBCluster
	[Pass]              	alert	(aws-s3): sending alert to 'sample_bucket'
	[Pass]   [trigger=1]	rule	(kinesis): CloudTrail - Critical API - StopConfigurationRecorder
	[Pass]              	alert	(aws-s3): sending alert to 'sample_bucket'
	[Pass]   [trigger=1]	rule	(kinesis): CloudTrail - Critical API - DeleteFlowLogs
	[Pass]              	alert	(aws-s3): sending alert to 'sample_bucket'
	[Pass]   [trigger=0]	rule	(kinesis): CloudTrail - Critical API - False Positive Case



(8/8)	Rule Tests Passed
(7/7)	Alert Tests Passed
```

Incoming PR for doc updates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ before_install:
  - sudo apt-get install -qq python-numpy python-scipy
 install:
   - pip install -r requirements.txt
-script: 
+script:
   - ./test/scripts/unit_tests.sh
-  - ./test/scripts/rule_test.sh
+  - ./stream_alert_cli.py lambda test --processor rule
+  - ./stream_alert_cli.py lambda test --processor alert
+  - ./stream_alert_cli.py lambda test --processor all
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,4 @@ install:
   - pip install -r requirements.txt
 script:
   - ./test/scripts/unit_tests.sh
-  - ./stream_alert_cli.py lambda test --processor rule
-  - ./stream_alert_cli.py lambda test --processor alert
   - ./stream_alert_cli.py lambda test --processor all
--- a/stream_alert_cli.py
+++ b/stream_alert_cli.py
@@ -80,7 +80,7 @@ def build_parser():
         choices=['new'],
         help=('new: create a new output to send alerts to\n')
     )
-    #
+    # output service options
     output_parser.add_argument(
         '--service',
         choices=['aws-lambda', 'aws-s3', 'pagerduty', 'phantom', 'slack'],
@@ -115,6 +115,13 @@ def build_parser():
         '--debug',
         action='store_true',
         help='Enable DEBUG logger output'
+    )
+
+    lambda_parser.add_argument(
+        '-r', '--rules',
+        type=str,
+        default='',
+        help='Comma separated names of rules to test'
     )
 
     # terraform parser and defaults
--- a/stream_alert_cli.py
+++ b/stream_alert_cli.py
@@ -119,9 +119,8 @@ def build_parser():
 
     lambda_parser.add_argument(
         '-r', '--rules',
-        type=str,
-        default='',
-        help='Comma separated names of rules to test'
+        nargs='+',
+        help='Names of rules to test, separated by spaces'
     )
 
     # terraform parser and defaults
--- a/stream_alert_cli/test.py
+++ b/stream_alert_cli/test.py
@@ -69,7 +69,16 @@ class RuleProcessorTester(object):
     """Class to encapsulate testing the rule processor"""
 
     def __init__(self, print_output):
-        super(self.__class__, self).__init__()
+        """RuleProcessorTester initializer
+
+        Args:
+            print_output [bool]: Boolean indicating whether this processor test
+                should print results to stdout. This is set to false when the
+                alert processor is explicitly being testing alone, and set to
+                true for rule processor tests and end-to-end tests.
+                Warnings and errors captrued during rule processor testing
+                will still be written to stdout regardless of this setting.
+        """
         # Create the topic used for the mocking of alert sending
         # This is used in stream_alert/rule_processor/sink.py to 'send' alerts
         sns_client = boto3.client('sns', region_name='us-east-1')
@@ -83,6 +92,10 @@ class RuleProcessorTester(object):
     def test_processor(self, rules):
         """Perform integration tests for the 'rule' Lambda function
 
+        Args:
+            rules [list or None]: Specific rule names (or None) to restrict
+                testing to. This is passed in from the CLI using the --rules option.
+
         Returns:
             [generator] yields a tuple containig a boolean of test status and
                 a list of alerts to run through the alert processor
@@ -93,7 +106,7 @@ class RuleProcessorTester(object):
             with open(os.path.join(DIR_RULES, rule_file), 'r') as rule_file_handle:
                 try:
                     contents = json.load(rule_file_handle)
-                except Exception as err:
+                except (ValueError, TypeError) as err:
                     all_tests_passed = False
                     message = 'Improperly formatted file - {}: {}'.format(
                         type(err).__name__, err)
@@ -135,10 +148,10 @@ class RuleProcessorTester(object):
                 if self.print_output:
                     report_output([
                         current_test_passed,
+                        '[trigger={}]'.format(expected_alerts),
                         'rule',
                         test_record['service'],
-                        '{} [trigger={}]'.format(test_record['description'],
-                                                 expected_alerts)])
+                        test_record['description']])
 
                 all_tests_passed = current_test_passed and all_tests_passed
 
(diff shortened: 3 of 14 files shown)


stream_alert_cli/test.py
```diff
@@ -48,338 +56,486 @@ class TestingSuppressFilter(logging.Filter):
...
-    # Create the topic used for the mocking of alert sending
-    boto3.client('sns', region_name='us-east-1').create_topic(Name='test_streamalerts')
+class RuleProcessorTester(object):
+    """Class to encapsulate testing the rule processor"""
+
+    def __init__(self, print_output):
+        super(self.__class__, self).__init__()
+        # Create the topic used for the mocking of alert sending
+        # This is used in stream_alert/rule_processor/sink.py to 'send' alerts
+        sns_client = boto3.client('sns', region_name='us-east-1')
```
jacknagz: I think I'm lacking the context where the mocking setup to prevent this from running on real infra

stream_alert_cli/test.py
```diff
@@ -48,338 +56,486 @@ class TestingSuppressFilter(logging.Filter):
...
+        sns_client.create_topic(Name='test_streamalerts')
+        # Create a list for pass/fails. The first value in the list is a
+        # list of tuples for failures, and the second is list of tuples for
+        # passes. Tuple is (rule_name, rule_description)
+        self.rules_fail_pass_warn = ([], [], [])
+        self.print_output = print_output
+
+    def test_processor(self, rules):
+        """Perform integration tests for the 'rule' Lambda function
```
jacknagz: add an Args block for `rules`

stream_alert_cli/test.py
```diff
@@ -48,338 +56,486 @@ class TestingSuppressFilter(logging.Filter):
...
 
-    # Start mocked sns
-    BOTO_MOCKER_SNS.start()
-
-    # Create the topic used for the mocking of alert sending
-    boto3.client('sns', region_name='us-east-1').create_topic(Name='test_streamalerts')
+class RuleProcessorTester(object):
+    """Class to encapsulate testing the rule processor"""
+
+    def __init__(self, print_output):
```
jacknagz: Pls add a docstring with Args description

stream_alert_cli/test.py
```diff
@@ -48,338 +56,486 @@ class TestingSuppressFilter(logging.Filter):
...
+            [generator] yields a tuple containig a boolean of test status and
+                a list of alerts to run through the alert processor
+        """
+        all_tests_passed = True
+
+        for rule_file, rule_name in get_rule_test_files(rules):
+            with open(os.path.join(DIR_RULES, rule_file), 'r') as rule_file_handle:
+                try:
+                    contents = json.load(rule_file_handle)
+                except Exception as err:
```
jacknagz: Which exceptions are we trying to catch here? It's generally bad practice to use `Exception`.  For JSON unmarshal, we should use `ValueError`

stream_alert_cli/test.py
```diff
@@ -48,338 +56,486 @@ class TestingSuppressFilter(logging.Filter):
...
+                    if alerts or self.print_output:
+                        print '\n{}'.format(rule_name)
+                        print_header = not print_header
+
+                if self.print_output:
+                    report_output([
+                        current_test_passed,
+                        'rule',
+                        test_record['service'],
+                        '{} [trigger={}]'.format(test_record['description'],
```
jacknagz: can we add this before service instead? the columns would line up nicer

stream_alert_cli/test.py
```diff
@@ -48,338 +56,486 @@ class TestingSuppressFilter(logging.Filter):
...
+                    (rule_name, 'Rule failure: {}'.format(test_record['description'])))
+
+        # Report on the final test results
+        if self.print_output:
+            self.report_output_summary()
+
+    def check_keys(self, rule_name, test_record):
+        """Check the test_record contains the required keys
+
+        Args:
```
jacknagz: missing an arg

stream_alert_cli/test.py
```diff
@@ -48,338 +56,493 @@ class TestingSuppressFilter(logging.Filter):
...
+                source)
         else:
-            kinesis_data = base64.b64encode(data)
+            LOGGER_CLI.info('Invalid service %s', service)
+
+        return template
+
+
+class AlertProcessorTester(object):
+    """Class to encapsulate testing the alert processor"""
```
jacknagz: So I'm understanding this right, the alerts used as input come from the rule processor, right? can we make that clear in this classes docstring?

stream_alert_cli.py
```diff
@@ -117,6 +117,13 @@ def build_parser():
         help='Enable DEBUG logger output'
     )
 
+    lambda_parser.add_argument(
+        '-r', '--rules',
+        type=str,
```
ryandeivert: I knew there had to be an easier way to do that. using `nargs='+'` since at least one should always be provided if this switch is used.

stream_alert_cli/test.py
ryandeivert: The beauty of the stacked decorators that wrap `def stream_alert_test(options):...`:

```
@mock_lambda
@mock_sns
@mock_s3
@mock_kms
```
These apply to _EVERY_ call that happens below their scope. This class is instantiated from within this function, so the `@mock_sns` automatically applies here.

stream_alert_cli/test.py
ryandeivert: That's not true. This can result in a number of errors, including `TypeError`. `ValueError` will not catch all possible errors here.

stream_alert_cli/test.py
ryandeivert: What item are you referring to? The `[trigger=?]` count? It isn't really added as a column (and I don't want it to be) since it would conflict with alert processor output column. Can you maybe paste an example?

stream_alert_cli/test.py
ryandeivert: Sure good catch.

stream_alert_cli/test.py
jacknagz: From the docs:

```If the data being deserialized is not a valid JSON document, a ValueError will be raised.```

Our error description does not match the type of Exceptions we are trying to catch.  Either make it more general and add `TypeError` and `OverflowError`, or just catch `ValueError` and call it a day

File: stream_alert_cli/test.py (hunk 1 above)
Comment (jacknagz) [inline stream_alert_cli/test.py]: I see, can we add a comment? this is a pretty important fact

File: stream_alert_cli/test.py (hunk 4 above)
Comment (ryandeivert) [inline stream_alert_cli/test.py]: Search that same page for `TypeError` ;). I'm changing it to catch both types of exception.

File: stream_alert_cli/test.py (hunk 1 above)
Comment (ryandeivert) [inline stream_alert_cli/test.py]: sure thing

**Comment by ryandeivert** [comment]
@jacknagz added updates PTAL

File: .travis.yml [hunk 9]
```diff
@@ -7,6 +7,8 @@ before_install:
  - sudo apt-get install -qq python-numpy python-scipy
 install:
   - pip install -r requirements.txt
-script: 
+script:
   - ./test/scripts/unit_tests.sh
-  - ./test/scripts/rule_test.sh
+  - ./stream_alert_cli.py lambda test --processor rule
+  - ./stream_alert_cli.py lambda test --processor alert
+  - ./stream_alert_cli.py lambda test --processor all
```
jacknagz: why do individually and all of them at the same time?

ryandeivert: @jacknagz amended to just have `all` option run

PR #181, opened 2017-06-16
[cli][test] live alert testing

ryandeivert:
to @jacknagz 

# we'll do it live #
* Adding a `-l/--live` option to the cli to allow for live alert dispatching to a specified cluster.
  * ie - `$ python stream_alert_cli.py live-test --cluster prod --rule cloudtrail_put_object_acl`
  * ~The "cluster" specified must exist in the `conf/clusters/` directory and should match the name exactly.~
  * The `--cluster/-c` argument is required and must be followed by a "cluster" name.
  * Available "clusters" to use for the `live-test` command are extracted from the `conf/clusters` directory (using the `<cluster>.json` files that exist there) and presented as `choices`.
* The cli package will now dynamically control if calls should be mocked out via a nested decorator, depending on if the `live` option is used.
* If the `--live` option is omitted, the mocking of all calls (boto3=moto and urllib2=patch) will happen.
* If the `--live` option is included, with a cluster, the alert processor with attempt to actually send alerts to the designated outputs for the triggered rule.
* The `CLIConfig` is passed to the cli package to extract values and construct a real context object containing an arn/function_name/etc.

# other updates #
* Changing rule processor to only try to "sink" (aka send to alert processor) only if the alerts are not being returned to the caller.
  * This was the original logic of the rule processor, but I had changed it previously. Now, this allows alerts to be returned to the caller for testing but sent to alert processor when not testing.
* Using `logging.exception` when trying to publish to an sns topic instead of raising the error. This avoids breaking execution.
* Example rules are now configured to send to all example outputs.

### DIFF
--- a/stream_alert_cli.py
+++ b/stream_alert_cli.py
@@ -24,6 +24,7 @@ To run terraform by hand, change to the terraform directory and run:
 
 terraform <cmd> -var-file=../terraform.tfvars -var-file=../variables.json
 '''
+import os
 
 from argparse import ArgumentParser, RawTextHelpFormatter
 
@@ -44,11 +45,12 @@ def build_parser():
     stream_alert_cli.py lambda deploy --processor 'all'
 
     Rolling Back:
-    stream_alert_cli.py lambda rollback --func 'rule'
+    stream_alert_cli.py lambda rollback --processor 'rule'
 
     Running Integration Tests:
     stream_alert_cli.py lambda test --processor 'rule'
     stream_alert_cli.py lambda test --processor 'alert'
+    stream_alert_cli.py lambda test --processor 'all'
 
     Building Infrastructure:
     stream_alert_cli.py terraform init
@@ -70,9 +72,9 @@ def build_parser():
         'output',
         help='Define a new output to send alerts to'
     )
-    output_parser.set_defaults(
-        command='output'
-    )
+
+    # set the name of this parser to 'output'
+    output_parser.set_defaults(command='output')
 
     # output parser arguments. the cli will handle the logic to set these up
     output_parser.add_argument(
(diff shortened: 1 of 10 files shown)


stream_alert_cli.py
```diff
@@ -123,6 +123,11 @@ def build_parser():
         help='Names of rules to test, separated by spaces'
     )
 
+    lambda_parser.add_argument(
+        '-l', '--live',
```
jacknagz: DOES THIS STORE A BOOLEAN? MAYBE YOU SHOULD USE STORE_TRUE

(My caps was on what I started typing and I thought it'd be funny to just keep going)

ryandeivert: lol troll

ryandeivert: NO IT'S STORES A STRING VALUE OF THE CLUSTER. CHECK THE HELP STRING.

ryandeivert: although I'm open to suggestions if you think it's confusing

jacknagz: Yeah let's make it two arguments, or a separate command altogether (`live-test` maybe?)

jacknagz: --live/--cluster or `live-test` with positional args

stream_alert_cli/test.py
```diff
@@ -523,47 +533,127 @@ def get_rule_test_files(filter_rules):
...
-@mock_lambda
-@mock_sns
-@mock_s3
-@mock_kms
-def stream_alert_test(options):
-    """Integration testing entry point. This function is wrapped with multiple
-    moto mock decorators to override any boto3 calls. Wrapping this function
-    here allows us to mock out all calls that happen below this scope.
+def mock_me(context):
+    """Decorator function for wrapping framework in mock calls
```
jacknagz: fAncY

stream_alert_cli/test.py
```diff
@@ -523,47 +533,127 @@ def get_rule_test_files(filter_rules):
...
+            an aws context object
+    """
+    if options.live and options.live not in config.clusters():
+        LOGGER_CLI.error('Specified cluster \'%s\' does not exist '
+                         'in config', options.live)
+        return
+
+    context = get_context_from_config(config, options.live)
+
+    @mock_me(context)
```
jacknagz: this is really cool, good work

stream_alert_cli.py
ryandeivert: I don't know how to best do this but still support the `--rules` feature, etc. Having multiple commands seems like overcomplicating this to me, especially since one command is just a boolean switch. That same logic is accomplished through the presence of a cluster value.

ryandeivert: We should discuss how to make this the best user experience.

ryandeivert:
@jacknagz PTAL at updates and see updated PR description (the strikethrough and following 2 bullets).

ryandeivert:
@jacknagz can I get 👀  on this?

jacknagz:
@ryandeivert you have a merge conflict

stream_alert_cli/test.py
```diff
@@ -38,7 +38,7 @@
 )
 
 from stream_alert_cli.outputs import load_outputs_config
-from stream_alert_cli.logger import LOGGER_CLI, LOGGER_SA
+from stream_alert_cli.logger import LOGGER_CLI, LOGGER_SA, LOGGER_SO
```
jacknagz: what is `LOGGER_SO` used for?

ryandeivert:
@jacknagz LOGGER_SO is used for the alert processor - just like you use LOGGER_SA for the rule processor and LOGGER_CLI for the CLI. It just helps differentiate between the two functions, since they'd log separately in production anyway.

mime-frame:
image

ryandeivert:
@mime-frame @jacknagz for clarity, I updated other aspects of the PR description to clear up the usage of the `live-test` sub command.

PR #183, opened 2017-06-16
[lambda][alert] fixing bug that caused logging to break with pagerduty

ryandeivert:
to @jacknagz 
size: small

## fixes ##
* Bug that caused logging to break with high throughput to PagerDuty
* The error occurs when attempting to send a large amount of consecutive alerts to PD.
* How the issue/fix was discovered
  * PagerDuty would begin to return a 403 Forbidden Request error
  * The HTTP response fp for the exception would return a weird 121 byte blob of binary data that could not be printed well.
  * AWS's logging choked on printing this, throwing the following error: `'utf8' codec can't decode byte 0x8b in position 93: invalid start byte: UnicodeDecodeError`
* The change that fixes this is not reading/printing the error response data and instead just printing the url and error code being returned. It is still unclear why the response returned by PagerDuty is _seemingly_ garbage.

## other changes ##
* Switching another place to use `logging.exception` so we have more insight into where errors occur.
* If an improperly formatted `output` is provided in a rule, we will now log the actual value of the output.

### DIFF
--- a/stream_alert/alert_processor/main.py
+++ b/stream_alert/alert_processor/main.py
@@ -117,8 +117,9 @@ def run(loaded_sns_message, region, function_name, config):
         try:
             service, descriptor = output.split(':')
         except ValueError:
-            LOGGER.error('Outputs for rules must be declared with both a service and a '
-                         'descriptor for the integration (ie: \'slack:my_channel\')')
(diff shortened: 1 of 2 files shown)


mime-frame:
Thanks for chasing this down @ryandeivert !

PR #191, opened 2017-06-21
Fix Travis

jacknagz:
to @ryandeivert 
size small

The new [Ubuntu image](https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch) broke our Travis build, this PR fixes it.

## Changes
* Add `sudo: required` to be more explicit 
* Ran `pip install coveralls` `pip install google-compute-engine` locally, followed by `pip freeze > requirements.txt` to update our dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 dist: trusty
 cache: pip
+sudo: required
 python:
   - 2.7
 before_install:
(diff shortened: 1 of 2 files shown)


austinbyers:
It's really unfortunate that we have to require `google-compute-engine` for the entire project. I wonder if there's a better way to separate test dependencies

jacknagz:
@austinbyers it's not a huge deal, it doesn't run in the Lambda functions, just locally and in CI

PR #193, opened 2017-06-22
[tests] testing fixes

ryandeivert:
to @austinbyers 
size: small

## changes ##
* Updating the boto3 lambda invocation to utilize a qualifier if one is provided in the output configuration.
* Updating lambda output/alert tests to go with above change.
* Patching a function to load the unit test outputs configuration during tests to avoid breaking tests with production configuration changes.

### DIFF
--- a/stream_alert/alert_processor/outputs.py
+++ b/stream_alert/alert_processor/outputs.py
@@ -581,12 +581,30 @@ class LambdaOutput(AWSOutput):
         alert = kwargs['alert']
         alert_string = json.dumps(alert['record'])
         function_name = self.config[self.__service__][kwargs['descriptor']]
+        parts = function_name.split(':')
+
+        # Check to see if there is an optional qualifier included here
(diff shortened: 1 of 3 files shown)


stream_alert/alert_processor/outputs.py
```diff
@@ -581,12 +581,30 @@ def dispatch(self, **kwargs):
         alert = kwargs['alert']
         alert_string = json.dumps(alert['record'])
         function_name = self.config[self.__service__][kwargs['descriptor']]
+        parts = function_name.split(':')
+
+        # Check to see if there is an optional qualifier included here
+        if len(parts) == 2 or len(parts) == 8:
```
austinbyers: Why are the two expected lengths 2 and 8? Can you provide examples of these two cases in the comments?

ryandeivert: @austinbyers added comment PTAL

PR #196, opened 2017-06-23
[lambda][alert] fixing bug in alert processor that cause lambda invocations to fail

ryandeivert:
to @austinbyers 

## bug fix ##
* Resolving an issue where lambda invocations of the alert processor function would fail because the return value of the main handler function is a `generator` object.

## context ##
* It appears that AWS lambda expects a return value from the functions main handler that is json serializable. Using a `generator` to yield results to the caller function breaks serialization, and thus breaks invoking of the lambda function.
* This change will create a list that is instead extended with the results of the `run` function, which is then returned to the caller. `list` objects are, in fact, serializable into json 😂

### DIFF
--- a/stream_alert/alert_processor/main.py
+++ b/stream_alert/alert_processor/main.py
@@ -35,7 +35,10 @@ def handler(event, context):
         context [AWSLambdaContext]: basically a namedtuple of properties from AWS
 
     Returns:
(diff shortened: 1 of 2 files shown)


ryandeivert:
Added test just now!

Issue #176, opened 2017-06-14
Follow Up: JSON Cluster Files


jacknagz:
Follow up to #174 

- [x] Documentation on setup, usage, options
- [x] Support S3 events module
- [x] Support custom Terraform as input
- [x] CloudTrail patterns as inputs
- [x] Investigate if we still need `variables.tf`

jacknagz:
> Support custom Terraform as input

A special filename can be reserved in the `terraform` directory for this purpose (like `misc.tf`, or `custom.tf`).   Documentation should also reflect this

jacknagz:
>  Investigate if we still need variables.tf

We still need it

PR #205, opened 2017-06-27
Bug Fixes (Initialization, AWS Perms, Config), CloudWatch Event Patterns, and More 

### LINKED ISSUE #201: KMS permissions not included in documentation
Trying to run `python stream_alert_cli.py terraform init` and failing due to KMS permissions.

I followed the instructions at https://streamalert.readthedocs.io/account.html for setting up my AWS IAM user.

Initially, I received the error:

```
* aws_kms_key.stream_alert_secrets: AccessDeniedException: User: arn:aws:iam::CENSORED:user/streamalert is not authorized to perform: kms:CreateKey
	status code: 400, request id: CENSORED
```

After adding the `AWSKeyManagementServicePowerUser` permission to the IAM user (not listed as a requirement in the docs), and running again, I received a different KMS-related permissions error:

```
* aws_kms_key.stream_alert_secrets: Failed to set key rotation for "CENSORED" to true: "AccessDeniedException: User: arn:aws:iam::CENSORED:user/streamalert is not authorized to perform: kms:EnableKeyRotation on resource: arn:aws:kms:us-east-1:CENSORED:key/CENSORED\n\tstatus code: 400, request id: CENSORED"
```

jacknagz:
to: @ryandeivert (code) @mime-frame (docs)
cc: @airbnb/streamalert-maintainers @zbuc
size: medium

A combination of various bug fixes and feature additions.

### Related Issues
* resolves #201 
* completes another task in #176

### Conf Changes
* Make sure we are starting with `"$LATEST"` as the AWS Lambda `current_version`.  This currently throws an error during the first deploy since the versions do not exist yet.
* Reset the Kinesis and Lambda settings to be a bare minimum.
* Enable `force_destroy` for prerequisite buckets.  This is makes destroying clusters easier, since this data is not of high value.

### CLI Changes
* Add a `stream_alert_cli.py terraform clean` command to simply remove any Terraform statefiles and other related files.  This is useful to have while troubleshooting issues with initialization.
* Fix a bug in the `init` target list: This list of Terraform resources is used during `init` to create the prerequisite infrastructure.  There was a resource naming issue here causing errors.
* Add a default to the `flow_logs` module for the log group name.  Adds some unit tests too.

### Terraform Changes
* Support all `tf_stream_alert_cloudtrail` module options.  This gives control over how users can configure CloudTrail to send to their StreamAlert cluster.  For more information, see the [module documentation.](https://github.com/airbnb/streamalert/tree/master/terraform/modules/tf_stream_alert_cloudtrail)

### Documentation Changes
* Update the Account setup page which explains how to configure your admin user
* Update the clusters section on how to configure `flow_logs` and `cloudtrail` modules

### DIFF
--- a/stream_alert_cli/runner.py
+++ b/stream_alert_cli/runner.py
@@ -137,10 +137,9 @@ def terraform_handler(options):
         LOGGER_CLI.info('Building Initial Infrastructure')
         init_targets = [
             'aws_s3_bucket.lambda_source',
-            'aws_s3_bucket.integration_testing',
-            'aws_s3_bucket.terraform_state',
-            'aws_s3_bucket.stream_alert_secrets',
             'aws_s3_bucket.logging_bucket',
+            'aws_s3_bucket.stream_alert_secrets',
+            'aws_s3_bucket.terraform_remote_state',
             'aws_kms_key.stream_alert_secrets',
             'aws_kms_alias.stream_alert_secrets'
         ]
--- a/stream_alert_cli/runner.py
+++ b/stream_alert_cli/runner.py
@@ -162,6 +162,9 @@ def terraform_handler(options):
         LOGGER_CLI.info('Building Remainder Infrastructure')
         tf_runner()
 
+    elif options.subcommand == 'clean':
+        terraform_clean()
+
     elif options.subcommand == 'destroy':
         if options.target:
             target = options.target
@@ -181,24 +184,32 @@ def terraform_handler(options):
             sys.exit(1)
 
         # Remove old Terraform files
-        LOGGER_CLI.info('Removing old Terraform files')
-        cleanup_files = ['{}.tf'.format(cluster) for cluster in CONFIG.clusters()]
-        cleanup_files.extend([
-            'main.tf',
-            'terraform.tfstate',
-            'terraform.tfstate.backup'
-        ])
-        for tf_file in cleanup_files:
-            file_to_remove = 'terraform/{}'.format(tf_file)
-            if not os.path.isfile(file_to_remove):
-                continue
-            os.remove(file_to_remove)
-        # Finally, delete the Terraform directory
-        shutil.rmtree('terraform/.terraform/')
+        terraform_clean()
 
     # get a quick status on our declared infrastructure
     elif options.subcommand == 'status':
         status()
+
+
+def terraform_clean():
+    """Remove leftover Terraform statefiles and main/cluster files"""
+    LOGGER_CLI.info('Cleaning Terraform files')
+
+    cleanup_files = ['{}.tf'.format(cluster) for cluster in CONFIG.clusters()]
+    cleanup_files.extend([
+        'main.tf',
+        'terraform.tfstate',
+        'terraform.tfstate.backup'
+    ])
+    for tf_file in cleanup_files:
+        file_to_remove = 'terraform/{}'.format(tf_file)
+        if not os.path.isfile(file_to_remove):
+            continue
+        os.remove(file_to_remove)
+
+    # Finally, delete the Terraform directory
+    if os.path.isdir('terraform/.terraform/'):
+        shutil.rmtree('terraform/.terraform/')
 
 
 def run_command(args=None, **kwargs):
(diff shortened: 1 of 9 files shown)


stream_alert_cli/terraform_generate.py
```diff
@@ -352,6 +352,32 @@ def generate_cloudtrail(cluster_name, cluster_dict, config):
     """
     modules = config['clusters'][cluster_name]['modules']
     cloudtrail_enabled = bool(modules['cloudtrail']['enabled'])
+    existing_trail_default = False
+    existing_trail = modules['cloudtrail'].get('existing_trail', existing_trail_default)
+    is_global_trail_default = True
```
ryandeivert: Maybe I'm just not seeing it elsewhere, but it looks like `existing_trail_default` and `is_global_trail_default` are just booleans and only used to hold default values for the `.get()` calls. Can we just place the boolean value directly in the get for brevity?

stream_alert_cli/terraform_generate.py
```diff
@@ -352,6 +352,32 @@ def generate_cloudtrail(cluster_name, cluster_dict, config):
...
+        'source',
+        'account',
+        'time',
+        'region',
+        'resources',
+        'detail'
+    }
+    if not set(event_pattern.keys()).issubset(valid_event_pattern_keys):
+        LOGGER_CLI.error('Invalid CloudWatch Event Pattern!')
+        sys.exit(1)
```
ryandeivert: We should probably start returning a value in places like this (maybe just a simple bool == False, or a more comprehensive solution would be returning a non-zero error number) that we could check up the call stack and exit on error. It would avoid having `sys.exit(1)` calls scattered about the CLI and could allow for us to start using useful exit codes for certain errors (that are not `1`). Not a necessary change now, but something to keep in mind.

stream_alert_cli/terraform_generate.py
```diff
@@ -360,7 +386,11 @@ def generate_cloudtrail(cluster_name, cluster_dict, config):
         'enable_logging': cloudtrail_enabled,
         'source': 'modules/tf_stream_alert_cloudtrail',
         's3_logging_bucket': '{}.streamalert.s3-logging'.format(
-            config['global']['account']['prefix'])}
+            config['global']['account']['prefix']),
+        'existing_trail': existing_trail,
+        'is_global_trail': is_global_trail,
+        'event_pattern': json.dumps(event_pattern)
```
ryandeivert: Why do we want `event_pattern` to be a json string within a dictionary as opposed to just another python dict? I assume this has something to do with the config stuff below?

stream_alert_cli/terraform_generate.py
jacknagz: Correct, they're just booleans.  I defined them as variables to be really explicit about what the defaults are for these options.

stream_alert_cli/terraform_generate.py
jacknagz: Agreed, I'll make some changes

stream_alert_cli/terraform_generate.py
jacknagz: https://www.terraform.io/docs/providers/aws/r/cloudwatch_event_rule.html

jacknagz:
PTAL @ryandeivert, tested all CLI commands by bringing up a test cluster, deploying, and destroying

stream_alert_cli/terraform_generate.py
```diff
@@ -303,6 +316,8 @@ def generate_kinesis(cluster_name, cluster_dict, config):
         's3_logging_bucket': logging_bucket
     }
 
+    return True
```
ryandeivert: @jacknagz  in a handful of these functions it looks like we return `True` and there's really no possibility of the function returning a falsey value. Am I overlooking something? If that's the case, there's not much benefit to having a check for the return value for these function calls.

That said, I do see a few functions below that can return `False`, so it definitely applies some places.

stream_alert_cli/runner.py
```diff
@@ -87,7 +89,7 @@ def terraform_check():
     prereqs_message = ('Terraform not found! Please install and add to '
                        'your $PATH:\n'
                        '\t$ export PATH=$PATH:/usr/local/terraform/bin')
-    run_command(['terraform', 'version'],
+    return run_command(['terraform', 'version'],
```
ryandeivert: I think the formatting/indent here got messed up when the `return` was added.

stream_alert_cli/terraform_generate.py
jacknagz: I'm doing it for consistency, this could change in the future as we add more checks for validity

PR #210, opened 2017-07-01
Reorganize Rules, Matchers, and Helpers

jacknagz:
to: @ryandeivert 
cc: @airbnb/streamalert-maintainers @zbuc
size: small

## Background
Currently, helpers, matchers, and rules are all nested within the `rules/` directory.  This is slightly confusing and also makes importing tricky.  This change moves all pieces into their own respective folder

## Changes
* Move matchers and helpers into their own folders
* Recursively import matchers and rules
* Update documentation


jacknagz:
re coveralls: @ryandeivert should we move our rules testing framework into `nose` to recognize the coverage?

ryandeivert:
@jacknagz I'm not sure I understand what you mean. `nose` and `coveralls` are really only for unit tests, they have no understanding of integration or functional tests.

Are you concerned about bits of code that may be duplicative? If so, you can do something similar to what I did with `test/unit/stream_alert_alert_processor/test_outputs.py`.. create helper functions in `stream_alert_cli/helpers.py` and import with:
```
from stream_alert_cli.helpers import (
    create_lambda_function,
    put_mock_creds
)
```

If that's not what you're getting at, let's talk more offline.

ryandeivert:
@jacknagz also RE your comment about increasing coverage, this is super helpful from the repo root:

`./test/scripts/unit_tests.sh; coverage html; open htmlcov/index.html`

Use the code block highlighting to see what's covered/missed. V nice.

stream_alert/rule_processor/main.py
```diff
@@ -18,25 +18,27 @@
...
-                                      not x.endswith('.pyc') and
-                                      not x.startswith('__init__'), files)
-    for import_file in filtered_files:
-        package_path = root.replace('/', '.')
-        import_module = os.path.splitext(import_file)[0]
-        modules_to_import.add('{}.{}'.format(package_path, import_module))
-
+for folder in ('matchers/', 'rules/'):
+    for root, dirs, files in os.walk(folder):
+        filtered_files = filter(lambda x: not x.startswith('.') and
```
ryandeivert: I know this isn't new code, but didn't you say pylint doesn't like `filter` anymore? Along with `map`, etc? Thinking forward with adding `pylint` to CI, we should probably knock these out as we see them and switch to list/dict comprehension.

stream_alert/rule_processor/main.py
```diff
@@ -18,25 +18,27 @@
...
+                                          not x.endswith('.pyc') and
+                                          not x.startswith('__init__'),
+                                          files)
+        for import_file in filtered_files:
+            # Greater than two because that indicates there's only one folder
+            if len(root.split('/')) > 2:
```
ryandeivert: You could avoid splitting this string into a list temporarily (which creates a new object in memory) by using `if root.count('/') > 1:...`. This should apply the same logic you're looking for.

stream_alert/rule_processor/main.py
```diff
@@ -18,25 +18,27 @@
...
+                package_path = root.replace('/', '.')
+            else:
+                package_path = root.rstrip('/')
+            import_module = os.path.splitext(import_file)[0]
+            print package_path, import_module
```
ryandeivert: Are these some rogue `print` statements that were left behind from testing? Or do we want to keep them here? There's this one and one on L41 below.

stream_alert/rule_processor/main.py
jacknagz: That's fair, I'll convert it (even though the ruby programmer in me really loves filter and map)

stream_alert/rule_processor/main.py
jacknagz: Good tip, I'll make that change

stream_alert/rule_processor/main.py
jacknagz: This is why you don't develop late on Fridays 🤣

PR #218, opened 2017-07-07
[ci] adding sphinx-build to CI to check for doc errors

ryandeivert:
to @jacknagz 

## Change
* Adding `sphinx-build` to travis so CI will fail if there are any warnings or errors when trying to build docs. This should help us avoid having to make massive fixes to docs in the future when updates are added.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,11 @@ cache: pip
 sudo: required
 python:
   - 2.7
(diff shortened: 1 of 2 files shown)


ryandeivert:
@jacknagz fixed build errors and removed an `before install` item from travis that seems unnecessary now

PR #222, opened 2017-07-11
[scripts] adding autoflake script to repo

ryandeivert:
to @jacknagz 

### change
* Adding `autoflake` to requirements.txt and adding helper script to repo to be used for checking files for unused imports.
  * Usage: `test/scripts/autoflake.sh /path/to/folder/or/python.py`

### DIFF
--- a/test/scripts/autoflake.sh
+++ b/test/scripts/autoflake.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+autoflake --recursive --in-place --remove-all-unused-imports $1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 appdirs==1.4.2
+autoflake==0.7
 autopep8==1.3.1
 awscli==1.10.5
 boto==2.46.1


test/scripts/autoflake.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+autoflake --recursive --in-place --remove-all-unused-imports $1
```
jacknagz: Add a check for the args to make the UX a little nicer

ryandeivert: @jacknagz  autoflake will report pretty much the same thing we would...:

```
$ test/scripts/autoflake.sh 
usage: autoflake [-h] [-i] [-r] [--imports IMPORTS]
                 [--remove-all-unused-imports] [--remove-unused-variables]
                 [--version]
                 files [files ...]
autoflake: error: too few arguments
```

ryandeivert: spoke offline - jack is okay with how this would be handled

PR #224, opened 2017-07-13
Consolidate StreamAlerts Buckets

jacknagz:
to @ryandeivert 

## Background
A current issue with per-cluster S3 Alert buckets is that a rule acting across multiple environments will throw a permissions error.  The reason is that a given Lambda function will try to send to another cluster's bucket, and no permission exists to allow that.

## Change
Move the creation of StreamAlerts buckets out of each cluster and into a single bucket per StreamAlert setup.

Also update unit tests accordingly.

## Tested
Verified in a side account with two clusters.  Alerts were sending to the same bucket from multiple Lambda functions.

## Important!
Before pulling this update and migrating to a unified bucket scheme, perform the following:

1. Copy all existing StreamAlerts to a temporary bucket.
2. Empty bucket/clear out old alerts.
3. Run `python stream_alert_cli.py terraform build` after pulling updates from this repo.
4. S3 sync temp bucket to new unified bucket `<prefix>.streamalerts`.

### DIFF
--- a/stream_alert_cli/terraform_generate.py
+++ b/stream_alert_cli/terraform_generate.py
@@ -138,6 +138,10 @@ def generate_main(**kwargs):
             acl='log-delivery-write',
             logging=logging_bucket,
             lifecycle_rule=logging_bucket_lifecycle
+        ),
+        'streamalerts': generate_s3_bucket(
+            bucket='{}.streamalerts'.format(config['global']['account']['prefix']),
+            logging=logging_bucket
         )
     }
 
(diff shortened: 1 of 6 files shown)


stream_alert_cli/terraform_generate.py
```diff
@@ -237,14 +240,19 @@ def generate_stream_alert(cluster_name, cluster_dict, config):
...
+    # Add Rule Processor input config from the loaded cluster file
     input_config = modules['stream_alert']['rule_processor'].get('inputs')
     if input_config:
-        cluster_dict['module']['stream_alert_{}'.format(cluster_name)].update({
-            'input_sns_topics': input_config['aws-sns']
-        })
+        input_mapping = {
+            'input_sns_topics': 'aws-sns'
+        }
+        for tf_key, input_key in input_mapping.iteritems():
```
ryandeivert: Why the `for` loop when there is only one item? Is there chance that this will be expanded in the future with more keys?

jacknagz: > Is there chance that this will be expanded in the future with more keys?

Yep, that's the idea.

Issue #189, opened 2017-06-20
Improvement: Invoke alert processor directly from rule processor


ryandeivert:
At the moment, we use AWS's SNS to publish alerts from the rule processor to an SNS topic. This in turn fires the alert processor lambda, sending the alert(s) to the alert processor in an SNS message payload.

We should migrate away from this to a model that uses boto3 to invoke the alert processor directly from the rule processor. See boto3 [docs](http://boto3.readthedocs.io/en/latest/reference/services/lambda.html#Lambda.Client.invoke) for more info.

As an added bonus, this cleans up a lot of code and makes a lot of 'formatting' requirements we currently have non-existent. For instance, we don't have to construct an SNS payload, and can send JSON data however we want to.

austinbyers:
👍  👍  👍

PR #225, opened 2017-07-14
[core] drop sns publish for lambda invoke

ryandeivert:
to @austinbyers 
cc @jacknagz , @airbnb/dashed-maintainers 
size: large
resolves: #189 

## Why
* We have been noticing an increased number of SNS message failures in cloudwatch, which means potentially meaningful alerts are being lost in transit between the Rule Processor and the Alert Processor. We currently have no insight into why these failure occur.
* Switching to lambda, with an asynchronous invocation, provides the benefit of automatic retry logic (see: [here](http://docs.aws.amazon.com/lambda/latest/dg/retries-on-errors.html)). The invocation will now retry twice if there was a failure.
* This greatly simplifies how the Rule Processor and Alert Processor interact, removing a good amount of previously necessary logic.

## Changes in collab branch with @jacknagz 
* Migrating to using direct lambda invocation of the alert processor function from the rule processor function.
* This is an async invocation ('Event' type) and sends the payload directly to the alert processor lambda (no 'sns' formatting required).
* The alert processor now receives the payload directly in the `event` parameter of the lambda handler.
* Updating the alert processor to perform some input validation of the alert before processing occurs. This should avoid errors due to a malformed payload.
  * Specific errors indicating what is wrong with the input payload will be logged as errors, and the alert payload itself will be reported after all other errors..
* Updating all unit tests and add new ones.
* Updating integration tests to conform to other updates.
* Thanks to @jacknagz for the necessary terraform, terraform generation, and doc updates. (please fill in details here as you see fit..).

## Fixes
* Resolving bug where importing `pip` as a module broke the CLI logging. The fix is in this [commit](https://github.com/airbnb/streamalert/pull/225/commits/bf8433e8cc96ec9c66c08f70d92087b6084bb893) and is described [here](https://stackoverflow.com/questions/38754432/pip-main-resetting-logging-settings-in-python).

## To consider:
* Setting up a [DLQ](http://docs.aws.amazon.com/lambda/latest/dg/dlq.html) (Dead Letter Queue) for our lambda functions that would give us insight into any lambda invocation failures in the future.
* Checking the payload size being sent in the Alert Processor lambda invocation, and segmenting messages if it exceeds the limit (see: [here](http://docs.aws.amazon.com/lambda/latest/dg/limits.html#limits-list)).


stream_alert/rule_processor/sink.py
```diff
@@ -13,121 +13,88 @@
...
+
+            except ClientError as err:
+                LOGGER.exception('An error occurred while sending alert to '
+                                 '\'%s\'. Error is: %s. Alert: %s',
+                                 self.function,
+                                 err.response,
+                                 data)
+                return
+
+            if response['ResponseMetadata']['HTTPStatusCode'] != 202:
```
jacknagz: Is this block slightly repetitive to the preceding one?

stream_alert/alert_processor/main.py
```diff
@@ -40,68 +41,41 @@ def handler(event, context):
...
-            continue
-
-        if not 'default' in loaded_sns_message:
-            # do not log for messages related to alarms
-            if not 'AlarmName' in loaded_sns_message:
-                LOGGER.error('Malformed SNS: %s', loaded_sns_message)
-            continue
-
-        status_values.extend(run(loaded_sns_message, region, function_name, config))
+    # Yield back the current status to the caller
```
jacknagz: Can this comment be a little more descriptive?

stream_alert/alert_processor/main.py
```diff
@@ -40,68 +41,41 @@ def handler(event, context):
...
+                    'type': payload.type,
+                    'source': {
+                        'service': payload.service,
+                        'entity': payload.entity
+                    }
                 }
             }
-        }
 
         region [string]: the AWS region being used
```
jacknagz: let's update this to: `The AWS region of the currently executing Lambda function`

stream_alert/rule_processor/sink.py
ryandeivert: Can you be more specific? What's repetitive about it (and are you referring to the try/except above?)? 

If so, I think an HTTP response of non-202 could possibly occur even if there is no ClientError occurred during invocation.  I could also be off on that.. We could also add `FunctionError` to this error logging.

test/unit/stream_alert_alert_processor/test_helpers.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # Test with invalid metadata outputs value
+    assert_false(validate_alert(invalid_metadata_outputs_value))
+
+    # metadata > non-string value validation
+    invalid_metadata_non_string = deepcopy(valid_alert)
+    invalid_metadata_non_string['metadata']['type'] = 4.5
+
+    # Test with invalid metadata non-string value
+    assert_false(validate_alert(invalid_metadata_non_string))
```
austinbyers: Can we break each of these up into smaller unit tests? Each `assert` condition with a comment can likely be its own self-contained test.

stream_alert/rule_processor/sink.py
```diff
@@ -13,121 +13,89 @@
...
 '''
-
 import json
 import logging
 import sys
 
 import boto3
 from botocore.exceptions import ClientError
 
-_SNS_MAX_SIZE = (256*1024)
```
austinbyers: There is still a [maximum Lambda invocation size](http://docs.aws.amazon.com/lambda/latest/dg/limits.html#limits-list) (6MB for asynchronous requests), but I don't think we'll be anywhere close to that limit. Just something to keep in mind

terraform/modules/tf_stream_alert/sns.tf
```diff
@@ -1,30 +1,18 @@
...
-// Subscribe the Rule Processor Lambda function to arbitrary SNS topics
+// SNS Topic Subscription: Subscribe the Rule Processor to configured SNS topics
 resource "aws_sns_topic_subscription" "input_topic_subscriptions" {
   count     = "${length(var.input_sns_topics)}"
   topic_arn = "${element(var.input_sns_topics, count.index)}"
   endpoint  = "${aws_lambda_function.streamalert_rule_processor.arn}:production"
   protocol  = "lambda"
 }
+
+// Lambda Permission: Allow SNS to invoke the Rule Processor
```
jacknagz: We had it, it just moved into this file from [here](https://github.com/airbnb/streamalert/pull/225/files/34853502dc286f165fa685af58d915eb1a13d33b#diff-ef4ea968687cd4397072c930bf5b0a50)

ryandeivert:
@jacknagz @austinbyers PTAL at new commit and comments

stream_alert/alert_processor/main.py
```diff
@@ -40,78 +42,52 @@ def handler(event, context):
...
-            continue
-
-        sns_message = sns_payload['Message']
-        try:
-            loaded_sns_message = json.loads(sns_message)
-        except ValueError as err:
-            LOGGER.error('An error occurred while decoding message to JSON: %s', err)
-            continue
+    # Return the current list of statuses back to the caller
+    return list(status for status in run(event, region, function_name, config))
```
austinbyers: You can simplify this further:
```python
return list(run(event, region, function_name, config))
```
`list` will automatically transform the generator into a list by iterating through it; you don't have to manually iterate

terraform/modules/tf_stream_alert/sns.tf
austinbyers: :+1:

test/unit/stream_alert_rule_processor/test_sink.py
```diff
@@ -78,16 +64,17 @@ def test_json_from_dict():
...
-            '"metadata": {"outputs": "rule.outputs_01", "type": "payload_type_01", ' \
-            '"log": "payload_data_01", "source": {"service": "payload_service_01", ' \
-            '"entity": "payload_entity_01"}}, "record": {"record_data_key02_01": ' \
-            '"record_data_value02_01", "record_data_key01_01": "record_data_value01_01"}}}')
+        assert_equal(json_message, '{"record": {"record_data_key02_01": ' \
+            '"record_data_value02_01", "record_data_key01_01": "record_data_value01_01"}, ' \
+            '"metadata": {"source": {"service": "payload_service_01", ' \
+            '"entity": "payload_entity_01"}, "rule_name": "test_rule_01", ' \
+            '"type": "payload_type_01", "log": "payload_data_01", ' \
+            '"outputs": "rule.outputs_01"}}')
```
austinbyers: You don't need the line continuation markers (`\`) (this is discouraged by Google Python style guide); Python has implicit string joining.

ryandeivert: not in py 2 :) you need to wrap in parenths for that to be effective, but I can add that

stream_alert/alert_processor/main.py
ryandeivert: ah sorry overlooked that bit.

test/unit/stream_alert_rule_processor/test_sink.py
austinbyers: Oh yes, you always need the parens (in both py2 and py3). Good catch

ryandeivert: @austinbyers  fixed in latest commit.. as a param in a function call it looks like it doesn't need to be encapsulated

PR #228, opened 2017-07-18
[logs] cb feed.query.hit.process opt key update

ryandeivert:
to @austinbyers 

## Changes
* Adding some optional keys to the `carbonblack:feed.query.hit.process` log schema.


jacknagz:
@ryandeivert ready to merge?

PR #229, opened 2017-07-20
Support Historical Search of Alerts, CLI Configure Command

jacknagz:
to: @ryandeivert 
size: medium

## Background

This PR marks the beginning of the Historical Search (#5) project with StreamAlert.  This set of changes focuses on refactoring the alert payload format to flatten the structure for searching with Amazon Athena.  It also includes a new CLI command for easily configuring your `prefix` and `aws_account_id`.

## Changes

Flatten the StreamAlert structure:

```
{
  "log_source": "cloudtrail:api_events",
  "log_type": "json",
  "outputs": [
    "aws-s3:main"
  ],
  "record": "{\"awsRegion\": \"...\", \"eventID\": \"...\", \"eventName\": \"DeleteVpc\", \"eventSource\": \"ec2.amazonaws.com\", \"eventTime\": \"...\", \"eventType\": \"AwsApiCall\", \"eventVersion\": \"...\", \"recipientAccountId\": \"12345\", \"requestID\": \"...\", \"requestParameters\": {\"vpcId\": \"...\"}, \"responseElements\": {\"_return\": true}, \"sourceIPAddress\": \"...\", \"userAgent\": \"...\", \"userIdentity\": {\"accessKeyId\": \"...\", \"accountId\": \"12345\", \"arn\": \"...\", \"principalId\": \"12345\", \"sessionContext\": {\"attributes\": {\"creationDate\": \"...\", \"mfaAuthenticated\": \"true\"}}, \"type\": \"...\"}}",
  "rule_description": "\n    author:           airbnb_csirt\n    description:      Alert on AWS API calls that stop or delete security/infrastructure logs.\n                      Additionally, alert on AWS API calls that delete critical resources (VPCs, Subnets, DB's, ...)\n    reference:        https://medium.com/@robwitoff/proactive-cloud-security-w-aws-organizations-d58695bcae16#.tx2e6iju0\n    playbook:         (a) identify the AWS account in the log\n                      (b) identify what resource(s) are impacted by the API call\n                      (c) determine if the intent is valid, malicious or accidental\n    ",
  "rule_name": "cloudtrail_critical_api",
  "source_entity": "prefix_cluster1_stream_alert_kinesis",
  "source_service": "kinesis"
}
```

In Athena, it's searchable via:

screen shot 2017-07-19 at 6 55 47 pm

Updates relevant unit tests

Add a new `configure` command to the StreamAlertCLI:

```
$ python stream_alert_cli.py configure prefix <prefix-goes-here>
```

## Next Steps

Following this PR, I'll start to add support for creation of the Athena streamalerts table, and create a scheduled lambda function for updating of the [partitions](http://docs.aws.amazon.com/athena/latest/ug/partitions.html) as new data comes in.  This is unfortunately required by Hive to update the metastore.


stream_alert/alert_processor/outputs.py
```diff
@@ -512,20 +513,33 @@ def dispatch(self, **kwargs):
...
+        #   alerts/dt=2017-01-25-00-15/kinesis_mystream_myrule_<uuid>.json
+        key = 'alerts/dt={}/{}_{}_{}_{}.json'.format(
+            current_date.strftime('%Y-%m-%d-%H-%M'),
             service,
             entity,
-            kwargs['rule_name'],
-            current_date.strftime('%Y-%m-%d-%H-%M'),
-            current_date.isoformat('-')
+            alert['rule_name'],
+            # keys need to be unique to avoid object overwriting
```
ryandeivert: Can this comment just get lumped in with the ones above. Placing it in the `format` params makes it look like a param and confusing/odd.

stream_alert_cli/runner.py
```diff
@@ -84,6 +88,25 @@ def lambda_handler(options):
...
+        options [named_tuple]: ArgParse command result
+    """
+    if options.config_key == 'prefix':
+        if type(options.config_key) not in (unicode, str):
+            LOGGER_CLI.error('Invalid prefix type, must be string')
+            return
+        CONFIG.set_prefix(options.config_value)
+
+    elif options.config_key == 'aws_account_id':
+        if not re.search(r'\A\d{12}\Z', options.config_value):
```
ryandeivert: Since regex is inherently slow, can't we just check for a len of 12 and str.isdigit()?

stream_alert_cli/config.py
```diff
@@ -47,6 +49,20 @@ def get(self, key):
     def clusters(self):
         return self.config['clusters'].keys()
 
+    def set_prefix(self, prefix):
```
ryandeivert: Love this! Great stuff. 👍 💯

stream_alert_cli/config.py
```diff
@@ -18,17 +18,19 @@
...
+
 from stream_alert_cli.logger import LOGGER_CLI
 
 
 class CLIConfig(object):
     '''Provide an object to load, modify, and display the StreamAlertCLI Config'''
 
     def __init__(self):
-        self.config_files = {
+        self.config_files = OrderedDict({
```
ryandeivert: I don't think you're actually using is as an OrderedDict now (so this could be removed) but I also thought I'd point out something I noticed a while back. This is a weird thing with OrderedDict - but you can't cast a dict to OrdererDict like this and have it maintain order (which I guess makes sense). For instance, if you run this code:

```

test_dict = OrderedDict({
    'global2': 'conf/global.json',
    'global1': 'conf/global.json',
    'lambda1': 'conf/lambda.json',
    'lambda2': 'conf/lambda.json',
    'lambda3': 'conf/lambda.json',
    'lambda4': 'conf/lambda.json',
    'lambda5': 'conf/lambda.json'
})

for key, value in test_dict.iteritems():
    print key, value
```
It will print not print in the order you expect. It will probably look _similar_ to:
```
global1 conf/global.json
global2 conf/global.json
lambda4 conf/lambda.json
lambda5 conf/lambda.json
lambda1 conf/lambda.json
lambda2 conf/lambda.json
lambda3 conf/lambda.json
```

The actual not-so-obvious way of doing this is through casting a list of ordered tuples:
```
OrderedDict([
    ('global', 'conf/global.json'),
    ('lambda', 'conf/lambda.json')
])
```

I think you just want to remove the cast to OrderedDict until we figure out how to better test this stuff.

stream_alert_cli/runner.py
jacknagz: is it _really_ that much slower though?

jacknagz:
```
>>> import cProfile
>>> import re
>>> account_id = "123456789123"
>>> def jack_test(account_id):
...     print re.search(r'\A\d{12}\Z', account_id)
>>> def ryan_test(account_id):
...     print len(account_id) == 12 and account_id.isdigit()
>>> cProfile.run('jack_test(account_id)')
<_sre.SRE_Match object at 0x106621718>
         6 function calls in 0.000 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 <stdin>:1(jack_test)
        1    0.000    0.000    0.000    0.000 <string>:1(<module>)
        1    0.000    0.000    0.000    0.000 re.py:143(search)
        1    0.000    0.000    0.000    0.000 re.py:230(_compile)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
        1    0.000    0.000    0.000    0.000 {method 'search' of '_sre.SRE_Pattern' objects}


>>> cProfile.run('ryan_test(account_id)')
True
         5 function calls in 0.000 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 <stdin>:1(ryan_test)
        1    0.000    0.000    0.000    0.000 <string>:1(<module>)
        1    0.000    0.000    0.000    0.000 {len}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
        1    0.000    0.000    0.000    0.000 {method 'isdigit' of 'str' objects}
```

🤣

stream_alert_cli/config.py
jacknagz: good callout, I'll make the fix

stream_alert_cli/config.py
jacknagz: 🥇

stream_alert/alert_processor/outputs.py
jacknagz: sure, I'll make the fix

stream_alert_cli/runner.py
ryandeivert: @jacknagz one call isn't going to be enough to clock anything 🤣 . Use `timeit`

```
$ python -m timeit --setup="import re; account_id='123456789123'" "re.search(r'\A\d{12}\Z', account_id)" 
1000000 loops, best of 3: 1.03 usec per loop
$ python -m timeit --setup="account_id='123456789123'" "len(account_id) == 12 and account_id.isdigit()"
10000000 loops, best of 3: 0.111 usec per loop
```
1.03 usec w/ `re` and 0.111 usec without

PR #234, opened 2017-07-31
Athena Partition Refresh

jacknagz:
to: 
@ryandeivert (CLI and Misc)
@chunyong-lin (Terraform)
@austinbyers (Lambda Function)

cc: 
@airbnb/streamalert-maintainers 

## Background

As a follow up to #229 and #5, this pull request includes automation for configuring and operationalizing a StreamAlert Athena database.  Currently, this focuses on supporting the `alerts` table, with automatic partition refreshing to enable searching of alerts right after they occur.

This includes initial setup of the Athena configuration via the CLI, creation of the `streamalert` database and `alerts` table, and configuring automatic partition refreshes (via Lambda) when new data is written to the `prefix.streamalerts` bucket or others defined in the config.

### Athena Terraform Modle
* Create a new Lambda function (`<prefix>_streamalert_athena_partition_refresh`).
* Configure S3 events from the configured data buckets to invoke this Lambda function.
* Allow the function to run Athena queries, get results, and write search results to an Athena bucket.
* Expire CloudWatch logs after 30 days.
* All generated Terraform writes to `terraform/athena.tf`.

### Athena Partition Refresh Lambda Function
* When invoked, first checks the `streamalert` database exists.  If it doesn't, the function exits and prints an error telling the user to create it (via the CLI).
* If the db exists, it loads the configured data buckets, and attempt to refresh all of them (I just realized this is a bug and I need to fix it, it should only refresh the bucket where the S3 event came from).
* When each query is ran, the function uses the `backoff` Python library to continuously check for the result of the query.   Results are printed to the CloudWatch log.
* Adds unit testing coverage (88%).

### Athena CLI Support
* Add a new `athena` command to initialize the Lambda configuration.  There is also an option of enabling/disabling the Lambda function, which creates (or removes) the generated Athena Terraform module.
* Generate the Athena Terraform module.
* Refactor versioning to support clustered or non-clustered Lambda functions.
* Update the deploy flow to support the newly created Athena function.
* Unit testing coverage.

### Misc
* Add a CLUSTER environment variable to the existing Lambda functions.  I would like to move towards a model where the function environment is more easily identifiable (vs chopping up the config or invoked function ARN).
* Add support for `--debug`in log output to all CLI parsers and move specific `logger.info` calls to `logger.debug`:

```
$ python stream_alert_cli.py lambda deploy --processor athena
StreamAlertCLI [INFO]: Creating package for athena_partition_refresh
StreamAlertCLI [INFO]: Installing third-party libraries: backoff
StreamAlertCLI [INFO]: Package successfully created
StreamAlertCLI [INFO]: Uploading StreamAlert package to S3
StreamAlertCLI [INFO]: Planning infrastructure
```

## Testing

* All testing has been performed in a test AWS account.
* Verified that S3 notifications are working, and the Lambda function is invoking and running commands.
* Verified the deploy flow for all Lambda functions.
* There's current an issue I'm experiencing with getting the repair table command failing to run via the Lambda function, I'm working on troubleshooting with AWS.

## To Do

- [ ] Documentation on setup and usage
- [ ] Use SQS for receiving bucket notifications from S3

### DIFF
--- a/stream_alert/athena_partition_refresh/main.py
+++ b/stream_alert/athena_partition_refresh/main.py
@@ -113,7 +113,7 @@ class StreamAlertAthenaClient(object):
 
         Returns:
             [string]: The result of the query, this can be SUCCEEDED, FAILED, or CANCELLED.
-                      Reference http://bit.ly/2uuRtda
+                      Reference https://bit.ly/2uuRtda
         """
         @backoff.on_predicate(backoff.fibo,
                               lambda status: status in ('QUEUED', 'RUNNING'),
@@ -166,7 +166,7 @@ class StreamAlertAthenaClient(object):
         # No data being returned isn't always an indication that something is wrong.
         # When handling the query result data, iterate over each element in the Row,
         # and parse the Data key.
-        # Reference: http://bit.ly/2tWOQ2N
+        # Reference: https://bit.ly/2tWOQ2N
         if not query_results_resp['ResultSet']['Rows']:
             LOGGER.debug('The query %s returned empty rows of data', kwargs['query'])
 
@@ -205,12 +205,12 @@ class StreamAlertAthenaClient(object):
                         table_name)
             return False
 
-    def normal_partition_refresh(self):
-        """Execute a normal MSCK REPAIR TABLE command on a given data bucket"""
+    def repair_hive_table(self):
+        """Execute a MSCK REPAIR TABLE on a given Athena table"""
         athena_config = self.config['lambda']['athena_partition_refresh_config']
-        normal_partition_config = athena_config['partitioning']['normal']
+        repair_hive_table_config = athena_config['refresh_type']['repair_hive_table']
 
-        for athena_table in normal_partition_config.itervalues():
+        for athena_table in repair_hive_table_config.itervalues():
             query_success, query_resp = self.run_athena_query(
                 query='MSCK REPAIR TABLE {};'.format(athena_table),
                 database=self.DATABASE_STREAMALERT
@@ -238,4 +238,4 @@ def handler(event, _):
     if not stream_alert_athena.check_database_exists():
         return
 
-    stream_alert_athena.normal_partition_refresh()
+    stream_alert_athena.repair_hive_table()
--- a/stream_alert/athena_partition_refresh/main.py
+++ b/stream_alert/athena_partition_refresh/main.py
@@ -28,179 +28,214 @@ LEVEL = os.environ.get('LOGGER_LEVEL', 'INFO')
 LOGGER = logging.getLogger('StreamAlertAthena')
 LOGGER.setLevel(LEVEL.upper())
 
-ATHENA_CLIENT = None
-DATABASE_DEFAULT = 'default'
+
+class ConfigError(Exception):
+    """Custom StreamAlertAthena Config Exception Class"""
+    pass
 
 
-def _load_config():
-    """Load the Athena Lambda configuration file
+class StreamAlertAthenaClient(object):
+    DATABASE_DEFAULT = 'default'
+    DATABASE_STREAMALERT = 'streamalert'
+    DEFAULT_S3_PREFIX = 'athena_partition_refresh'
 
-    Returns:
-        [dict] Configuration settings by file, includes two keys:
-            lambda: All lambda function settings
-            global: StreamAlert global settings
-    """
-    config_files = ('lambda', 'global')
-    config = {}
-    for config_file in config_files:
-        with open('conf/{}.json'.format(config_file)) as config_fh:
-            try:
-                config[config_file] = json.load(config_fh)
-            except ValueError:
-                LOGGER.error('The \'%s\' file could not be loaded into json', config_file)
-                return
+    def __init__(self, **kwargs):
+        """Initialize the Boto3 Athena Client, and S3 results bucket/key
 
-    return config
+        Keyword Arguments:
+            config [CLIConfig]: Loaded StreamAlert configuration (optional)
+            results_key_prefix [string]: The S3 key prefix to store Athena results (optional)
+        """
+        # Load the config from files or accept it as an argument
+        self.config = kwargs.get('config') or self._load_config()
+        region = self.config['global']['account']['region']
+        self.athena_client = boto3.client('athena', region_name=region)
 
+        # Format the S3 bucket to store Athena query results
+        self.athena_results_bucket = 's3://aws-athena-query-results-{}-{}'.format(
+            self.config['global']['account']['aws_account_id'],
+            self.config['global']['account']['region'])
 
-def _backoff_handler(details):
-    """Simple logging handler for when polling backoff occurs."""
-    LOGGER.debug('Trying again in %f seconds after %d tries calling %s',
-                 details['wait'],
-                 details['tries'],
-                 details['target'])
+        # Format the S3 key to store specific objects
+        results_key_prefix = kwargs.get('results_key_prefix', self.DEFAULT_S3_PREFIX)
+        # Produces athena_partition_refresh/YYYY/MM/DD S3 keys
+        self.athena_results_key = os.path.join(
+            results_key_prefix,
+            datetime.now().strftime('%Y/%m/%d'))
 
+    @staticmethod
+    def _load_config():
+        """Load the StreamAlert Athena configuration files
 
-def _success_handler(details):
-    """Simple logging handler for when polling backoff occurs."""
-    LOGGER.debug('Completed after %d tries calling %s',
-                 details['tries'],
-                 details['target'])
+        Returns:
+            [dict] Configuration settings by file, includes two keys:
+                lambda: All lambda function settings
+                global: StreamAlert global settings
 
+        Raises:
+            [ConfigError] For invalid or missing configuration files.
+        """
+        config_files = ('lambda', 'global')
+        config = {}
+        for config_file in config_files:
+            config_file_path = 'conf/{}.json'.format(config_file)
 
-def check_query_status(query_execution_id):
-    """Check in on the running query, back off if the job is running or queued
+            if not os.path.exists(config_file_path):
+                raise ConfigError('The \'{}\' config file was not found'.format(
+                    config_file_path))
 
-    Returns:
-        [string]: The result of the query, this can be SUCCEEDED, FAILED, or CANCELLED.
-                  Reference http://bit.ly/2uuRtda
-    """
-    @backoff.on_predicate(backoff.fibo,
-                          lambda status: status in ('QUEUED', 'RUNNING'),
-                          max_value=10,
-                          jitter=backoff.full_jitter,
-                          on_backoff=_backoff_handler,
-                          on_success=_success_handler)
-    def _get_query_execution(query_execution_id):
-        return ATHENA_CLIENT.get_query_execution(
-            QueryExecutionId=query_execution_id
-        )['QueryExecution']['Status']['State']
+            with open(config_file_path) as config_fh:
+                try:
+                    config[config_file] = json.load(config_fh)
+                except ValueError:
+                    raise ConfigError('The \'{}\' config file is not valid JSON'.format(
+                        config_file))
 
-    return _get_query_execution(query_execution_id)
+        return config
 
+    @staticmethod
+    def _backoff_handler(details):
+        """Simple logging handler for when polling backoff occurs."""
+        LOGGER.debug('[Backoff]: Trying again in %f seconds after %d tries calling %s',
+                     details['wait'],
+                     details['tries'],
+                     details['target'])
 
-def run_athena_query(**kwargs):
-    """Helper function to run Athena queries
+    @staticmethod
+    def _success_handler(details):
+        """Simple logging handler for when polling backoff occurs."""
+        LOGGER.debug('[Backoff]: Completed after %d tries calling %s',
+                     details['tries'],
+                     details['target'])
 
-    Keyword Args:
-        query [string]: The SQL query to execute
-        database [string]: The database to execute the query against
-        results_bucket [string]: The S3 bucket to store query results
-        results_path [string]: The S3 key to store results
+    def check_query_status(self, query_execution_id):
+        """Check in on the running query, back off if the job is running or queued
 
-    Returns (one or the other):
-        [bool]: Query success
-        [dict]: Query result response
-    """
-    LOGGER.debug('Executing query: %s', kwargs['query'])
-    query_execution_resp = ATHENA_CLIENT.start_query_execution(
-        QueryString=kwargs['query'],
-        QueryExecutionContext={'Database': kwargs.get('database', DATABASE_DEFAULT)},
-        ResultConfiguration={'OutputLocation': '{}/{}'.format(
-            kwargs['results_bucket'],
-            kwargs['results_path']
-        )}
-    )
-    query_execution_result = check_query_status(query_execution_resp['QueryExecutionId'])
-    if query_execution_result != 'SUCCEEDED':
-        LOGGER.error(
-            'The query %s returned %s, exiting!',
-            kwargs['query'],
-            query_execution_result)
-        return False
+        Returns:
+            [string]: The result of the query, this can be SUCCEEDED, FAILED, or CANCELLED.
+                      Reference http://bit.ly/2uuRtda
+        """
+        @backoff.on_predicate(backoff.fibo,
+                              lambda status: status in ('QUEUED', 'RUNNING'),
+                              max_value=10,
+                              jitter=backoff.full_jitter,
+                              on_backoff=self._backoff_handler,
+                              on_success=self._success_handler)
+        def _get_query_execution(query_execution_id):
+            return self.athena_client.get_query_execution(
+                QueryExecutionId=query_execution_id
+            )['QueryExecution']['Status']['State']
 
-    query_results_resp = ATHENA_CLIENT.get_query_results(
-        QueryExecutionId=query_execution_resp['QueryExecutionId'],
-    )
+        return _get_query_execution(query_execution_id)
 
-    # The idea here is to leave the processing logic to the calling functions.
-    # No data being returned isn't always an indication that something is wrong.
-    if not query_results_resp['ResultSet']['Rows']:
-        LOGGER.debug('The query %s returned no data', kwargs['query'])
+    def run_athena_query(self, **kwargs):
+        """Helper function to run Athena queries
 
-    return query_results_resp
+        Keyword Args:
+            query [string]: The SQL query to execute
+            database [string]: The database context to execute the query in
 
-
-def check_database_exists(results_bucket, results_path):
-    """Verify the StreamAlert Athena database exists."""
-    query_resp = run_athena_query(
-        query='SHOW DATABASES LIKE \'streamalert\';',
-        results_bucket=results_bucket,
-        results_path=results_path
-    )
-    if isinstance(query_resp, dict) and not query_resp['ResultSet']['Rows']:
-        LOGGER.info('The \'streamalert\' database does not exist, please create it.')
-        return False
-
-    return True
-
-
-def check_table_exists(results_bucket, results_path, table_name):
-    """Verify a given StreamAlert Athena table exists."""
-    query_resp = run_athena_query(
-        query='SHOW TABLES LIKE \'{}\';'.format(table_name),
-        database='streamalert',
-        results_bucket=results_bucket,
-        results_path=results_path
-    )
-    if isinstance(query_resp, dict) and not query_resp['ResultSet']['Rows']:
-        LOGGER.info('The streamalert table \'%s\' does not exist, please create it.', table_name)
-        return False
-
-    return True
-
-
-def normal_partition_refresh(config, athena_results_bucket, athena_results_path):
-    normal_partition_config = config['lambda']['athena_partition_refresh_config']['partitioning']['normal']
-    for _, athena_table in normal_partition_config.iteritems():
-        resp = run_athena_query(
-            query='MSCK REPAIR TABLE {};'.format(athena_table),
-            database='streamalert',
-            results_bucket=athena_results_bucket,
-            results_path=athena_results_path
+        Returns:
+            tuple: [bool]: Query success, [dict]: Query result response
+        """
+        LOGGER.debug('Executing query: %s', kwargs['query'])
+        query_execution_resp = self.athena_client.start_query_execution(
+            QueryString=kwargs['query'],
+            QueryExecutionContext={'Database': kwargs.get('database', self.DATABASE_DEFAULT)},
+            ResultConfiguration={'OutputLocation': '{}/{}'.format(
+                self.athena_results_bucket,
+                self.athena_results_key
+            )}
         )
-        if resp:
-            LOGGER.info('Query results:')
-            for row in resp['ResultSet']['Rows']:
-                LOGGER.info(row)
+
+        query_execution_result = self.check_query_status(
+            query_execution_resp['QueryExecutionId'])
+
+        if query_execution_result != 'SUCCEEDED':
+            LOGGER.error(
+                'The query %s returned %s, exiting!',
+                kwargs['query'],
+                query_execution_result)
+            return False, {}
+
+        query_results_resp = self.athena_client.get_query_results(
+            QueryExecutionId=query_execution_resp['QueryExecutionId'],
+        )
+
+        # The idea here is to leave the processing logic to the calling functions.
+        # No data being returned isn't always an indication that something is wrong.
+        # When handling the query result data, iterate over each element in the Row,
+        # and parse the Data key.
+        # Reference: http://bit.ly/2tWOQ2N
+        if not query_results_resp['ResultSet']['Rows']:
+            LOGGER.debug('The query %s returned empty rows of data', kwargs['query'])
+
+        return True, query_results_resp
+
+    def check_database_exists(self, **kwargs):
+        """Verify the StreamAlert Athena database exists."""
+        database = kwargs.get('database', self.DATABASE_STREAMALERT)
+        query_success, query_resp = self.run_athena_query(
+            query='SHOW DATABASES LIKE \'{}\';'.format(database),
+        )
+
+        if query_success and query_resp['ResultSet']['Rows']:
+            return True
+        else:
+            LOGGER.error('The \'%s\' database does not exist. '
+                         'Create it with the following command: \n'
+                         '$ python stream_alert_cli.py athena create-db',
+                         database)
+            return False
+
+    def check_table_exists(self, table_name):
+        """Verify a given StreamAlert Athena table exists."""
+        query_success, query_resp = self.run_athena_query(
+            query='SHOW TABLES LIKE \'{}\';'.format(table_name),
+            database=self.DATABASE_STREAMALERT
+        )
+
+        if query_success and query_resp['ResultSet']['Rows']:
+            return True
+        else:
+            LOGGER.info('The streamalert table \'%s\' does not exist. '
+                        'For alert buckets, create it with the following command: \n'
+                        '$ python stream_alert_cli.py athena create-table '
+                        '--type alerts --bucket s3.bucket.id',
+                        table_name)
+            return False
+
+    def normal_partition_refresh(self):
+        """Execute a normal MSCK REPAIR TABLE command on a given data bucket"""
+        athena_config = self.config['lambda']['athena_partition_refresh_config']
+        normal_partition_config = athena_config['partitioning']['normal']
+
+        for athena_table in normal_partition_config.itervalues():
+            query_success, query_resp = self.run_athena_query(
+                query='MSCK REPAIR TABLE {};'.format(athena_table),
+                database=self.DATABASE_STREAMALERT
+            )
+            if query_success:
+                LOGGER.info('Query results:')
+                for row in query_resp['ResultSet']['Rows']:
+                    LOGGER.info(row['Data'])
+            else:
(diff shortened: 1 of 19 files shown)


stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+
+def _load_config():
+    """Load the Athena Lambda configuration file
+
+    Returns:
+        [dict] Configuration settings by file, includes two keys:
+            lambda: All lambda function settings
+            global: StreamAlert global settings
+    """
+    config_files = ('lambda', 'global')
```
mime-frame: `lambda_settings` and `global_settings` - more accurately describes what the key's point to

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+            try:
+                config[config_file] = json.load(config_fh)
+            except ValueError:
+                LOGGER.error('The \'%s\' file could not be loaded into json', config_file)
+                return
+
+    return config
+
+
+def _backoff_handler(details):
```
mime-frame: `_backoff_logging_handler` - more accurately describes what the function does

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+    """Simple logging handler for when polling backoff occurs."""
+    LOGGER.debug('Trying again in %f seconds after %d tries calling %s',
+                 details['wait'],
+                 details['tries'],
+                 details['target'])
+
+
+def _success_handler(details):
```
mime-frame: `_success_logging_handler`

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+
+def check_database_exists(results_bucket, results_path):
+    """Verify the StreamAlert Athena database exists."""
+    query_resp = run_athena_query(
+        query='SHOW DATABASES LIKE \'streamalert\';',
+        results_bucket=results_bucket,
+        results_path=results_path
+    )
+    if isinstance(query_resp, dict) and not query_resp['ResultSet']['Rows']:
+        LOGGER.info('The \'streamalert\' database does not exist, please create it.')
```
mime-frame: Why `info` vs. `error`? Also, could this happen if the right policies/permissions aren't applied?

mime-frame: "please create it." - should TF be doing this? If so, maybe give more direct instructions on what command they should run

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+def check_table_exists(results_bucket, results_path, table_name):
+    """Verify a given StreamAlert Athena table exists."""
+    query_resp = run_athena_query(
+        query='SHOW TABLES LIKE \'{}\';'.format(table_name),
+        database='streamalert',
+        results_bucket=results_bucket,
+        results_path=results_path
+    )
+    if isinstance(query_resp, dict) and not query_resp['ResultSet']['Rows']:
+        LOGGER.info('The streamalert table \'%s\' does not exist, please create it.', table_name)
```
mime-frame: same comment - if TF should be doing this, maybe give more direct instructions on what command they should run

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def run_athena_query(**kwargs):
+    """Helper function to run Athena queries
+
+    Keyword Args:
+        query [string]: The SQL query to execute
+        database [string]: The database to execute the query against
+        results_bucket [string]: The S3 bucket to store query results
+        results_path [string]: The S3 key to store results
```
mime-frame: I would suggest changing this to `results_bucket_key` and refactoring the remaining code - no need to abstract away s3 concepts and terminology, in fact it can make it more confusing

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+            kwargs['query'],
+            query_execution_result)
+        return False
+
+    query_results_resp = ATHENA_CLIENT.get_query_results(
+        QueryExecutionId=query_execution_resp['QueryExecutionId'],
+    )
+
+    # The idea here is to leave the processing logic to the calling functions.
+    # No data being returned isn't always an indication that something is wrong.
```
mime-frame: expound upon this, what should the caller do if there is no data returned? And to clarify, data is returned, it's just that ['ResultSet']['Rows'] is an empty set or hash?

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+    if not query_results_resp['ResultSet']['Rows']:
+        LOGGER.debug('The query %s returned no data', kwargs['query'])
+
+    return query_results_resp
+
+
+def check_database_exists(results_bucket, results_path):
+    """Verify the StreamAlert Athena database exists."""
+    query_resp = run_athena_query(
+        query='SHOW DATABASES LIKE \'streamalert\';',
```
mime-frame: probably should be a const or loaded from a config vs. hardcoded string literal? (same thought for all other instances where you do this)

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+
+    global ATHENA_CLIENT
+    ATHENA_CLIENT = boto3.client('athena', region_name=config['global']['account']['region'])
+
+    # Athena queries need an S3 bucket for results
+    athena_results_bucket = 's3://aws-athena-query-results-{}-{}'.format(
+        config['global']['account']['aws_account_id'],
+        config['global']['account']['region']
+    )
+    # Produces athena_partition_refresh/2017/01/01 keys
```
mime-frame: update code comment - it makes it sound like we're only updating keys for `2017/01/01`

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+def run_athena_query(**kwargs):
+    """Helper function to run Athena queries
+
+    Keyword Args:
+        query [string]: The SQL query to execute
+        database [string]: The database to execute the query against
+        results_bucket [string]: The S3 bucket to store query results
+        results_path [string]: The S3 key to store results
+
+    Returns (one or the other):
```
mime-frame: consider creating a new object that has a member field that denotes if the query is complete or not, and another member field that has the query results. i can see us wanting to store other data in that object as well (total time, errors, ...)

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+
+    # The idea here is to leave the processing logic to the calling functions.
+    # No data being returned isn't always an indication that something is wrong.
+    if not query_results_resp['ResultSet']['Rows']:
+        LOGGER.debug('The query %s returned no data', kwargs['query'])
+
+    return query_results_resp
+
+
+def check_database_exists(results_bucket, results_path):
```
mime-frame: function does not take in a database name and should be denoted as private via a `_` prefix?

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+        results_path=results_path
+    )
+    if isinstance(query_resp, dict) and not query_resp['ResultSet']['Rows']:
+        LOGGER.info('The streamalert table \'%s\' does not exist, please create it.', table_name)
+        return False
+
+    return True
+
+
+def normal_partition_refresh(config, athena_results_bucket, athena_results_path):
```
mime-frame: where did the `normal` terminology come from? I searched the docs, couldn't find it. If made up, what's normal about this and not normal about others?

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+            except ValueError:
+                LOGGER.error('The \'%s\' file could not be loaded into json', config_file)
+                return
+
+    return config
+
+
+def _backoff_handler(details):
+    """Simple logging handler for when polling backoff occurs."""
+    LOGGER.debug('Trying again in %f seconds after %d tries calling %s',
```
mime-frame: This message will appear in console with no context, the string should denote something about polling backoff

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+                 details['wait'],
+                 details['tries'],
+                 details['target'])
+
+
+def _success_handler(details):
+    """Simple logging handler for when polling backoff occurs."""
+    LOGGER.debug('Completed after %d tries calling %s',
```
mime-frame: Be more descriptive since this will show up in console w/o context. E.g. what was completed

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+        return False
+
+    query_results_resp = ATHENA_CLIENT.get_query_results(
+        QueryExecutionId=query_execution_resp['QueryExecutionId'],
+    )
+
+    # The idea here is to leave the processing logic to the calling functions.
+    # No data being returned isn't always an indication that something is wrong.
+    if not query_results_resp['ResultSet']['Rows']:
+        LOGGER.debug('The query %s returned no data', kwargs['query'])
```
mime-frame: `The query %s returned no data` or `The query %s returned no matching results` ?

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+def firehose_partition_refresh(_):
+    LOGGER.error('Firehose partition refresh is not yet supported, exiting!')
+    raise NotImplementedError
+
+
+def handler(event, context):
+    """Athena Partition Refresher Handler Function"""
+    config = _load_config()
+    if not config:
+        LOGGER.error('No config found, exiting!')
```
mime-frame: Will appear in console w/o context, list the path or be more helpful

stream_alert_cli/package.py
```diff
@@ -193,25 +205,27 @@ def _upload(self, package_path):
             [boolean] Indicating a successful upload
 
         Raises:
-            Exception if client put_object fails
+            Exception if the Lambda package put object fails to S3
```
ryandeivert: This no longer raises an exception and only returns a boolean now - maybe either update the docstring or still raise the exception below.

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+limitations under the License.
+'''
+
+import json
+import logging
+import os
+
+from datetime import datetime
+
+import backoff
```
austinbyers: I did not know about the `backoff` library; very interesting!

austinbyers: I didn't know about the `backoff` library, very interesting!

stream_alert_cli/terraform_generate.py
```diff
@@ -601,16 +634,16 @@ def terraform_generate(**kwargs):
...
-    # Break out early during the init process, clusters aren't needed yet
+    # Return early during the init process, clusters are not needed yet
     if init:
         return True
 
-    # Setup clusters
+    # Setup cluster files
     for cluster in config.clusters():
-        if cluster == 'main':
+        if cluster in RESTRICTED_CLUSTER_NAMES:
```
ryandeivert: You're now restricting using the cluster name of `athena` also - update the exceptions message to state the actual cluster name (formatted with `main` or `athena`)

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+
+def check_query_status(query_execution_id):
+    """Check in on the running query, back off if the job is running or queued
+
+    Returns:
+        [string]: The result of the query, this can be SUCCEEDED, FAILED, or CANCELLED.
+                  Reference http://bit.ly/2uuRtda
+    """
+    @backoff.on_predicate(backoff.fibo,
+                          lambda status: status in ('QUEUED', 'RUNNING'),
```
austinbyers: Use `set` instead of `tuple`:
```python
status in {'QUEUED', 'RUNNING'}
```

terraform/modules/tf_stream_alert_athena/main.tf
```diff
@@ -0,0 +1,137 @@
...
+      "s3:GetBucketLocation",
+      "s3:GetObject",
+      "s3:ListBucket",
+      "s3:ListBucketMultipartUploads",
+      "s3:ListMultipartUploadParts",
+      "s3:PutObject",
+    ]
+
+    resources = [
+      "arn:aws:s3:::aws-athena-query-results-*",
```
chunyong-lin: Just want to confirm that the resources has hyphen sign after `results` (``).
[AWS docs](http://docs.aws.amazon.com/athena/latest/ug/access.html) shows it wildcards everything after results which has wider range
```
"Resource": [
        "arn:aws:s3:::aws-athena-query-results*"
      ]
```

stream_alert/athena_partition_refresh/main.py
austinbyers: It's also not really true; `_load_config` returns `None` if there was an error in parsing. If the config file doesn't exist, it throws an uncaught IOError.

Since an error reading the config file is so damaging that the entire function has to halt its execution, this is a great example of an error which should bubble all the way up, breaking the function execution. In other words, instead of trying to catch `IOError` and/or a parsing error, just let config errors bubble up. The code is simpler and the error will be caught faster

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+
+import backoff
+import boto3
+
+logging.basicConfig()
+LEVEL = os.environ.get('LOGGER_LEVEL', 'INFO')
+LOGGER = logging.getLogger('StreamAlertAthena')
+LOGGER.setLevel(LEVEL.upper())
+
+ATHENA_CLIENT = None
```
austinbyers: For even faster execution, initialize the client at import time:

```python
ATHENA_CLIENT = boto3.client('athena', ...)
```

In order to get around failing tests, you just have to mock out `boto3` before importing the test file (for example, in `tests/__init__.py` )

stream_alert/athena_partition_refresh/main.py
austinbyers: Agreed - should be `error` or even `critical`

stream_alert/athena_partition_refresh/main.py
austinbyers: Agreed - a `namedtuple` might suffice.

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+        results_path=results_path
+    )
+    if isinstance(query_resp, dict) and not query_resp['ResultSet']['Rows']:
+        LOGGER.info('The \'streamalert\' database does not exist, please create it.')
+        return False
+
+    return True
+
+
+def check_table_exists(results_bucket, results_path, table_name):
```
austinbyers: It looks like this function is only used in the CLI (not in the Lambda function). If that's the case, consider factoring it out into the CLI code (unless you plan on using this later)

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+    if isinstance(query_resp, dict) and not query_resp['ResultSet']['Rows']:
+        LOGGER.info('The streamalert table \'%s\' does not exist, please create it.', table_name)
+        return False
+
+    return True
+
+
+def normal_partition_refresh(config, athena_results_bucket, athena_results_path):
+    normal_partition_config = config['lambda']['athena_partition_refresh_config']['partitioning']['normal']
+    for _, athena_table in normal_partition_config.iteritems():
```
austinbyers: Since you don't need the keys, you can simply do:

```python
for athena_table in normal_partition_config.itervalues():
    ...
```

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,206 @@
...
+        config['global']['account']['aws_account_id'],
+        config['global']['account']['region']
+    )
+    # Produces athena_partition_refresh/2017/01/01 keys
+    athena_results_path = 'athena_partition_refresh/{}'.format(
+        datetime.now().strftime('%Y/%m/%d'))
+
+    # The StreamAlert database needs to exist before we run queries.
+    if not check_database_exists(athena_results_bucket, athena_results_path):
+        return
```
austinbyers: Another example of a critical error that should probably bubble all the way up instead of failing silently.

stream_alert/athena_partition_refresh/main.py
jacknagz: Is that necessary in this use case? The possible statuses aren't dynamic.

stream_alert/athena_partition_refresh/main.py
jacknagz: I'm planning on using it later, but I also like the idea of keeping all of this code in the same library (it doesn't hurt if it's unused)

terraform/modules/tf_stream_alert_athena/main.tf
```diff
@@ -0,0 +1,137 @@
...
+  statement {
+    effect = "Allow"
+
+    actions = [
+      "s3:GetBucketLocation",
+      "s3:GetObject",
+      "s3:ListBucket",
+      "s3:ListBucketMultipartUploads",
+      "s3:ListMultipartUploadParts",
+      "s3:PutObject",
```
jacknagz: It's for writing Athena results

terraform/modules/tf_stream_alert_athena/main.tf
jacknagz: Either would work the same way if you look at the generated bucket name

stream_alert/athena_partition_refresh/main.py
jacknagz: they aren't called that in the folder structure

jacknagz: this is also a common paradigm in all of the code base

stream_alert/athena_partition_refresh/main.py
jacknagz: I like the clarity of `_backoff_handler`.  The implementation can also change over time

stream_alert/athena_partition_refresh/main.py
jacknagz: Note: It only shows up when you're debugging

stream_alert/athena_partition_refresh/main.py
jacknagz: It's executing in the `default` context to look for the `streamalert` db

stream_alert/athena_partition_refresh/main.py
jacknagz: > Why info vs. error? 

I'm changing it to `error`

> Could this happen if the right policies/permissions aren't applied?

Maybe? But the application would fail in that context anyway

stream_alert/athena_partition_refresh/main.py
jacknagz: TF can't create tables or databases, it happens with the CLI

stream_alert/athena_partition_refresh/main.py
jacknagz: I made it up.  `Normal` indicates a simple `MSCK REPAIR TABLE <table>` command, while `firehose` will indicate a more complex refresh command specific to Firehose partitions.

stream_alert/athena_partition_refresh/main.py
jacknagz: It depends on the situation.  If you are querying a lot of data and you know there should be results, you would handle more aggressively than a result that can be OK to not have results.

jacknagz: It is returned as an empty list in the response

stream_alert/athena_partition_refresh/main.py
jacknagz: The returned object has all this information:

http://boto3.readthedocs.io/en/latest/reference/services/athena.html#Athena.Client.get_query_execution

stream_alert_cli/runner.py
```diff
@@ -68,6 +70,77 @@ def cli_runner(options):
...
+                query=query,
+                database='streamalert',
+                results_bucket=athena_results_bucket,
+                results_path=athena_results_path
+            )
+
+            if create_table_result:
+                CONFIG['lambda']['athena_partition_refresh_config']['partitioning']['normal'][options.bucket] = 'alerts'
+                CONFIG.write()
+                LOGGER_CLI.info('alerts table successfully created!')
```
jacknagz: the library accounts for failures already

stream_alert/athena_partition_refresh/main.py
jacknagz: Not sure that'll work, it's throwing errors in CI.  I'm going to go with [this](https://github.com/airbnb/streamalert/pull/234#discussion_r130403033)

stream_alert/athena_partition_refresh/main.py
jacknagz: 👍 👍 👍

stream_alert/athena_partition_refresh/main.py
jacknagz: just changed it to `results_key`

stream_alert/athena_partition_refresh/main.py
mime-frame: Which is when you need context :) It's better if someone can reason about a debug statement without having to reference the code to understand it

stream_alert/athena_partition_refresh/main.py
mime-frame: to clarify - expound upon this *in the code comment* for future engineers or future self

stream_alert/athena_partition_refresh/main.py
mime-frame: Clarify this in the string

stream_alert/athena_partition_refresh/main.py
ryandeivert: late to the party, but just some additional insight: a namedtuple is immutable and thus wouldn't work from time tracking, etc. it could be recreated each time with an update value, but reusing the object itself and updating the stored values wouldn't work.

stream_alert/athena_partition_refresh/main.py
mime-frame: You have to imagine that the only time this issue will come about is usually for brand new users of StreamAlert or brand new users of historical search. Saying "please create it" doesn't cut it - try and have the string that's output to the user help them as much as possible in terms of next steps

stream_alert/athena_partition_refresh/main.py
mime-frame: Ok, then tell the user more explicitly how to create it - presumably they misunderstood the steps, or something broke, which is how they arrived at this error in the first place

stream_alert/athena_partition_refresh/main.py
mime-frame: I would suggest using different terminology to be clearer. perhaps `msck_repair_table` or `partition_refresh`

stream_alert/athena_partition_refresh/main.py
mime-frame: I think the point Austin and I were trying to get at is it'd be nice if a single type is returned vs. a bool or a dict. This could be done via a namedtuple, or by returning the object you reference.

stream_alert_cli/package.py
```diff
@@ -175,45 +186,47 @@ def _resolve_third_party(self, temp_package_path):
...
+            [boolean] Indicating a successful S3 upload
         """
         LOGGER_CLI.info('Uploading StreamAlert package to S3')
         client = boto3.client(
             's3', region_name=self.config['global']['account']['region'])
-        # the zip and the checksum file
+
         for package_file in (package_path, '{}.sha256'.format(package_path)):
             package_name = package_file.split('/')[-1]
             package_fh = open(package_file, 'r')
```
mime-frame: `open()` can throw an IOError, which could presumably happen due to file/path permission issues. Put inside the `try`?

stream_alert_cli/runner.py
```diff
@@ -68,6 +73,66 @@ def cli_runner(options):
...
+
+    if options.subcommand == 'init':
+        CONFIG.generate_athena()
+
+    elif options.subcommand == 'enable':
+        CONFIG.set_athena_lambda_enable()
+
+    elif options.subcommand == 'create-db':
+        if athena_client.check_database_exists():
+            LOGGER_CLI.info('The `streamalert` database exists, nothing to do')
```
mime-frame: seems more like a `warn` vs. an `info`?

stream_alert_cli/runner.py
```diff
@@ -68,6 +73,66 @@ def cli_runner(options):
...
+                     'ROW FORMAT SERDE \'org.openx.data.jsonserde.JsonSerDe\''
+                     'LOCATION \'s3://{bucket}/alerts/\''.format(bucket=options.bucket))
+
+            create_table_success, create_table_result = athena_client.run_athena_query(
+                query=query,
+                database='streamalert'
+            )
+
+            if create_table_success:
+                CONFIG['lambda']['athena_partition_refresh_config']['partitioning']['normal'][options.bucket] = 'alerts'
```
mime-frame: I'm still not a fan of the use of `normal` :(

terraform/modules/tf_stream_alert_athena/README.md
```diff
@@ -0,0 +1,83 @@
...
+  </tr>
+  <tr>
+    <td>lambda_memory</td>
+    <td>The memory allocation in MB for the lambda function</td>
+    <td>128MB</td>
+    <td>False</td>
+  </tr>
+  <tr>
+    <td>lambda_s3_bucket</td>
+    <td>The name of the S3 bucket to store lambda deployment packages</td>
```
mime-frame: stick with uppercase or lowercase for `lambda`

terraform/modules/tf_stream_alert_athena/main.tf
```diff
@@ -0,0 +1,167 @@
...
+    resources = [
+      "arn:aws:logs:*:*:*",
+    ]
+  }
+
+  statement {
+    effect = "Allow"
+
+    actions = [
+      "athena:*",
```
mime-frame: it needs to be `*`?

mime-frame:
Looking good! A few nits @jacknagz

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,241 @@
...
+        LOGGER.debug('[Backoff]: Completed after %d tries calling %s',
+                     details['tries'],
+                     details['target'])
+
+    def check_query_status(self, query_execution_id):
+        """Check in on the running query, back off if the job is running or queued
+
+        Returns:
+            [string]: The result of the query, this can be SUCCEEDED, FAILED, or CANCELLED.
+                      Reference http://bit.ly/2uuRtda
```
jacknagz: #security

stream_alert_cli/package.py
jacknagz: we haven't ran into this issue on any conf file loading yet, it seems v unlikely

terraform/modules/tf_stream_alert_athena/main.tf
jacknagz: http://docs.aws.amazon.com/athena/latest/ug/access.html

terraform/modules/tf_stream_alert_athena/README.md
jacknagz: LAMBDA?

stream_alert_cli/runner.py
jacknagz: since there's no action required by the user, I think `info` is fine

stream_alert_cli/runner.py
jacknagz: ok, any suggestion on adjusted wording?

terraform/modules/tf_stream_alert_athena/main.tf
mime-frame: @jacknagz - include as a code comment? :)

stream_alert/athena_partition_refresh/main.py
```diff
@@ -0,0 +1,241 @@
...
+            return True
+        else:
+            LOGGER.info('The streamalert table \'%s\' does not exist. '
+                        'For alert buckets, create it with the following command: \n'
+                        '$ python stream_alert_cli.py athena create-table '
+                        '--type alerts --bucket s3.bucket.id',
+                        table_name)
+            return False
+
+    def normal_partition_refresh(self):
```
mime-frame: Proposal @jacknagz: The intent of `MSCK REPAIR TABLE` is to identify new partitions that have been added to the table and subsequently incorporate them (metadata, etc). Therefore, I propose one of the function names:

```
hive_partition_refresh
msck_repair_hive_table
incorporate_new_hive_partitions
```
Or something similar. I personally like the first one.

Similarly, switch the use of `normal` throughout the code to `hive`. Examples:

```
athena_config['partitioning']['hive']
...
stream_alert_athena.hive_partition_refresh()
....
````

jacknagz: The only issue is that both are updating hive partitions, it's the way in which it happens that's different.  `MSCK REPAIR TABLE` vs `ADD PARTITION`

stream_alert/athena_partition_refresh/main.py
austinbyers: The `set` is just a faster lookup time. The difference is of course marginal

PR #239, opened 2017-08-03
[lambda][rule] logger centralizing and config loading improvements

ryandeivert:
to @jacknagz 
size: small

### Changes
* Rule processor `logging` object is now instantiated within the package's init so it can be imported throughout the package. Using the object is now as easy as importing with:
  * `from stream_alert.rule_processor import LOGGER`
* If any errors occur during the loading of the config, they will now provide more context as to what failed within a `ConfigError` exception.
  * For instance, we will report if the `schema`, `parser`, etc is missing from `logs` or if there ware invalid keys in `sources`.
* Updating config tests to be more concise/less code. 
* Small tweak to `sink` code that just passes a lambda function to the `json.dumps` call to serialize objects instead of creating a whole new function to do this.
  * Updating `sink.py` tests to increase coverage to 💯 
* Adding module docstring that was not present.
* Various `autopep8` nits.


stream_alert/rule_processor/config.py
```diff
@@ -20,7 +20,7 @@
 
 
 class ConfigError(Exception):
-    pass
+    """Exception class for config file errors"""
```
jacknagz: isn't this a syntax error? what happened to `pass`

stream_alert/rule_processor/config.py
```diff
@@ -35,22 +35,23 @@ def load_config(conf_dir='conf/'):
...
     rules to log fields.
     """
-    conf_files = {
-        'sources': 'sources.json',
-        'logs': 'logs.json'
-    }
-    config = {}
-    for desc, filename in conf_files.iteritems():
-        with open(os.path.join(conf_dir, filename)) as data:
+    conf_files = {'sources', 'logs'}
```
jacknagz: this can be a tuple - this allows us to preserve the order and it's not going to be used with other `set` methods

stream_alert/rule_processor/config.py
```diff
@@ -100,21 +102,18 @@ def load_env(context):
         context: The AWS Lambda context object.
 
     Returns:
-        {'lambda_region': 'region_name',
-         'account_id': <ACCOUNT_ID>,
-         'lambda_function_name': 'function_name',
-         'lambda_alias': 'qualifier'}
+        [dict] including the following items:
```
jacknagz: super nit: should be `(dict)` per [this](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/#type-annotations)

test/unit/stream_alert_rule_processor/test_helpers.py
```diff
@@ -0,0 +1,65 @@
...
+    """Create a fake context object using Mock"""
+    arn = 'arn:aws:lambda:{}:123456789012:function:{}:development'
+    context = Mock(invoked_function_arn=(arn.format(REGION, FUNCTION_NAME)),
+                   function_name=FUNCTION_NAME)
+
+    return context
+
+
+def _get_valid_config():
+    """Helper function to return a config that is valid
```
jacknagz: specify what type of config it's returning - the dict below is actually non representative of a full config

stream_alert/rule_processor/sink.py
```diff
@@ -70,7 +59,12 @@ def sink(self, alerts):
             }
         """
         for alert in alerts:
-            data = _json_dump(alert)
+            try:
+                data = json.dumps(alert, default=lambda o: o.__dict__)
+            except AttributeError as err:
+                LOGGER.error('An error occurred while dumping object to JSON: %s',
```
jacknagz: should we also print the `alert` in the case of failure

stream_alert/rule_processor/config.py
```diff
@@ -58,23 +59,30 @@ def validate_config(config):
...
+            raise ConfigError('The \'schema\' is missing for {}'.format(log))
+
+        if 'parser' not in attrs:
+            raise ConfigError('The \'parser\' is missing for {}'.format(log))
+
+    # check sources attributes
+    if not set(config['sources']).issubset({'kinesis', 's3', 'sns'}):
+        missing_sources = {'kinesis', 's3', 'sns'} - set(config['sources'])
+        raise ConfigError(
+            'Sources contains invalid key(s): %s',
```
ryandeivert: Adding some additional info here

stream_alert/rule_processor/config.py
ryandeivert: lol not fixing this now since this is a common thing across the code base right now and I don't want to introduce inconsistency. When we go to support the napoleon stuff we should update all of these in one fell swoop

Issue #241, opened 2017-08-07
KeyError when creating package for athena_partition_refresh


VVMichaelSawyer:
When executing `TF_LOG=DEBUG python ./stream_alert_cli.py terraform init` with the latest code, I receive the following error message:

```
StreamAlertCLI [INFO]: Creating package for athena_partition_refresh
Traceback (most recent call last):
  File "./stream_alert_cli.py", line 436, in <module>
    main()
  File "./stream_alert_cli.py", line 431, in main
    cli_runner(options)
  File "/Users/michaelsawyer/dev/streamalert/stream_alert_cli/runner.py", line 71, in cli_runner
    terraform_handler(options)
  File "/Users/michaelsawyer/dev/streamalert/stream_alert_cli/runner.py", line 255, in terraform_handler
    deploy(deploy_opts('all'))
  File "/Users/michaelsawyer/dev/streamalert/stream_alert_cli/runner.py", line 527, in deploy
    packages.append(_deploy_athena_partition_refresh())
  File "/Users/michaelsawyer/dev/streamalert/stream_alert_cli/runner.py", line 499, in _deploy_athena_partition_refresh
    athena_package.create_and_upload()
  File "/Users/michaelsawyer/dev/streamalert/stream_alert_cli/package.py", line 68, in create_and_upload
    if not self._resolve_third_party(temp_package_path):
  File "/Users/michaelsawyer/dev/streamalert/stream_alert_cli/package.py", line 183, in _resolve_third_party
    third_party_libs = self.config['lambda'][self.config_key]['third_party_libraries']
KeyError: 'athena_partition_refresh_config'
```

jacknagz:
Thanks @VVMichaelSawyer, I'll push a fix today.  In the meantime, please checkout a tagged release:

`$ git checkout tags/1.4.0`

PR #242, opened 2017-08-07
[cli] allow a list of lambda functions to be deployed

jacknagz:
to: @ryandeivert 
cc: @airbnb/streamalert-maintainers @VVMichaelSawyer 
resolves: #241 
size: small

Convert the `--processor` command line option type from a string to a list, which allows users to cherry pick which Lambda function is deployed.

This fixes a bug identified in #241 where the CLI would try to deploy a Lambda function that is optional (Athena Refresh).

Tested and verified by walking through the `$ python stream_alert_cli.py terraform init` workflow


ryandeivert:
LGTM 🚢

PR #243, opened 2017-08-08
Athena SQS Support

jacknagz:
**to**: @chunyong-lin (Terraform, Lambda), @ryandeivert (Lambda, CLI, Tests)
**cc**: @airbnb/streamalert-maintainers 
**size**: large

## Summary

Currently, the Athena Partition Refresh Lambda function is invoked directly (by s3) when new data is written to a bucket.  If the write volume is low, this isn't an issue, but if it's very high, we have a potential to refresh the same Athena table multiple times a second/minute.  This is unnecessary since the smallest data partition is in one minute intervals. 

## Change Overview

To fix this issue, I have added an SQS queue to be the recipient of s3 event notifications.  This also modifies how Lambda is invoked, since it now runs on a CloudWatch event timer of `10 minutes`.  

When the Lambda is now invoked, it fetches up to `50` messages from SQS, and deduplicates the s3 bucket ids.  Using this set of buckets, only the tables found in the `athena` configuration are refreshed:

```
# conf/lambda.json

  "athena_partition_refresh_config": {
    "current_version": 9,
    "log_level": "debug",
    "enabled": true,
    "handler": "stream_alert.athena_partition_refresh.main.handler",
    "memory": 128,
    "refresh_type": {
      "add_hive_partition": {},
      "repair_hive_table": {
        "my.s3.alerts.bucket": "alerts"
      }
    },
   ...
  },
```

After hive tables are refreshed, the messages are deleted off of the queue.

## Code Changes

* The Terraform Athena module has been modified to configure a SQS Queue, CloudWatch for periodic invocations of Lambda, and other necessary changes to wire everything together.
* Adds the ability to pass in a `log_level` to the Athena Lambda function via the CLI config.
* Refactor of the `athena_partition_refresh` Lambda function (maybe I should break this file out into multiple?)
* Refactor unit tests

## To Do (this PR)
- [x] Wrap up unit tests
- [x] Final AWS deployment and verify

## Next Steps (next PR)
- [ ] Write up initial Athena documentation
- [ ] Check deleted messages for success/failure
- [ ] SQS cloudwatch monitoring

### DIFF
--- a/stream_alert/athena_partition_refresh/main.py
+++ b/stream_alert/athena_partition_refresh/main.py
@@ -58,12 +58,12 @@ def _load_config():
     """Load the StreamAlert Athena configuration files
 
     Returns:
-        [dict] Configuration settings by file, includes two keys:
-            lambda: All lambda function settings
-            global: StreamAlert global settings
+        dict: Configuration settings by file, includes two keys:
+            lambda, All lambda function settings
+            global, StreamAlert global settings
 
     Raises:
-        [ConfigError] For invalid or missing configuration files.
+        ConfigError: For invalid or missing configuration files.
     """
     config_files = ('lambda', 'global')
     config = {}
@@ -107,14 +107,16 @@ class StreamAlertAthenaClient(object):
     DATABASE_STREAMALERT = 'streamalert'
     DEFAULT_S3_PREFIX = 'athena_partition_refresh'
 
-    def __init__(self, **kwargs):
+    def __init__(self, config, **kwargs):
         """Initialize the Boto3 Athena Client, and S3 results bucket/key
 
-        Keyword Args:
+        Args:
             config (CLIConfig): Loaded StreamAlert configuration
-            results_key_prefix (string): The S3 key prefix to store Athena results (optional)
+
+        Keyword Args:
+            results_key_prefix (string): The S3 key prefix to store Athena results
         """
-        self.config = kwargs['config']
+        self.config = config
         region = self.config['global']['account']['region']
         self.athena_client = boto3.client('athena', region_name=region)
 
@@ -281,8 +283,13 @@ class StreamAlertSQSClient(object):
     """
     QUEUENAME = 'streamalert_athena_data_bucket_notifications'
 
-    def __init__(self, **kwargs):
-        self.config = kwargs['config']
+    def __init__(self, config):
+        """Initialize the StreamAlertSQS Client
+
+        Args:
+            config (CLIConfig): Loaded StreamAlert configuration
+        """
+        self.config = config
         self.received_messages = []
         self.processed_messages = []
 
@@ -314,10 +321,9 @@ class StreamAlertSQSClient(object):
                 MaxNumberOfMessages=10
             )
 
-            if 'Messages' in polled_messages:
-                self.received_messages.extend(polled_messages['Messages'])
-            else:
+            if not 'Messages' in polled_messages:
                 return False
+            self.received_messages.extend(polled_messages['Messages'])
 
         _receive_messages()
         LOGGER.info('Received %s messages', len(self.received_messages))
@@ -348,7 +354,7 @@ class StreamAlertSQSClient(object):
         """Filter a list of unique s3 buckets from the received messages
 
         Returns:
-            (set) Unique s3 buckets derived from s3 event notifications
+            set: Unique s3 buckets derived from s3 event notifications
         """
         buckets = set()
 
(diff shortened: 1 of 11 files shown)


stream_alert/athena_partition_refresh/main.py
```diff
@@ -29,25 +29,92 @@
...
+    """
+    LOGGER.debug('[Backoff]: Completed after %d tries calling %s',
+                 details['tries'],
+                 details['target'])
+
+def _load_config():
+    """Load the StreamAlert Athena configuration files
+
+    Returns:
+        [dict] Configuration settings by file, includes two keys:
```
ryandeivert: if we're going to switch to the `(dict)` style I'd suggest doing it everywhere in this file so it's not forgotten about in the future

stream_alert/athena_partition_refresh/main.py
```diff
@@ -29,25 +29,92 @@
...
-        Keyword Arguments:
-            config [CLIConfig]: Loaded StreamAlert configuration (optional)
-            results_key_prefix [string]: The S3 key prefix to store Athena results (optional)
+        Keyword Args:
+            config (CLIConfig): Loaded StreamAlert configuration
+            results_key_prefix (string): The S3 key prefix to store Athena results (optional)
         """
-        # Load the config from files or accept it as an argument
-        self.config = kwargs.get('config') or self._load_config()
+        self.config = kwargs['config']
```
ryandeivert: I know you're a big fan of kwargs, but I worry when we use it in cases where the parameter is very much so required. For instance, there is nothing in the current interface to enforce that the initializer receives a `config=...` argument, even though this class would fail miserably without it. If someone wanted to use this class and instantiated `StreamAlertAthenaClient()` without any args, it wouldn't fail explicitly until the kwargs were accessed and wouldn't even result in a pylint error telling them something was wrong (http://pylint-messages.wikidot.com/messages:e1120).

kwargs are great for optionals and maybe this is a personal preference, but positionals help enforce required params and make the interface easier to use/understand without having to rely on an exception (when the kwargs['config'] fails) to know what's going on..

/end rant.. no need to change here, just some insight on why I tend to not use them for functions with only a handful of args.

That said - you can still call a function with the `arg=value` syntax even without using kwargs, if that's any consolation.

stream_alert/athena_partition_refresh/main.py
```diff
@@ -204,37 +233,178 @@ def check_table_exists(self, table_name):
...
+                              on_backoff=_backoff_handler,
+                              on_success=_success_handler)
+        def _receive_messages():
+            polled_messages = self.sqs_client.receive_message(
+                QueueUrl=self.athena_sqs_url,
+                MaxNumberOfMessages=10
+            )
+
+            if 'Messages' in polled_messages:
+                self.received_messages.extend(polled_messages['Messages'])
```
ryandeivert: suggest getting rid of the `else` by doing `if not 'Messages' in polled_messages: return False`

stream_alert/athena_partition_refresh/main.py
```diff
@@ -204,37 +233,178 @@ def check_table_exists(self, table_name):
...
+
+            loaded_message = json.loads(message['Body'])
+
+            if 'Records' not in loaded_message:
+                LOGGER.error('Missing `Records` key, trying next SQS message')
+                continue
+
+            for record in loaded_message['Records']:
+                if 's3' not in record:
+                    LOGGER.info('Skipping non s3 bucket notification message')
```
ryandeivert: nit: `non-s3`

stream_alert/athena_partition_refresh/main.py
jacknagz: I can agree with the required vs optional viewpoint for kwargs.  I just modified the initializer to accept a `config` first argument, and an arbitrary amount of kwargs after that

mime-frame: Nice! Good call out, agreed. Function declaration defines the contract that you want others to follow.

terraform/modules/tf_stream_alert_athena/variables.tf
```diff
@@ -37,3 +37,8 @@ variable "athena_data_buckets" {
 variable "prefix" {
   type = "string"
 }
+
+variable "refresh_interval" {
+  type    = "string"
+  default = "rate(10 minutes)"
```
chunyong-lin: Want to confirm that is it 10 or 1 minutes for cloudwatch event refresh interval?  In PR description, it mentions `This also modifies how Lambda is invoked, since it now runs on a CloudWatch event timer of 1 minute`.

jacknagz: good catch, I actually just made this change and forgot to add it to the PR body

jacknagz:
FYI: Going to merge this and open a second PR for documentation - which is a WIP

PR #244, opened 2017-08-08
[lambda][rule] payload refactor in preparation for better logging/metrics

ryandeivert:
#### to:
@jacknagz for cli & lambda updates
@austinbyers for unit tests
size: large

### Changes:
* The `StreamClassifier` class will now be instantiated once and cached on the `StreamAlert` class for reuse.
  * Previously this was instantiated within a loop with a new config each time.
  * By caching on the `StreamAlert` class, we now dynamically load logs for a given service/entity on the fly.
* Subclassing the `StreamPayload` class to encapsulate service-specific operations. For instance, the pre-parser methods have been migrated to within each of these respective subclasses. The subclasses have a method of `pre_parse` that conforms to the interface of returning a generator. Any additional subclasses added in the future should do the same.
  * The main reason for returning a generator is that s3 files are multiple records within one file (unlike SNS or Kinesis, where we really only process one record at a time). By having all `pre_parse` methods return a generator, we can duck type the call in a `for ... in payload.pre_parse()...` and just know it will work.
* Simplifying the `_convert_type` method by removing some used params, etc and adding some additional context to error logging if there is a type mismatch.
* Each payload also has a `service` method that is abstract and must be implemented on subclasses. By making this an abstractmethod, we reduce that chance of a simple `service` property being omitted on a future implementation of the class.
* Various unit tests to go with all of the changes.


stream_alert/rule_processor/handler.py
```diff
@@ -61,22 +62,14 @@ def run(self, event):
...
 
         for record in event.get('Records', []):
-            payload = StreamPayload(raw_record=record)
             classifier = StreamClassifier(config=config)
 
-            # If the kinesis stream, s3 bucket, or sns topic is not in our config,
-            # go onto the next record
-            if not classifier.map_source(payload):
+            # Create the StreamPayload to use for encapsulating parsed info
+            payload = load_stream_payload(service, entity, raw_record)
```
jacknagz: where are these args coming from?

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+
+        Used mainly when S3 is used as a source, due to looping over files
+        downloadd from S3 events versus all records being readily available
+        from a Kinesis stream.
+
+        Args:
+            new_record [string]: A new raw record to be parsed
+        """
+        self.pre_parsed_record = new_record
+        self.type = None
```
jacknagz: these initializations should be encapsulated into a method, it's the same thing as in `__init__`

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+
+        Returns:
+            [generator] Yields instances of `self` back to the caller with the
+                proper `pre_parsed_record` set. Conforms to the interface of
+                returning a generator, providing the ability to support
+                multi-record like this (s3).
+        """
+        s3_file = self._get_object()
+        line_num, processed_size = 0, 0
+        for line_num, data in S3Payload._read_s3_file(s3_file):
```
jacknagz: why not just call `self._read_s3_file(s3_file)`?

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+
+            self.refresh_record(data)
+            yield self
+
+            # Only do the extra calculations below if debug logging is enabled
+            if not LOGGER.isEnabledFor(log_level_debug):
+                continue
+
+            # Add the current data to the total processed size, +1 to account for line
+            # feed
```
jacknagz: condense this comment or split everything after the comma onto the second line

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+
+        return downloaded_s3_object
+
+    def _get_object(self):
+        """Given an S3 record, download and parse the data.
+
+        Returns:
+            [string] Path to the downloaded s3 object.
+        """
+        unquoted = lambda data: unquote(data).decode('utf8')
```
jacknagz: can you add a quick note explaining this `lambda`

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+        region = self.raw_record['awsRegion']
+        bucket = unquoted(self.raw_record['s3']['bucket']['name'])
+        key = unquoted(self.raw_record['s3']['object']['key'])
+        self.s3_object_size = int(self.raw_record['s3']['object']['size'])
+
+        LOGGER.debug('Pre-parsing record from S3. Bucket: %s, Key: %s, Size: %d',
+                     bucket, key, self.s3_object_size)
+
+        downloaded_s3_object = self._download_object(region, bucket, key)
```
jacknagz: maybe just `return self._download_object(region, bucket, key)`, that var isn't used elsewhere

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+
+        return downloaded_s3_object
+
+    @staticmethod
+    def _read_s3_file(s3_object):
```
jacknagz: rename to `_read_downloaded_s3_object()`?

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+            client = boto3.client('s3', region_name=region)
+            start_time = time.time()
+            client.download_fileobj(bucket, key, data)
+
+        total_time = time.time() - start_time
+        LOGGER.info('Completed download in %s seconds', round(total_time, 2))
+
+        return downloaded_s3_object
+
+    def _get_object(self):
```
jacknagz: what if we standardize on 's3 object' wording for these methods? potentially rename to `_get_s3_object()`

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+                    approx_record_count = self.s3_object_size / avg_record_size
+                    LOGGER.debug(
+                        'Processed %s records out of an approximate total of %s '
+                        '(average record size: %s bytes, total size: %s bytes)',
+                        line_num,
+                        approx_record_count,
+                        avg_record_size,
+                        self.s3_object_size)
+
+    def _download_object(self, region, bucket, key):
```
jacknagz: See other comments, but this might be renamed to `_downloaded_s3_object()`

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+        """Download an object from S3.
+
+        Verifies the S3 object is less than or equal to 128MB, and
+        stores into a temp file.  Lambda can only execute for a
+        maximum of 300 seconds, and the file to download
+        greatly impacts that time.
+
+        Args:
+            region [string]: AWS region to use for boto client instance.
+            bucket (string): S3 bucket to download object from.
```
jacknagz: `(string)`? or `[string]` :trollface:

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+        """
+        size_kb = self.s3_object_size / 1024.0
+        size_mb = size_kb / 1024.0
+        if size_mb > 128:
+            raise S3ObjectSizeError('S3 object to download is above 128MB')
+
+        LOGGER.debug('/tmp directory contents:%s ', os.listdir('/tmp'))
+        LOGGER.debug(os.popen('df -h /tmp | tail -1').read().strip())
+
+        if size_mb:
```
jacknagz: we can condense this into one line:

`display_size = '{}MB'.format(size_mb) if size_mb else '{}KB'.format(size_kb)`

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+class S3ObjectSizeError(Exception):
+    """Exception indicating the S3 object is too large to process"""
+    pass
+
+
+class S3Payload(StreamPayload):
+    """S3Payload class"""
+    s3_object_size = 0
+
+    def service(self):
```
jacknagz: let's condense into a single line (for this and other classes)

stream_alert/rule_processor/classifier.py
```diff
@@ -89,26 +89,25 @@ def refresh_record(self, new_record):
...
-    def __init__(self, **kwargs):
-        self.config = kwargs['config']
+
+    def __init__(self, config):
+        self._config = config
         self._entity_log_sources = []
 
-    def map_source(self, payload):
+    @staticmethod
+    def extract_service_and_entity(raw_record):
```
jacknagz: the `map` term is used throughout this method, I'd keep it in the method name unless we change the references below

stream_alert/rule_processor/classifier.py
```diff
@@ -118,29 +117,50 @@ def map_source(self, payload):
...
+        Args:
+            payload: A StreamPayload object
+
+        Returns:
+            [boolean] True if the entity's log sources loaded properly
+        """
+        # Clear the list from any previous runs
+        del self._entity_log_sources[:]
+
+        # get all logs for the configured service/entity (s3, kinesis, or sns)
```
jacknagz: nit: capitalization

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,328 @@
...
+                   'kinesis': KinesisPayload}
+
+    if service not in payload_map:
+        LOGGER.error('Service payload not supported: %s', service)
+        return
+
+    return payload_map[service](raw_record=raw_record, entity=entity)
+
+
+class StreamPayload(object):
```
jacknagz: doesn't this class still exist in `classifier.py`?

test/unit/stream_alert_rule_processor/test_classifier.py
```diff
@@ -13,148 +13,270 @@
...
     def setup_class(cls):
         """Setup the class before any methods"""
-        cls.env = {
-            'lambda_region': 'us-east-1',
-            'account_id': '123456789012',
-            'lambda_function_name': 'test_kinesis_stream',
-            'lambda_alias': 'production'
-        }
+        config = load_config('test/unit/conf')
+        cls.classifier = sa_classifier.StreamClassifier(config)
```
austinbyers: Because the classifier stores generated state (namely `self._entity_log_sources`), I would suggest creating the StreamClassifier for every test (i.e. in `setup`, not `setup_class`). This will help avoid weird bugs with persistent state between different tests

test/unit/stream_alert_rule_processor/test_classifier.py
```diff
@@ -13,148 +13,270 @@
...
 
     @classmethod
     def teardown_class(cls):
         """Teardown the class after all methods"""
-        cls.env = None
-
-    def setup(self):
-        """Setup before each method"""
-        self.config = load_config('test/unit/conf')
+        cls.classifier = None
```
austinbyers: Unless `nose` requires this method, it's not necessary

test/unit/stream_alert_rule_processor/test_classifier.py
```diff
@@ -13,148 +13,270 @@
...
 
     def teardown(self):
         """Teardown after each method"""
-        self.config = None
+        del self.classifier._entity_log_sources[:]
+        sa_classifier.SUPPORT_MULTIPLE_SCHEMA_MATCHING = False
```
austinbyers: Once the Classifier is built in `setup`, this method won't be necessary

test/unit/stream_alert_rule_processor/test_classifier.py
```diff
@@ -13,148 +13,270 @@
...
+        """StreamClassifier - Convert Type, Invalid Int"""
+        payload = {'key_01': 'NotInt'}
+        schema = {'key_01': 'integer'}
+
+        self.classifier._convert_type(payload, schema)
+
+        log_mock.assert_called_with(
+            'Invalid schema. Value for key [%s] is not an int: %s',
+            'key_01',
+            'NotInt')
```
austinbyers: 🎉  for log mock!

test/unit/stream_alert_rule_processor/test_classifier.py
```diff
@@ -255,46 +410,41 @@ def test_classify_record_kinesis_json_optional(self):
...
-        data = self.pre_parse_kinesis(payload)
-        classifier.classify_record(payload, data)
+        service, entity = 'kinesis', 'test_kinesis_stream'
+        raw_record = _make_kinesis_raw_record(entity, kinesis_data)
+        payload = self._prepare_and_classify_payload(service, entity, raw_record)
 
         # valid record test
         assert_equal(payload.valid, True)
-        assert_equal(type(payload.records[0]), dict)
+        assert_is_instance(payload.records[0], dict)
```
austinbyers: It is OK to check the type directly if you want; `isinstance` is a looser check because it will work for subclasses as well

```python
d = collections.OrderedDict()
type(d) == dict  # False
isinstance(d, dict)  # True
```

stream_alert/rule_processor/handler.py
ryandeivert: Squashed some commits so you should see the entire context now.

stream_alert/rule_processor/handler.py
```diff
@@ -91,60 +84,35 @@ def get_alerts(self):
...
-            data [string]: Pre parsed data string from a raw_event to be parsed
         """
-        classifier.classify_record(payload, data)
-        if not payload.valid:
-            if self.env['lambda_alias'] != 'development':
-                LOGGER.error('Record does not match any defined schemas: %s\n%s', payload, data)
-            self._failed_record_count += 1
-            return
+        for record in payload.pre_parse():
+            self.classifier.classify_record(record)
```
ryandeivert: again - squashed commits

test/unit/stream_alert_rule_processor/test_handler.py
```diff
@@ -0,0 +1,231 @@
...
+    def teardown_class(cls):
+        """Teardown the class after all methods"""
+        cls.__sa_handler = None
+
+    def teardown(self):
+        """Teardown the class after each methods"""
+        del self.__sa_handler._alerts[:]
+        self.__sa_handler._failed_record_count = 0
+        self.__sa_handler.env['lambda_alias'] = 'development'
+        self.__sa_handler.enable_alert_processor = False
```
austinbyers: Again, all of this teardown suggests that the `StreamAlert` should be built for each test (in `setup`), not for the whole class (`setup_class`). The extra initialization cost is minimal compared to the benefit of guaranteeing a consistent environment for testing.

TL;DR avoid `setup_class` unless you absolutely need state saved across multiple test runs.

stream_alert/rule_processor/payload.py
ryandeivert: This is within the s3payload class now and is implicit that it's downloading an s3 object - going to leave it as is

test/unit/stream_alert_rule_processor/test_main.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_imports():
+    """Rule Processor Main - Test Imports"""
+    imports = {'matchers.sample',
+               'rules.community.cloudtrail.cloudtrail_critical_api',
+               'rules.community.cloudtrail.cloudtrail_root_account',
+               'rules.community.cloudtrail.cloudtrail_put_bucket_acl',
+               'rules.community.cloudtrail.cloudtrail_put_object_acl'}
+
+
```
austinbyers: Nit: there should never be two blank lines within a function

test/unit/stream_alert_rule_processor/test_parsers.py
```diff
@@ -77,18 +154,17 @@ def test_inspec(self):
...
                                            schema=schema,
                                            options=options)
 
         assert_equal(len(parsed_result), 2)
-        inspec_keys = (u'impact', u'code', u'tags', u'source_location', u'refs',
-                       u'title', u'results', u'id', u'desc')
-        assert_equal(sorted((inspec_keys)), sorted(parsed_result[0].keys()))
+        inspec_keys = ['impact', 'code', 'tags', 'source_location', 'refs',
+                       'title', 'results', 'id', 'desc']
+        assert_items_equal(parsed_result[0].keys(), inspec_keys)
```
austinbyers: `assert_equal` should automatically invoke `assert_items_equal` for sets of things

stream_alert/rule_processor/payload.py
ryandeivert: see: https://github.com/airbnb/streamalert/pull/244#discussion_r132000335

ryandeivert:
@jacknagz pushed a commit to address your feedback - @austinbyers  upcoming commit with your requests

stream_alert/rule_processor/handler.py
```diff
@@ -56,30 +70,47 @@ def run(self, event):
...
+            if not (service and entity):
+                LOGGER.error('Skipping record: %s', raw_record)
+                continue
 
-        for record in event.get('Records', []):
-            payload = StreamPayload(raw_record=record)
-            classifier = StreamClassifier(config=config)
+            # If the payload's service and entity are found in the config and
+            # contains logs then load the sources for this log
+            if not self.classifier.load_sources(service, entity):
```
mime-frame: this statement seems to match the first part of your code comment, but not the `and contains logs` portion?

stream_alert/rule_processor/handler.py
```diff
@@ -91,60 +122,35 @@ def get_alerts(self):
...
-        if not payload.valid:
-            if self.env['lambda_alias'] != 'development':
-                LOGGER.error('Record does not match any defined schemas: %s\n%s', payload, data)
-            self._failed_record_count += 1
-            return
+        for record in payload.pre_parse():
+            self.classifier.classify_record(record)
+            if not record.valid:
+                if self.env['lambda_alias'] != 'development':
+                    LOGGER.error('Record does not match any defined schemas: %s\n%s',
```
mime-frame: why wouldn't we want to log this in the development environment?

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,310 @@
...
+    """Container class for the StreamAlert payload object.
+
+    Attributes:
+        entity: The instance of the sending service. Can be either a
```
mime-frame: `instance`? Like an instance of an object or a string literal?

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,310 @@
...
+                proper `pre_parsed_record` set. Conforming to the interface of
+                returning a generator provides the ability to support multi-record
+                payloads, such as those similar to S3.
+        """
+
+    def _refresh_record(self, new_record):
+        """Replace the currently loaded record with a new one.
+
+        Used mainly when S3 is used as a source, due to looping over files
+        downloadd from S3 events versus all records being readily available
```
mime-frame: `download`

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,310 @@
...
+            self._refresh_record(data)
+            yield self
+
+            # Only do the extra calculations below if debug logging is enabled
+            if not LOGGER.isEnabledFor(log_level_debug):
+                continue
+
+            # Add the current data to the total processed size
+            # +1 to account for line feed
+            processed_size += (len(data) + 1)
```
mime-frame: `len()` doesn't account for LF?

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,310 @@
...
+                        line_num,
+                        approx_record_count,
+                        avg_record_size,
+                        self.s3_object_size)
+
+    def _download_object(self, region, bucket, key):
+        """Download an object from S3.
+
+        Verifies the S3 object is less than or equal to 128MB, and
+        stores into a temp file.  Lambda can only execute for a
```
mime-frame: stores it into

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,310 @@
...
+                yield num, line.rstrip()
+        else:
+            for num, line in enumerate(open(s3_object, 'r'), start=1):
+                yield num, line.rstrip()
+
+        # AWS Lambda apparently does not reallocate disk space when files are
+        # removed using os.remove(), so we must truncate them before removal
+        open(s3_object, 'w')
+
+        # Remove the file
```
mime-frame: duplicate comment as the one above it

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,310 @@
...
+        os.remove(s3_object)
+        if not os.path.exists(s3_object):
+            LOGGER.debug('Removed temp file: %s', s3_object)
```
mime-frame: `Removed temp S3 file:`

stream_alert/rule_processor/payload.py
```diff
@@ -0,0 +1,310 @@
...
+        else:
+            LOGGER.error('Failed to remove temp file: %s', s3_object)
```
mime-frame: `Failed to remove temp S3 file:`

test/unit/stream_alert_rule_processor/test_payload.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+@with_setup(setup=None, teardown=teardown_s3)
+@patch('stream_alert.rule_processor.payload.S3Payload._get_object')
+@patch('logging.Logger.debug')
+@patch('stream_alert.rule_processor.payload.S3Payload._read_downloaded_s3_object')
+def test_pre_parse_s3_debug(s3_mock, log_mock, _):
+    """S3Payload - Pre Parse, Debug On"""
+    # Cache the logger level
+    lvl = LOGGER.getEffectiveLevel()
```
mime-frame: poor variable name, maybe `log_level`

test/unit/stream_alert_rule_processor/test_classifier.py
ryandeivert: Couldn't I also just explicitly check that it was an OrderedDict using isinstance? Type checking with `type` is discouraged in practice so I figured I would also move away from it in tests

test/unit/stream_alert_rule_processor/test_parsers.py
ryandeivert: This is inaccurate since `assert_equal` cares about order here and we do not. Here is the result of using `assert_equal`:
```
First differing element 0:
city
name

- [u'city', u'age', u'name', u'state']
+ ['name', 'age', 'city', 'state']
```

ryandeivert:
@austinbyers push a commit that addresses your feedback. @mime-frame is up next!

stream_alert/rule_processor/handler.py
ryandeivert: Because it will get lost in the shuffle when printed in-line with testing output. The variable below (`self._failed_record_count += 1`) is what is used by the testing framework to handle error reporting for invalid records/logs. Basically, a similar messages gets printed when running in a development env, just not at this time and at the end of testing instead.

The alternative to this would be having a list of errors that get appended to, and printing them all at the end.. **However** this would be bad practice, since these errors could potentially never get printed if the functions times out or an exception occurs down the line. It's safer to log them as they occur in production.

stream_alert/rule_processor/payload.py
ryandeivert: @mime-frame `len()` would, if the LF were there.. the LF is not returned as part of the line, since the func `_read_downloaded_s3_object` does a `line.rstrip()` on the line. We could return the line, get the size, then call `rstrip()` on it, but it felt safer to keep string manipulation within the function doing the reading.

stream_alert/rule_processor/handler.py
ryandeivert: updated comment

ryandeivert:
@mime-frame , @austinbyers , @jacknagz please take another look when you have a second. I've pushed changes to address all of your feedback

test/unit/stream_alert_rule_processor/test_classifier.py
austinbyers: My understanding is that `type` is discouraged because it is too strict - it will not match subclasses, which you will often want to do.

However, in a test, you may want to know exactly what type something is. Using `type` vs `isinstance` is saying "I assert this is a dict" vs. "I assert this is *like* a dict (a subclass)." Whichever one is more useful to you; I just wanted to make sure we were aware of the difference. I don't think there's any subclassing going on in right now where this would matter.

test/unit/stream_alert_rule_processor/test_parsers.py
austinbyers: Ah, nice! It turns out raw `unittest` does not have such a method; I was thinking of `assertListEqual`. Learn something new every day!

test/unit/stream_alert_rule_processor/test_classifier.py
```diff
@@ -13,148 +13,263 @@
...
 
 
-class TestStreamPayload(object):
-    """Test class for StreamPayload"""
+class TestStreamClassifier(object):
+    """Test class for StreamClassifier"""
+
     def __init__(self):
-        self.config = {}
+        self.classifier = None
```
austinbyers: Does `pylint` complain if you leave this method out? No need to initialize it here since it's set in `setup`.

mime-frame:
Nice work @ryandeivert , I'm happy with my portion of the review. It's left to @jacknagz and @austinbyers

test/unit/stream_alert_rule_processor/test_handler.py
```diff
@@ -0,0 +1,222 @@
...
+    _get_mock_context,
+    _get_valid_event
+)
+
+
+class TestStreamAlert(object):
+    """Test class for StreamAlert class"""
+
+    def __init__(self):
+        self.__sa_handler = None
```
austinbyers: Any particular reason this is a double underscore when other private variables are single underscore? Doesn't really matter either way, just be consistent

test/unit/stream_alert_rule_processor/test_payload.py
```diff
@@ -0,0 +1,303 @@
...
+    records = ['_first_line_test_' * 10,
+               '_second_line_test_' * 10]
+
+    s3_mock.side_effect = [((100, records[0]), (200, records[1]))]
+
+    raw_record = _make_s3_raw_record('unit_bucket_name', 'unit_key_name')
+    s3_payload = load_stream_payload('s3', 'unit_key_name', raw_record)
+    S3Payload.s3_object_size = 350
+
+    _ = [_ for _ in s3_payload.pre_parse()]
```
austinbyers: How about

```python
for _ in s3_payload.pre_parse():
    pass
```

test/unit/stream_alert_rule_processor/test_classifier.py
ryandeivert: That's a great point. And you're right, in the case there isn't any subclassing but that's a good thing to keep in mind.

test/unit/stream_alert_rule_processor/test_parsers.py
ryandeivert: 🎉   This was actually a new thing I discovered recently too.. It's nice with cases of unordered objects.

stream_alert/rule_processor/handler.py
```diff
@@ -56,30 +70,46 @@ def run(self, event):
...
+            service, entity = self.classifier.extract_service_and_entity(raw_record)
+            if not service:
+                LOGGER.error('No valid service found in payload\'s raw record')
+
+            if not entity:
+                LOGGER.error(
+                    'Unable to map entity from payload\'s raw record for service %s',
+                    service)
+
+            if not (service and entity):
```
jacknagz: Why not just `continue` in each `service`/`entity` check above?  You could also print the raw record there too and then just omit this block.

ryandeivert: Updated this block to be a bit less confusing, nice catch.

PR #247, opened 2017-08-09
[metrics] initial metrics support

ryandeivert:
to @austinbyers , @jacknagz 

## Changes
* Adding a metrics class to support publishing metrics to cloudwatch
* Including a few metrics for now, which will be added to later. Those include:
  * Total records being processed
  * Number of logs that failed to parse
  * Number of alerts triggered
  * S3 file download time (for latency visibility)
  * Total number of records from S3 files
* Updating tests to cover metrics.

### DIFF
--- a/stream_alert/rule_processor/metrics.py
+++ b/stream_alert/rule_processor/metrics.py
@@ -1,0 +1,122 @@
+'''
+Copyright 2017-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+'''
+import json
+
+from datetime import datetime
+
+import boto3
+
+from botocore.exceptions import ClientError
+
+from stream_alert.rule_processor import LOGGER
+
+
+class Metrics(object):
+    """Class to hold Rule Processor metric name and unit constants
+    This basically acts as an enum, allowing for the use of dot notation for
+    accessing properties and avoids doing dict lookups a ton.
+    """
+
+    def __init__(self, region):
+        self.boto_cloudwatch = boto3.client('cloudwatch', region_name=region)
+
+    class Name(object):
+        """Constant metric names used in the rule processor"""
+        FAILED_PARSES = 'RuleProcessorFailedParses'
+        S3_DOWNLOAD_TIME = 'RuleProcessorS3DownloadTime'
+        TOTAL_RECORDS = 'RuleProcessorTotalRecords'
+        TOTAL_S3_RECORDS = 'RuleProcessorTotalS3Records'
+        TRIGGERED_ALERTS = 'RuleProcessorTriggeredAlerts'
+
+    class Unit(object):
+        """Unit names for metrics. These are taken from the boto3 CloudWatch page"""
+        SECONDS = 'Seconds'
+        MICROSECONDS = 'Microseconds'
+        MILLISECONDS = 'Milliseconds'
+        BYTES = 'Bytes'
+        KILOBYTES = 'Kilobytes'
+        MEGABYTES = 'Megabytes'
+        GIGABYTES = 'Gigabytes'
+        TERABYTES = 'Terabytes'
+        BITS = 'Bits'
+        KILOBITS = 'Kilobits'
+        MEGABITS = 'Megabits'
+        GIGABITS = 'Gigabits'
+        TERABITS = 'Terabits'
+        PERCENT = 'Percent'
+        COUNT = 'Count'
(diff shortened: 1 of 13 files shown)


stream_alert/rule_processor/metrics.py
```diff
@@ -0,0 +1,122 @@
...
+from botocore.exceptions import ClientError
+
+from stream_alert.rule_processor import LOGGER
+
+
+class Metrics(object):
+    """Class to hold Rule Processor metric name and unit constants
+    This basically acts as an enum, allowing for the use of dot notation for
+    accessing properties and avoids doing dict lookups a ton.
+    """
```
austinbyers: Python2 has a backported [`enum` package](https://pypi.python.org/pypi/enum), would that be simpler?

test/unit/stream_alert_rule_processor/test_metrics.py
```diff
@@ -0,0 +1,83 @@
...
+        # Add ClientError side_effect to mock
+        boto_mock.client.return_value.put_metric_data.side_effect = ClientError(
+            err_response, 'operation')
+
+        Metrics(REGION)._put_metric([{'test': 'info'}])
+
+        log_mock.assert_called_with(
+            'Failed to send metric to CloudWatch. Error: %s\nMetric data:\n%s',
+            err_response,
+            '[\n  {\n    "test": "info"\n  }\n]')
```
austinbyers: `moto` does have support for `put_metric_data` (though not `get_metric_data`); if you use moto elsewhere you can use `moto` to verify that `boto3` accepts the arguments as valid

stream_alert/rule_processor/metrics.py
```diff
@@ -0,0 +1,122 @@
...
+        TERABYTES_PER_SECOND = 'Terabytes/Second'
+        BITS_PER_SECOND = 'Bits/Second'
+        KILOBITS_PER_SECOND = 'Kilobits/Second'
+        MEGABITS_PER_SECOND = 'Megabits/Second'
+        GIGABITS_PER_SECOND = 'Gigabits/Second'
+        TERABITS_PER_SECOND = 'Terabits/Second'
+        COUNT_PER_SECOND = 'Count/Second'
+        NONE = 'None'
+
+    def put_metric_data(self, metric_name, value, unit):
```
austinbyers: You can actually publish multiple metrics at once ([example](https://github.com/airbnb/binaryalert/blob/master/lambda_functions/analyzer/analyzer_aws_lib.py#L102-L124)), which I strongly recommend to reduce the number of API calls

stream_alert/rule_processor/metrics.py
ryandeivert: I debated using enum but that solution actually seemed more complicated for such a simple use case. Can you see any real benefits to it?

stream_alert/rule_processor/metrics.py
ryandeivert: I actually modeled some of this off of your code, but metric publishing happens at a number of spots throughout code. I have an idea for centralizing them, would you mind sitting down with me to discuss?

stream_alert/rule_processor/metrics.py
```diff
@@ -0,0 +1,122 @@
...
+from datetime import datetime
+
+import boto3
+
+from botocore.exceptions import ClientError
+
+from stream_alert.rule_processor import LOGGER
+
+
+class Metrics(object):
```
jacknagz: Would it make sense to encapsulate this module into a new package under `stream_alert`? Like a helpers library that could be shared across all Lambdas. We could even move config and loading and other duplicated code into that as well.

stream_alert/rule_processor/metrics.py
austinbyers: Nah, this is fine. I just saw `enum` in a comment and was making sure we knew about it.

stream_alert/rule_processor/metrics.py
austinbyers: For example, `FailedParses` and `TriggeredAlerts` are sent right after each other; these could be consolidated. 

My strategy was to publish all the metrics at the end, unless there are important metrics that might be lost because an exception was raised somewhere.

Happy to chat offline!

mime-frame:
+1 for doing metrics on the # of logs for each log type (carbonblack, osquery, ...), so we can detect/reason about dips

ryandeivert:
@jacknagz PTAL at recent commit - this moves the metrics stuff to a `shared` sub package and updates the rule processor lambda packaging. Please make sure I didn't miss anything.

stream_alert_cli/package.py
```diff
@@ -235,6 +235,7 @@ class RuleProcessorPackage(LambdaPackage):
     """Deployment package class for the StreamAlert Rule Processor function"""
     package_folders = {
         'stream_alert/rule_processor',
+        'stream_alert/shared',
```
jacknagz: this should go in all lambda function packages

jacknagz:
👍  after my 1 comment

stream_alert_cli/package.py
ryandeivert: Metrics are only used in the rule processor right now. Why add it to others? 🤔

ryandeivert:
@austinbyers I added a new commit that adds support for caching metrics. for some reason it doesn't look like it's showing up yet, but PTAL when you get a second

Issue #236, opened 2017-08-01
Improvement: pylint compliance & enforcement


ryandeivert:
The StreamAlert project needs to be thoroughly pylint'd and then pylint should be enforced through travis CI. A `.pylintrc` needs to be added to aid in this.

The following line needs to be added to `script` section of travis after SA is pylint compliant:
`"find . -name '*.py' -exec pylint '{}' +"`

ryandeivert:
This feature should also encompass local testing through the CLI. Meaning, all rules, etc should be pylint'd once we have enforcement across the rest of the code base.

PR #250, opened 2017-08-10
Pylint cleanup

austinbyers:
Adds `.pylintrc` configuration and pylint enforcement in CI. I also used PyCharm IDE to find PEP8 violations not caught by pylint (e.g. number of blank lines).

Resolves: #236 

## Renaming
- `stream_alert_cli.py` was renamed to `manage.py` because `pylint` couldn't distinguish between `stream_alert_cli.py` (the file) and `stream_alert_cli` (the module). Both were individually importable; now they have separate names
- `test\` is renamed to `tests\`, because `test` is a standard Python library. This is also more consistent with other folder names (`docs`, `rules`, etc)

## Style Notes
- Imports are in 3 distinct groups: standard libraries, `pip` packages, and local module imports. `from collections import namedtuple` is a standard import and shouldn't be in a separate group
- Docstrings should always be triple-quoted (PEP8) and have been standardized according to [this style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/#type-annotations)

## Tested
- CI checks unit tests, integration tests, docs, and pylint
- Searched for `"test/"` and `"stream_alert_cli.py"` to verify that nothing was missed during the renaming
- Tested the help strings for every option in the CLI (since the argparser was refactored)
- Deployed in an empty test account and tested both the rule and alert processor on empty input

## Reviewers
to: @jacknagz CLI
to: @ryandeivert Lambda code and unit tests


stream_alert_cli_module/test.py
```diff
@@ -436,23 +429,23 @@ def setup_outputs(self, alert):
...
+                output_name = '/'.join([service, descriptor])
                 creds = {'service_key': '247b97499078a015cc6c586bc0a92de6'}
                 helpers.put_mock_creds(output_name, creds, self.secrets_bucket,
                                        'us-east-1', self.kms_alias)
 
                 # Set the patched urlopen.getcode return value to 200
                 url_mock.return_value.getcode.return_value = 200
             elif service == 'phantom':
-                output_name = ('/').join([service, descriptor])
+                output_name = '/'.join([service, descriptor])
```
austinbyers: This can be changed to simple string formatting (as in line 459)

tests/unit/stream_alert_cli/test_config.py
```diff
@@ -12,14 +12,13 @@
...
-
+"""
 import json
 
 from nose.tools import assert_equal
 
-from unit.helpers.base import mock_open
-from stream_alert_cli.config import CLIConfig
+from tests.unit.helpers.base import mock_open
+from stream_alert_cli_module.config import CLIConfig
```
austinbyers: Note to self: wrong import order

stream_alert_cli_module/runner.py
```diff
@@ -604,6 +596,8 @@ def configure_output(options):
     props = output.get_user_defined_properties()
 
     for name, prop in props.iteritems():
+        # TODO: Don't use protected method here
+        # pylint: disable=protected-access
```
ryandeivert: We could create a new copy of the namedtuple here, but I would be 100% okay with adding _replace to the ignored protected warnings.

stream_alert_cli_module/test.py
```diff
@@ -168,7 +161,7 @@ def check_keys(self, rule_name, test_record):
         Args:
             rule_name: The name of the rule being tested. This is passed in
                 here strictly for reporting any errors with key checks.
-            test_record [dict]: Test record metadata dict
+            test_record: [dict] Test record metadata dict
```
jacknagz: Can we stick to [this style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/#type-annotations) going forward?

stream_alert/alert_processor/output_base.py
```diff
@@ -249,17 +248,17 @@ def output_cred_name(self, descriptor):
...
 
-    def format_output_config(self, config, props):
+    def format_output_config(self, service_config, values):
         """Add this descriptor to the list of descriptor this service
            If the service doesn't exist, a new entry is added to an empty list
 
         Args:
-            config [dict]: Loaded configuration as a dictionary
-            props [OrderedDict]: Contains various OutputProperty items
+            service_config [dict]: Loaded configuration as a dictionary
```
jacknagz: again, let's please use [this style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/#type-annotations)

stream_alert/rule_processor/classifier.py
```diff
@@ -69,7 +69,8 @@ def load_sources(self, service, entity):
         """Load the sources for this payload.
 
         Args:
-            payload: A StreamPayload object
+            service: [String] Source service
```
jacknagz: again, docstrings style

stream_alert_cli_module/terraform_generate.py
```diff
@@ -331,7 +331,8 @@ def generate_kinesis(cluster_name, cluster_dict, config):
     """
     logging_bucket = '{}.streamalert.s3-logging'.format(
         config['global']['account']['prefix'])
-    firehose_suffix = config['clusters'][cluster_name]['modules']['kinesis']['firehose']['s3_bucket_suffix']
+    firehose_suffix = config['clusters'][cluster_name]['modules']['kinesis']['firehose'][
```
jacknagz: can we standardize on the `\` across lines instead? I believe we are doing that elsewhere in the codebase

austinbyers: The standard is to never use `\`; the only such cases were these long dict accesses and I switched all of them to this new format

stream_alert_cli_module/runner.py
austinbyers: I would prefer a new copy of the namedtuple rather than accessing an internal method in the namedtuple class. I just wanted to verify that this was expected behavior

austinbyers: Actually, I'll just leave it as is. It's very convenient, and people do seem to use it. Not sure why Python wouldn't expose it as a normal method

stream_alert_cli_module/test.py
austinbyers: Sure thing! It turns out that specific style also works better for at least my IDE: it started flagging misnamed arguments, etc

austinbyers:
@jacknagz @ryandeivert PTAL
The docstrings have been standardized and I renamed the CLI as requested. The PR comment was updated with the changes.

I'm working on a test deploy now

jacknagz:
🚢  after verified deploy

austinbyers:
Update: test deploy was successful. I tested the alert_processor and rule_processor and everything seems to be in working order. @ryandeivert can test in his test account if needed, but I think we should be good

jacknagz:
@austinbyers can you provide more details on what was tested (which event sources did you use, etc)

austinbyers:
@jacknagz I'm not sure how to test specific event sources, I just verified that the Lambda functions executed individually.

austinbyers:
@ryandeivert @jacknagz rebased and up-to-date. 95% of the changes are docstrings, comments, and import orders. Test deploy is pending

austinbyers:
@jacknagz @ryandeivert Test deploy finished successfully; CloudTrail alerts are rolling into S3

ryandeivert:
Thanks again @austinbyers !

Issue #256, opened 2017-08-11
Can't set prefix: "Invalid prefix type, must be string"


austinbyers:
The CLI can't be used to set a prefix:

```
$ python stream_alert_cli.py configure prefix my_name
StreamAlertCLI [ERROR]: Invalid prefix type, must be string

$ python stream_alert_cli.py configure prefix "my_name"
StreamAlertCLI [ERROR]: Invalid prefix type, must be string
```

After some blame-digging, we found that #239 accidentally [inverted the type-checking logic](https://github.com/airbnb/streamalert/commit/4094ba073705b4d443a2184c92d9d87859710087#diff-18657a05bb7b57424976b92346d9e512R162)

PR #257, opened 2017-08-11
Fix 'stream_alert_cli.py configure prefix'

austinbyers:
Fixes `stream_alert_cli.py configure prefix` to error only if the argument is *not* a string (and to check the config value)

Resolves: #256 

to: @ryandeivert 

### Before
```
$ python stream_alert_cli.py configure prefix hello_world
StreamAlertCLI [ERROR]: Invalid prefix type, must be string
```
and `conf/global.json` is not modified.

### After
```
$ python stream_alert_cli.py configure prefix hello_world
```
and `conf/global.json` is updated with the new prefix


jacknagz:
🚢

PR #259, opened 2017-08-12
[metrics] cloudwatch permissions and metric dimensions

ryandeivert:
to @jacknagz , @austinbyers 

## Changes
* Fixing bug where the rule processor did not have permissions to publish metric data to CloudWatch
* Adding dimensions to CloudWatch metrics to allow for better attribution to cluster/function
* Adding logging environment variables to both alert processor and rule processor.
  * The rule processor has been updated to safely call `logging.setLevel`, accounting for poorly set log levels by users.
* Updating unit tests

## Testing steps
* With a test AWS account, I used `python stream_alert_cli.py configure` to add my account ID and prefix. I then enabled the `cloudtrail` module in the cluster conf by adding:

```
    "cloudtrail": {
      "enabled": true
    }
```
* I then updated the `sources.json` file to include my kinesis stream name of `rdmetrics_prod_stream_alert_kinesis`.
* Updated `outputs.json` to include `aws-s3` output:
```
  "aws-s3": {
    "alerts": "rdmetrics.streamalerts"
  }
```
* Updated the `cloudtrail_critical_api.py` rule to be super noisy by just returning `True`.
  * Also updated `outputs` list in rule to include `aws-s3:alerts` to send to `rdmetrics.streamalerts` bucket.
* Ran `python stream_alert_cli.py terraform init`
* Watched metrics pile up 💯 

## To Do
- [ ] Update alert processor and athena refresh function to have centralized logging and institute safety check for `logging.setLevel`
- [x] Add more test coverage for changes where necessary.


conf/clusters/prod.json
```diff
@@ -20,11 +23,13 @@
     "stream_alert": {
       "alert_processor": {
         "current_version": "$LATEST",
+        "log_level": "info",
```
jacknagz: Please update docs on usage

stream_alert/shared/metrics.py
```diff
@@ -23,24 +24,36 @@
 
 from stream_alert.shared import LOGGER
 
+CLUSTER = os.environ.get('CLUSTER', 'unknown_cluster')
```
jacknagz: 👏

stream_alert/shared/metrics.py
```diff
@@ -23,24 +24,36 @@
...
 
 class Metrics(object):
-    """Class to hold Rule Processor metric name and unit constants
+    """Class to hold metric names and unit constants
     This basically acts as an enum, allowing for the use of dot notation for
     accessing properties and avoids doing dict lookups a ton.
     """
 
-    def __init__(self, region):
+    def __init__(self, function, region):
```
jacknagz: can `function` be renamed to something more specific like`lambda_function`?

terraform/modules/tf_stream_alert/iam.tf
```diff
@@ -42,6 +42,29 @@ data "aws_iam_policy_document" "rule_processor_invoke_alert_proc" {
   }
 }
 
+// IAM Role Policy: Allow the Rule Processor to publish CloudWatch metrics
+resource "aws_iam_role_policy" "streamalert_rule_processor_cloudwatch_put_metric_data" {
```
jacknagz: Please add this policy to the `alert_processor` as well

austinbyers: Instead of creating a separate policy for metrics, all of the Lambda functions could inherit the same base policy with permission to publish logs and metrics (e.g. https://github.com/airbnb/binaryalert/blob/master/terraform/lambda_iam.tf#L3-L17)

ryandeivert: @austinbyers I think this is a great idea, but for now we'll just get this change in and then think about a larger refactor of shared perms across the 3 functions

ryandeivert:
@jacknagz updated PR body and made requested changes. PTAL

docs/source/lambda.rst
```diff
@@ -62,10 +64,11 @@ Per Cluster Lambda Settings
 ===================  ========  ===========
 Key                  Required  Description
 -------------------  --------  -----------
-``timeout``          ``true``  The time (in seconds) the Lambda function is allowed to process an incoming record. The timeout can be set to any value between 1 and 300 seconds.
-``memory``           ``true``  The amount of memory allocated for the Lambda function execution.
-``current_version``  ``true``  The most current published version of the Lambda function.
-``outputs``          ``true``  A collection of S3 bucket IDs or AWS Lambda function names to configure as valid outputs.  By default, ``aws-s3`` should contain the bucket created by the ``stream_alert`` module: ``prefix.cluster.streamalerts``.  Optionally, if the alert processor needs to invoke other Lambda functions from within your AWS account, specify a list of function names.
+``timeout``          ``Yes``   The time (in seconds) the Lambda function is allowed to process an incoming record. The timeout can be set to any value between 1 and 300 seconds.
```
jacknagz: Is the `true` vs `Yes` present elsewhere in the docs? can you make the change if it is?

ryandeivert: @jacknagz I was following the updates you made in the athena stuff. I can check else where but :damn:

ryandeivert: @jacknagz  updated a few others spots in docs. we should get better about consistency when changes are first made instead of having to go back retroactively and update

jacknagz:
Thanks for making the change on the docs @ryandeivert.  You have done the community a great service 😆

PR #260, opened 2017-08-13
[cli][bug] fixing bug with rollback function cli

ryandeivert:
to @jacknagz 

## Fixes
* Fixing bug in `runner.py` that was preventing rollbacks from happening through the cli.

Traceback:
```
  File "stream_alert_cli.py", line 437, in <module>
    main()
  File "stream_alert_cli.py", line 432, in main
    cli_runner(options)
  File "/Users/<USERNAME>/repos/streamalert/stream_alert_cli/runner.py", line 63, in cli_runner
    lambda_handler(options)
  File "/Users/<USERNAME>/repos/streamalert/stream_alert_cli/runner.py", line 149, in lambda_handler
    rollback(options)
  File "/Users/<USERNAME>/repos/streamalert/stream_alert_cli/runner.py", line 423, in rollback
    current_vers = stream_alert_key[lambda_function]['current_version']
KeyError: "['rule']_processor"
```

## Other change
* Adding `athena_partition_refresh` function rollback functionality
* Fixing bug where multiple `--processor` flag could result in duplicate deployments:
  * For instance: `python stream_alert_cli.py lambda deploy --processor all --processor rule --processor alert` would result in `_deploy_alert_processor` and `_deploy_rule_processor` both being called twice.

### DIFF
--- a/stream_alert_cli/runner.py
+++ b/stream_alert_cli/runner.py
@@ -412,10 +412,14 @@ def rollback(options):
         Only rollsback if published version is greater than 1
     """
     clusters = CONFIG.clusters()
-    if options.processor == 'all':
-        lambda_functions = {'rule_processor', 'alert_processor'}
+
+    if 'all' in options.processor:
+        lambda_functions = {'rule_processor', 'alert_processor', 'athena_partition_refresh'}
     else:
-        lambda_functions = {'{}_processor'.format(options.processor)}
+        lambda_functions = {'{}_processor'.format(proc) for proc in options.processor
+                            if proc != 'athena'}
+        if 'athena' in options.processor:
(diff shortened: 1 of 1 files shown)


stream_alert_cli/runner.py
```diff
@@ -412,10 +412,14 @@ def rollback(options):
         Only rollsback if published version is greater than 1
     """
     clusters = CONFIG.clusters()
-    if options.processor == 'all':
-        lambda_functions = {'rule_processor', 'alert_processor'}
+
+    if 'all' in options.processor:
+        lambda_functions = {'rule_processor', 'alert_processor', 'athena_partition_refresh'}
```
jacknagz: this can be a tuple (we aren't using set methods or adding/removing elements)

PR #265, opened 2017-08-16
[lambda][athena] athena minor tf updates and tf destroy fix

ryandeivert:
to @jacknagz 
size: small

## Changes
* Updating backoff logging handler to use name of function instead of function object which does not look nice in logs + updating tests
* Bug fix in cli terraform destroy command
* Fixing metrics variable name enforcing some dependencies that could cause a race condition

### DIFF
--- a/stream_alert_cli/runner.py
+++ b/stream_alert_cli/runner.py
@@ -262,9 +262,14 @@ def terraform_handler(options):
 
     elif options.subcommand == 'destroy':
         if options.target:
-            target = options.target
-            targets = ['module.{}_{}'.format(target, cluster)
-                       for cluster in CONFIG.clusters()]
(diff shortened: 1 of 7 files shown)


stream_alert_cli/runner.py
```diff
@@ -262,9 +262,14 @@ def terraform_handler(options):
 
     elif options.subcommand == 'destroy':
         if options.target:
-            target = options.target
-            targets = ['module.{}_{}'.format(target, cluster)
-                       for cluster in CONFIG.clusters()]
+            targets = []
+            for target in options.target:
+                if target == 'athena':
```
jacknagz: can we add a comment here about global vs cluster based resources

PR #269, opened 2017-08-16
[shared][stats] adding timing decorator to log stats on methods

ryandeivert:
to @austinbyers, @jacknagz 

## Changes
* Adding new `stats` module within shared to hold a timing decorator (and potentially more internal statistic related stuff in the future).
* classifier.classify_record, classifier._process_log_schemas now provide timing info
* JSON parser's parse_records now provides timing info
* This is enabled in debug mode only

## Example output
`StreamAlertShared [DEBUG]: (module) stream_alert.rule_processor.classifier (method) _process_log_schemas (time): 0.7188ms`

## Caveat
* The decorator has no way to determine the **class** a function being timed is tied to. I even investigated hacky ways to try to determine this using the `inspect` package but came up short. I'm open to suggestions, but logging the module may be enough for our purposes for now.

### DIFF
--- a/stream_alert/rule_processor/classifier.py
+++ b/stream_alert/rule_processor/classifier.py
@@ -18,7 +18,7 @@ from collections import namedtuple, OrderedDict
 
 from stream_alert.rule_processor import LOGGER
 from stream_alert.rule_processor.parsers import get_parser
-from stream_alert.shared.stats import timeme
+from stream_alert.shared.stats import time_me
 
 # Set the below to True when we want to support matching on multiple schemas
 # and then log_patterns will be used as a fall back for key/value matching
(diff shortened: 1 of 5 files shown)


stream_alert/rule_processor/classifier.py
```diff
@@ -17,6 +17,7 @@
 
 from stream_alert.rule_processor import LOGGER
 from stream_alert.rule_processor.parsers import get_parser
+from stream_alert.shared.stats import timeme
```
austinbyers: I keep reading this as `ti_meme`. As silly as that is, I would recommend renaming to `time_me`to be consistent

mime-frame:
lol @ ti_meme. Can we not use @timer ?

mime-frame:
image

jacknagz:
+1 to TI Meme name change

stream_alert/rule_processor/classifier.py
ryandeivert: LOLOLOL

ryandeivert:
Changed the name as requested 😂

PR #271, opened 2017-08-17
[cli] support athena refresh interval option

jacknagz:
to @chunyong-lin 

* Add an option to change the Athena Lambda function invocation rate via the StreamAlert CLI config
* Update tests
* Update docs


chunyong-lin:
hi @jacknagz what's the current athena refresh interval? Is it same thing with [this interval setting](https://github.com/airbnb/streamalert/blob/master/terraform/modules/tf_stream_alert_athena/variables.tf#L43)?

jacknagz:
@chunyong-lin yep

PR #273, opened 2017-08-19
[multiproc] EXPERIMENTAL: rule processor multiprocessing proof of concept

ryandeivert:

## Changes
* Creating a pool of workers to process records on parallel. This is mostly advantageous for compound files where there is more than one record (such as S3 files).
* The current amount of processes that will be spawned is set to 6 but can be changed.
* Each worker will get an equal portion of the total logs to process (or as close to it as possible)
* This enhacement helps most with compound files that contain many
  records (like those from s3 for instance)
* Each worker appends alerts to the shared alerts list that can be returned to the caller
* This has been tested locally with reading a file of 19993 lines/records.

## Lambda Experimentation
* In a test AWS account, I create a function that has a very simple purpose (putting objects into an S3 bucket) and used it to experiment with various worker/task operations.
* Below are the high level results of those tests with a function configured with 512 MB of memory an a timeout of 5 min.

#### Total of 100 operations (5 tests):
- 1 worker doing 100 operations:
  - Test run 1 in 1.84 sec
  - Test run 2 in 1.62 sec
  - Test run 3 in 1.86 sec
  - Test run 4 in 1.76 sec
  - Test run 5 in 1.76 sec
  - Average time: 1.77 
  - Lambda function ran in 8.85 sec (total execution)

- 10 workers doing 10 operations each: 
  - Test run 1 in 1.64 sec
  - Test run 2 in 2.03 sec
  - Test run 3 in 1.78 sec
  - Test run 4 in 2.07 sec
  - Test run 5 in 1.63 sec
  - Average time: 1.83 sec (0.06 sec diminishment)
  - Lambda function ran in 9.15 sec (total execution)

#### Total of 1000 operations (5 tests):
- 1 worker doing 1000 operations:
  - Test run 1 in 19.42 sec
  - Test run 2 in 21.17 sec
  - Test run 3 in 22.90 sec
  - Test run 4 in 17.67 sec
  - Test run 5 in 18.88 sec
  - Average time: 20.00 sec
  - Lambda function ran in 100.05 sec (total execution)

- 10 workers doing 100 operations each:
  - Test run 1 in 8.46 sec
  - Test run 2 in 8.68 sec
  - Test run 3 in 8.79 sec
  - Test run 4 in 8.63 sec
  - Test run 5 in 8.50 sec
  - Average time: 8.61 sec (11.49 sec improvement)
  - Lambda function ran in 43.06 sec (total execution)

#### Total of 5000 operations (3 tests due to timeout exceeding):
- 1 workers doing 5000 operations:
  - Test run 1 in 98.49 sec
  - Test run 2 in 98.71 sec
  - Test run 3 in 95.76 sec
  - Average time: 97.65 sec
  - Lambda function ran in 292.95 sec (total execution)

- 10 workers doing 500 operations each:
  - Test run 1 in 41.45 sec
  - Test run 2 in 41.92 sec
  - Test run 3 in 42.63 sec
  - Average time: 42.00 sec (55.65 sec improvement)
  - Lambda function ran in 126.01 sec (total execution)

## Performance improvements
* Profiling locally, with code in master processing the 19993 records:
```
Start time 1503108573.75
StreamAlertShared [INFO]: [stats.timed] (module) stream_alert.rule_processor.handler (method) _process_alerts (time): 26575.2771ms
End time 1503108600.33
Ran in 26.5756158829 seconds
```
* Notice: **Ran in 26.5756158829 seconds**
* Profiling locally with this multiprocessing code:
```
Start time 1503108143.73
StreamAlert [DEBUG]: [handler._run_batches] Running worker #1 for 1-3333 of 19993 records
StreamAlert [DEBUG]: [handler._run_batches] Running worker #2 for 3334-6666 of 19993 records
StreamAlert [DEBUG]: [handler._run_batches] Running worker #3 for 6667-9999 of 19993 records
StreamAlert [DEBUG]: [handler._run_batches] Running worker #4 for 10000-13332 of 19993 records
StreamAlert [DEBUG]: [handler._run_batches] Running worker #5 for 13333-16665 of 19993 records
StreamAlert [DEBUG]: [handler._run_batches] Running worker #6 for 16666-19993 of 19993 records
StreamAlertShared [INFO]: [stats.timed] (module) stream_alert.rule_processor.handler (method) _process_alerts (time): 6664.8021ms
StreamAlertShared [INFO]: [stats.timed] (module) stream_alert.rule_processor.handler (method) _process_alerts (time): 6761.4338ms
StreamAlertShared [INFO]: [stats.timed] (module) stream_alert.rule_processor.handler (method) _process_alerts (time): 6767.2961ms
StreamAlertShared [INFO]: [stats.timed] (module) stream_alert.rule_processor.handler (method) _process_alerts (time): 6834.5852ms
StreamAlertShared [INFO]: [stats.timed] (module) stream_alert.rule_processor.handler (method) _process_alerts (time): 6849.8139ms
StreamAlertShared [INFO]: [stats.timed] (module) stream_alert.rule_processor.handler (method) _process_alerts (time): 6880.0972ms
StreamAlert [INFO]: [handler.run] 1 alerts triggered
End time 1503108151.08
Ran in 7.35581588745 seconds
```
* Notice: **Ran in 7.35581588745 seconds**

#### More than 3.5x improvement in processing speed

### DIFF
--- a/stream_alert/rule_processor/handler.py
+++ b/stream_alert/rule_processor/handler.py
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 '''
 import json
+import multiprocessing as multiproc
 
 from logging import DEBUG as log_level_debug
 
@@ -24,6 +25,7 @@ from stream_alert.rule_processor.payload import load_stream_payload
 from stream_alert.rule_processor.rules_engine import StreamRules
 from stream_alert.rule_processor.sink import StreamSink
 from stream_alert.shared.metrics import Metrics
+PROC_MANAGER = multiproc.Manager()
 
 
 class StreamAlert(object):
@@ -55,7 +57,7 @@ class StreamAlert(object):
         self.metrics = Metrics('RuleProcessor', self.env['lambda_region'])
         self.enable_alert_processor = enable_alert_processor
         self._failed_record_count = 0
-        self._alerts = []
+        self._alerts = PROC_MANAGER.list()
 
     def run(self, event):
         """StreamAlert Lambda function handler.
@@ -124,7 +126,7 @@ class StreamAlert(object):
         # Check if debugging logging is on before json dumping alerts since
         # this can be time consuming if there are a lot of alerts
         if self._alerts and LOGGER.isEnabledFor(log_level_debug):
-            LOGGER.debug('Alerts:\n%s', json.dumps(self._alerts, indent=2))
+            LOGGER.debug('Alerts:\n%s', json.dumps(self.get_alerts(), indent=2))
 
         # Send any cached metrics to CloudWatch before returning
         self.metrics.send_metrics()
@@ -137,7 +139,7 @@ class StreamAlert(object):
         Returns:
             [list] list of alerts as dictionaries
         """
-        return self._alerts
+        return self._alerts._getvalue()
 
     def _process_alerts(self, payload):
         """Process records for alerts and send them to the correct places
--- a/stream_alert/rule_processor/handler.py
+++ b/stream_alert/rule_processor/handler.py
@@ -18,7 +18,7 @@ import math
 
 from copy import copy
 from logging import DEBUG as log_level_debug
-from multiprocessing import Process, Manager, Value
+import multiprocessing as multiproc
 
 from stream_alert.rule_processor import LOGGER
 from stream_alert.rule_processor.config import load_config, load_env
@@ -29,8 +29,7 @@ from stream_alert.rule_processor.sink import StreamSink
 from stream_alert.shared.metrics import Metrics
 from stream_alert.shared.stats import time_me
 
-NUM_WORKERS = 6
-PROC_MANAGER = Manager()
+PROC_MANAGER = multiproc.Manager()
 
 
 class StreamAlert(object):
@@ -61,7 +60,7 @@ class StreamAlert(object):
 
         self.metrics = Metrics('RuleProcessor', self.env['lambda_region'])
         self.enable_alert_processor = enable_alert_processor
-        self._failed_record_count = Value('I', 0)
+        self._failed_record_count = multiproc.Value('I', 0)
         self._alerts = PROC_MANAGER.list()
 
     def run(self, event):
@@ -151,6 +150,7 @@ class StreamAlert(object):
         Returns:
             [list] list of alerts as dictionaries
         """
+        # Call through to return the underlying list value from the ListProxy
         return self._alerts._getvalue()
 
     def _run_batches(self, payload, workers):
@@ -162,16 +162,21 @@ class StreamAlert(object):
             workers [list<mutliprocessing.Process>]: The complete list of
                 multiprocessing jobs to append this job to
         """
-        # Get all of the pre-parsed records for this payload
-        payload_records = [record for record in payload.pre_parse()]
+        # Get all of the pre-parsed records from the generator for this payload
+        payload_records = list(payload.pre_parse())
+
+        # Default to spawning 2 processes per available CPU
+        num_workers = multiproc.cpu_count() * 2
+
+        LOGGER.debug('Creating a maximum of %d workers for processing', num_workers)
 
         # Get the interval that we should segment the list of alerts on and
         # arrange them into sublists of this size. Default to minimum of 1
-        interval = int(math.ceil(len(payload_records)/float(NUM_WORKERS))) or 1
-        record_sublists = [payload_records[index:interval+index]
-                           for index in range(0, len(payload_records), interval)]
+        interval = int(math.ceil(len(payload_records)/float(num_workers))) or 1
+        record_groups = [payload_records[index:interval+index]
+                         for index in range(0, len(payload_records), interval)]
 
-        for worker_index, sublist in enumerate(record_sublists, start=1):
+        for worker_index, sublist in enumerate(record_groups, start=1):
             offset = (worker_index - 1) * interval
             LOGGER.debug('Running worker #%d for %d-%d of %d records',
                          worker_index, offset + 1, offset + len(sublist),
--- a/stream_alert/rule_processor/handler.py
+++ b/stream_alert/rule_processor/handler.py
@@ -129,7 +129,7 @@ class StreamAlert(object):
             self._failed_record_count,
             Metrics.Unit.COUNT)
 
-        LOGGER.debug('%s alerts triggered', len(self._alerts))
+        LOGGER.info('%s alerts triggered', len(self._alerts))
 
         self.metrics.add_metric(
             Metrics.Name.TRIGGERED_ALERTS, len(
@@ -210,7 +210,7 @@ class StreamAlert(object):
             record_alerts = StreamRules.process(record)
 
             LOGGER.debug('Processed %d valid record(s) that resulted in %d alert(s).',
-                         len(payload.records),
+                         len(record.records),
                          len(record_alerts))
 
             if not record_alerts:
--- a/stream_alert/rule_processor/handler.py
+++ b/stream_alert/rule_processor/handler.py
@@ -60,7 +60,7 @@ class StreamAlert(object):
 
         self.metrics = Metrics('RuleProcessor', self.env['lambda_region'])
         self.enable_alert_processor = enable_alert_processor
-        self._failed_record_count = multiproc.Value('I', 0)
+        self._unmatched_record_count = multiproc.Value('i', 0)
         self._alerts = PROC_MANAGER.list()
 
     def run(self, event):
@@ -121,11 +121,12 @@ class StreamAlert(object):
         for worker in workers:
             worker.join()
 
-        LOGGER.debug('Invalid record count: %d', self._failed_record_count.value)
+        if self._unmatched_record_count.value:
+            LOGGER.debug('Invalid record count: %d', self._unmatched_record_count.value)
 
         self.metrics.add_metric(
             Metrics.Name.FAILED_PARSES,
-            self._failed_record_count.value,
+            self._unmatched_record_count.value,
             Metrics.Unit.COUNT)
 
         LOGGER.info('%s alerts triggered', len(self._alerts))
@@ -142,7 +143,7 @@ class StreamAlert(object):
         # Send any cached metrics to CloudWatch before returning
         self.metrics.send_metrics()
 
-        return self._failed_record_count.value == 0
+        return self._unmatched_record_count.value == 0
 
     def get_alerts(self):
         """Public method to return alerts from class. Useful for testing.
@@ -209,8 +210,8 @@ class StreamAlert(object):
                     LOGGER.error('Record does not match any defined schemas: %s\n%s',
                                  record, record.pre_parsed_record)
 
-                with self._failed_record_count.get_lock():
-                    self._failed_record_count.value += 1
+                with self._unmatched_record_count.get_lock():
+                    self._unmatched_record_count.value += 1
                 continue
 
             LOGGER.debug('Classified and Parsed Payload: <Valid: %s, Log Source: %s, Entity: %s>',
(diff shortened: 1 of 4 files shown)


stream_alert/rule_processor/handler.py
```diff
@@ -54,8 +61,8 @@ def __init__(self, context, enable_alert_processor=True):
 
         self.metrics = Metrics('RuleProcessor', self.env['lambda_region'])
         self.enable_alert_processor = enable_alert_processor
-        self._failed_record_count = 0
-        self._alerts = []
+        self._failed_record_count = Value('I', 0)
```
austinbyers: This is an example of why I personally prefer to import only top-level modules when possible. In other words, I prefer to just `import multiprocessing` instead of `from multiprocessing import Value`. Here, it's not clear where `Value` comes from; it's a very generic name. `multiprocessing.Value` would be more clear IMHO.

However, I know this is more consistent with the rest of the StreamAlert code. Ultimately up to you; just something to think about

stream_alert/rule_processor/handler.py
```diff
@@ -24,6 +27,10 @@
 from stream_alert.rule_processor.rules_engine import StreamRules
 from stream_alert.rule_processor.sink import StreamSink
 from stream_alert.shared.metrics import Metrics
+from stream_alert.shared.stats import time_me
+
+NUM_WORKERS = 6
```
austinbyers: Some experimentation might be necessary to know what size works best in Lambda. The standard when creating multiprocessing pools is to default to `multiprocessing.cpu_count()`. I wonder what Lambda would return for `multiprocessing.cpu_count()`?

stream_alert/rule_processor/handler.py
```diff
@@ -124,35 +138,71 @@ def run(self, event):
...
+        return self._failed_record_count.value == 0
 
     def get_alerts(self):
         """Public method to return alerts from class. Useful for testing.
 
         Returns:
             [list] list of alerts as dictionaries
         """
-        return self._alerts
+        return self._alerts._getvalue()
```
austinbyers: `pylint` will complain about accessing an internal member here; is this required?

stream_alert/rule_processor/handler.py
```diff
@@ -124,35 +138,71 @@ def run(self, event):
...
         Args:
             payload [StreamPayload]: StreamAlert payload object being processed
+            workers [list<mutliprocessing.Process>]: The complete list of
+                multiprocessing jobs to append this job to
+        """
+        # Get all of the pre-parsed records for this payload
+        payload_records = [record for record in payload.pre_parse()]
+
+        # Get the interval that we should segment the list of alerts on and
+        # arrange them into sublists of this size. Default to minimum of 1
```
austinbyers: If you create a `multiprocessing.Pool`, I don't think you need to explicitly segment the list; it should handle it automatically.

stream_alert/rule_processor/handler.py
```diff
@@ -124,35 +138,71 @@ def run(self, event):
...
+            LOGGER.debug('Running worker #%d for %d-%d of %d records',
+                         worker_index, offset + 1, offset + len(sublist),
+                         len(payload_records))
+            # Create a copy of the classifier for each worker being created
+            # to avoid conflicts with the workers sharing the cached classifier
+            worker_classifier = copy(self.classifier)
+            processor = Process(
+                target=self._process_alerts, args=(worker_classifier, sublist))
+            workers.append(processor)
+            processor.start()
```
austinbyers: The segmentation logic is definitely complex enough to need its own unit tests

stream_alert/rule_processor/handler.py
ryandeivert: This is a good tip and I was debating trying to use a pool. I will experiment with this

stream_alert/rule_processor/handler.py
ryandeivert: I initially had it importing the whole package, but you know how I hate annoyingly long package names, so I switched it :).  I can add the namespace back for clarity

ryandeivert:
@austinbyers multiprocessing does in fact work in Lambda - I will update the PR to describe some tests I ran

stream_alert/rule_processor/handler.py
ryandeivert: @austinbyers I read up on this yesterday - and it's the only way to get the actual underlying shared list back from the shared list. The `Manager().list()` returns a `ListProxy` class that cannot be serialized to json, etc and we must get the actual underlying list. I also read that although this is a protected method, it is documented as safe to use. I will add a comment above this as to why it's used and we can add a pylint exception for it.

ryandeivert: Reference: https://docs.python.org/3.1/library/multiprocessing.html#multiprocessing.managers.BaseProxy._getvalue

stream_alert/rule_processor/handler.py
ryandeivert: It looks like this returns 2 cpus in a lambda configured with 512 mb ram.. may be different for larger ram allocations.

ryandeivert: Update: just tested this with  1536 MB of memory and it still only gives 2 processors.. :sadparrot:. Testing with 2 workers (obviously) gets us about a 2x speed improvements. I think we should discuss this more and use at least 4 workers.

stream_alert/rule_processor/handler.py
ryandeivert: @austinbyers turns out `Pool` is not an option in Lambda (and I think we already knew this...):

```
    sl = self._semlock = _multiprocessing.SemLock(kind, value, maxvalue)
OSError: [Errno 38] Function not implemented
```

We can, however, continue to use `Process` and not have any issues.

### Other Reading:
https://stackoverflow.com/questions/34005930/multiprocessing-semlock-is-not-implemented-when-running-on-aws-lambda

austinbyers: Ah, that makes sense. I didn't realize the limitation was only for multiprocess pools

stream_alert/rule_processor/handler.py
```diff
@@ -124,35 +138,71 @@ def run(self, event):
...
+        """Get all of the pre-parsed records from the payload and segment them
+        into chunks to be spread accross multiprocessing workers.
 
         Args:
             payload [StreamPayload]: StreamAlert payload object being processed
+            workers [list<mutliprocessing.Process>]: The complete list of
+                multiprocessing jobs to append this job to
+        """
+        # Get all of the pre-parsed records for this payload
+        payload_records = [record for record in payload.pre_parse()]
```
austinbyers: More simply,

```python
payload_records = list(payload.pre_parse())
```

stream_alert/rule_processor/handler.py
jacknagz: this should reference an environment variable IMO

stream_alert/rule_processor/handler.py
jacknagz: `_failed_record_count` is a slightly confusing name, pls add a comment or make it more descriptive

stream_alert/rule_processor/handler.py
ryandeivert: @jacknagz I was going to do away with this and always default to 2x the available CPUs (in lambda this will always == 4). I think it could be v bad to allow users to set this to whatever they 'think' is appropriate

stream_alert/rule_processor/handler.py
```diff
@@ -106,16 +114,22 @@ def run(self, event):
...
+            self._run_batches(payload, workers)
+
+        LOGGER.debug('Number of running workers: %d', len(workers))
 
-        LOGGER.debug('Invalid record count: %d', self._failed_record_count)
+        # Wait for all of the workers to finish before continuing
+        for worker in workers:
+            worker.join()
+
+        LOGGER.debug('Invalid record count: %d', self._failed_record_count.value)
```
jacknagz: wouldn't you only want to print this if it's nonzero?

stream_alert/rule_processor/handler.py
```diff
@@ -124,35 +138,71 @@ def run(self, event):
...
+            workers [list<mutliprocessing.Process>]: The complete list of
+                multiprocessing jobs to append this job to
+        """
+        # Get all of the pre-parsed records for this payload
+        payload_records = [record for record in payload.pre_parse()]
+
+        # Get the interval that we should segment the list of alerts on and
+        # arrange them into sublists of this size. Default to minimum of 1
+        interval = int(math.ceil(len(payload_records)/float(NUM_WORKERS))) or 1
+        record_sublists = [payload_records[index:interval+index]
```
jacknagz: `record_groups`?

stream_alert/rule_processor/handler.py
jacknagz: @ryandeivert for sure, that's a good point

stream_alert/rule_processor/handler.py
ryandeivert: This line is not new and is how we have it configured right now. @mime-frame is not a fan of conditional logging statements so if it's not a huge deal I'll leave it like this

jacknagz: interesting bc we do a lot of conditional logging

stream_alert/rule_processor/handler.py
ryandeivert: @jacknagz  this is not new code, do you have a suggestion for any better names?

ryandeivert: I'm coming up short

ryandeivert:
@austinbyers Adding a test to validate grouping of the list and addressing @jacknagz's concerns

stream_alert/rule_processor/handler.py
```diff
@@ -14,8 +14,11 @@
 limitations under the License.
 '''
 import json
+import math
 
+from copy import copy
 from logging import DEBUG as log_level_debug
+import multiprocessing as multiproc
```
austinbyers: Haha 6 characters shaved!

test/unit/stream_alert_rule_processor/test_handler.py
```diff
@@ -206,6 +211,52 @@ def test_run_debug_log_alert(self, extract_mock, rules_mock, _, log_mock):
...
+        # Force copy to return the actual object instead of a copy
+        copy_mock.return_value = self.__sa_handler.classifier
+
+        self.__sa_handler._run_batches(payload_mock, [])
+
+        # Set the groups we expect to be created from the segmentation logic
+        groups = [[0, 1, 2, 3, 4, 5],
+                  [6, 7, 8, 9, 10, 11],
+                  [12, 13, 14, 15, 16, 17],
+                  [18, 19, 20, 21]]
```
austinbyers: 🙌

javefang:
Great work @ryandeivert, is this going to help the case where logs from fed from Kinesis?

ryandeivert:
Hi @javefang -

Let me preface this with a disclosure that these changes will almost certainly never merge into master (and are wildly out of date now due to neglect on my part).

After looking at the change again, the answer is yes and no :). Theoretically, this would also apply to nested records _within_ a single kinesis record. For instance, if the incoming record in kinesis was a log that used the `json_path` attribute to extract/parse an array of nest records, this would apply the multiprocessing to that array of nested records.

However, if we go one step up and think about the entire 'records' array in a kinesis batch, this logic will **not** apply multiprocessing across the batch. For instance, the default for a kinesis batch size (in streamalert) is 100 records. The multiprocessing could apply to _nested_ records **within** each of those 100 records, but not the the 100 records themselves (if that makes sense? 🤔 ).

To further the discussion, we are likely going to address this problem through a variety of other changes in the next few months. I'm happy to discuss more if you have more questions.

javefang:
Thanks for the explanation @ryandeivert, I found my solution here to speed up the logs processing: add more shards (which will run more lambdas in parallel) :)

ryandeivert:
Closing this as it's stale and will likely not be an approach we take for optimization

PR #275, opened 2017-08-22
Athena Refresh 2.0

jacknagz:
to: @ryandeivert @chunyong-lin (both pls review all code)
size: medium

### Main Changes
* Hive Add Partition support: Instead of searching an entire bucket for new partitions, add them based on S3 event notifications.  This works by parsing the S3 key using a regular expression to extract the YYYY-MM-DD-HH information from the key.  This is specific to Firehose and StreamAlert's S3 Output format
* Add an option for async Athena query execution: I plan on using this option in the future for long running processes.  To maintain state, I plan on using DynamoDB.  This iteration may come sooner if the `ADD PARTITION` command starts taking too long.
* Deduplicate partitions before issuing Athena queries.
* Add clearer log messages.

### CLI Changes
* UX improvements: Moar logger messages indicating things went as expected.  
* Allow for custom KMS alias names (this was previously hardcoded, which is inflexible when a single AWS account has multiple StreamAlert deployments)

### CloudWatch Changes
* Add CloudWatch alarms for the Athena Refresh function via TF generate
* Tune params for WriteThroughputExceeded alarm

## Testing

All Athena and Outputs changes deployed and verified in a side account.  I configured CloudTrail and CloudWatch events to send dummy alerts to S3 with some basic added rules.

screen shot 2017-08-22 at 5 07 19 pm

### DIFF
--- a/stream_alert/athena_partition_refresh/main.py
+++ b/stream_alert/athena_partition_refresh/main.py
@@ -115,7 +115,7 @@ class StreamAlertAthenaClient(object):
                                     r'\-(?P<month>\d{2})'
                                     r'\-(?P<day>\d{2})'
                                     r'\-(?P<hour>\d{2})'
-                                    r'\-.*.json')
+                                    r'\/.*.json')
     FIREHOSE_REGEX = re.compile(r'(?P<year>\d{4})'
                                 r'\/(?P<month>\d{2})'
                                 r'\/(?P<day>\d{2})'
@@ -270,6 +270,7 @@ class StreamAlertAthenaClient(object):
         athena_config = self.config['lambda']['athena_partition_refresh_config']
         repair_hive_table_config = athena_config['refresh_type']['repair_hive_table']
 
+        LOGGER.info('Processing Hive repair table...')
         for data_bucket in unique_buckets:
             athena_table = repair_hive_table_config.get(data_bucket)
             if not athena_table:
@@ -305,8 +306,9 @@ class StreamAlertAthenaClient(object):
         """
         athena_config = self.config['lambda']['athena_partition_refresh_config']
         add_hive_partition_config = athena_config['refresh_type']['add_hive_partition']
-        partitions = set()
+        partitions = {}
 
+        LOGGER.info('Processing new Hive partitions...')
         for bucket, keys in s3_buckets_and_keys.iteritems():
             athena_table = add_hive_partition_config.get(bucket)
             if not athena_table:
@@ -320,9 +322,9 @@ class StreamAlertAthenaClient(object):
             s3_key_regex = self.STREAMALERTS_REGEX if athena_table == 'alerts' else self.FIREHOSE_REGEX
             for key in keys:
                 match = s3_key_regex.search(key)
-                if len(match.groups()) != 4:
-                    LOGGER.warning('%s does not match the regex: %s',
-                                   key, s3_key_regex.pattern)
+                if not match:
+                    LOGGER.error('The key %s does not match the regex %s, skipping',
+                                 key, s3_key_regex.pattern)
                     continue
 
                 # Convert the match groups to a dict for easy access
@@ -330,32 +332,39 @@ class StreamAlertAthenaClient(object):
                 # Get the path to the objects in S3
                 path = '/'.join(key.split('/')[:-1])
 
-                partitions.add("PARTITION (dt = '{year}-{month}-{day}-{hour}') "
-                               "LOCATION 's3://{bucket}/{path}'".format(
-                                   year=match_dict['year'],
-                                   month=match_dict['month'],
-                                   day=match_dict['day'],
-                                   hour=match_dict['hour'],
-                                   bucket=bucket,
-                                   path=path
-                               ))
+                # PARTITION (dt = '2017-01-01-01') LOCATION 's3://bucket/path/'
+                partition = '(dt = \'{year}-{month}-{day}-{hour}\')'.format(
+                    year=match_dict['year'],
+                    month=match_dict['month'],
+                    day=match_dict['day'],
+                    hour=match_dict['hour'])
+                location = '\'s3://{bucket}/{path}\''.format(
+                    bucket=bucket,
+                    path=path)
+                # By using the partition as the dict key, this ensures that
+                # Athena will not try to add the same partition twice.
+                partitions[partition] = location
 
+            partition_statement = ' '.join(
+                ['PARTITION {0} LOCATION {1}'.format(
+                 partition, location) for partition, location in partitions.iteritems()])
             query = ('ALTER TABLE {athena_table} '
-                     'ADD IF NOT EXISTS {partition_statements};'.format(
+                     'ADD IF NOT EXISTS {partition_statement};'.format(
                          athena_table=athena_table,
-                         partition_statements=' '.join(partitions)))
+                         partition_statement=partition_statement))
+
             query_success, query_resp = self.run_athena_query(
                 query=query,
                 database=self.DATABASE_STREAMALERT
             )
 
             if not query_success:
-                LOGGER.error('The following query has failed: %s', query)
+                LOGGER.error('The following query has failed:\n%s', query)
                 return False
 
-            LOGGER.info('Query results:')
-            for row in query_resp['ResultSet']['Rows']:
-                LOGGER.info(row.get('Data'))
+            LOGGER.info('Successfully added the following partitions:\n%s',
+                        '\n'.join(partitions))
+
             return True
 
 
@@ -392,21 +401,33 @@ class StreamAlertSQSClient(object):
             QueueNamePrefix=self.QUEUENAME
         )['QueueUrls'][0]
 
-    def get_messages(self):
-        """Poll the queue for messages"""
+    def get_messages(self, **kwargs):
+        """Poll the SQS queue for new messages
+
+        Keyword Args:
+            max_tries (int): The number of times to backoff
+            max_value (int): The max wait interval between backoffs
+            max_messages (int): The max number of messages to get from SQS
+        """
+        start_message_count = len(self.received_messages)
+
+        max_tries = kwargs.get('max_tries', 5)
+        max_value = kwargs.get('max_value', 5)
+        max_messages = kwargs.get('max_messages', 10)
+
         @backoff.on_predicate(backoff.fibo,
                               # Backoff up to 5 times
-                              max_tries=5,
+                              max_tries=max_tries,
                               # Don't backoff for longer than 5 seconds
                               # This constrains the total max backoff to 25 seconds
-                              max_value=5,
+                              max_value=max_value,
                               jitter=backoff.full_jitter,
                               on_backoff=_backoff_handler,
                               on_success=_success_handler)
         def _receive_messages():
             polled_messages = self.sqs_client.receive_message(
                 QueueUrl=self.athena_sqs_url,
-                MaxNumberOfMessages=10
+                MaxNumberOfMessages=max_messages
             )
 
             if 'Messages' not in polled_messages:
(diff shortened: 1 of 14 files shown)


stream_alert_cli/outputs.py
```diff
@@ -97,23 +98,24 @@ def encrypt_and_push_creds_to_s3(region, bucket, key, props):
...
         data [string]: json string to be encrypted
+        kms_key_alias [string]: The KMS key alias to use for encryption of S3 objects
 
     Returns:
         [string] Encrypted ciphertext data blob
     """
     try:
         client = boto3.client('kms', region_name=region)
-        response = client.encrypt(KeyId='alias/stream_alert_secrets',
+        response = client.encrypt(KeyId=kms_key_alias,
```
ryandeivert: ++ for this!!
Question though - should we keep the `alias/` prefix stuff here to avoid having to have it in a bunch of different places?

stream_alert/athena_partition_refresh/main.py
```diff
@@ -243,7 +257,14 @@ def check_table_exists(self, table_name):
         return False
 
     def repair_hive_table(self, unique_buckets):
-        """Execute a MSCK REPAIR TABLE on a given Athena table"""
+        """Execute a MSCK REPAIR TABLE on a given Athena table
+
+        Args:
+            unique_buckets (list): S3 buckets to repair
```
ryandeivert: So we're going to this format now? :P

stream_alert/athena_partition_refresh/main.py
```diff
@@ -271,11 +292,69 @@ def repair_hive_table(self, unique_buckets):
...
+            (bool): If the repair was successful for not
+        """
+        athena_config = self.config['lambda']['athena_partition_refresh_config']
+        add_hive_partition_config = athena_config['refresh_type']['add_hive_partition']
+        partitions = set()
+
+        for bucket, keys in s3_buckets_and_keys.iteritems():
+            athena_table = add_hive_partition_config.get(bucket)
+            if not athena_table:
+                LOGGER.error('%s not found in add_hive_partition config. '
```
ryandeivert: Can we quote `add_hive_partition` to make it more clear what this is

stream_alert/athena_partition_refresh/main.py
```diff
@@ -271,11 +292,69 @@ def repair_hive_table(self, unique_buckets):
...
+                match = s3_key_regex.search(key)
+                if len(match.groups()) != 4:
+                    LOGGER.warning('%s does not match the regex: %s',
+                                   key, s3_key_regex.pattern)
+                    continue
+
+                # Convert the match groups to a dict for easy access
+                match_dict = match.groupdict()
+                # Get the path to the objects in S3
+                path = '/'.join(key.split('/')[:-1])
```
ryandeivert: Is the goal here to just strip everything after and including the last '/'? If so, and if I'm understanding this correctly, maybe consider `os.path.dirname(key)` or `os.path.split(key)[0]`

stream_alert/athena_partition_refresh/main.py
```diff
@@ -386,11 +465,15 @@ def unique_buckets_from_messages(self):
                     LOGGER.debug(record)
                     continue
 
-                buckets.add(record['s3']['bucket']['name'])
+                bucket_name = record['s3']['bucket']['name']
+                # Account for special characters in the S3 object key
```
ryandeivert: What are these special characters you speak of 🤔  Consider being consistent with [this comment ](https://github.com/airbnb/streamalert/blob/afd19b203d529d03b2e32542889eda3fc6984fe6/stream_alert/rule_processor/payload.py#L230)

stream_alert/athena_partition_refresh/main.py
```diff
@@ -450,13 +475,18 @@ def unique_s3_buckets_and_keys(self):
...
         for message in self.received_messages:
             if 'Body' not in message:
-                LOGGER.error('Missing \'Body\' key, trying next SQS message')
+                LOGGER.error('Missing \'Body\' key in SQS message, skipping')
                 continue
 
             loaded_message = json.loads(message['Body'])
 
+            if loaded_message.get('Event') == 'S3:TestEvent':
+                LOGGER.debug('Skipping S3:TestEvent')
```
ryandeivert: Is this something that will happen in production? Curious why this code exists here. If it should remain please add a comment.

stream_alert/athena_partition_refresh/main.py
```diff
@@ -482,15 +512,22 @@ def handler(*_):
...
+    stream_alert_sqs.get_messages(max_tries=2)
 
     if not stream_alert_sqs.received_messages:
-        LOGGER.info('No messages recieved, exiting')
+        LOGGER.info('No SQS messages recieved, exiting')
         return
 
+    # If the max amount of messages was initially returned,
+    # then get the next batch of messages
+    if len(stream_alert_sqs.received_messages) == 20:
```
ryandeivert: Your comment seems to indicate that the max amount of messages here == 20, but [this code above](https://github.com/airbnb/streamalert/pull/275/commits/ba11a9d10af46736ea225d60c53743dfd47ec65a#diff-3159770d8fb42356c5e4163f106792c1R414) looks like it sets the default to `10`.. 

As an aside, could we set the `max_tries`, `max_values`, and `max_messages` default values as class constants that can be referenced throughout.

stream_alert_cli/outputs.py
jacknagz: `alias/` is a prefix added by Amazon to identify key aliases by name vs UUID

stream_alert/athena_partition_refresh/main.py
jacknagz: There's an existing PR that does this, I'm going to just wait for that

stream_alert_cli/outputs.py
jacknagz: yeah that works, I'll make the change

stream_alert/athena_partition_refresh/main.py
jacknagz: Added some context

stream_alert/athena_partition_refresh/main.py
jacknagz: Added a little comment

stream_alert/athena_partition_refresh/main.py
jacknagz: > Your comment seems to indicate that the max amount of messages here == 20, but this code above looks like it sets the default to 10

There are two attempts, so it would be max 20 total messages with 10 each try.

> As an aside, could we set the max_tries, max_values, and max_messages default values as class constants that can be referenced throughout.

I could, but it seems like overkill since it's only used in that one method

stream_alert/athena_partition_refresh/main.py
```diff
@@ -392,25 +524,32 @@ def handler(*_):
...
-        LOGGER.info('No messages recieved, exiting')
+        LOGGER.info('No SQS messages recieved, exiting')
         return
 
-    unique_buckets = stream_alert_sqs.unique_buckets_from_messages()
-    if not unique_buckets:
-        LOGGER.error('No s3 buckets to refresh, exiting')
+    # If the max amount of messages was initially returned,
+    # then get the next batch of messages
+    if len(stream_alert_sqs.received_messages) == 20:
```
chunyong-lin: How do you come up `20` here?

jacknagz: @chunyong-lin  https://github.com/airbnb/streamalert/pull/275#discussion_r134801556

stream_alert/athena_partition_refresh/main.py
ryandeivert: @jacknagz it's incredibly unclear where this value comes from though - a named constant (or even a property) would help a ton here, and it should be clear that it's a multiple of `max_message * max_tries` Also why do you perform `get_messages` with `max_tries=2` once and then `max_tries=8` next?

jacknagz: @ryandeivert I'll add clarification in the comments.  And `max_tries` and `max_value` are determined by us, I can set the default `max_messages` to a constant.

> Also why do you perform get_messages with max_tries=2 once and then max_tries=8 next?

Did you see the comment above that block?

```
    # Get the first batch of messages from SQS.  If there are no
    # messages, this will exit early.
```

jacknagz:
@ryandeivert @chunyong-lin pls take another look, I added additional clarification and addressed your concerns

jacknagz:
Here's an example of the SQS message sampling in real life:

```
START RequestId: 8889c324-8845-11e7-9138-7d37eaafb948 Version: 8
[ERROR]	2017-08-23T20:56:29.552Z	8889c324-8845-11e7-9138-7d37eaafb948	Giving up _receive_messages(...) after 2 tries (None)
[INFO]	2017-08-23T20:56:29.552Z	8889c324-8845-11e7-9138-7d37eaafb948	Received 20 message(s) from SQS
[ERROR]	2017-08-23T20:56:36.189Z	8889c324-8845-11e7-9138-7d37eaafb948	Giving up _receive_messages(...) after 8 tries (None)
[INFO]	2017-08-23T20:56:36.189Z	8889c324-8845-11e7-9138-7d37eaafb948	Received 48 message(s) from SQS
[INFO]	2017-08-23T20:56:37.966Z	8889c324-8845-11e7-9138-7d37eaafb948	Processing new Hive partitions...
[INFO]	2017-08-23T20:56:39.800Z	8889c324-8845-11e7-9138-7d37eaafb948	Successfully added the following partitions:
(dt = '2017-08-23-20')
[INFO]	2017-08-23T20:56:49.666Z	8889c324-8845-11e7-9138-7d37eaafb948	Deleted 68 messages from SQS
END RequestId: 8889c324-8845-11e7-9138-7d37eaafb948
```

stream_alert/athena_partition_refresh/main.py
```diff
@@ -409,19 +424,26 @@ def get_messages(self, **kwargs):
...
+        # relative to the entire Lambda duration.
         max_tries = kwargs.get('max_tries', 5)
+        # This value restricts the max time of backoff each try.
+        # This means the total backoff time for one function call is:
+        #   max_tries (attempts) * max_value (seconds)
         max_value = kwargs.get('max_value', 5)
-        max_messages = kwargs.get('max_messages', 10)
+        # Number of messages to poll from the stream.
+        max_messages = kwargs.get('max_messages', self.MAX_SQS_GET_MESSAGE_COUNT)
+        if max_messages > 10:
```
ryandeivert: Please also use `self.MAX_SQS_GET_MESSAGE_COUNT` here

ryandeivert:
Also super sad to see coverage drop so much with this PR - can you work to add coverage for the new stuff in the next ~day?

stream_alert/athena_partition_refresh/main.py
chunyong-lin: Use `self.MAX_SQS_GET_MESSAGE_COUNT` to replace `10` here.

stream_alert/athena_partition_refresh/main.py
```diff
@@ -442,24 +464,40 @@ def delete_messages(self):
             LOGGER.error('No processed messages to delete')
             return
 
-        deleted_messages = 0
-        while self.processed_messages:
+        @backoff.on_predicate(backoff.fibo,
+                              lambda len_messages: len_messages > 0,
+                              max_value=10,
```
chunyong-lin: Why not use `MAX_SQS_GET_MESSAGE_COUNT ` here also?

jacknagz: `max_value` is not the same as `max_messages`

ryandeivert:
@jacknagz new changes look a-okay! 🚢

Issue #245, opened 2017-08-08
Bug: Live-test should fail when no test events exist


jacknagz:
## Background

When testing a rule, if no related test event exists, nothing happens.  It's difficult to diagnose or troubleshoot why.

```
$ python stream_alert_cli.py live-test --cluster prod --rule <my-rule-name>
StreamAlertCLI [INFO]: Issues? Report here: https://github.com/airbnb/streamalert/issues



(0/0)	Rule Tests Passed
(0/0)	Alert Tests Passed
StreamAlertCLI [INFO]: Completed
```

## Desired Outcome

The CLI should check first for an existing test event, and then throw an error if one doesn't exist.  Also, we could probably check coverage based on all loaded rules and # of test events.

Issue #154, opened 2017-05-17
Allow schema validation testing without the need for defined rules


ryandeivert:
**Problem**

When a new log schema is added to logs.json, we must currently write a rule (inside of `/rules/`) that will apply to that specific schema in order to simply test that we have it defined properly.

It would be ideal if we were able to define a schema in logs.json, and add a sample log to `/test/integration/rules/` that we believe corresponds to this new schema.  Once these two steps are complete, we could run a `schema` test (either through a flag with stream_alert_cli, or some other means) that would simply validate that any test files match to a schema. Preferably, we would output _which_ schema the test rule file matched to as well. This feature would allow us to define schemas before we have a real need to run rules against said schema, but still validate that we have defined the schema properly.

I will update this issue if the context of this changes.

Issue #149, opened 2017-05-15
Improvement: Make `Invalid data` debugging easier


mime-frame:
**What**

When running `./test/scripts/rule_test.sh` you may run into an error like the following:

`StreamAlert [ERROR]: Invalid data: <StreamPayload valid:False log_source:None entity:blah.us-east-1 type:None record:None>`

This is difficult to debug and the error could be made more useful.

Examples of how this error could be made better:
- output the missing/incorrect keys
- output the incorrect types

Issue #276, opened 2017-08-23
Improvement: Fail harder for StreamAlertOutput errors


ghost:
**Overview**

When a rule has something defined in `outputs` that does not exist in `outputs.json`, it does output the following on stdout:

`StreamAlertOutput [ERROR]: The output 'pagerduty:example' does not exist!`

However:
- It is not in red font
- It is not output at the end (aka it hides amongst hundreds of lines of text)
- It does not cause `python stream_alert_cli.py lambda test --processor all` to fail; e.g. it can show everything as passing

I managed to catch this by accident when redirecting output like so:
- `python stream_alert_cli.py lambda test --processor all > foo.txt`

**Reproduction steps**

- Modify one of your rules' outputs - take off the last letter for example
- Re-run your tests, notice everything as passing and notice how you see nothing red

ryandeivert:
the way the alert processor is now architected, this is no longer an issue

PR #277, opened 2017-08-24
[cli][test] validation of schemas without writing a rule

ryandeivert:
to @austinbyers for manage.py updates and various logging changes
to @jacknagz and @chunyong-lin for cli refactor
to @mime-frame: please confirm that log output examples below are sensible
size: medium
resolves: #149 
resolves: #154 
resolves: #245
contributes to: #276 

### Changes
* Testing via `manage.py` now supports a `validate-schemas` command.
  * An optional argument for this command is `--test-files` which accepts a space separated list of files (or, alternatively, rule names) to validate.
  * For example: `python manage.py validate-schemas --test-files cloudtrail_put_bucket_acl`
  * This command does not perform any actual rule engine operations and will simply attempt to 'classify' the record against the defined log schemas.
* All summary output is now printed through the logging object with their respective levels (ie: warning, error, info) instead of through print statements.

##### `$ python manage.py validate-schemas --test-files cloudtrail_put_object_acl.json`
```
$ python manage.py validate-schemas --test-files cloudtrail_put_object_acl.json

cloudtrail_put_object_acl
       [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - PutObjectAcl - True Positive
       [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - PutObjectAcl - False Positive

```

##### `$ python manage.py validate-schemas --test-files cloudtrail_put_object_acl.json`
```
$ python manage.py validate-schemas --test-files cloudtrail_put_object_acl.json

cloudtrail_put_object_acl
       [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - PutObjectAcl - True Positive
       [Fail]  [log='unknown']               validation  (s3): CloudTrail - PutObjectAcl - False Positive

StreamAlertCLI [INFO]: (1/2) Successful Tests
StreamAlertCLI [ERROR]: (1/2) Failures
StreamAlertCLI [ERROR]: (1/1) [cloudtrail_put_object_acl] Data is invalid due to missing key(s) in test record: 'eventVersion'. Rule: 'cloudtrail_put_object_acl'. Description: 'CloudTrail - PutObjectAcl - False Positive'
```

### Other updates
* Adding the ability to detect any error messages that are logged during execution
  * By adding a MemoryHandler to the root logger and filtering out all messages that are not of level 'error', we can store any errors that occur during test runs in a buffer.
  * The number of errors that occurred will be reported at the end, and the errors themselves will be logged so they can be addressed.
* We will now log warnings to stdout for any rule that exists without a corresponding integration test, like so:
```
StreamAlertCLI [WARNING]: No tests configured for rule: 'rule_without_a_test'. Please add a corresponding test file for this rule in 'tests/integration/rules' with the name 'rule_without_a_test.json' to avoid seeing this warning
```
* A similar warning like the one above will be logged if a integration test file exists, but there is no rule for said file. However, this will **only** occur if the command is `lambda test` and not with `validate-schemas`. The output is like so:
```
StreamAlertCLI [WARNING]: No rules configured for test file: 'test_without_rule.json'. Please add a corresponding rule for this test file in 'rules/' with the name 'test_without_rule.py' to avoid seeing this warning and any associated errors above
```
* Significantly changing how status messages are stored and output during a test run.

### Bug fixes
* Resolving bug with improper set logic that could cause the alert processor to get ran inadvertently. See [this commit](https://github.com/airbnb/streamalert/commit/817d9e348300fd718498ab0adb67cfed0070fde6).
* Resolving bug that could cause the `analyze_record_delta` function to report that keys defined as `optional_top_level_keys` in a log's schema were in fact required. See [this commit](https://github.com/airbnb/streamalert/commit/c9c493579403108a31bf6292143fcab81c41c741).

### DIFF
--- a/docs/source/rule-testing.rst
+++ b/docs/source/rule-testing.rst
@@ -14,8 +14,8 @@ To test a new rule, first create a new file under ``tests/integration/rules`` na
         "data": {} or "",
         "description": "of the test",
         "trigger": true or false,
-        "source": "kinesis_stream_name" or "s3_bucket_id",
-        "service": "kinesis" or "s3"
+        "source": "kinesis_stream_name" or "s3_bucket_id" or "sns_topic_name",
+        "service": "kinesis" or "s3" or "sns"
       }
     ]
   }
@@ -74,6 +74,84 @@ For example, to replace a time based field with ``last_hour``:
       }
     ]
   }
+
+
+Validating Log Schemas
+~~~~~~~~~~~~~~~~~~~~~~
+
+In some cases, there may be logs coming into StreamAlert that are of a known type, but you do not have a current need to apply rule
+logic to them. However, it is still advantageous to add schemas for these logs and *test* to ensure your new schemas are valid.
+
+This is possible through the addition of log schema(s) to ``conf/logs.json`` and creation of test record(s) in ``tests/integration/rules/``
+using known samples of logs (without actually adding a corresponding rule). Running the ``manage.py`` script with the ``validate-schemas``
+option will iterate over all json test files and attempt to match records within them to a defined log schema.
+
+To simply run schema validation on all test files:
+
+.. code-block:: bash
+
+  $ python manage.py validate-schemas
+
+
+To run schema validation on a specific test file within ``tests/integration/rules/``:
+
+.. code-block:: bash
+
+  $ python manage.py validate-schemas --test-files <test_rule_file.json>
+
+Or:
+
+.. code-block:: bash
+
+  $ python manage.py validate-schemas --test-files <test_rule_file>
+
+
+Example of running schema validation on two valid test files:
+
+.. code-block:: bash
+
+  $ python manage.py validate-schemas --test-files cloudtrail_critical_api_calls cloudtrail_put_bucket_acl.json
+
+This will produce output similar to the following::
+
+  cloudtrail_critical_api_calls
+         [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - Critical API - DeleteSubnet
+         [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - Critical API - DeleteVpc
+         [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - Critical API - UpdateTrail
+         [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - Critical API - StopLogging
+         [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - Critical API - DeleteDBCluster
+         [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - Critical API - StopConfigurationRecorder
+         [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - Critical API - DeleteFlowLogs
+         [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - Critical API - False Positive Case
+
+  cloudtrail_put_bucket_acl
+         [Pass]  [log='cloudwatch:events']     validation  (kinesis): CloudTrail - PutBucketAcl - True Positive
+         [Pass]  [log='cloudwatch:events']     validation  (kinesis): CloudTrail - PutBucketAcl - False Positive
+
+
+
+  StreamAlertCLI [INFO]: (10/10) Successful Tests
+  StreamAlertCLI [INFO]: Completed
+
+
+Example of running schema validation on a test file that contains one valid record and one invalid record:
+
+.. code-block:: bash
+
+  $ python manage.py validate-schemas --test-files cloudtrail_put_object_acl.json
+
+
+This will produce output similar to the following::
+
+  cloudtrail_put_object_acl
+         [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - PutObjectAcl - True Positive
+         [Fail]  [log='unknown']               validation  (s3): CloudTrail - PutObjectAcl - False Positive
+
+
+
+  StreamAlertCLI [INFO]: (1/2) Successful Tests
+  StreamAlertCLI [ERROR]: (1/2) Failures
+  StreamAlertCLI [ERROR]: (1/1) [cloudtrail_put_object_acl] Data is invalid due to missing key(s) in test record: 'eventVersion'. Rule: 'cloudtrail_put_object_acl'. Description: 'CloudTrail - PutObjectAcl - False Positive'
 
 
 Running Tests
@@ -141,7 +219,6 @@ This will produce output similar to the following::
   	[Pass]              	alert	(pagerduty): sending alert to 'sample_integration'
   	[Pass]              	alert	(aws-s3): sending alert to 'sample_bucket'
   	[Pass]   [trigger=0]	rule	(kinesis): CloudTrail - Root Account Usage - False Positive
-
 
 
   (4/4)	Rule Tests Passed
--- a/docs/source/rule-testing.rst
+++ b/docs/source/rule-testing.rst
@@ -76,17 +76,17 @@ For example, to replace a time based field with ``last_hour``:
   }
 
 
-Validating Log Schemas
-~~~~~~~~~~~~~~~~~~~~~~
+Validate Log Schemas
+~~~~~~~~~~~~~~~~~~~~
 
-In some cases, there may be logs coming into StreamAlert that are of a known type, but you do not have a current need to apply rule
-logic to them. However, it is still advantageous to add schemas for these logs and *test* to ensure your new schemas are valid.
+In some cases, there may be incoming logs to StreamAlert with a known type, but without specific rules that apply to them.
+However, it is best practice to write schemas for these logs and *verify* that they are valid.
 
-This is possible through the addition of log schema(s) to ``conf/logs.json`` and creation of test record(s) in ``tests/integration/rules/``
-using known samples of logs (without actually adding a corresponding rule). Running the ``manage.py`` script with the ``validate-schemas``
-option will iterate over all json test files and attempt to match records within them to a defined log schema.
+This is possible by first adding the new schema(s) to ``conf/logs.json`` along with creation of test record(s) in ``tests/integration/rules/``
+containing samples of real logs (without actually adding a corresponding rule). Running the ``manage.py`` script with the ``validate-schemas``
+option will iterate over all json test files and attempt to classify each record.
 
-To simply run schema validation on all test files:
+To run schema validation on all test files:
 
 .. code-block:: bash
 
@@ -106,7 +106,7 @@ Or:
   $ python manage.py validate-schemas --test-files <test_rule_file>
 
 
-Example of running schema validation on two valid test files:
+Schema validation on two valid test files:
 
 .. code-block:: bash
 
@@ -134,7 +134,7 @@ This will produce output similar to the following::
   StreamAlertCLI [INFO]: Completed
 
 
-Example of running schema validation on a test file that contains one valid record and one invalid record:
+Schema validation failure on a test file containing one valid record and one invalid record:
 
 .. code-block:: bash
 
--- a/manage.py
+++ b/manage.py
@@ -153,7 +153,7 @@ Available Options:
 
 Examples:
 
-    stream_alert_cli.py validate-schemas --files
+    manage.py validate-schemas --files
 
 """.format(version))
     schema_validation_parser = subparsers.add_parser(
--- a/manage.py
+++ b/manage.py
@@ -140,6 +140,48 @@ Examples:
     )
 
 
+def _add_validate_schema_subparser(subparsers):
+    schema_validation_usage = 'manage.py validate-schemas [options]'
+    schema_validation_description = ("""
+StreamAlertCLI v{}
+Run end-to-end tests that will attempt to send alerts
+
+Available Options:
+
+    --files                 Name (not full path) of test file(s) to validate, separated by spaces
+    --debug                 Enable Debug logger output
+
+Examples:
+
+    stream_alert_cli.py validate-schemas --files
+
+""".format(version))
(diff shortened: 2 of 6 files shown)


stream_alert_cli/test.py
```diff
@@ -86,81 +86,132 @@ def __init__(self, context, print_output):
         self.print_output = print_output
         self.invalid_log_messages = []
 
-    def test_processor(self, rules):
+    def test_processor(self, filter_rules):
         """Perform integration tests for the 'rule' Lambda function
 
         Args:
-            rules (list|None): Specific rule names (or None) to restrict
+            filter_rules [list or None]: Specific rule names (or None) to restrict
```
jacknagz: Update the docstrings to use `(type)` per our new convention pls

stream_alert_cli/test.py
```diff
@@ -86,81 +86,132 @@ def __init__(self, context, print_output):
...
-                message = ('' if current_test_passed else
-                           'Rule failure: {}'.format(test_record['description']))
-                self.status_messages.append(StatusMessage(message_type, rule_name, message))
+                yield self._run_rule_tests(rule_name, test_record, print_header_line)
 
         # Report on the final test results
         self.report_output_summary()
 
+    def _run_rule_tests(self, rule_name, test_record, print_header_line):
+        """Run tests on a test record for a given rule
```
jacknagz: take another look at this docstring description

stream_alert_cli/test.py
```diff
@@ -86,81 +86,132 @@ def __init__(self, context, print_output):
...
+            [generator] Yields back the rule name and the json loaded contents of
+                the respective test event file.
+        """
+        for _, _, test_rule_files in os.walk(DIR_RULES):
+            for rule_file in test_rule_files:
+                rule_name = rule_file.split('.')[0]
+
+                # If only specific rules are being tested,
+                # skip files that do not match those rules
+                if filter_rules:
```
jacknagz: chain these two `if`s together

manage.py
```diff
@@ -140,6 +140,48 @@ def _add_live_test_subparser(subparsers):
     )
 
 
+def _add_validate_schema_subparser(subparsers):
```
jacknagz: pylint doesn't complain about no docstring?

manage.py
```diff
@@ -140,6 +140,48 @@ def _add_live_test_subparser(subparsers):
...
+    schema_validation_usage = 'manage.py validate-schemas [options]'
+    schema_validation_description = ("""
+StreamAlertCLI v{}
+Run end-to-end tests that will attempt to send alerts
+
+Available Options:
+
+    --files                 Name (not full path) of test file(s) to validate, separated by spaces
```
jacknagz: Should we call this `--test-files`? I'd also add a little description of what these files should contain

manage.py
ryandeivert: Non-public mehods/functions don't actually _need_ docstrings so it wouldn't complain here.. but I didn't realize this was missing one so I'll add it for consistency

manage.py
ryandeivert: Sure that's a good change - I'll update the comment to clarify some stuff

ryandeivert:
@jacknagz  PTAL at new commits - one for your feedback and one for doc updates

mime-frame:
It took me awhile to understand what exactly was being passed to `--files` in this example:

`$ python manage.py validate-schemas --files cloudtrail_put_object_acl`

Proposal: We change it from `--files` to `--log-files`

austinbyers:
@ryandeivert Just a note that you can't actually use a comma-separated list for "Resolves" in the PR description :( If you hover over it, it only shows that it will resolve the first listed issue

mime-frame:
I'm a bit confused. We state this:

> This command does not perform any actual rule engine operations and will simply attempt to 'classify' the record against the defined log schemas.

Yet, the output references rule integration tests and whether they passed?:

```
cloudtrail_put_object_acl
       [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - PutObjectAcl - True Positive
       [Pass]  [log='cloudtrail:events']     validation  (s3): CloudTrail - PutObjectAcl - False Positive
```

manage.py
austinbyers: Good to know

manage.py
```diff
@@ -140,6 +140,56 @@ def _add_live_test_subparser(subparsers):
...
+def _add_validate_schema_subparser(subparsers):
+    """Add the validate-schemas subparser: manage.py validate-schemas [options]"""
+    schema_validation_usage = 'manage.py validate-schemas [options]'
+    schema_validation_description = ("""
+StreamAlertCLI v{}
+Run end-to-end tests that will attempt to send alerts
+
+Available Options:
+
+    --test-files         Name(s) of test files to validate, separated by spaces (not full path)
```
austinbyers: Isn't it usually the standard to separate multiple arguments in CLI by commas instead of spaces? Each space-delimited token should be its own separate argument

stream_alert_cli/helpers.py
```diff
@@ -23,6 +24,7 @@
 import zlib
 
 import boto3
+from moto import mock_cloudwatch, mock_lambda, mock_kms, mock_s3
```
austinbyers: Nit: alphabetical import `mock_kms` before `mock_lambda`

stream_alert_cli/logger.py
```diff
@@ -33,3 +54,14 @@
...
+# Get a logging MemoryHandler with a default buffer size of 1000
+# We don't care about assigning a target to this handler since these logs
+# will not actually be written out to disk, etc
+LOG_ERROR_HANDLER = logging.handlers.MemoryHandler(1000)
+
+# Add a filter to suppress everything that is not an error
+LOG_ERROR_HANDLER.addFilter(SuppressNonErrors())
+
+# Add the MemoryHandler to the root logger to capture all logs in all loggers
+logging.getLogger().addHandler(LOG_ERROR_HANDLER)
```
austinbyers: If this extra log handler is only used for unit tests, maybe we should only add it if the user is running tests? Instead of adding an extra in-memory logging handler for all StreamAlert operations

ryandeivert: This isn't unit tests and doesn't get added for all streamalert operations - this happens only in the CLI package - but you are right I can update to only happen if integration tests are being run! making that change

ryandeivert:
@mime-frame idk how you commented but I am not able to reply directly to you. re this comment: https://github.com/airbnb/streamalert/pull/277#issuecomment-324720347

I think your preexisting knowledge of output from other commands (`python manage.py lambda test ...`) may be skewing how you are interpreting this output. The `Pass` and `Fail` in this case is related to matching a defined schema (not triggering a rule)..

For instance, we print `[Pass]  [log='cloudtrail:events']     validation ...` meaning "this validation test passed, and the log type matched to is `cloudtrail:events`"

The alternative was to print nothing at all :)

Also RE this comment: https://github.com/airbnb/streamalert/pull/277#issuecomment-324719919  there was a new commit pushed a few hours ago that changes this to `--test-files`

mime-frame:
Cool, thanks @ryandeivert - my review is complete

manage.py
ryandeivert: Yes I agree with you here, but argparse doesn't :) List supplied as args to argparse are expected to be space delimited. We could change this to expect a string that is a comma separated and handle all of the logic in code, but then we lose the validation/err handling baked into argparse. At one point (with the --rules flag) I was doing comma separated but we decided to do spaces since it was safer with less overhead

austinbyers: Oh interesting. Ok, sounds good! Thanks for the note

manage.py
```diff
@@ -140,6 +140,56 @@ def _add_live_test_subparser(subparsers):
...
+    schema_validation_parser.set_defaults(command='validate-schemas')
+
+    # add the optional ability to test against specific files
+    schema_validation_parser.add_argument(
+        '-f', '--test-files',
+        nargs='+',
+        help=ARGPARSE_SUPPRESS
+    )
+
+    # allow verbose output for the CLI with te --debug option
```
austinbyers: nit: typo (`with te --debug option`)

stream_alert_cli/helpers.py
```diff
@@ -203,3 +205,68 @@ def put_mock_s3_object(bucket, key, data, region):
...
+
+    Args:
+        context (namedtuple): A constructed aws context object
+    """
+    def wrap(func):
+        """Wrap the returned function with or without mocks"""
+        if context.mocked:
+            @mock_cloudwatch
+            @mock_lambda
+            @mock_s3
```
austinbyers: Just beware of https://github.com/spulec/moto/issues/1026; the mock may not be restored correctly every time

stream_alert_cli/test.py
```diff
@@ -68,98 +71,223 @@ def __init__(self, context, print_output):
...
 
-    def test_processor(self, rules):
+    def test_processor(self, filter_rules, validate_only=False):
         """Perform integration tests for the 'rule' Lambda function
 
         Args:
-            rules (list|None): Specific rule names (or None) to restrict
+            filter_rules (list|None): Specific rule names (or None) to restrict
                 testing to. This is passed in from the CLI using the --rules option.
+            validate_only [bool=False]: If true, validation of test records will occur
```
austinbyers: Nit: should be `validate_only (bool)`. Types should be in parens (not brackets), and the default value need not be specified since it's in the function definition

docs/source/rule-testing.rst
```diff
@@ -76,6 +76,84 @@ For example, to replace a time based field with ``last_hour``:
   }
 
 
+Validating Log Schemas
```
jacknagz: `Validate Log Schemas`

docs/source/rule-testing.rst
```diff
@@ -76,6 +76,84 @@ For example, to replace a time based field with ``last_hour``:
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+In some cases, there may be logs coming into StreamAlert that are of a known type, but you do not have a current need to apply rule
```
jacknagz: `may be incoming logs to StreamAlert with a known type but without specific rules`

docs/source/rule-testing.rst
```diff
@@ -76,6 +76,84 @@ For example, to replace a time based field with ``last_hour``:
...
+logic to them. However, it is still advantageous to add schemas for these logs and *test* to ensure your new schemas are valid.
```
jacknagz: `However, it is best practice to write schemas for these logs and *verify* that they are valid.`

docs/source/rule-testing.rst
```diff
@@ -76,6 +76,84 @@ For example, to replace a time based field with ``last_hour``:
...
+
+This is possible through the addition of log schema(s) to ``conf/logs.json`` and creation of test record(s) in ``tests/integration/rules/``
```
jacknagz: `This is possible by first adding the new schema(s) to ``conf/logs.json`` along with creation of test record(s)...`

docs/source/rule-testing.rst
```diff
@@ -76,6 +76,84 @@ For example, to replace a time based field with ``last_hour``:
...
+using known samples of logs (without actually adding a corresponding rule). Running the ``manage.py`` script with the ``validate-schemas``
```
jacknagz: `containing samples of real logs.`

docs/source/rule-testing.rst
```diff
@@ -76,6 +76,84 @@ For example, to replace a time based field with ``last_hour``:
...
+option will iterate over all json test files and attempt to match records within them to a defined log schema.
```
jacknagz: I think it's cool to use `classify records` here

docs/source/rule-testing.rst
```diff
@@ -76,6 +76,84 @@ For example, to replace a time based field with ``last_hour``:
...
+
+To simply run schema validation on all test files:
```
jacknagz: nix `simply`

docs/source/rule-testing.rst
```diff
@@ -76,6 +76,84 @@ For example, to replace a time based field with ``last_hour``:
...
+  $ python manage.py validate-schemas --test-files <test_rule_file.json>
+
+Or:
+
+.. code-block:: bash
+
+  $ python manage.py validate-schemas --test-files <test_rule_file>
+
+
+Example of running schema validation on two valid test files:
```
jacknagz: `s/Example of running//`

docs/source/rule-testing.rst
```diff
@@ -76,6 +76,84 @@ For example, to replace a time based field with ``last_hour``:
...
+         [Pass]  [log='cloudwatch:events']     validation  (kinesis): CloudTrail - PutBucketAcl - True Positive
+         [Pass]  [log='cloudwatch:events']     validation  (kinesis): CloudTrail - PutBucketAcl - False Positive
+
+
+
+  StreamAlertCLI [INFO]: (10/10) Successful Tests
+  StreamAlertCLI [INFO]: Completed
+
+
+Example of running schema validation on a test file that contains one valid record and one invalid record:
```
jacknagz: maybe shorten this to `failed record` or `failed test` example?

manage.py
ryandeivert: good catch!

stream_alert_cli/helpers.py
ryandeivert: These tests don't actually run on deploy/in production but good to keep in mind - thanks!

docs/source/rule-testing.rst
ryandeivert: unclear what you're looking for here? just `Schema validation on two valid test files:`?

docs/source/rule-testing.rst
ryandeivert: I think explaining what is actually in the output makes it more obvious to the user what is happening. Saying `Failed test` doesn't really explain what failed very well

PR #281, opened 2017-08-28
[streamalert] logs - update CB and osquery schemas

mime-frame:
to: @ryandeivert 

**What**

Commit schema changes upstream:
- CarbonBlack schema updates
- Support for osquery snapshots

**Testing**

```
StreamAlertCLI [INFO]: (14/14) Successful Tests
StreamAlertCLI [INFO]: (30/30) Alert Tests Passed
```

### DIFF
--- a/conf/logs.json
+++ b/conf/logs.json
@@ -7,7 +7,7 @@
       "md5": "string",
       "node_id": "string",
       "size": "string",
-      "timestamp": "float",
+      "timestamp": "string",
       "type": "string"
     },
     "parser": "json"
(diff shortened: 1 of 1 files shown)


conf/logs.json
```diff
@@ -835,6 +845,24 @@
       ]
     }
   },
+  "osquery_snapshot": {
```
jacknagz: why aren't we just namespacing like all the other types? 

`osquery:snapshot`

jacknagz: we should also be doing this for `osquery:status`

mime-frame:
@ryandeivert - can you stamp this, i'll merge, then i'll put up another PR that migrates to the `osquery:` prefix? The latter is a sizeable change that I'd like to do separately

Issue #157, opened 2017-05-23
Feature: Alert statistics via Cloudwatch


mime-frame:
Per suggestion by scott.piper:

> You want stats on your alerts.  How often did they fire? When? To whom?  You could probably fiddle with JIRA/PagerDuty to do this, but it'd be nice for something to automatically aggregate these into a report for you daily or weekly.

Goal:
- Use Cloudwatch metrics to capture this information

jacknagz:
This will most likely be achieved via QuickSight instead of CloudWatch.  

Athena tables will already have the data, which can be loaded and analyzed with QuickSight.  If we used CloudWatch we would potentially be duplicating stats.

mime-frame:
Closing as a result of #282

PR #282, opened 2017-08-29
[metrics] v2 of metrics support using metric filters

ryandeivert:
to @austinbyers or @jacknagz 
cc @airbnb/streamalert-maintainers & @mime-frame 
size: medium
contributes to: #157 

## Background
* After a v1 attempt in https://github.com/airbnb/streamalert/pull/247, it was discovered that the implementation had a major set back:
  * We quickly discovered that we were exceeding the 150 transactions-per-second allowed by [CloudWatch's PutMetricData limits](http://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_limits.html) and thus were being throttled practically immediately.
  * This is more speculative, but the latency of calling boto3 with the metric data (exacerbated by being throttled) may also cause a slight increase in the rule processor invocation time.

## Changes
* Migrate to utilizing CloudWatch Metric Filters to allow for filtering CloudWatch logs for specific terms and extracting values from those log entries.
* Metrics will now be compiled per-cluster and also sent to an aggregate metric that encompasses all clusters. For instance, the metric for `S3DownloadTime` will appear as:
  * `RuleProcessor-S3DownloadTime` (aggregate)
  * `RuleProcessor-S3DownloadTime-PROD` (prod cluster)
  * `RuleProcessor-S3DownloadTime-CORP` (corp cluster)
* Logs with metrics will default to logging json using: `json.dumps({'metric_name': metric_name, 'metric_value': value})`, resulting in lines like:
 ```
[INFO]	2017-08-29T06:12:52.56Z	89bd8174-cc61-4e64-862b-4b084de2281f
{
    "metric_name": "TriggeredAlerts",
    "metric_value": 2
}
```
* The new `MetricLogger` class can be extended to use non-default logging formats and specify different filter patterns/value lookup - but the defaults should cover most of our use cases for now.
* Example of the new metrics, namespaced under `StreamAlert` in CloudWatch metrics:
image
* Graph showing failed log parsing from data coming into 2 rule processors in different clusters:
image

### Caveat(s)
* We lose the ability to assign 'dimensions' to our CloudWatch metrics being posted, which allows for better segmentation of metric data in dashboard. For instance, we could break out metrics by functions name and cluster using dimensions.
  * The band-aid approach to make up for the lack of dimensions is prefixing metric names with the function name and adding a suffix for the cluster. We now will have metric names similar to: `RuleProcessor-FailedParses-PROD`

### Testing
* Deployed these changes in a test AWS account, utilizing a multiple cluster configuration to ensure all the metrics played well together. 

### Other
* Migrating some logging instantiation to within init - this is most in preparation of moving all logging into the `shared` package in the near future

### Upcoming
* Doc changes to demonstrate how to enabled/disable metrics
* CloudWatch alarm creation through the cli for any available metric filter.

### DIFF
--- a/stream_alert/athena_partition_refresh/__init__.py
+++ b/stream_alert/athena_partition_refresh/__init__.py
@@ -2,7 +2,7 @@
 import logging
 import os
 
-FUNCTION_NAME = 'athena_partition_refresh'
+from stream_alert.shared import ATHENA_PARTITION_REFRESH_NAME as FUNCTION_NAME
 
 # Create a package level logger to import
 LEVEL = os.environ.get('LOGGER_LEVEL', 'INFO').upper()
--- a/stream_alert/athena_partition_refresh/__init__.py
+++ b/stream_alert/athena_partition_refresh/__init__.py
@@ -1,0 +1,21 @@
+"""Initialize logging for the athena partition refresh function."""
+import logging
+import os
+
+FUNCTION_NAME = 'athena_partition_refresh'
+
+# Create a package level logger to import
+LEVEL = os.environ.get('LOGGER_LEVEL', 'INFO').upper()
+
+# Cast integer levels to avoid a ValueError
+if LEVEL.isdigit():
+    LEVEL = int(LEVEL)
+
+logging.basicConfig(format='%(name)s [%(levelname)s]: [%(module)s.%(funcName)s] %(message)s')
+
+LOGGER = logging.getLogger('StreamAlertAthena')
+try:
+    LOGGER.setLevel(LEVEL)
+except (TypeError, ValueError) as err:
+    LOGGER.setLevel('INFO')
+    LOGGER.error('Defaulting to INFO logging: %s', err)
--- a/stream_alert/shared/metrics.py
+++ b/stream_alert/shared/metrics.py
@@ -13,14 +13,25 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-from datetime import datetime
 import json
 import os
 
-import boto3
-from botocore.exceptions import ClientError
-
 from stream_alert.shared import LOGGER
+# Since this is a shared package, try to import the relevant processor name
+try:
+    from stream_alert.alert_processor import FUNCTION_NAME as ALERT_PROCESSOR
+except ImportError:
+    ALERT_PROCESSOR = 'alert_processor'
+
+try:
+    from stream_alert.athena_partition_refresh import FUNCTION_NAME as ATHENA_PROCESSOR
+except ImportError:
+    ATHENA_PROCESSOR = 'athena_partition_refresh'
+
+try:
+    from stream_alert.rule_processor import FUNCTION_NAME as RULE_PROCESSOR
+except ImportError:
+    RULE_PROCESSOR = 'rule_processor'
 
 CLUSTER = os.environ.get('CLUSTER', 'unknown_cluster')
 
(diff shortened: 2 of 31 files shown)

### CODE REVIEW

File: stream_alert_cli/terraform_generate.py [hunk 1]
```diff
@@ -18,6 +18,7 @@
 import os
 
 from stream_alert_cli.logger import LOGGER_CLI
+from stream_alert.shared import metrics
```
austinbyers: Nit: this would imported before `stream_alert_cli.logger`

terraform/modules/tf_stream_alert_athena/main.tf
```diff
@@ -97,3 +97,18 @@ resource "aws_cloudwatch_log_group" "athena" {
   name              = "/aws/lambda/${var.prefix}_streamalert_athena_partition_refresh"
   retention_in_days = 60
 }
+
+// CloudWatch metric filters for the athena partition refresh function
+// The split list is made up of: <filter_name>, <filter_pattern>, <value>
+resource "aws_cloudwatch_log_metric_filter" "athena_partition_refresh_cw_metric_filters" {
+  count          = "${length(var.athena_metric_filters)}"
+  name           = "${element(split(",", var.athena_metric_filters[count.index]), 0)}"
+  pattern        = "${element(split(",", var.athena_metric_filters[count.index]), 1)}"
```
austinbyers: Nice use of the `count`; this will be really helpful for me later

stream_alert_cli/terraform_generate.py
```diff
@@ -280,6 +281,56 @@ def generate_stream_alert(cluster_name, cluster_dict, config):
...
+def generate_cloudwatch_log_metrics(cluster_name, cluster_dict, config):
+    """Add the CloudWatch Metric Filters module to the Terraform cluster dict.
+
+    Args:
+        cluster_name (str): The name of the currently generating cluster
+        cluster_dict (defaultdict): The dict containing all Terraform config for a given cluster.
+        config (dict): The loaded config from the 'conf/' directory
+
+    Returns:
+        bool: Result of applying the cloudwatch metric filters to the stream_alert module
```
austinbyers: The function always returns True right now. Is that intended? Maybe you can clarify this

stream_alert_cli/terraform_generate.py
```diff
@@ -629,6 +683,27 @@ def generate_athena(config):
...
+    metric_prefix = 'AthenaRefresh'
+    filter_pattern_idx, filter_value_idx = 0, 1
+
+    # Add filters for the cluster and aggregate
+    # Use a list of strings that represnt the following comma separated values:
+    #   <filter_name>,<filter_pattern>,<value>
+    filters = ['{},{},{}'.format('{}-{}'.format(metric_prefix, metric),
+                                 settings[filter_pattern_idx],
+                                 settings[filter_value_idx])
+               for metric, settings in current_metrics[metrics.ATHENA_PROCESSOR].items()]
```
austinbyers: Nit: `.iteritems()`

stream_alert/shared/metrics.py
```diff
@@ -13,14 +13,25 @@
...
+
+try:
+    from stream_alert.athena_partition_refresh import FUNCTION_NAME as ATHENA_PROCESSOR
+except ImportError:
+    ATHENA_PROCESSOR = 'athena_partition_refresh'
+
+try:
+    from stream_alert.rule_processor import FUNCTION_NAME as RULE_PROCESSOR
+except ImportError:
+    RULE_PROCESSOR = 'rule_processor'
```
austinbyers: I don't really like these `try: except` imports with default overrides. This could lead to an inconsistency where the`MetricLogger._available_metrics` dict has different keys in each Lambda function if the default values here differ from the `FUNCTION_NAME`. That might not matter, but it makes it more difficult to reason about.

The `shared` code is uploaded with every Lambda function, right? Maybe we could define all function names in the shared module?

stream_alert/athena_partition_refresh/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+import os
+
+FUNCTION_NAME = 'athena_partition_refresh'
+
+# Create a package level logger to import
+LEVEL = os.environ.get('LOGGER_LEVEL', 'INFO').upper()
+
+# Cast integer levels to avoid a ValueError
+if LEVEL.isdigit():
+    LEVEL = int(LEVEL)
```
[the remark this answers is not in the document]
ryandeivert: Maybe not but this is just consistent with what we do in the other packages right now. Ideally we should validate the value set by the user within their cluster config (but that will come in a future PR for improved logging throughout). Good callout!

stream_alert_cli/terraform_generate.py
ryandeivert: @austinbyers I had suggested using `.items()` since `.iteritems()` is not a py3 thing. However you raise a good point that I didn't think of when I originally suggested this to @jacknagz . `.iteritems()` returns a generator while `.items()` returns a copy of the items as tuples.. Since that's the case we're probably better off using `.iteritems()` for the time being (so I take back what I previously said @jacknagz sorry about this !! :/)

stream_alert/shared/metrics.py
ryandeivert: That's not a bad suggestion - let me look into placing in `shared`. I was super confident with this approach anyway

stream_alert/shared/metrics.py
```diff
@@ -32,120 +43,74 @@
...
-                'Value': value,
-                "Dimensions": self._dimensions
-            }
-        )
-
-    def send_metrics(self):
-        """Public method for publishing custom metric data to CloudWatch."""
         if not ENABLE_METRICS:
-            LOGGER.debug('Sending of metric data is currently disabled.')
+            LOGGER.debug('Logging of metric data is currently disabled.')
```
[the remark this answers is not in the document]
ryandeivert: Not a bad idea - I'll make that change

stream_alert/shared/metrics.py
```diff
@@ -32,120 +43,74 @@
...
+            LOGGER.error('Metric name (\'%s\') not defined for \'%s\' function. Options are: %s',
+                         metric_name,
+                         lambda_function,
+                         ', '.join('\'{}\''.format(value)
+                                   for value in cls._available_metrics[lambda_function]))
+            return
 
-        self._put_metrics()
+        # Use a default format for logging this metric that will get picked up by the filters
+        LOGGER.info('%s', json.dumps({'metric_name': metric_name, 'metric_value': value}))
```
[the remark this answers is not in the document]
ryandeivert: @jacknagz Yes I have an s3 datasource now but haven't done any profiling to see how big the impact is. This json is simple enough that we could format it on our own (and that's actually the approach I was originally taking). I'll just switch to that to save the overhead

stream_alert_cli/terraform_generate.py
ryandeivert: @austinbyers @jacknagz  this is to conform to the same precedent set by Jack in all of the `generate...` functions. A good amount of them only return True, and I recall jack wanting to do it for consistency. Can you clarify @jacknagz

stream_alert_cli/terraform_generate.py
```diff
@@ -280,6 +281,56 @@ def generate_stream_alert(cluster_name, cluster_dict, config):
...
+        cluster_dict (defaultdict): The dict containing all Terraform config for a given cluster.
+        config (dict): The loaded config from the 'conf/' directory
+
+    Returns:
+        bool: Result of applying the cloudwatch metric filters to the stream_alert module
+    """
+    enable_metrics = config['global'].get('infrastructure',
+                                          {}).get('metrics', {}).get('enabled', False)
+    if not enable_metrics:
+        return True
```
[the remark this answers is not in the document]
ryandeivert: Because metrics are optional and returning `False` or `None` would exit the `generate_cluster` function above early which would be bad/no intended. IMHO I don't think the generate functions should return anything if they don't need to. A good amount `return True` with no possibility of ever returning false. See [this comment](https://github.com/airbnb/streamalert/pull/282/files/cdb9197a9ff92e4f2c94ba355882ea5916c62ae7..380819d181c93d456128179932996d9425ac3edb#r136194177)

stream_alert/shared/metrics.py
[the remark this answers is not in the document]
jacknagz: > we could format it on our own

Just so I'm clear, you mean just interpolate a string as JSON format?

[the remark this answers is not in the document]
ryandeivert: Yes @jacknagz exactly. Like this:

`LOGGER.info('{"metric_name": "%s", "metric_value": %s}', metric_name, value)`

stream_alert_cli/terraform_generate.py
austinbyers: @jacknagz pylint cares about the order of each import group (stdlib then third-party packages), but I don't think it cares about the order within each group

Issue #283, opened 2017-08-30
StreamAlertOutput [ERROR]: an error occurred during credentials decryption (Slack)


JJediny:
`
StreamAlertOutput [ERROR]: an error occurred during credentials decryption: {'ResponseMetadata': {'RetryAttempts': 0, 'HTTPStatusCode': 400, 'RequestId': 'e6a571c3-8da2-11e7-b06a-4573c2972fd5', 'HTTPHeaders': {'x-amzn-requestid': 'e6a571c3-8da2-11e7-b06a-4573c2972fd5', 'content-length': '39', 'server': 'Server', 'connection': 'keep-alive', 'date': 'Wed, 30 Aug 2017 16:47:25 GMT', 'content-type': 'application/x-amz-json-1.1'}}, 'Error': {'Message': '', 'Code': u'InvalidCiphertextException'}}
StreamAlertOutput [ERROR]: Failed to send alert to slack
`
### Steps to recreate
python manage.py lambda test --processor all :+1: 
python manage.py output new --service slack :+1: 
python manage.py lambda deploy --processor all :+1: 
python manage.py live-test --cluster prod :-1: 

This looks like a KMS error? is there a good way to debug output errors? Couldn't find any in the docs

ryandeivert:
hey @JJediny thanks for submitting the issue! I'm going to take a look at this today and will reach out to you on the StreamAlert slack with questions

JJediny:
To wrap this up - the issue was `python manage.py output new --service slack` was created to a different KMS store than was stored in tf.state so lookup/decrypt failed. This was caused by having to rename KMS twice - due to having to `terraform destroy` for unrelated reasons. AWS places a delete policy on KMS with a 30-day deletion protection, which doesn't allow for streamalert to reuse/re-create an already created KMS store (hence having to create an entirely new store). @ryandeivert good to close from me, really appreciate all the help from the team!

ryandeivert:
Thanks for the update @JJediny! Closing out

PR #284, opened 2017-08-30
[cli] raising exception if output creds could not be encrypted

ryandeivert:
to @jacknagz 

### Fix
* Raising an exception if credentials fail to encrypt with KMS. This used to allow execution to continue to the point of trying to upload an empty blob to S3, and then would fail with a different error. This helps differentiate failures.
* Updating unit tests for new exceptions being raised.


ryandeivert:
This wasn't actually related to a filed issue but was discovered when troubleshooting  #283

Issue #105, opened 2017-04-10
[q3] Feature: Normalization of Consumed Data


armtash:
From what I can see, data that is consumed is checked against the logs.json to validate the schema of the consumed data, however no mechanism exists to map data being consumed to a "common" or "standard" format.  Common formats include [CEF](https://www.protect724.hpe.com/docs/DOC-1072) (ArcSight standard) and [CIM](http://docs.splunk.com/Documentation/CIM/4.7.0/User/Overview) (Splunk standard).

Splunk's justification for CIM:
>  CIM helps you to normalize your data to match a common standard, using the same field names and event tags for equivalent events from different sources or vendors.

Mapping data to common format would make things like rule creation and integration with other tools more simplistic.  It could also help with IOC matching in https://github.com/airbnb/streamalert/issues/2.

From a rule creation perspective, rather than having to explicitly match against all possible keys: 

```
def bad_ip_address(rec):
    """Malicious IP Detected"""
    return (
        rec['destip'] == 8.8.8.8 or rec['destination_ip'] == 8.8.8.8 or rec['dst_ip'] == 8.8.8.8
    )
```
you could match against a single standard field (CEF):
```
def bad_ip_address(rec):
    """Malicious IP Detected"""
    return (
        rec['destinationAddress'] == 8.8.8.8
    )
```

mime-frame:
We'll be working on this, this quarter :)

armtash:
Awesome!

mime-frame:
Completed via https://github.com/airbnb/streamalert/pull/285

PR #285, opened 2017-08-31
[rule processor] Support data normalization

chunyong-lin:
to @ryandeivert @mime-frame 
cc @austinbyers 
size: medium
contributes to: #105 

## Background
Currently, logs from different log sources (CarbonBlack, Osquery etc) can capture the same types of information, but using different schemas or different key names. Thus, it requires writing a rule for every log source and referencing the specific key-names used. 

Example:
* Log1 from LogSource1: {"host": "bob-pc", "cmdline": "wget 'evil.com'"}
* Log2 from LogSource2: {"src_host": "bob-pc", "command": "wget 'hacked.com'"}

Using example above, it requires to write a rule for each log if you want to alert on any suspicious use of `wget`.  Example rules are like so:
```
Rule 1 for Log1 (pseudocode):
if rec['cmdline'] matches "wget *":
	return true
return false
```

```
Rule 2 for Log2 (pseudocode):
if rec['command'] matches "wget *":
	return true
return false
```

Data normalization feature will help you achieve writing one rule against all relevant logs. Applying data normalization, you can write one rule to alert on `wget` usage on logs shown above.
```
"""This is sample rule to alert on any suspicious use of wget"""
import fnmatch

@rule(datatypes=['command'])
def alert_suspecious_wget(rec):
    for result in results:
        if fnmatch(result, "wget *"):
            return true
    return false
```

We are using CEF (Common Event Format) standard to define normalized types. Please refer to [CEF documentation (PDF)](https://kc.mcafee.com/resources/sites/MCAFEE/content/live/CORP_KNOWLEDGEBASE/78000/KB78712/en_US/CEF_White_Paper_20100722.pdf) if you want to define your custom normalized types.

## Changes
* New configuration file types.json
 This configuration file defines normalized types for each log source. This example shows the normalized types define for logs from cloudwatch. 
```
{
  "cloudwatch":{
    "destinationAccount": ["recipientAccountId"],
    "destinationAddress": ["destination"],
    "destinationPort": ["destport"],
    "eventName": ["eventName"],
    "eventType": ["eventType"],
    "region": ["region"],
    "sourceAccount": ["account"],
    "sourceAddress": ["source", "sourceIPAddress"],
    "sourcePort": ["srcport"],
    "transportProtocol": ["protocol"],
    "userAgent": ["userAgent"],
    "userName": ["userName", "owner", "invokedBy"]
  }
}
```
* Data normalization logic is applied in rule processor. 

## Testing
* Add sample rules using data normalization feature and integration test. 
  * Integration test was passed.
* Add unit test. 
  * Unit test was passed. 
* Also it was tested in testing AWS account. 
  * Normalized type `sourceAddress` was working correctly
  * Normalized type `eventType` was working correctly
  * Normalized type `userName` was working correctly
### Upcoming
* Types validation (validate value of normalized types, `ipv4`, `domain`, `cmd`, etc.).
* Documentation about Data Normalization in details.


mime-frame:
For the `types`, I'm in favor of adopting a spec/standard vs. creating one ourselves.

Example: https://cee.mitre.org/language/1.0-beta1/core-profile.html

Granted it may not have everything we need, but we can borrow 90%+ of it.

Other possibilities: https://github.com/vz-risk/veris, ...

helpers/base.py
```diff
@@ -87,3 +87,30 @@ def in_network(ip_address, cidrs):
...
+
+    Args:
+        rec (dict): parsed payload of any log
+        datatype (str): normalized type user interested
+
+    Returns:
+        (list) The values of normalized types
+    """
+    results = []
+    if not datatype in rec['normalized_types'].keys():
```
austinbyers: Nit: the `.keys()` is superfluous and you can remove it

rules/community/cloudtrail/cloudtrail_aws_access_by_evil.py
```diff
@@ -0,0 +1,26 @@
+"""Alert on matching IP address from aws access."""
+from stream_alert.rule_processor.rules_engine import StreamRules
+from helpers.base import fetch_values_by_datatype
```
austinbyers: Nit: To be alphabetical, the `helpers.base` import should come first

rules/community/cloudtrail/cloudtrail_aws_access_by_evil.py
```diff
@@ -0,0 +1,26 @@
...
+    description:      This is sample rule to get alert by using normalized type
+                      "ipaddress".
+    """
+
+    results = fetch_values_by_datatype(rec, 'ipv4')
+
+    for result in results:
+        if result == '1.1.1.2':
+            return True
+    return False
```
austinbyers: 🎉  This is a great example that helped me understand how the normalization can be used

rules/community/cloudtrail/cloudtrail_aws_access_by_evil.py
```diff
@@ -0,0 +1,26 @@
...
+rule = StreamRules.rule
+
+
+@rule(logs=['cloudwatch:events'],
+      matchers=[],
+      datatypes=['ipv4'],
+      outputs=['aws-s3:sample-bucket',
+               'pagerduty:sample-integration',
+               'slack:sample-channel'])
+def cloudtrail_aws_access_by_evil(rec):
```
mime-frame: We no longer put "sample" rules in the public repository, only real ones. I'll work with you offline to choose which ones to open source :)

stream_alert/rule_processor/config.py
```diff
@@ -88,6 +88,14 @@ def _validate_config(config):
                 raise ConfigError(
                     'List of \'logs\' is empty for entity: {}'.format(entity))
 
+    # validate supported normalized types
+    supported_logs = [
+        'carbonblack', 'cloudwatch', 'cloudtrail', 'ghe', 'osquery', 'pan'
```
mime-frame: a hardcoded list? Doesn't seem right... why not pull from the config?

stream_alert/rule_processor/rules_engine.py
```diff
@@ -154,6 +157,81 @@ def match_event(cls, record, rule):
...
+                {
+                    "defined_type1": [original_key1],
+                    "defined_type2": [[original_key2, sub_key2], original_key3]
+                }
+        """
+        results = dict()
+        if not (datatypes and cls.validate_datatypes(normalized_types, datatypes)):
+            return results
+
+        for key, val in record.iteritems(): # pylint: disable=too-many-nested-blocks
```
mime-frame: hah, that pylint warning exists for a reason :)

Here's how to avoid having too much nesting: https://blog.rburchell.com/2010/11/coding-antipatterns-excessive-nesting.html

stream_alert/rule_processor/rules_engine.py
```diff
@@ -154,6 +157,81 @@ def match_event(cls, record, rule):
...
+                    if key in original_key_names:
+                        if not datatype in results.keys():
+                            results[datatype] = [key]
+                        else:
+                            results[datatype].append(key)
+        return results
+
+    @classmethod
+    def validate_datatypes(cls, normalized_types, datatypes):
+        """validate if datatypes valid in normalized_types for certain log
```
mime-frame: grammar, not sure what this is trying to say

helpers/base.py
```diff
@@ -87,3 +87,30 @@ def in_network(ip_address, cidrs):
...
+        return results
+
+    for key in rec['normalized_types'][datatype]:
+        # Normalized type may be in nested subkeys, we only support one level of
```
mime-frame: To confirm, this will work for cases where `envelope_keys` is used, correct?

stream_alert/rule_processor/config.py
chunyong-lin: This code block is to validate if the log sources defined in the `types.json` supported. I will remove this code, looks redundant. We assume the log sources defined in `types.json` are supported.

stream_alert/rule_processor/rules_engine.py
```diff
@@ -250,6 +328,11 @@ def process(cls, input_payload):
                 if not matcher_result:
                     continue
 
+                types_result = cls.match_types(record,
+                                               payload.normalized_types,
+                                               rule.datatypes)
+                record.update({'normalized_types': types_result})
```
ryandeivert: You don't have to use `.update` here. `update` is useful if you have a preexisting dictionary that you want to 'append' to another dict, but since you're only adding one key/value item to the dictionary just use:

`record['normalized_types'] = types_result`

stream_alert/rule_processor/rules_engine.py
```diff
@@ -154,6 +157,81 @@ def match_event(cls, record, rule):
...
+
+        Args:
+            normalized_types (dict): normalized_types for certain log
+            datatypes (list): defined in rule options, users interested types
+
+        Returns:
+            (boolean): return true if all datatypes are defined
+        """
+        if not normalized_types:
+            LOGGER.error('Normalized_types is empty.')
```
ryandeivert: Please use a human readable string in the logger statement here ('Normalized types dictionary is empty')

stream_alert/rule_processor/rules_engine.py
```diff
@@ -154,6 +157,81 @@ def match_event(cls, record, rule):
...
+            return False
+
+        for datatype in datatypes:
+            if not datatype in normalized_types.keys():
```
ryandeivert: The `.keys()` call here is unnecessary. Please utilize `if not datatype in normalized_types`

stream_alert/rule_processor/config.py
```diff
@@ -88,6 +88,14 @@ def _validate_config(config):
...
+    ]
+    for log_type in config['types'].keys():
```
ryandeivert: The `.keys()` call here is not required

helpers/base.py
chunyong-lin: Based on my live test (which has `envelope_keys` defined), it works for cases where `envelope_keys` is used.

chunyong-lin:
@ryandeivert @mime-frame @austinbyers PTAL.

conf/types.json
```diff
@@ -0,0 +1,69 @@
...
+    "sourceAccount": ["account", "accountId"],
+    "destinationAccount": ["recipientAccountId"],
+    "eventType": ["eventType"],
+    "eventName": ["eventName"],
+    "region": ["region", "awsRegion"],
+    "userAgent": ["userAgent"],
+    "sourceAddress": ["sourceIPAddress"]
+  },
+  "ghe": {
+    "processName": ["program"],
```
mime-frame: In GHE's case, `program` is not a real process, please remove this

conf/types.json
```diff
@@ -0,0 +1,69 @@
...
+    "userName": ["current_user"],
+    "destinationAddress": ["remote_address"],
```
mime-frame: I don't see an example GHE log with `remote_address` - lets chat offline

conf/types.json
```diff
@@ -0,0 +1,69 @@
...
+    "sourcePort": ["port"]
+  },
+  "osquery": {
+    "userName": ["username", "user"],
+    "filePath": ["path"],
+    "transportProtocol": ["protocol"],
+    "severity": ["severity"],
+    "environmentIdentifier": ["envIdentifier"],
+    "roleIdentifier": ["roleIdentifier"],
```
mime-frame: this is a custom decoration that only we do - remove

conf/types.json
```diff
@@ -0,0 +1,69 @@
...
+    "userName": ["current_user"],
+    "destinationAddress": ["remote_address"],
+    "sourcePort": ["port"]
+  },
+  "osquery": {
+    "userName": ["username", "user"],
+    "filePath": ["path"],
+    "transportProtocol": ["protocol"],
+    "severity": ["severity"],
+    "environmentIdentifier": ["envIdentifier"],
```
mime-frame: this is a custom decoration that only we do - remove

conf/types.json
```diff
@@ -0,0 +1,69 @@
...
+    "roleIdentifier": ["roleIdentifier"],
+    "command": ["cmdline", "command"],
+    "message": ["message"],
+    "sourceAddress": ["host", "source", "local_address", "address"],
+    "destinationAddress": ["destination", "remote_address", "gateway"],
```
mime-frame: make sure to account for values like `::1` when you write the validation classes

conf/types.json
chunyong-lin: It is defined in the [GHE schema](https://github.com/airbnb/streamalert/blob/master/conf/logs.json#L785).

conf/types.json
```diff
@@ -0,0 +1,66 @@
...
+    "region": ["region", "awsRegion"],
+    "userAgent": ["userAgent"],
+    "sourceAddress": ["sourceIPAddress"]
+  },
+  "ghe": {
+    "userName": ["current_user"],
+    "destinationAddress": ["remote_address"],
+    "sourcePort": ["port"]
+  },
+  "osquery": {
```
mime-frame: add `"fileHash": ["md5", "sha1", "sha256"]` (see https://osquery.io/docs/tables/#hash)

conf/types.json
```diff
@@ -0,0 +1,66 @@
...
+    "userName": ["username", "user"],
+    "filePath": ["path"],
```
mime-frame: add `directory` to the `filePath` array (see https://osquery.io/docs/tables/#hash)

conf/types.json
mime-frame: add `sourceUserId`, have the array contain `uid` (see https://osquery.io/docs/tables/#users)

mime-frame: add `receiptTime`, have the array contain `unixTime` (all osquery logs have this field, it denotes when the info was collected)

mime-frame: add `fileSize`, have it contain `size`. ex: https://osquery.io/docs/tables/#file_events

stream_alert/rule_processor/rules_engine.py
```diff
@@ -154,6 +157,127 @@ def match_event(cls, record, rule):
...
+                for datatype in datatypes:
+                    if key in normalized_types[datatype]:
+                        if not datatype in results:
+                            results[datatype] = [[key]]
+                        else:
+                            results[datatype].append([key])
+        return results
+
+    @classmethod
+    def update(cls, results, parent_key, nested_results):
```
ryandeivert: Can you add a description of the args to this function's docstring?

tests/unit/stream_alert_rule_processor/test_rule_helpers.py
```diff
@@ -73,3 +73,53 @@ def test_in_network():
 
     ip_not_in_cidr = '10.0.15.24'
     assert_equal(base.in_network(ip_not_in_cidr, cidrs), False)
+
+def test_fetch_values_by_datatype():
+    """Helpers - Fetch values from a record by normalized type"""
+    rec = {
```
ryandeivert: Can we remove the unicode (`u'..`) prefix from these strings, or are they required for your tests? They typically aren't necessary unless you explicitly need the strings as unicode. Please also remove from your test on line 125 below if you can.

stream_alert/rule_processor/rules_engine.py
```diff
@@ -154,6 +157,127 @@ def match_event(cls, record, rule):
...
+            normalized_types (dict): Normalized types
+            datatypes (list): normalized types users interested in.
+
+        Returns:
+            (dict): A dict of normalized_types with original key names
+        """
+        results = dict()
+        for key, val in record.iteritems():
+            if isinstance(val, dict):
+                nested_results = cls.match_types_helper(val, normalized_types, datatypes)
```
ryandeivert: Well done! 😸

stream_alert/rule_processor/rules_engine.py
```diff
@@ -219,6 +219,15 @@ def update(cls, results, parent_key, nested_results):
         """Update nested_results by inserting parent key to beginning of list.
         Also combine results and nested_results into one dictionary
 
+        Args:
+            results (dict): A dict of normalized_types with original key names
+            parent_key (str): Parent key of values in nested_results. The values
+                in nested_results are original keys of normalized types.
+            nested_results (dict): A dict of normalized_types from nested record
+
+        Returns:
```
ryandeivert: @chunyong-lin  this function doesn't actually 'return' anything. please remove this since it alters a dictionary by reference

chunyong-lin: Good catch! Thanks!

PR #286, opened 2017-08-31
[outputs] changing phantom container name to be rule name

ryandeivert:
to @mime-frame 
cc @airbnb/streamalert-maintainers , @fusionrace 

## Change
* Removing `StreamAlert Rule Triggered - ` prefix from container names in Phantom.

## Testing
* Unit tests passing locally


mime-frame:
image

Issue #287, opened 2017-08-31
Improvement: Phantom should write alerts to preexisting containers for triggering rule


ryandeivert:
### Issue
Currently, the PhantomOutput class will create a new container for every single alert that fires. This results in one artifact per container relationship and a ton of seemingly duplicate containers. 

### Proposal
Each rule would ideally have it's own container (full of alerts that fired for that rule). A container lookup should occur to determine if any containers for the triggering rule exist. If none already exist, a new one should be created.

PR #288, opened 2017-08-31
[lambda][alert] phantom container lookup by rule name

ryandeivert:
to @austinbyers 
size: small
resolves: #287

### Why
* The current implementation of the phantom output class will create a new container for every single alert that fires. This results in one artifact per container relationship, where each rule would ideally have it's own container (full of alerts that fired for that rule). (See #287)

### Changes
* This change performs a query using Phantom's api to filter on the name of the container.
  * In conjunction with the change made in #286, this will now do a container lookup via the 'name' value which will be the name of the rule that triggered the alert and return the ID for the first matching container.
* If not container exists with the name of this rule, a new one will be created and that ID will be returned. The 'artifact' will then be placed in the newly created container instead of the preexisting one.
* This query will be limited to 1 page to optimize the execution of this lookup. One page is all we need to determine 1) if a container/containers exists with this name and 2) get the ID of said container.

### Testing
* Tested locally and added unit test to ensure the proper URL is crafted for querying.

### DIFF
--- a/stream_alert/alert_processor/output_base.py
+++ b/stream_alert/alert_processor/output_base.py
@@ -197,7 +197,14 @@ class StreamOutputBase(object):
                 context = ssl.create_default_context()
                 context.check_hostname = False
                 context.verify_mode = ssl.CERT_NONE
-            request = urllib2.Request(url, data=data, headers=headers or {})
+
+            http_headers = headers or {}
+
+            # Omitting data means a GET request should occur, not POST
+            if not data:
+                request = urllib2.Request(url, headers=http_headers)
+            else:
+                request = urllib2.Request(url, data=data, headers=http_headers)
             resp = urllib2.urlopen(request, context=context)
             return resp
         except urllib2.HTTPError as err:
--- a/stream_alert/alert_processor/outputs.py
+++ b/stream_alert/alert_processor/outputs.py
@@ -132,8 +132,8 @@ class PagerDutyOutput(StreamOutputBase):
 class PhantomOutput(StreamOutputBase):
     """PhantomOutput handles all alert dispatching for Phantom"""
     __service__ = 'phantom'
-    CONTAINER_ENDPOINT = 'rest/container/'
-    ARTIFACT_ENDPOINT = 'rest/artifact/'
+    CONTAINER_ENDPOINT = 'rest/container'
+    ARTIFACT_ENDPOINT = 'rest/artifact'
 
     def get_user_defined_properties(self):
         """Get properties that must be asssigned by the user when configuring a new Phantom
--- a/stream_alert/alert_processor/outputs.py
+++ b/stream_alert/alert_processor/outputs.py
@@ -19,6 +19,7 @@ from collections import OrderedDict
 from datetime import datetime
 import json
 import os
+import urllib
 import uuid
 
 import boto3
@@ -164,8 +165,45 @@ class PhantomOutput(StreamOutputBase):
                             cred_requirement=True))
         ])
 
+    def _check_container_exists(self, rule_name, container_url, headers):
+        """Check to see if a Phantom container already exists for this rule
+
+        Args:
+            rule_name (str): The name of the rule that triggered the alert
+            container_url (str): The constructed container url for this Phantom instance
+            headers (dict): A dictionary containing header parameters
+
+        Returns:
+            int: ID of an existing Phantom container for this rule where the alerts
+                will be sent or False if a matching container does not yet exists
+        """
+        # Limit the query to 1 page, since we only care if one container exists with
+        # this name. This should not be a problem, but utilize urllib.quote to
+        # replace any unsafe characters in the rule_name
+        query_url = '{}?_filter_name="{}"&page_size=1'.format(container_url,
+                                                              urllib.quote(rule_name))
+
+        # Passing None for the data param ensures a GET request happens
+        resp = self._request_helper(query_url, None, headers, False)
+
+        if not self._check_http_response(resp):
+            return False
+
+        try:
+            resp_dict = json.loads(resp.read())
+        except ValueError as err:
+            LOGGER.error('An error occurred while decoding Phantom container query '
+                         'response to JSON: %s', err)
+            return False
+
+        # If the count == 0 then we know there are no containers with this name and this
+        # will evaluate to False. Otherwise there is at least one item in the list
+        # of 'data' with a container id we can use
+        return resp_dict and resp_dict['count'] and resp_dict['data'][0]['id']
+
     def _setup_container(self, rule_name, rule_description, base_url, headers):
-        """Establish a Phantom container to write the alerts to
+        """Establish a Phantom container to write the alerts to. This checks to see
+        if an appropriate containers exists first and returns the ID if so.
 
         Args:
             rule_name (str): The name of the rule that triggered the alert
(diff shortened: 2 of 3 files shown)


armtash:
Nice work Ryan.  I think that the Phantom cases might be of interest to you.  It would be something to allow for you to associate multiple containers into a single case.  

In a scenario with two similar streamalert matches, the process could be:

1. Create phantom container for the first rule fire.
2. Upon the second streamalert rule fire, perform query to determine if existing open containers exist.
3. If existing containers exist, create the 2nd container, then associate both the 1st and 2nd containers into a new case.

mime-frame:
Hey @armtash, thanks for the feedback! To help solidify our understanding, why would one want to create a new container for every alert fired? Granted, you can associate the containers for a given rule within a single case, but that seems overkill, whereas having all alerts from a given rule landing in the same container made more sense to us.

That being said, we're likely missing some knowledge that you have :) Any additional details or thoughts you can share would be very helpful :) Thanks!

armtash:
Hey @mime-frame,

Please let me know if you spot anything below that could be improved.

I suppose what I’m curious about is what happens to the results from StreamAlert that are condensed into a single container?  Does your analyst monitor this single container for changes?  Is there an additional ticketing system that the analyst interacts with outside of Phantom?  Is there a phantom playbook that was written to account for artifacts being condensed into a single container?  If you were feeding data to a ticketing system like Service Now or Jira, would you also attempt to contain artifacts into a single ticket?

Unfortunately, its really hard for me to say whether or not I think containing all rule matches into a single container is a good idea, because this is something that is totally subjective and dependent on your own use cases, processes, and downstream tools.  

With that said, I don’t like the idea of containing all rule matches into a single container, because our processes might be different from some other StreamAlert user.  For example, if I wanted to leverage case management within Phantom, and I establish a StreamAlert feed, I don’t necessary want a single container to continuously have artifacts appended to it.  If you guys do want to introduce this as a feature, I would recommend giving the user of StreamAlert the ability to turn this on or off.  This could be the default behavior, but when configuring the StreamAlert output, the user can specify whether or not to enable the aggregation of artifacts.

Case management might not be of interest to everyone, but in a scenario where I have an IDS like StreamAlert, and I leverage Phantom cases to also serve as my “ticketing system” (meaning no downstream tools), the process would look like this: 

1. An active playbook called “Find Similar Containers” exists, which runs against all containers created by StreamAlert.
2. The playbook will look at the current container, and query other Phantom cases to see if some criteria matches.  It could be querying against something like the StreamAlert rule name, or something a little more complicated like what I described in “Attempt 4” below.
3. If a suitable Phantom case does not exist, create it.  If similar containers are created after case creation, append to the existing case.

Now hypothetically speaking, if you did want to introduce a downstream ticketing system, you can essentially maintain a 1-1 relationship between the phantom case and a single ticket.


I spent a good chunk of time last year automating the ticket creation from our SIEM into our incident response ticketing platform, and I can see some overlap between these topics.  What became pretty obvious back then was the need to condense similar events, in order to make things more manageable for the incident responders.  I’m assuming that you guys have experienced some of the same things, which is why you've made the recent changes to condense the alerts into a single container.

Here is the what I did:

Attempt <span>#</span>1:  Generate a ticket for every SIEM alert.
Outcome: 

- Team hated me.  
- Too many tickets created.  
- Difficult associating tickets together after creation within ticketing system.  
- Not practical solution with rules that could generate multiple alerts per day.

Attempt <span>#</span>2:  For every alert from our SIEM, first check to see if an existing ticket exists which matches rule <span>#</span>/name.  If ticket exists, append to ticket.  Otherwise, create a new ticket.  This might sound familiar, as its pretty similar to the recent changes introduced to the Phantom output.
Outcome: 

- Tickets associated with a rule <span>#</span> were having events appended to them.
- Tickets that were months old were getting reactivated due to events that were not necessarily related to the original ticket’s events. 

Attempt <span>#</span>3: Everything in attempt <span>#</span>2, however we add some additional criteria when performing the query.  If an existing ticket exists, and the ticket status is “Closed”, then create a new ticket.  Only append to existing tickets if the rule<span>#</span> matches, and the status is “Open”.
Outcome: 

- A slight improvement, as unrelated closed tickets would not be “reactivated” by the new alerts.
- Still leaves the possibility for having only 1 open ticket at a time associated with a rule.  
- Unrelated events could still be associated within the single open ticket, requiring analysts to do extra work to make sense of things.

Now at this point, we had a better idea of how things should be grouped together.  What we realized though, was that its probably not enough to make a blanket statement and say, everything with X matching criteria should be bundled together.  Each use case might need to be treated in a slightly different way.  For example, Rule <span>#</span>1 might require us to bundle events based on the destination hostname.  Rule <span>#</span>2 might require us to bundle events together based on a matching CVE, so that all endpoints associated with said CVE would be grouped together.  Rule <span>#</span>3 might require us to bundle multiple criteria together, like the destination + http user agent or something like that.

The end product looked like this:
Attempt <span>#</span>4:
1.	For a particular rule, determine which fields you want to group things together by (i.e. dest ip, CVE, running process, etc.)  
2.	Check the ticketing system to see if there is an open ticket associated that matches said criteria.
3.	If open ticket exists, append to existing ticket.  If not, create new ticket.  Never append to closed.

Lets say you have 50 alerts generated by a single rule called “Possible C2 Communication”.  If all alerts for 50 devices are associated with the same C2 address, lets include those all in the same ticket. If we have 25 devices associated with  C2 address A, and 25 devices are associated with a different C2 address B, lets create two tickets.

Here is what the that could have translated from the ticketing perspective:

**Rule <span>#</span>1: “Possible C2 Communication” Grouping Criteria: Group by destination address.**

Ticketing system:
```
Ticket #1:          
Associated Rule: “Possible C2 Communication”  
Dest IP:  8.8.8.8
Source IP:  10.1.1.1, 10.2.2.2, 10.3.3.3, 10.4.4.4
Status: CLOSED
```
```
Ticket #2:          
Associated Rule: “Possible C2 Communication”  
Dest IP: 5.5.5.5
Source IP:  10.3.5.6, 10.2.7.4, 10.8.3.4
Status: OPEN
```
**“Possible C2 Communication” Hits:**
```
Match #1:
Dest IP:  5.5.5.5
Source IP:  10.5.2.6
Outcome: Append to ticket #1, as dst ip matches existing ticket.  Append 10.5.2.6 to the src IP field.
```
```
Match #2: 
Dest IP: 8.8.8.8
Source IP: 10.4.2.7
Outcome: Create a new ticket, Ticket #3, as an existing ticket exists for 8.8.8.8, but is closed.
```
```
Match #3: 
Dest IP: 8.8.8.8
Source IP: 10.4.2.7
Outcome: Append to existing ticket #3.
```

While this was an improvement from our original process, there were definitely areas that could be improved:
- Part of what I was doing was trying to replicate some of the functionality of a correlation engine.  Much of the heavy lifting associated with grouping specific items together could have more easily been done in a SIEM.  At this point, I don’t see StreamAlert as something that attempts to replicate this functionality, as you don’t have access to any historical data when writing a rule.  _It would definitely be cool if StreamAlert could correlate multiple types of data somehow.  A rule that a correlation engine could potentially support would be something like “In a 48 hour period, if Sally visits suspicious domain X, and Sally modifies sudoers on her device, generate an alert”._ 
- You may have noticed that events were directly associated with tickets.  This is something that really depends on your rules/use cases to be on point, otherwise you might be causing a lot of pain on the ticketing/analysis side of things.
- What we wanted was to introduce some automation to potentially do most of the low level analysis, even before a ticket was created.  This is partly why Phantom is an important component of our future IR processes.

I realize that I may have gone off on a tangent, but hopefully there is something in that blob of text that's helpful.

PR #292, opened 2017-09-01
[lambda][rule] logging rules that trigger

ryandeivert:
to @chunyong-lin 

## Change
* Adding logger info statement to log any rules that trigger.
* The statement will include the name of the rule, the log type, the entity and the service

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -253,6 +253,9 @@ class StreamRules(object):
                 # rule analysis
                 rule_result = cls.process_rule(record, rule)
                 if rule_result:
+                    LOGGER.info('Rule [%s] triggered an alert on log type [%s] from entity \'%s\' '
+                                'in service \'%s\'', rule.rule_name, payload.log_source,
+                                payload.entity, payload.service())
                     alert = {
                         'record': record,
                         'rule_name': rule.rule_name,


stream_alert/rule_processor/rules_engine.py
```diff
@@ -253,6 +253,9 @@ def process(cls, input_payload):
                 # rule analysis
                 rule_result = cls.process_rule(record, rule)
                 if rule_result:
+                    LOGGER.info('Rule triggered [%s] on log type [%s] from entity \'%s\' '
```
mime-frame: What does "triggered" mean? The rule was evaluated, or the rule evaluated to True?

ryandeivert: @mime-frame  In this case yeah - do you have a suggestion for better verbiage?

ryandeivert: Alert fired?

ryandeivert: Rule fired?

mime-frame: > In this case yeah

lol, which one :)

ryandeivert: Sorry missed that bit - what is `function decl`?

mime-frame: function decl = function declaration = all of the requirements list in the function declaration of the rule = `req_subkeys`, `logs`, ...

ryandeivert: gotcha - in that case, b - since b cannot be true without a. This log message is supposed to indicate that an alert has been fired and should be making its way to the alert processor via the rule > alert lambda invocation. There hasn't been a logger statement that does this until now.

mime-frame: Awesome, thanks for clarifying!  Perhaps then `Rule [%s] triggered an alert on log type [%s]...`

ryandeivert: okay - that works for me!

ryandeivert:
@chunyong-lin @mime-frame amended last commit with suggestion from @mime-frame

PR #303, opened 2017-09-08
streamalert - types - wrong command value for CB

mime-frame:
to: @chunyong-lin 

**What**

CB supports both `cmdline` and `command_line`

`carbonblack:feed.query.hit.process` -> `cmdline`

& 

`carbonblack:ingress.event.procstart` -> `command_line`

### DIFF
--- a/conf/types.json
+++ b/conf/types.json
@@ -1,6 +1,6 @@
 {
   "carbonblack":{
-    "command": ["cmdline"],
+    "command": ["command_line"],
     "destinationAddress": ["remote_ip"],
     "destinationDomain": ["domain"],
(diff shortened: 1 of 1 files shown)


ryandeivert:
@mime-frame I don't know if that's entirely true: https://github.com/airbnb/streamalert/blob/bc6e37156bc8860d930a5f9e94b7fe194e3a4c37/conf/logs.json#L650

mime-frame:
@ryandeivert , @chunyong-lin - updated PR description, it supports/uses both

ryandeivert:
@mime-frame cool! thanks :)

Issue #304, opened 2017-09-08
Bug: `logs` key required for rules that use `datatypes`


mime-frame:
**What**

https://github.com/airbnb/streamalert/pull/285 introduced data normalization.

An intended side effect was the the `logs` key was not required if the `datatypes` key is provided.

Example:

```
@rule(datatypes=['command'])
def alert_suspicious_wget(rec):
    results = fetch_values_by_datatype(rec, 'command')
    for result in results:
        if fnmatch(result, "wget *"):
            return true
    return false
```

However, this doesn't match reality. If you don't specify `logs`, you get the following:

`StreamAlert [ERROR]: Invalid rule [...] - rule must have 'logs' declared`

**Workaround**

Define all relevant logs in the `logs` key for rules that use data normalization.

**Expectations**

- if the `datatypes` key exists, `logs` is *not* required, _it's optional_
- if the `datatypes` key does *not* exist, `logs` is required

mime-frame:
This is now fixed

PR #307, opened 2017-09-11
[Data Normalization] Make `logs` optional when `datatypes` is defined

chunyong-lin:
to @ryandeivert @mime-frame 
size: tiny
contributes to: #304 

## Summary
Remove `log_source` constrain which means `logs` keyword is optional when we define a rule if `datatypes` keyword present. Otherwise `logs` keyword is required. 
Apply Data Normalization feature to a rule, we can define a rule in this format
```

@rule(datatypes=['command'])
def alert_suspicious_wget(rec):
    for result in results:
        if fnmatch(result, "wget *"):
            return true
    return false
```

## Testing
* Integration test is passed, it covers three cases
  * `logs` is present, `datatypes` is not
  * `logs` is present, `datatypes` is
  * `logs` is not present, `datatypes` is
* Unit test is passed, and it covers four cases
  * `logs` is present, `datatypes` is not - working as expected
  * `logs` is present, `datatypes` is - working as expected
  * `logs` is not present, `datatypes` is not - log an ERROR
  * `logs` is not present, `datatypes` is - working as expected.
* Testing in AWS testing account is passed

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -70,9 +70,10 @@ class StreamRules(object):
             datatypes = opts.get('datatypes')
             req_subkeys = opts.get('req_subkeys')
 
-            if not logs:
+            if not (logs or datatypes):
                 LOGGER.error(
-                    'Invalid rule [%s] - rule must have \'logs\' declared',
+                    'Invalid rule [%s] - rule must have either \'logs\' or \''
+                    'datatypes\' declared',
                     rule_name)
                 return
 
@@ -202,6 +203,8 @@ class StreamRules(object):
         """
         results = dict()
         for key, val in record.iteritems():
+            if key == 'normalized_types':
+                continue
             if isinstance(val, dict):
                 nested_results = cls.match_types_helper(val, normalized_types, datatypes)
                 cls.update(results, key, nested_results)
@@ -274,12 +277,10 @@ class StreamRules(object):
             (boolean): return true if all datatypes are defined
         """
         if not normalized_types:
-            LOGGER.error('Normalized types dictionary is empty.')
             return False
 
         for datatype in datatypes:
             if not datatype in normalized_types:
-                LOGGER.error('The datatype [%s] is not defined!', datatype)
                 return False
         return True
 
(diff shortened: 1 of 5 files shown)


stream_alert/rule_processor/rules_engine.py
```diff
@@ -379,11 +380,11 @@ def process(cls, input_payload):
                 matcher_result = cls.match_event(record, rule)
                 if not matcher_result:
                     continue
-
-                types_result = cls.match_types(record,
-                                               payload.normalized_types,
-                                               rule.datatypes)
-                record['normalized_types'] = types_result
+                if rule.datatypes is not None:
```
ryandeivert: Optional: The `is not None` here is implicit with just `if rule.datatypes:`

stream_alert/rule_processor/rules_engine.py
```diff
@@ -274,12 +277,10 @@ def validate_datatypes(cls, normalized_types, datatypes):
             (boolean): return true if all datatypes are defined
         """
         if not normalized_types:
-            LOGGER.error('Normalized types dictionary is empty.')
             return False
 
         for datatype in datatypes:
             if not datatype in normalized_types:
-                LOGGER.error('The datatype [%s] is not defined!', datatype)
```
ryandeivert: Is this logging statement no longer helpful? I understand why the one above was removed but can you explain why this one was also?

mime-frame:
@chunyong-lin - For the testing, please ensure there is a test for all possible options and update the PR description:

1) `logs` is present, `datatypes` is not
2) `logs` is present, `datatypes` is
3) `logs` is not present, `datatypes` is not - should log an ERROR
4) `logs` is not present, `datatypes` is

stream_alert/rule_processor/rules_engine.py
chunyong-lin: Since we removed `log_sources` constrain, one rule will matched to multiple records. Normalized types dictionaries are defined based on each log source though. It is not an error if normalized type `foo` doesn't defined in Record A, it just doesn't match and return `False` is enough. Does it make sense to you?

chunyong-lin:
@mime-frame @ryandeivert PTAL, I also updated my test status to PR description. Thanks 😸

mime-frame:
@chunyong-lin - to confirm, this is ready for review?

chunyong-lin:
@mime-frame Yes, it is ready for review.

PR #310, opened 2017-09-11
[lambda][rule] EXPERIMENTAL: schema matching now has better support log patterns

ryandeivert:
size: small

## This is *EXPERIMENTAL* for now so please do not merge
* We (and others) should test with a more impactful dataset before deploying this

## Changes 
* Enabling multiple schema matching by default and removing fallback logic that was used before.
* Type conversion now happens during schema matching
  * This change is in place to allow for type conversion to mark a schema as invalid for this log type and continue on to subsequent schemas.
  * Prior to this change, if type conversion failed there would be no way to continue to try other schemas and we would just assume this log was invalid.
* Log pattern matching is also enforced during schema matching now if `log_patterns` are defined for a schema instead of occurring after all matches were determined.

## Testing
* Updating unit tests and ran rule processing tests locally. This will have to be tested in a more realistic setup to determine if it performs well enough (ie - does not impact processing time too much).

### DIFF
--- a/stream_alert/rule_processor/classifier.py
+++ b/stream_alert/rule_processor/classifier.py
@@ -20,10 +20,6 @@ from stream_alert.rule_processor import LOGGER, LOGGER_DEBUG_ENABLED
 from stream_alert.rule_processor.parsers import get_parser
 from stream_alert.shared.stats import time_me
 
-# Set the below to True when we want to support matching on multiple schemas
-# and then log_patterns will be used as a fall back for key/value matching
-SUPPORT_MULTIPLE_SCHEMA_MATCHING = False
-
 
 class StreamClassifier(object):
     """Classify, map source, and parse a raw record into its declared type."""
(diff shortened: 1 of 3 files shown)


mime-frame:
Thanks for exploring this @ryandeivert

ryandeivert:
## Note: Execution time in local testing against a file with 19k+ records (s3 input)
### master
14.9212501049 seconds

### branch
29.085807085 seconds

ryandeivert:
When I get a chance I'll do some profiling and see where the biggest impacted areas are in code. This could still be a viable fix that just needs some additional tweaking.

austinbyers:
I would bet the type conversion is slowing it down, but I like the simpler logic. We could imagine a sort of pre-processing stage where a Lambda function is responsible for converting types and annotating which schema(s) a log could have matched.

Thanks for looking into this!

ryandeivert:
Closing this as it's stale and we can investigate further or other options in the future.

PR #320, opened 2017-09-25
Lambda to Firehose: Part 2

jacknagz:
to: @chunyong-lin  @ryandeivert 
size: medium
reference issue: #5 

## Background

The Part 1 changes (#280) introduced the CLI and Terraform code to enable creation of a Kinesis Firehose Delivery Stream per StreamAlert log type.

Part 2 focuses on the application logic from the Lambda functions to PutRecords into the delivery stream.

The desired flow is demonstrated below:

arch diagrams

### Infrastructure

* Attach a new policy to the `rule_processor` allowing write only privileges to all `streamalert_data_*` Firehose Delivery Streams.

### Application

* Load Firehose settings from `conf/global.json`
* Send batches of classified payloads to their respective Firehose Delivery Stream.  For example, if a record is classified as the type `cloudtrail:events`, it will be sent to the `streamalert_data_cloudtrail_events` Delivery Stream.
* Log metrics to understand Firehose send failures and throughput.
* Add unit tests to verify the functionality above.

### Other

* @chunyong-lin PTAL at https://github.com/airbnb/streamalert/commit/533bfbecfac556610a117496211897e0386abb7c
* Bump the library version to `1.5.0`.
* Fix two regressions: Removal of the Kinesis user/terraform outputs, and duplicate tf generate.


jacknagz:
note: the decreased coverage is due to a flakey unit test that has been disabled

terraform/modules/tf_stream_alert_kinesis_streams/main.tf
```diff
@@ -19,3 +19,37 @@ resource "aws_kinesis_stream" "stream_alert_stream" {
     Cluster = "${var.cluster_name}"
   }
 }
+
+// IAM User: stream_alert user for systems to send data to the stream
+resource "aws_iam_user" "stream_alert" {
+  name = "stream_alert_writeonly"
```
chunyong-lin: Can you reason about naming this IAM user name here? Is it renamed? I could not find the original name but the code been deleted is [here](https://github.com/airbnb/streamalert/blob/b3d62705293228f7aae1763d4675b7113dc8f17d/terraform/modules/tf_stream_alert_kinesis/users.tf)

jacknagz: It's renamed for clarity

jacknagz:
@chunyong-lin just added some clarifying commits, PTAL

jacknagz:
I have additional changes to be made, but I'll leave it at a Part 3 instead of appending more commits here:

- [ ] Allow only specific logs to be sent to Firehose (either in the form of inclusions/exclusions)
- [ ]  Sanitize special characters in key names 
- [ ] Size check the entire payload before sending

mime-frame:
🚢

stream_alert/rule_processor/handler.py
```diff
@@ -25,21 +28,26 @@
 from stream_alert.shared.metrics import MetricLogger
 
 
+MAX_RECORD_SIZE = 1000 * 1000
+MAX_BATCH_SIZE = 500
+
+
 class StreamAlert(object):
-    """Wrapper class for handling all StreamAlert classificaiton and processing"""
+    """Fascade for handling StreamAlert classificaiton and processing"""
```
ryandeivert: Why `Fascade`?

stream_alert/rule_processor/handler.py
```diff
@@ -25,21 +28,26 @@
...
+            context (dict): An AWS context object which provides metadata on the currently
                 executing lambda function.
             enable_alert_processor (bool): If the user wants to send the alerts using their
                 own methods, 'enable_alert_processor' can be set to False to suppress
                 sending with the StreamAlert alert processor.
         """
         # Load the config. Validation occurs during load, which will
         # raise exceptions on any ConfigErrors
-        config = load_config()
+        self.config = load_config()
```
ryandeivert: If we're going to start saving this on init, can we move it to a class level property and only call `load_config` if the `StreamAlert._config` isn't populated?

stream_alert/rule_processor/handler.py
```diff
@@ -133,6 +160,69 @@ def get_alerts(self):
         """
         return self._alerts
 
+    def _send_to_firehose(self):
+        """Put all classified records into its respective Firehose Delivery Stream
+
+        Args:
+            categorized_payloads (defaultdict): Categorized set of payloads by classified type
```
ryandeivert: I think this args list is outdated - it looks like you've made this an instance property instead

stream_alert/rule_processor/handler.py
```diff
@@ -133,6 +160,69 @@ def get_alerts(self):
...
+            # This same method is used when naming the Delivery Streams
+            formatted_log_type = log_type.replace(':', '_')
+
+            for record_batch in _chunk(records, MAX_BATCH_SIZE):
+                stream_name = delivery_stream_name_pattern.format(formatted_log_type)
+                _check_record_batch(record_batch)
+
+                resp = self.firehose_client.put_record_batch(
+                    DeliveryStreamName=stream_name,
+                    Records=[{'Data': json.dumps(record, separators=(",", ":")) + '\n'}
```
ryandeivert: Can you add a comment about the newline that's being appended here? Curious why this would be needed? Since strings are immutable, this will be creating a copy of the string dumped json every time just to add a new line to it, causing potentially unwanted memory usage and processing time. Just something to keep in mind if is nice to have but not required

stream_alert/rule_processor/handler.py
```diff
@@ -133,6 +160,69 @@ def get_alerts(self):
...
+        """
+        def _chunk(record_list, chunk_size):
+            """Helper function to chunk payloads"""
+            for item in range(0, len(record_list), chunk_size):
+                yield record_list[item:item + chunk_size]
+
+        def _check_record_batch(batch):
+            """Helper function to verify record size"""
+            for index, record in enumerate(batch):
+                if len(str(record).encode('utf-8')) > MAX_RECORD_SIZE:
```
ryandeivert: Is the encoding here required? Please add a comment if so, since it's not clear and seems to be only used for the size check (the encoded string isn't saved).

stream_alert/rule_processor/handler.py
```diff
@@ -133,6 +160,69 @@ def get_alerts(self):
...
+                                      for failed
+                                      in resp['RequestResponses']
+                                      if failed.get('ErrorCode')]
+                    MetricLogger.log_metric(FUNCTION_NAME,
+                                            MetricLogger.FIREHOSE_FAILED_RECORDS,
+                                            resp['FailedPutCount'])
+                    LOGGER.error('The following records failed to Put to the'
+                                 'Delivery stream %s: %s',
+                                 stream_name,
+                                 json.dumps(failed_records, indent=2))
```
ryandeivert: Can we do a check here to make sure we aren't trying to dump 5k failed records to the logger before doing it? Maybe add an error log with the total failed record count and then json dump a subset (first 5/last 5) or something?

stream_alert/rule_processor/handler.py
```diff
@@ -133,6 +160,69 @@ def get_alerts(self):
...
+                else:
+                    MetricLogger.log_metric(FUNCTION_NAME,
```
ryandeivert: Nice job on adding the custom metrics right away, this will be sick!

stream_alert_cli/terraform/streamalert.py
```diff
@@ -76,17 +74,19 @@ def generate_stream_alert(cluster_name, cluster_dict, config):
         'prefix': account['prefix'],
         'cluster': cluster_name,
         'kms_key_arn': '${aws_kms_key.stream_alert_secrets.arn}',
-        'rule_processor_enable_metrics': enable_metrics,
+        'rule_processor_enable_metrics': modules['stream_alert'] \
```
ryandeivert: Thanks for this fix 👍

stream_alert/rule_processor/handler.py
```diff
@@ -49,24 +57,38 @@ def __init__(self, context, enable_alert_processor=True):
...
         self._processed_size = 0
         self._alerts = []
 
+        # Create a dictionary to hold parsed payloads by log type.
+        # Firehose needs this information to send to its corresponding
+        # delivery stream.
+        self.categorized_payloads = defaultdict(list)
+
+        # Firehose initialization
+        firehose_config = self.config['global'].get(
```
ryandeivert: Can you just create the `self.firehose_client` object on the `init` and then have a method to handle instantiation during the `run` (instead of putting the if/else block in the init)

stream_alert/rule_processor/handler.py
jacknagz: Because the class exists to include other classes

stream_alert/rule_processor/handler.py
jacknagz: I'm actually not sure this needs to even be an instance attribute anymore, but I can still make that change

stream_alert/rule_processor/handler.py
jacknagz: Good catch

stream_alert/rule_processor/handler.py
jacknagz: It's just to ensure there's no weird characters in the log file

stream_alert/rule_processor/handler.py
ryandeivert: but fascade isn't a word? And also not a common term, while wrapper is both common and explicit

stream_alert/rule_processor/handler.py
jacknagz: It's required by Firehose or else all records will be written to a single line

stream_alert/rule_processor/handler.py
```diff
@@ -47,7 +48,7 @@ def __init__(self, context, enable_alert_processor=True):
         """
         # Load the config. Validation occurs during load, which will
         # raise exceptions on any ConfigErrors
-        self.config = load_config()
+        self.__config = getattr(StreamAlert, '__config', load_config())
```
ryandeivert: I don't think this code will ever actually sets the class property of `StreamAlert.__config` and thus it will always be empty and load_config() will still get called each time regardless.

Suggestion: avoid actually setting this value on the _instance_ and do a simple ternary-esque operator check like:

`StreamAlert.__config = StreamAlert.__config or load_config()`

And then always reference the config on the class, not instance

stream_alert/rule_processor/handler.py
ryandeivert: Can you add a comment saying it's required? I know I'll see that in the future and totally nix it without a comment

PR #322, opened 2017-09-26
Threat Intelligence Integration (alpha) with StreamAlert

chunyong-lin:
to: (@ryandeivert or @jacknagz) @mime-frame 
size: medium

## Background
StreamAlert is processing data (IP addresses, Domains, File Hashes, etc) from different log sources across production infrastructure, but it doesn't check if these data are known indicators of compromise (IOC) or not. This feature proposes to integrate Threat Intelligence (aka TI) to enable StreamAlert to alert on any data are matching to known indicators. The integration will help us get more proactive about potential security threats and inform us almost realtime if any indicator managed into production infrastructure. 

## Changes
* Intelligence (aka IOC) are saved in csv file and compressed in gz format.
* Rule processor load compressed intelligence file into memory.
* Add  configuration file for Threat Intel feature, where can enable and disable this feature (disabled by default). Also define a mapping between normalized types and IOC types.
* If there is IOC match, IOC information will be added to the record and sent to outputs. User is able to write a rule by checking if IOC field `rec['streamalert:ioc']` existed. 

## Important Note

This is an alpha feature that is gated by a flag - be very careful if you use this prior to an official cut release.

## Testing
* Add sample rules to detect on IOC
```

rule = StreamRules.rule


@rule(datatypes=['sourceAddress'],
      outputs=['aws-s3:sample-bucket'])
def test_detect_ioc(rec):
    """Testing rule to find is there any ip IOC matching"""
```
* Add unit test
  * Unit test was passed.
* It was tested in AWS testing account.
  * Testing rule was able to catch testing IP which is testing IOC.
#### Caveat
If Threat Intel feature is enabled, we should also bump memory size and timeout for rule processor lambda function. For example, if you have 1 million IOC on csv file (uncompressed), recommend to set rule processor `memory` to 256MB and `timeout` to 60 seconds at least in `conf/clusters/prod.json`.


conf/clusters/prod.json
```diff
@@ -31,8 +31,8 @@
         "current_version": "$LATEST",
         "enable_metrics": true,
         "log_level": "info",
-        "memory": 128,
-        "timeout": 10
+        "memory": 512,
```
mime-frame: Can we make this configurable, depending on whether threat intel is enabled, since you pay for this on every invocation? See: https://aws.amazon.com/lambda/pricing/

stream_alert/rule_processor/threat_intel.py
```diff
@@ -0,0 +1,87 @@
...
+    def read_compressed_files(self):
+        """Read all intelligence from csv.gz files located in threat_intel
+        directory into a dictionary. csv filename should follow the convention
+        that <ioc_type>.csv.gz. The basename (without extension) of csv file
+        will be the key in return dictionary.
+
+        Returns:
+            (dict): A dictionary stores intelligence in format
+                {
+                	"domain": {
```
mime-frame: spacing (this is indented more than `ip` and others

stream_alert/rule_processor/threat_intel.py
```diff
@@ -0,0 +1,87 @@
...
+                    },
+                    "ip": {
+                    	"1.1.1.2": ["scan_ip", "source reported ip1"],
+                        "2.2.2.2": ["scan_ip", "source reported ip2"]
+                    },
+                    "url": {
+                        "www.hacker.com/evil_page": ["mal_url", "source_foo"]
+                    },
+                    "md5": {
+                        "deadbeefdeadbeef": ["mal_md5", "soure_bar"]
```
mime-frame: please put an arbitrary md5 hash so it's more readable; also `source_bar` vs. `soure_var`

stream_alert/rule_processor/threat_intel.py
```diff
@@ -0,0 +1,87 @@
...
+
+        for gz_file in gz_files:
+            with gzip.open(gz_file, 'rb') as f:
+                csv_reader = csv.reader(f, delimiter=self._delimiter)
+                ioc_type = os.path.basename(gz_file).split('.')[0]
+                if ioc_type not in self._intel_dict:
+                    self._intel_dict[ioc_type] = dict()
+                for row in csv_reader:
+                    if len(row) < 2:
+                        LOGGER.debug('Warming, each row in CSV file should \
```
mime-frame: `Warming` -> `Warning`

tests/unit/stream_alert_rule_processor/test_config.py
```diff
@@ -112,3 +119,28 @@ def test_config_valid_types():
...
+        'deviceAddress': 'ip',
+        'fileHash': 'md5',
+        'sourceAddress': 'ip'
+    }
+    assert_equal(enable_threat_intel, False)
+    assert_equal(mapping, expected_mapping)
+
+@raises(ConfigError)
+def test_load_threat_intel_conf_nonexist():
+    """Config Error - Threat Intel configuration file is not existing"""
```
mime-frame: `Threat Intel configuration file is not existing` -> `Threat Intel configuration file does not exist`

conf/threat_intel.json
```diff
@@ -0,0 +1,10 @@
+{
+  "enabled": false,
+  "mapping": {
+    "destinationAddress": "ip",
+    "destinationDomain": "domain",
+    "deviceAddress": "ip",
+    "fileHash": "md5",
```
mime-frame: is `fileHash` always an MD5, or could it be a SHA1 or SHA256 hash?

conf/clusters/prod.json
chunyong-lin: You bring up a good point. This parameter is defined in the json format and used by terraform. We can not add logic here. Since Threat Intel feature is disabled by default, I will change `memory` size back to 128.

conf/threat_intel.json
chunyong-lin: Yes, it can be a SHA1 or SHA256, and `fileHash` is defined in [conf/types.com](https://github.com/airbnb/streamalert/blob/master/conf/types.json#L50). The IOC source uses `md5` designation for SHA and MD5.

stream_alert/rule_processor/threat_intel.py
chunyong-lin: 🤔  The alignment looks totally fine in my editor. I will re-format it.

conf/threat_intel.json
mime-frame: `The IOC source uses md5 designation for SHA and MD5.` - What do you mean, I don't follow :)

stream_alert/rule_processor/rules_engine.py
```diff
@@ -403,3 +418,36 @@ def process(cls, input_payload):
                     alerts.append(alert)
 
         return alerts
+
+    @classmethod
+    def get_intelligence(cls, intel_dir='threat_intel'):
```
ryandeivert: I'd suggest renaming this method from being prefixed with `get_` to being something like `load_intelligence` or `populate_intelligence`, since `get_..` implies it will return something

stream_alert/rule_processor/rules_engine.py
```diff
@@ -403,3 +418,36 @@ def process(cls, input_payload):
...
+        """Load intelligence from csv.gz files into a dictionary
+
+        Args:
+            intel_dir (str): Location where stores compressed intelligence
+        """
+        threat_intel = ThreatIntel(intel_dir)
```
ryandeivert: Move the instantiation of `ThreatIntel` to after the `if ... return` check since the object doesn't even need created if the dict is already present

stream_alert/rule_processor/threat_intel.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        if not os.path.exists(self._intel_dir):
+            return
+
+        gz_files = [os.path.join(self._intel_dir, gz_file)
+                    for gz_file in os.listdir(self._intel_dir)
+                    if gz_file.endswith('.gz')]
+
+        for gz_file in gz_files:
+            with gzip.open(gz_file, 'rb') as f:
```
ryandeivert: You can remove the 'b' from the open flags, since this file should not be binary and should only contain ASCII chars

conf/threat_intel.json
chunyong-lin: I will do move `enable_threat_intel` parameter into rule processor settings in each cluster right now. So that Threat Intel can be toggled per cluster based.

stream_alert/rule_processor/rules_engine.py
```diff
@@ -403,3 +419,36 @@ def process(cls, input_payload):
...
+        compromise) which store in class variable intelligence dictionary.
+
+        Args:
+            record (dict): Payload record to process
+
+        Return:
+            (tuple): (True or False, IOC type, IOC value)
+        """
+        _, datatypes_ioc_mapping = load_threat_intel_conf()
+        for datatype in rec['normalized_types']:
```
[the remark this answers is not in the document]
chunyong-lin: * `is_ioc` function is only called when there is `datatypes` defined in a rule.
* `rec` argument is a parsed and "half-processed" record, there is `normalized_types` key but its value could be empty when this record is passed to `is_ioc` function.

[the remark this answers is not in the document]
mime-frame: I'm pretty confused. Is the idea that

1) `is_ioc()` will be called from rules themselves, or

2) `is_ioc()` is handled internally within the processor, and it's just the responsibility of the user to define types and their mappings in JSON files and it just magically works

3) both (1) and (2)?

From your implementation, it seems like you went the route of `(2)`, since the only `is_ioc()` call can be found in `rules_engine.py`, and, it doesn't accept arguments that would support a use-case like `is_ioc(record['foobar'], IOC_TYPE::DOMAIN)` from within a rule.

If this is the case, your first bullet in your response to Jack is confusing. Why would `datatypes` need to be set in a rule in order for this to work?

Please clarify

stream_alert/rule_processor/threat_intel.py
```diff
@@ -0,0 +1,87 @@
...
+                        "2.2.2.2": ["scan_ip", "source reported ip2"]
+                    },
+                    "url": {
+                        "www.hacker.com/evil_page": ["mal_url", "source_foo"]
+                    },
+                    "md5": {
+                        "0123456789abcdef0123456789abcdef": ["mal_md5", "source_bar"]
+                    }
+                }
+            None: if location storing intelligence is not existed
```
mime-frame: `None: if the intelligence directory does not exist`

stream_alert/rule_processor/threat_intel.py
```diff
@@ -0,0 +1,87 @@
...
+                        "deadbeefdeadbeef": ["mal_md5", "soure_bar"]
+                    }
+                }
+            None: if location storing intelligence is not existed
+        """
+        if not os.path.exists(self._intel_dir):
+            return
+
+        gz_files = [os.path.join(self._intel_dir, gz_file)
+                    for gz_file in os.listdir(self._intel_dir)
```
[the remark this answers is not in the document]
mime-frame: also the indentation of this is larger than normal?

stream_alert/rule_processor/threat_intel.py
```diff
@@ -0,0 +1,87 @@
...
+        Returns:
+            (dict): A dictionary stores intelligence in format
+                {
+                    "domain": {
+                        "evil1.com": ["apt_domain", "source1 reported evil1.com"],
+                        "evil2.com": ["c2_domain", "source2 reported evil2.com"]
+                    },
+                    "ip": {
+                    	"1.1.1.2": ["scan_ip", "source reported ip1"],
+                        "2.2.2.2": ["scan_ip", "source reported ip2"]
```
mime-frame: extra indentation/space exists (doesn't line up with the line above it)

stream_alert/rule_processor/threat_intel.py
```diff
@@ -0,0 +1,87 @@
...
+        self._delimiter = delimiter
+
+    def read_compressed_files(self):
+        """Read all intelligence from csv.gz files located in threat_intel
+        directory into a dictionary. csv filename should follow the convention
+        that <ioc_type>.csv.gz. The basename (without extension) of csv file
+        will be the key in return dictionary.
+
+        Returns:
+            (dict): A dictionary stores intelligence in format
```
mime-frame: `(dict): Threat intelligence in the following format:`

stream_alert/rule_processor/threat_intel.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(self, intel_dir, delimiter=','):
+        """Initialize class instance variables
+        Args:
+            intel_dir (str): Location where stores compressed intelligence files
+
+        Instance variables:
+            self._intel_dict (dict): A dictionary to store all intelligence read
+                from *.csv.gz files
+            self._intel_dir (str):  Location where stores compressed intelligence
```
mime-frame: `self._intel_dir (str):  Directory path that contains the threat intelligence files`

stream_alert/rule_processor/threat_intel.py
chunyong-lin: Atom is super confusing these days 😢  Alignment looks well from in Atom. Just reformat it again.

stream_alert/rule_processor/threat_intel.py
chunyong-lin: pylint doesn't complaint this. Replace `f` with `ioc_file`.

stream_alert/rule_processor/threat_intel.py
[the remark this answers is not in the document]
chunyong-lin: Yep, pylint suggests this indentation.

stream_alert/rule_processor/rules_engine.py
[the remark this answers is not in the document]
chunyong-lin: @mime-frame route of (2) is the one I implemented. `is_ioc()` is handled internally, and it will not called from rules themselves. Users need to define types and their mappings in the JSON files as you mentioned in route (2). We can improve the user experience in near future. 

Currently, it doesn't support user-case like `is_ioc(record['foobar'], IOC_TYPE::DOMAIN)` within a rule.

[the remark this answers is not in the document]
mime-frame: Which means this comment:

> is_ioc function is only called when there is datatypes defined in a rule.

Is factually incorrect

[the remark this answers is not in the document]
mime-frame: Just wanted to ensure we're all on the same page.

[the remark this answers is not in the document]
chunyong-lin: > is_ioc function is only called when there is datatypes defined in a rule.

I probably should reword this comment to `is_ioc function is working with Data Normalization currently.`
In current implementation stage, Threat Intel is strongly tight with Data Normalization. Without Data Normalization we have to define a bigger map between original keys in each schema and IOC types. 

Does it make sense to you?

chunyong-lin:
@ryandeivert @jacknagz @mime-frame PTAL, thanks!

tests/unit/stream_alert_rule_processor/test_config.py
```diff
@@ -114,3 +120,15 @@ def test_config_valid_types():
     config = load_config()
 
     _validate_config(config)
+
+# def test_threat_intel_is_enabled():
```
mime-frame: ??

stream_alert/rule_processor/handler.py
```diff
@@ -93,6 +97,12 @@ def run(self, event):
         if not records:
             return False
 
+        threat_intel_config = dict()
+        if (self.__config.get('threat_intel')
```
jacknagz: Do you need those parens?

stream_alert/rule_processor/rules_engine.py
```diff
@@ -385,6 +388,24 @@ def process(cls, input_payload):
...
+                    # to outputs.
+                    if threat_intel_config and threat_intel_config.get('enabled'):
+                        ioc_result, ioc_type, ioc_value = cls.is_ioc(
+                            record,
+                            threat_intel_config.get('mapping'))
+
+                        # if there is ioc mathing, append this ioc_type and
+                        # ioc_value to the record which will be sent to outputs
+                        if ioc_result:
+                            record['streamalert_ioc'] = {
```
jacknagz: use `streamalert:ioc` to be consistent with other `streamalert` appended keys (as mentioned in my previous comment)

tests/unit/stream_alert_rule_processor/test_handler.py
```diff
@@ -360,3 +360,11 @@ def put_record_batch(**kwargs):
 
         self.__sa_handler.run(test_event)
         assert_true(mock_logging.error.called)
+
+    @patch('stream_alert.rule_processor.handler.StreamRules.load_intelligence')
```
jacknagz: Side note: we need to figure out a better story with end-to-end unit tests.  We only use a single config ('/test/unit/conf') in our test cases, however that doesn't represent all the possible combinations in the real world, which affects how logic is forked.

stream_alert/rule_processor/handler.py
chunyong-lin: Yes, I do need those parenthesizes. Pylint is very helpful and it will warm if there is unnecessary parenthesizes.

stream_alert/rule_processor/rules_engine.py
chunyong-lin: Yes, I tried to use `streamalert:ioc` in the beginning, and I do agree we should keep naming consistent. But I encountered strange situation that key `streamalert:ioc` can not be inserted to the record unless I have backslash before colon, `streamalert\:ioc`. However, pylint complaint `Anomalous backslash in string`. So I gave up and used `streamalert_ioc` instead.

stream_alert/rule_processor/rules_engine.py
```diff
@@ -420,34 +426,36 @@ def process(cls, input_payload):
...
         """
-        _, datatypes_ioc_mapping = load_threat_intel_conf()
-        for datatype in rec['normalized_types']:
-            if datatype not in datatypes_ioc_mapping:
-                continue
-            results = fetch_values_by_datatype(rec, datatype)
-            for result in results:
-                if result in cls.__intelligence[datatypes_ioc_mapping[datatype]]:
-                    return True, datatypes_ioc_mapping[datatype], result
+        if datatypes_ioc_mapping:
```
ryandeivert: suggestion here to avoid multi-level nesting so much would be to return early if possible. ie:
```
if not datatypes_ioc_mapping:
    return False, None, None
```

stream_alert/rule_processor/threat_intel.py
```diff
@@ -61,27 +63,27 @@ def read_compressed_files(self):
...
+            with gzip.open(gz_file, 'r') as ioc_file:
+                csv_reader = csv.reader(ioc_file, delimiter=self._delimiter)
                 ioc_type = os.path.basename(gz_file).split('.')[0]
-                if ioc_type not in self._intel_dict:
-                    self._intel_dict[ioc_type] = dict()
+                if ioc_type not in self._intel_dictionary:
+                    self._intel_dictionary[ioc_type] = dict()
                 for row in csv_reader:
                     if len(row) < 2:
                         LOGGER.debug('Warning, each row in CSV file should \
```
ryandeivert: The `\` in the line above should be avoided like so:
```
                    LOGGER.debug('Warning, each row in CSV file should'
                                 'contain at least two fields. Bad row [%s]', row)
```
Also, can the hanging `row)` just be placed on the preceding line, like above ^ ?

stream_alert/rule_processor/threat_intel.py
```diff
@@ -0,0 +1,89 @@
...
+
+        for gz_file in gz_files:
+            with gzip.open(gz_file, 'r') as ioc_file:
+                csv_reader = csv.reader(ioc_file, delimiter=self._delimiter)
+                ioc_type = os.path.basename(gz_file).split('.')[0]
+                if ioc_type not in self._intel_dictionary:
+                    self._intel_dictionary[ioc_type] = dict()
+                for row in csv_reader:
+                    if len(row) < 2:
+                        LOGGER.debug('Warning, each row in CSV file should \
```
ryandeivert: The `\` in the line above should be avoided like so:
```
                    LOGGER.debug('Warning, each row in CSV file should'
                                 'contain at least two fields. Bad row [%s]', row)
```
Also, can the hanging `row)` just be placed on the preceding line, like above ^ ?

stream_alert/rule_processor/rules_engine.py
```diff
@@ -403,3 +426,38 @@ def process(cls, input_payload):
...
+        A class variable is defined to store IOCs (Indicator of Compromise) in a
+        dictionary.
+
+        Args:
+            rec (dict): Payload record to process
+
+        Return:
+            (tuple): (True or False, IOC type, IOC value)
+        """
+        if datatypes_ioc_mapping:
```
ryandeivert: suggestion here to avoid multi-level nesting so much would be to return early if possible. ie:
```
if not datatypes_ioc_mapping:
    return False, None, None
```

stream_alert/rule_processor/threat_intel.py
chunyong-lin: Thanks for the suggestion. I removed `\`. Unfortunately, `row)` will make preceding line too long, I will keep it hanging there 😄

ryandeivert: move `contain` to the previous line? 🤣

Issue #278, opened 2017-08-24
Support ingestion of logs via fetching/polling from RESTful APIs


ryandeivert:
In its current state, StreamAlert is able to receive log data via a *push*-like approach, wherein other services (AWS SNS, Kinesis, S3) are able to *send* data into the Rule Processor lambda function for processing.

This proposal is to incorporate a flexible *poll*-like model into StreamAlert that would allow for scheduled fetching of logs via RESTful APIs that could then be handed off to the Rule Processor lambda function for processing.

The idea is that this would be extensible and offer the ability to support practically any generic SaaS application, RESTful endpoint, etc. A few bullet point use cases could include retrieving logs from services provides such as OneLogin, DUO Security, Google Admin, etc and batching/formatting them in a way that the Rule Processor could easily ingest & process.

glaslos:
Shouldn't a scheduled event triggering lambda to fetch and prepare data be something that I create to feed data into "Stream"Alert? This proposal sounds a bit against the concept of serverless and add a data producer to something I considered a pure data consumer.

jacknagz:
@ryandeivert:

> In its current state, StreamAlert is able to receive log data via a push-like approach, wherein other services (AWS SNS, Kinesis, S3) are able to send data into the Rule Processor lambda function for processing.

It's worth noting that Kinesis is actually a polling method.  Lambda checks the stream every 250ms and a static [batch of records](https://github.com/airbnb/streamalert/blob/master/terraform/modules/tf_stream_alert_kinesis_events/main.tf#L30) is analyzed.

@glaslos: 
> Shouldn't a scheduled event triggering lambda to fetch and prepare data be something that I create to feed data into "Stream"Alert? 

I believe @ryandeivert is suggesting the creation of a framework that would be generic enough for most use cases to avoid duplication of effort.  Users always have the option of creating their own producer by whichever preferred means.

> This proposal sounds a bit against the concept of serverless and add a data producer to something I considered a pure data consumer.

Scheduled jobs can run in Lambda just fine via [CloudWatch Scheduled Events](http://docs.aws.amazon.com/lambda/latest/dg/with-scheduled-events.html), this is actually how you connect a Lambda function to a SQS Queue.

PR #323, opened 2017-09-27
[apps] log poll function initial implementation

ryandeivert:
to @jacknagz & @austinbyers for py module code, to @chunyong-lin  for tf and py cli code
cc @airbnb/streamalert-maintainers , @mime-frame 
size: muy large
resolves: #278 

### Background
* See #278 for context. In its current state, StreamAlert is only able to receive logs from services that offer a push model. This adds the ability to 'poll/fetch' logs on a given interval and have them forwarded to the rule processor for processing.

### Things to look closely for
* Please make sure there is no copy/pasta hanging around. All of this code went through a massive rename from gatherer --> app (sigh) and there's a chance some cruft was left behind.
  * **NOTE**: Function names like `gather_logs` are meant to be that way. I'm more concerned about comments/dictionary key names/etc.

### Changes
* Updating Terraform minimum version from 0.9.6  --> 0.10.6 and adding AWS provider version of 0.1.4.
* Adding a new top level package (separate from the `stream_alert` package) named `app_integrations` to house the various 'apps' and relevant code.
* The `apps` are located in `app_integrations/apps`
  * Each app (or, more importantly, service) should reside in its own file for that specific services (ie: `duo`, `onelogin`, etc)
  * There is a supporting base class named `AppIntegration` within `app_integrations/apps/app_base` that all new apps should be subclassed from, and decorated with the `@app` decorator. No additional code is necessary to make sure the apps are loaded.
* The `config` file house an `AppConfig` class that handles all the config related operations. This is a subclass of `dict` with helper methods for saving state to parameter store, etc.
* The 'batcher' file contains a class (aptly named `Batcher`) that will performed log segmentation to ensure the App is never trying to send data payloads that are too large the the Rule Processor Lambda function.
* Updated the Rule Processor function with small changes to support incoming data from `apps` (ie: `StreamAlertAppPayload` for pre-parsing these payloads, etc).
* Added a `tf_stream_alert_app` Terraform module for handling all Terraform operations for these app functions.
* Updated `manage.py` with code necessary for adding, listing, and updating any configured App integrations.
  * The CLI package has been updated to handle creating/updating the config for `stream_alert_apps`.

### How
* **IMPORTANT**: An initial deploy of `StreamAlert` should be performed before configuring any app integrations.
* An App integration configuration is add via the `manage.py` cli script, like so:
```
     manage.py app new \
      --type duo_auth \
      --cluster prod \
      --name duo_prod_collector \
      --interval 'rate(2 hours)' \
      --timeout 80 \
      --memory 128
```
* The app integrations configured per-cluster and multiple 'types' for each cluster can be added.
* Adding a configuration will walk the user through required authentication/secret information needed from the user. This info is sent to Parameter Store as type `SecureString` for storage. That process looks similar to this:
```
$ python manage.py app new --type duo_auth --cluster prod --name duo_prod_collector --interval 'rate(2 hours)' --timeout 80 --memory 128

Please supply the API url for this duosecurity instance. This should be in a format similar to 'api-abcdef12.duosecurity.com': api-abcdef12.duosecurity.com

Please supply the secret key for this duosecurity Admin API. This should a string of 40 alphanumeric characters: 123424af2ae101d47d9704b783c940dffa825678

Please supply the integration key for this duosecurity Admin API. This should be in a format similar to 'DIABCDEFGHIJKLMN1234': DIABCDEFGHIJKLMN1234

StreamAlertCLI [INFO]: App authentication info successfully saved to parameter store.
StreamAlertCLI [INFO]: Successfully added 'duo_prod_collector' app integration to 'conf/clusters/prod.json' for service 'duo_auth'.
```
* After adding a configuration, the app function itself must be deployed. This is accomplished via: `python manage.py lambda deploy --processor apps`
* Configuring an app will automatically update the `sources.json` with the proper corresponding source/log info. However, in order for the `sources.json` to be updated within the current rule processor, the rule processor must also now be re-deployed.
* Two other `manage.py` cli commands have been added with this PR. They are:
  * `$ python manage.py app update-auth` which outputs prompts to update auth information in Parameter Store. And,
  * `$ python manage.py app list`, outputting a list of configured apps grouped by cluster:
```
Cluster: prod

	Name: duo_prod_collector
		log_level:           	info
		interval:           	rate(2 hours)
		timeout:           	60
		memory:           	256
		current_version:        $LATEST
		type:           	duo_auth
```

### Testing
* This has been deployed in a test AWS account with a mocked Duo Admin API to retrieve logs to be sent to the Rule Processor. The end-to-end fetching/forwarding on the designated interval is working as desired.
* Some unit tests have been added now, but more will be coming in the next day or so to increase coverage on the function.

#### Caveats 
* Users are limited to using AWS `rate` intervals for their apps and cannot use `cron` intervals at this time.
* Adding a new App configuration updates the `conf/sources.json` automatically, but the Rule Processor does not get deployed every time the Apps function(s) are deployed. For the rule processor to be able to properly parse logs coming from an App integration, it must be redeployed.
  * Therefore, a deploy targeting both `apps` and `rule` as 'processors' is recommended.

#### Future Work
* Adding a local integration testing pipeline to ensure the App integrations are sending properly formatted data to the Rule Processor function for processing.


app_integrations/apps/duo.py
```diff
@@ -0,0 +1,209 @@
...
+        if not self._check_http_response(response):
+            return False
+
+        logs = response.json()['response']
+
+        # Get the timestamp from the latest event. Duo produces these sequentially
+        # so we can just extract the timestamp from the last item in the list
+        self._last_timestamp = logs[-1]['timestamp']
+
+        # Check if the max amount of logs was returned with this request. If the value
```
mime-frame: > If the value is not the max, then we are done polling logs for this timeframe`

Did you mean to say `If the value is the max`?

app_integrations/apps/duo.py
```diff
@@ -0,0 +1,209 @@
...
+        self._more_to_poll = len(logs) >= self._MAX_RESPONSE_LOGS
+
+        # Return the list of logs to the caller so they can be send to the batcher
+        return logs
+
+    def required_auth_keys(self):
+        return {
+            'api_hostname':
+                {
+                    'description': ('the API url for this duosecurity instance. This should '
```
mime-frame: `url` -> `hostname`

app_integrations/batcher.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # Try to send all of the logs in one fell swoop
+        if self._send_logs_to_stream_alert(source_function, logs):
+            return
+
+        # Fall back on segmenting the list of logs into multiple requests
+        # if they could not be sent at once
+        return self._segement_and_send(source_function, logs)
+
+    def _segement_and_send(self, source_function, logs):
```
mime-frame: `segment` vs. `segement` (update all other references)

app_integrations/batcher.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        log_count = len(logs)
+        LOGGER.debug('Segmenting %d logs into subsets', log_count)
+
+        segment_size = int(math.ceil(log_count / 2.0))
+        for index in range(0, log_count, segment_size):
+            subset = logs[index:segment_size + index]
+            # Try to send this current subset to the rule processor
+            # and segment again if they are too large to be sent at once
+            if not self._send_logs_to_stream_alert(source_function, subset):
```
mime-frame: If a log exceeds the size of `MAX_LAMBDA_PAYLOAD_SIZE`, `_send_logs_to_stream_alert` will return False. `not` of `False` is True, so `_segement_and_send` will be called, and that function doesn't check MAX_LAMBDA_PAYLOAD_SIZE?

app_integrations/batcher.py
```diff
@@ -0,0 +1,122 @@
...
+            response = Batcher.LAMBDA_CLIENT.invoke(
+                FunctionName=self.rp_function,
+                InvocationType='Event',
+                Payload=payload_json,
+                Qualifier='production'
+            )
+
+        except ClientError as err:
+            LOGGER.error('An error occurred while sending logs to '
+                         '\'%s:production\'. Error is: %s',
```
mime-frame: why does this have a `:production` suffix?

stream_alert/shared/metrics.py
```diff
@@ -54,6 +54,7 @@ class MetricLogger(object):
     TOTAL_PROCESSED_SIZE = 'TotalProcessedSize'
     TOTAL_RECORDS = 'TotalRecords'
     TOTAL_S3_RECORDS = 'TotalS3Records'
+    TOTAL_STREAM_ALERT_APP_RECORDS = 'TotalStreamAlertAppRecords'
```
mime-frame: Awesome, love that we'll have metrics!

app_integrations/config.py
```diff
@@ -0,0 +1,362 @@
...
+        return {
+            'region': arn[3],
+            'account_id': arn[4],
+            'function_name': arn[6],
+            'qualifier': arn[7]
+        }
+
+    def _determine_last_time(self):
+        """Determine the last time this function was executed and fallback on
+        evaluting the rate value if there is no last timestamp available
```
mime-frame: `evaluating`

app_integrations/main.py
```diff
@@ -0,0 +1,47 @@
...
+
+def handler(event, context):
+    """Main lambda handler use as the entry point
+
+    Args:
+        event (dict): Always empty (for now) event object
+        context (LambdaContxt): AWS LambdaContext object
+    """
+    if event and 'full_run' in event:
+        # TODO: implement support for historical runs via input events
```
mime-frame: i think we've traditionally raised an exception when an unsupported/future feature was used?

ryandeivert: There's really no point in raising the exception right now, since if it _was_ used it just won't have an effect (also note: there there is no way this _could_ be used right now and this is mostly in place so I remember how I wanted to implement this).

app_integrations/batcher.py
ryandeivert: We always invoke the 'production' alias for Lambda functions. See Line 107 above (https://github.com/airbnb/streamalert/pull/323/files/071565d00c53c208aba3ca5e60b6f55bdf19f3ef#diff-dd9370bfe3012b45b5d3cec0e771cc18R107)

ryandeivert: Also see: https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/sink.py#L80

app_integrations/batcher.py
ryandeivert: The call to`_segement_and_send` is this a recursive call to this same function (which then splits the list and calls `_send_logs_to_stream_alert` again with smaller lists where the size is checked). Let me know if it needs more explanation. See `test_batcher.py` for tests related to this.

app_integrations/apps/duo.py
ryandeivert: Nope - it's meant to be how it's worded. Duo's only indicator of there being _more_ logs to pull is whether or not the **max** amount of logs possible was returned with a request. So, if less than the max was returned, this means we've collected all the logs up to this timeframe.

stream_alert_cli/package.py
```diff
@@ -220,7 +220,7 @@ def _upload(self, package_path):
                 return False
 
             package_fh.close()
-            LOGGER_CLI.debug('Uploaded %s to S3', package_name)
+            LOGGER_CLI.info('Uploaded %s to S3', package_name)
```
jacknagz: Why make this `info`?

stream_alert_cli/terraform/app_integrations.py
```diff
@@ -0,0 +1,69 @@
...
+    dlq_topic = (DEFAULT_SNS_MONITORING_TOPIC
+                 if infrastructure_config.get('monitoring', {}).get('create_sns_topic')
+                 else infrastructure_config.get('monitoring', {}).get('sns_topic_name',
+                                                                      DEFAULT_SNS_MONITORING_TOPIC))
+
+    prefix = config['global']['account']['prefix']
+
+    # App integration modules
+    for app_name, app_info in config['clusters'][cluster_name] \
+        ['modules'].get('stream_alert_apps', {}).iteritems():
```
jacknagz: IIRC you gave me a nit on usage of `iteritems` in favor of `items` 😄

stream_alert_cli/version.py
```diff
@@ -106,30 +104,53 @@ def _publish_helper(self, **kwargs):
         client = boto3.client('lambda', region_name=region)
         code_sha_256 = self.config['lambda'][self.package.config_key]['source_current_hash']
 
-        # Publish the function
+        # Publish the function(s)
+        if self.package.package_name == 'stream_alert_app':
+            for app_name, app_info in self.config['clusters'][cluster]['modules'] \
+                ['stream_alert_apps'].iteritems():
```
jacknagz: items vs iteritems 🤔

stream_alert_cli/version.py
```diff
@@ -106,30 +104,53 @@ def _publish_helper(self, **kwargs):
...
+                # Name follows format: '<prefix>_<cluster>_<service>_<app_name>_app'
+                function_name = '_'.join([self.config['global']['account']['prefix'], cluster,
+                                          app_info['type'], app_name, 'app'])
+                new_version = self._publish(client, function_name, code_sha_256)
+                if not new_version:
```
jacknagz: why wouldn't it be a new version?

stream_alert_cli/version.py
```diff
@@ -106,30 +104,53 @@ def _publish_helper(self, **kwargs):
         client = boto3.client('lambda', region_name=region)
         code_sha_256 = self.config['lambda'][self.package.config_key]['source_current_hash']
 
-        # Publish the function
+        # Publish the function(s)
+        if self.package.package_name == 'stream_alert_app':
```
jacknagz: can we keep this class as generic as possible and modify the logic above to work with any `function_name`?

jacknagz: for Athena, the clustered vs non clustered logic lives in `deploy.py`

stream_alert_cli/config.py
```diff
@@ -397,10 +397,11 @@ def _config_reader(self, key, file_path, **kwargs):
     @staticmethod
     def _config_writer(config, path, **kwargs):
         with open(path, 'r+') as conf_file:
-            conf_file.write(json.dumps(config,
-                                       indent=2,
-                                       separators=(',', ': '),
-                                       sort_keys=kwargs.get('sort_keys', True)))
+            json.dump(config,
```
jacknagz: 👍 👍 👍 👍 👍

stream_alert_cli/terraform/generate.py
```diff
@@ -340,20 +340,17 @@ def terraform_generate(**kwargs):
...
     # Setup the main.tf.json file
     LOGGER_CLI.debug('Generating cluster file: main.tf.json')
-    main_json = json.dumps(
-        generate_main(init=init, config=config),
-        indent=2,
-        sort_keys=True
-    )
     with open('terraform/main.tf.json', 'w') as tf_file:
-        tf_file.write(main_json)
+        json.dump(
```
jacknagz: can we decide on an indentation style? we seem to be all over the place:

```
json.dump(arg,
          arg2,
          arg3)

or

json.dump(
  arg,
  arg2,
  arg3
)
```

terraform/modules/tf_stream_alert_app/main.tf
```diff
@@ -0,0 +1,69 @@
...
+      LOGGER_LEVEL = "${var.log_level}"
+    }
+  }
+
+  dead_letter_config {
+    target_arn = "arn:aws:sns:${var.region}:${var.account_id}:${var.monitoring_sns_topic}"
+  }
+
+  tags {
+    Name = "App:${var.type}"
```
jacknagz: This should still have a StreamAlert name tag similar to the others for billing purpose.  Additionally, you can add a new App key

terraform/modules/tf_stream_alert_app/iam.tf
```diff
@@ -0,0 +1,117 @@
...
+
+// IAM Role Policy: Allow the StreamAlert App function to invoke the Rule Processor
+resource "aws_iam_role_policy" "stream_alert_app_role_policy" {
+  name   = "${var.function_prefix}_app_role_policy"
+  role   = "${aws_iam_role.stream_alert_app_role.id}"
+  policy = "${data.aws_iam_policy_document.app_invoke_rule_proc.json}"
+}
+
+// IAM Policy Doc: Allow the StreamAlert App function to invoke the Rule Processor
+data "aws_iam_policy_document" "app_invoke_rule_proc" {
```
jacknagz: this policy does more than just invoking the rule processor, I'd split them out into 3 separate policies for clarity and to easily understand the permissions model from console

app_integrations/apps/app_base.py
```diff
@@ -0,0 +1,258 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+from abc import ABCMeta, abstractmethod, abstractproperty
+from decimal import Decimal
+import time
+from timeit import timeit
```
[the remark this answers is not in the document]
ryandeivert: According to @austinbyers  and pylint, `from` doesn't take precedence over alphabetical, and this is how it should actually read. Austin, please clarify if I'm getting it wrong

app_integrations/apps/app_base.py
```diff
@@ -0,0 +1,258 @@
...
+    def _gather(self):
+        """Protected entry point for the beginning of polling"""
+
+        # Make this request sleep if the API throttles requests
+        self._sleep()
+        def do_gather():
+            """Perform the gather using this scoped method so we can time it"""
+            logs = self._gather_logs()
+
+            # Make sure there are logs, this can be False if there was an issue polling
```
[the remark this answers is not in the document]
ryandeivert: Yes, that's possible, but it will just exit cleanly in that case. I will update the comment since there are multiple reasons this could be fale

app_integrations/batcher.py
```diff
@@ -0,0 +1,122 @@
...
+        """Protected method for sending logs to the rule processor lambda
+        function for processing. This performs some size checks before sending.
+
+        Args:
+            source_function (str): The app function name from which the logs came
+            logs (list): List of the logs that have been gathered
+        """
+        # Create a payload to be sent to the rule processor that contains the
+        # service these logs were collected from and the list of logs
+        payload = {'Records': [{'stream_alert_app': source_function, 'gathered_logs': logs}]}
```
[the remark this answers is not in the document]
ryandeivert: Yeah sadly - this is the format that the Rule Processor is expecting all input to come in as. See: https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/handler.py#L74

We can make more drastic change to the input flow in the RP, but this is what required the least amount of manipulation on the RP side.

app_integrations/batcher.py
```diff
@@ -0,0 +1,122 @@
...
+        payload_json = json.dumps(payload, separators=(',', ':'))
+        if len(payload_json) > MAX_LAMBDA_PAYLOAD_SIZE:
```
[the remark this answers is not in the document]
ryandeivert: I'm not sure I'm following - let's touch base tomorrow about this

app_integrations/batcher.py
```diff
@@ -0,0 +1,122 @@
...
+            LOGGER.debug('Log payload size for %d logs exceeds limit and will be '
+                         'segmented (%d > %d max).', len(logs), len(payload_json),
+                         MAX_LAMBDA_PAYLOAD_SIZE)
+            return False
+
+        LOGGER.debug('Sending %d logs to rule processor with payload size %d',
+                     len(logs), len(payload_json))
+
+        try:
+            response = Batcher.LAMBDA_CLIENT.invoke(
```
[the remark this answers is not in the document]
ryandeivert: Theoretically, yes - _but_ class properties will be overridden by instance properties if they are set anywhere and references with the instance (ie: `self`). If `self.LAMBDA_CLIENT` ever got set to, say, `None` for some reason (shouldn't happen) then this would break. Not to say the class property couldn't also get set to None for some reason, but keeping it namespaced to the class with `Batcher` makes it clear that this is not an instance property being referenced.

app_integrations/config.py
```diff
@@ -0,0 +1,362 @@
...
+        LOGGER.debug('Parameter suffixes and names: %s', param_names)
+
+        # Get the loaded parameters and a list of any invalid ones from parameter store
+        params, invalid_params = AppConfig._get_parameters(param_names.values())
+        LOGGER.debug('Retrieved parameters from parameter store: %s', params)
+        LOGGER.debug('Invalid parameters could not be retrieved from parameter store: %s',
+                     invalid_params)
+
+        # Check to see if there are any required parameters in the invalid params list
+        missing_required_params = [param for param in invalid_params
```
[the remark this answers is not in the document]
ryandeivert: 🤔

app_integrations/config.py
```diff
@@ -0,0 +1,362 @@
...
+                    }
+        """
+        # Load the base config from the context that will get updated with other info
+        base_config = AppConfig._parse_context(context)
+
+        LOGGER.debug('Loaded env config: %s', base_config)
+
+        # Create the ssm boto3 client that will be cached and used throughout this execution
+        # if one does not exist already
+        if AppConfig.SSM_CLIENT is None:
```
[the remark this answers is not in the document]
ryandeivert: This all happens within a class method, and the ssm client is used for loading the config parameters from parameter store before the actual init to create the config happens a bit further down (https://github.com/airbnb/streamalert/pull/323/files/b54909a70c2ac5a0ff96a9230aadeb13ff277cab..c5e9b7f2c0afe67a93f16e63e2aed61e718e5272#diff-067a5757e3a3e84a44a80fb17190570aR158).

Think of this as a shared method on the class that creates an instance of itself (so meta..)

stream_alert/rule_processor/payload.py
```diff
@@ -323,3 +324,28 @@ def pre_parse(self):
...
+        """Pre-parsing method for incoming app records that iterates over all the
+        incoming logs in the 'gathered_logs' list.
+
+        Yields:
+            Instances of `self` back to the caller with the proper
+                `pre_parsed_record` set to the current log data. This conforms
+                to the interface of returning a generator, providing the ability
+                to support multiple records like this.
+        """
+        for data in self.raw_record['gathered_logs']:
```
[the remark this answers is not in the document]
ryandeivert: Sure, this key can be updated

stream_alert_cli/terraform/app_integrations.py
ryandeivert: Haha - I did do that, but then I recanted my statement in that same PR once I realized that `items` (while forward compatible to py3) actually makes a _copy_ of the key/value pairs as a list of tuples instead of a generator-style iterator. Therefore `iteritems` is actually much more performant - sorry to be a hypocrite.

stream_alert_cli/version.py
ryandeivert: See other comment

stream_alert_cli/version.py
ryandeivert: I'll try to make this change. I saw that athena handled this in deploy.py but it I couldn't see a reason why forking logic there is better than just forking it here. tbh - the LambdaPackage subclasses should really override methods like this one to handle cases that require different logic.

stream_alert_cli/version.py
ryandeivert: The `_version_helper` method that this calls is within a try/except for a boto3 call. on exception, it will return False: https://github.com/airbnb/streamalert/pull/323/files/91d7ef4c86983517af3119584bf5462c1cc87b4f..d465fc724eebf0f838e8750cc1d787b0b5f8f592#diff-106ec47af4d3b5071339461a157d8e14R74

This _should not_ happen, but I noticed that when trying to have app integrations pre-configured (before the `terraform init` happens) this would fail and return False due to not being able to find the function in S3. After speaking with you, that error is obviously expected, but adding this check seemed like it would still be a good idea. Thoughts?

terraform/modules/tf_stream_alert_app/main.tf
ryandeivert: Oh nice - I didn't realize that was the purpose. Would that just look similar to the following?
```
tags {
    Name  = "StreamAlert"
    App   = ${var.type}
}
```

app_integrations/batcher.py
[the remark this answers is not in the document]
jacknagz: Makes sense, that should be fine given this format is standard across most services.

app_integrations/config.py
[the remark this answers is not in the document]
jacknagz: Something I read recently when trying to understand how to break up a list comprehension across lines

stream_alert_cli/package.py
jacknagz: It's to reduce the amount of messages printed, it was way too verbose before in info mode

stream_alert_cli/version.py
jacknagz: yeah, that's a better solution

app_integrations/batcher.py
ryandeivert: LOL it's hilarious that I typed that wrong so many times without catching this. thanks!

app_integrations/apps/app_base.py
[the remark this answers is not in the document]
ryandeivert: Actually neither pylint nor pep cares how these are ordered.. I did it like this to conform to previous PR feeback from austin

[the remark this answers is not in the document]
jacknagz: It was my impression all the `from`s would be grouped alphabetically, followed by the `import`s without a line dividing them:

app_integrations/apps/app_base.py
```diff
@@ -0,0 +1,258 @@
...
+        """Method for validating the authentication dictionary retrieved from
+        AWS Parameter Store
+
+        Returns:
+            bool: Inidcator of successful validation
+        """
+        if not self._config:
+            raise AppIntegrationConfigError('Config for service \'{}\' is empty', self.type())
+
+        if not 'auth' in self._config:
```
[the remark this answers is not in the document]
jacknagz: perhaps add a comment explaining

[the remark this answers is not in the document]
ryandeivert: The docstring mentions:
> Method for validating the authentication dictionary retrieved from
        AWS Parameter Store

app_integrations/config.py
```diff
@@ -0,0 +1,362 @@
...
+                                      {cls._STATE_KEY: None, cls._TIME_KEY: None}))
+
+        # Add the auth config info to the 'auth' key since these key/values can vary
+        # from service to service
+        base_config[cls.AUTH_CONFIG_SUFFIX] = params[param_names[cls.AUTH_CONFIG_SUFFIX]]
+
+        return AppConfig(base_config)
+
+    @staticmethod
+    def _get_parameters(names):
```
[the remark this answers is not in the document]
ryandeivert: @jacknagz can you clarify what you mean here?

stream_alert_cli/terraform/generate.py
ryandeivert: I'm not seeing anywhere that uses the first format? there is a one-liner with `json.dump(...)` but can't find anywhere that uses the format:
```
json.dump(arg,
          arg2,
          arg3)

```

ryandeivert:
@jacknagz , @mime-frame pushed a commit with some changes - PTAL

terraform/modules/tf_stream_alert_app/main.tf
jacknagz: Yep

stream_alert_cli/terraform/generate.py
jacknagz: I'm not speaking to just `json.dump`, just speaking generally

app_integrations/config.py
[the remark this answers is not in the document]
jacknagz: can we isolate all the SSM methods into an inner class similar to how the states are contained

terraform/modules/tf_stream_alert_app/iam.tf
```diff
@@ -0,0 +1,117 @@
...
+
+    resources = [
+      "arn:aws:logs:${var.region}:${var.account_id}:log-group:${aws_cloudwatch_log_group.stream_alert_app.name}:log-stream:*",
+    ]
+  }
+}
+
+// Lambda Permission: Allow Cloudwatch Scheduled Events to invoke StreamAlert App Lambda
+resource "aws_lambda_permission" "allow_cloudwatch_events_invocation" {
+  statement_id  = "CloudwatchEventsInvokeAthenaRefresh"
```
chunyong-lin: `AthenaRefresh`?

ryandeivert: oooooh good catch!!!

stream_alert_cli/terraform/generate.py
chunyong-lin: Does python[ PEP 8](https://www.python.org/dev/peps/pep-0008/#indentation) answer this question?
```
Yes:

# Aligned with opening delimiter.
foo = long_function_name(var_one, var_two,
                         var_three, var_four)

# More indentation included to distinguish this from the rest.
def long_function_name(
        var_one, var_two, var_three,
        var_four):
    print(var_one)

# Hanging indents should add a level.
foo = long_function_name(
    var_one, var_two,
    var_three, var_four)
```

ryandeivert: So is that saying they are all acceptable?

austinbyers: Yes. I prefer styles which don't require changes if the function were renamed (i.e. I don't like the first one in the PEP8 list)

terraform/modules/tf_stream_alert_app/README.md
```diff
@@ -0,0 +1,162 @@
+# StreamAlert App Integration Terraform Module
```
austinbyers: Don't we want this to be a `.rst` file to be consistent with the rest of the files? GitHub renders `.rst` just fine

austinbyers: `.rst` would also make the table a lot easier to understand in the raw code

ryandeivert: @austinbyers all of the TF readmes are md, so I followed that trend. Not changing it now :)

stream_alert_cli/helpers.py
```diff
@@ -52,6 +54,12 @@ def run_command(runner_args, **kwargs):
     default_cwd = 'terraform'
     cwd = kwargs.get('cwd', default_cwd)
 
+    # Add the -force-copy flag for s3 state copying to suppress dialogs that
+    # the user must type 'yes' into.
+    if runner_args[0] == 'terraform':
+        if runner_args[1] == 'init':
+            runner_args.append('-force-copy')
```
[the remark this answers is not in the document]
ryandeivert: Would we want to set that variable when it's invoked from `manage.py` though? Anything dangerous about ti?

app_integrations/apps/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Create some package level items to make this nicer to use"""
+import importlib
+import os
+
+# Import all files containing subclasses of AppIntegration, skipping the common base class
+for app_file in os.listdir(os.path.dirname(__file__)):
+    # Skip the common base file and any non-py files
+    if app_file.startswith(('__init__', 'app_base')) or not app_file.endswith('py'):
```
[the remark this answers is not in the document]
ryandeivert: @austinbyers  the logic is `not` - this skips files that _don't_ end in py

stream_alert/rule_processor/classifier.py
```diff
@@ -53,7 +53,8 @@ def extract_service_and_entity(raw_record):
         entity_mapper = {
             'kinesis': lambda r: r['eventSourceARN'].split('/')[1],
             's3': lambda r: r['s3']['bucket']['name'],
-            'Sns': lambda r: r['EventSubscriptionArn'].split(':')[5]
+            'Sns': lambda r: r['EventSubscriptionArn'].split(':')[5],
```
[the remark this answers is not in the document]
ryandeivert: Weirdly, yes - see the line above: https://github.com/airbnb/streamalert/pull/323/files#diff-c26614beb435b88a022830d4ed7c088fR51

app_integrations/main.py
```diff
@@ -0,0 +1,47 @@
...
+        # Load the config from this context object, pulling info from parameter store
+        config = AppConfig.load_config(context)
+
+        # The config specifies what app this function is supposed to run
+        app = get_app(config)
+
+        # Run the gather operation
+        app.gather()
+    except:
+        raise
```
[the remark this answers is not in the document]
ryandeivert: Thanks - I added this because I wasn't sure if the finally got called without it, but it seems it does

manage.py
```diff
@@ -23,13 +23,17 @@
 
 terraform <cmd> -var-file=../terraform.tfvars -var-file=../variables.json
 """
+# pylint: disable=too-many-lines
```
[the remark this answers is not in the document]
ryandeivert: Haha I know - but there is _literally_ nothing in here besides argparse stuff right now. I'm open to suggestions for how to split that, but splitting up the argparse logic would be a PITA and cause weird fragmentation.

app_integrations/apps/app_base.py
```diff
@@ -0,0 +1,282 @@
...
+    Args:
+        config (AppConfig): Loaded configuration with service, etc
+
+    Returns:
+        AppIntegration: Subclass of AppIntegration
+    """
+    try:
+        return STREAMALERT_APPS[config['type']](config)
+    except AppIntegrationException:
+        raise
```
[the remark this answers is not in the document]
ryandeivert: Yes - this has been removed, sorry!

app_integrations/apps/__init__.py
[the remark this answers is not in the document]
austinbyers: Sorry, I wasn't clear - I meant why check for `not app_file.endswith('py')` instead of `not app_file.endswith('.py')`? Because right now, a file called `happy` would be considered a "Python file"

app_integrations/apps/app_base.py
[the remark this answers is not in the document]
austinbyers: No worries, I was just making sure I wasn't missing something!

manage.py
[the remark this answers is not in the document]
austinbyers: Yeah, that's fine. So instead of disabling the pylint check, change the `.pylintrc` to allow longer files. Either way is fine tho

stream_alert_cli/helpers.py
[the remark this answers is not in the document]
austinbyers: The variable is designed to be set when you're running Terraform from a management script and the user is not invoking `terraform` commands directly. I haven't tried it (and now might not be the time), but just FYI that Terraform has a mode to support exactly what we're doing (running it in automation)

terraform/modules/tf_stream_alert_app/README.md
austinbyers: Ok! Didn't know the TF stuff were READMEs

app_integrations/apps/__init__.py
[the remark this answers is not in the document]
ryandeivert: ooooooh I hear you - will make the change -__-

manage.py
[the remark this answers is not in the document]
ryandeivert: @austinbyers I didn't add it because I don't really want to promote longer files project wide, and would rather do case-by-case exceptions like this. Is there a way to do that in the rc?

PR #324, opened 2017-09-28
Lambda to Firehose: Part 3

jacknagz:
to: @austinbyers 
size: medium
references: #5, #280, #320 

### Background

Part 3 completes the Lambda and CLI development necessary to send logs directly to Firehose.  At a high level, it accomplishes the following:

* Strip special characters from record keys: This enables records to be compliant with loading data into Hive.
* Payload size checks: Ensure that the max payload size for Firehose is not surpassed.
* Disable log types: For whatever reason, if a user does not want to send a top level log type (`cloudwatch` for example), it can be added to the config.  This will also remove the Firehose Delivery Stream to keep costs down

### Testing

Deployed in a test AWS account to verify end to end workflows

### DIFF
--- a/stream_alert/rule_processor/handler.py
+++ b/stream_alert/rule_processor/handler.py
@@ -19,6 +19,7 @@ import json
 import re
 
 import boto3
+from botocore.exceptions import ClientError
 
 from stream_alert.rule_processor import FUNCTION_NAME, LOGGER
 from stream_alert.rule_processor.classifier import StreamClassifier
@@ -28,16 +29,19 @@ from stream_alert.rule_processor.rules_engine import StreamRules
 from stream_alert.rule_processor.sink import StreamSink
 from stream_alert.shared.metrics import MetricLogger
 
+# Firehose Limits: http://bit.ly/2fw5UY2
 MAX_BATCH_COUNT = 500
 MAX_BATCH_SIZE = 4000 * 1000
-MAX_RECORD_SIZE = 1000 * 1000
+# The subtraction of 2 accounts for the newline at the end
+MAX_RECORD_SIZE = 1000 * 1000 - 2
 
 
 class StreamAlert(object):
     """Wrapper class for handling StreamAlert classificaiton and processing"""
     config = {}
-    special_char_regex = re.compile(r'[\`\~\!\@\#\$\%\^\&\*\(\)\-\+\
-                                       \=\[\]\\\{\}\;\'\:\"\,\.\/\<\>\?]')
+    # Used to detect special characters in payload keys.
+    # This is necessary for sanitization of data prior to searching in Athena.
+    special_char_regex = re.compile(r'\W')
     special_char_sub = '_'
 
     def __init__(self, context, enable_alert_processor=True):
@@ -179,8 +183,8 @@ class StreamAlert(object):
             record_list (list): The original records list to be segmented
             max_count (int): The max amount of records to yield per group
         """
-        for item in range(0, len(record_list), max_count):
-            yield record_list[item:item + max_count]
+        for index in range(0, len(record_list), max_count):
+            yield record_list[index:index + max_count]
 
     def _segment_records_by_size(self, record_batch):
         """Segment record groups by size
@@ -194,11 +198,12 @@ class StreamAlert(object):
         split_factor = 1
         len_batch = len(record_batch)
 
-        while True:
-            if len(str(record_batch[:len_batch / split_factor])) > MAX_BATCH_SIZE:
-                split_factor += 1
-            else:
-                break
+        # Sample the first batch of records to determine the split factor.
+        # Generally, it's very rare for a group of records to have
+        # drastically different sizes in a single Lambda invocation.
+        while len(json.dumps(record_batch[:len_batch / split_factor],
+                             separators=(",", ":"))) > MAX_BATCH_SIZE:
+            split_factor += 1
 
         return self._segment_records_by_count(record_batch, len_batch / split_factor)
 
@@ -210,7 +215,7 @@ class StreamAlert(object):
             batch (list): Record batch to iterate on
         """
         for index, record in enumerate(batch):
-            if len(str(record)) > MAX_RECORD_SIZE:
+            if len(json.dumps(record, separators=(",", ":"))) > MAX_RECORD_SIZE:
                 # Show the first 1k bytes in order to not overload
                 # CloudWatch logs
                 LOGGER.error('The following record is too large'
@@ -234,12 +239,9 @@ class StreamAlert(object):
         """
         new_record = {}
         for key, value in record.iteritems():
-            # Set a default value of the original key
-            sanitized_key = key
-            if re.search(self.special_char_regex, key):
-                sanitized_key = re.sub(self.special_char_regex,
-                                       self.special_char_sub,
-                                       key)
+            sanitized_key = re.sub(self.special_char_regex,
+                                   self.special_char_sub,
+                                   key)
 
             # Handle nested objects
             if isinstance(value, dict):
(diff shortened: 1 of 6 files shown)


stream_alert/rule_processor/handler.py
```diff
@@ -27,14 +28,17 @@
...
 MAX_RECORD_SIZE = 1000 * 1000
-MAX_BATCH_SIZE = 500
 
 
 class StreamAlert(object):
     """Wrapper class for handling StreamAlert classificaiton and processing"""
-    __config = {}
+    config = {}
+    special_char_regex = re.compile(r'[\`\~\!\@\#\$\%\^\&\*\(\)\-\+\
+                                       \=\[\]\\\{\}\;\'\:\"\,\.\/\<\>\?]')
```
austinbyers: Instead of trying to list every possible special character, `\W` matches any non-alphanumeric character (i.e. it is equivalent to `[^a-zA-Z0-9_]`). This should work quite nicely for you

stream_alert/rule_processor/handler.py
```diff
@@ -27,14 +28,17 @@
 from stream_alert.rule_processor.sink import StreamSink
 from stream_alert.shared.metrics import MetricLogger
 
-
+MAX_BATCH_COUNT = 500
+MAX_BATCH_SIZE = 4000 * 1000
```
austinbyers: Can you add a comment about what this is? Is this a limit imposed by AWS?

stream_alert/rule_processor/handler.py
```diff
@@ -167,68 +171,141 @@ def get_alerts(self):
...
+    @staticmethod
+    def _segment_records_by_count(record_list, max_count):
+        """Segment records by length
+
+        Args:
+            record_list (list): The original records list to be segmented
+            max_count (int): The max amount of records to yield per group
+        """
+        for item in range(0, len(record_list), max_count):
+            yield record_list[item:item + max_count]
```
austinbyers: I would rename `item` to `index` for clarity

stream_alert/rule_processor/handler.py
```diff
@@ -167,68 +171,141 @@ def get_alerts(self):
...
+            generator: Used to iterate on each newly segmented group
+        """
+        split_factor = 1
+        len_batch = len(record_batch)
+
+        while True:
+            if len(str(record_batch[:len_batch / split_factor])) > MAX_BATCH_SIZE:
+                split_factor += 1
+            else:
+                break
```
austinbyers: More simply:

```python
# Calculate the appropriate split factor
while len(str(record_batch[:len_batch / split_factor])) > MAX_BATCH_SIZE:
    split_factor += 1
```

austinbyers: This also only checks the size of the first batch and assumes every other chunk will be about the same size. Since it's possible that different chunks may be different sizes, can you add a comment explaining why you feel that possibility is unlikely?

stream_alert/rule_processor/handler.py
```diff
@@ -167,68 +171,141 @@ def get_alerts(self):
...
+
+    @staticmethod
+    def _limit_record_size(batch):
+        """Limit the record size to be sent to Firehose
+
+        Args:
+            batch (list): Record batch to iterate on
+        """
+        for index, record in enumerate(batch):
+            if len(str(record)) > MAX_RECORD_SIZE:
```
austinbyers: What is the type of `record`? If it's a dictionary, than `str(record)` will include whitespace between keys and values, which you might not have in the actual data you send over the wire

stream_alert/rule_processor/handler.py
```diff
@@ -167,68 +171,141 @@ def get_alerts(self):
...
+            dict: A sanitized record
+        """
+        new_record = {}
+        for key, value in record.iteritems():
+            # Set a default value of the original key
+            sanitized_key = key
+            if re.search(self.special_char_regex, key):
+                sanitized_key = re.sub(self.special_char_regex,
+                                       self.special_char_sub,
+                                       key)
```
austinbyers: You don't actually have to check for existence with `re.search`; `re.sub` will leave the original unchanged if there were no matches. So lines 238-242 can be simply:

```python
sanitized_key = re.sub(self.special_char_regex, self.special_char_sub, key)
```

stream_alert/rule_processor/handler.py
```diff
@@ -167,68 +171,141 @@ def get_alerts(self):
...
+
+            # Handle nested objects
+            if isinstance(value, dict):
+                new_record[sanitized_key] = self._sanitize_keys(record[key])
+            else:
+                new_record[sanitized_key] = record[key]
```
austinbyers: What if the record contains a list?

stream_alert/rule_processor/handler.py
jacknagz: @austinbyers it's a `dict`, what would you suggest in this case then?

stream_alert/rule_processor/handler.py
jacknagz: @austinbyers It wouldn't matter because there are no keys with a `list`.

stream_alert/rule_processor/handler.py
austinbyers: Since `str` will take up more space, it's definitely safe as is. If you want to pack in more data, you could `json.dumps` to calculate its size as if you were about to send it. I don't imagine `json.dumps` takes that much more time than `str`

stream_alert/rule_processor/handler.py
jacknagz: Good point, adding more clarificaiton

stream_alert/rule_processor/handler.py
austinbyers: You can't have a list of dicts or anything like that? E.g.
```
{
  'results': [
    {'hi': 'there'},
    {'special*&^char': 'fdad'}
  ]
}
```

stream_alert/rule_processor/handler.py
jacknagz: Sure, good point!

stream_alert/rule_processor/handler.py
jacknagz: This shouldn't be too much of an issue, we generally load nested lists like this with JSONPath

stream_alert/rule_processor/handler.py
jacknagz: Just timed it, `json.dumps` is faster

austinbyers: Cool! Good to know

stream_alert/rule_processor/handler.py
austinbyers: 👍

stream_alert/rule_processor/handler.py
```diff
@@ -28,16 +28,19 @@
...
 
 
 class StreamAlert(object):
     """Wrapper class for handling StreamAlert classificaiton and processing"""
     config = {}
-    special_char_regex = re.compile(r'[\`\~\!\@\#\$\%\^\&\*\(\)\-\+\
-                                       \=\[\]\\\{\}\;\'\:\"\,\.\/\<\>\?]')
+    # Used to detect special characters in payload keys.
+    # This is necessary for sanitization of data prior to searching in Athena.
+    special_char_regex = re.compile(r'[^a-zA-Z0-9_]')
```
austinbyers: You can literally just do
```python
special_char_regex = re.compile(r'\W')
```

stream_alert/rule_processor/handler.py
```diff
@@ -167,68 +174,138 @@ def get_alerts(self):
...
+        Returns:
+            generator: Used to iterate on each newly segmented group
+        """
+        split_factor = 1
+        len_batch = len(record_batch)
+
+        # Sample the first batch of records to determine the split factor.
+        # Generally, it's very rare for a group of records to have
+        # drastically different sizes in a single Lambda invocation.
+        while len(str(record_batch[:len_batch / split_factor])) > MAX_BATCH_SIZE:
```
austinbyers: This has the same length-check discrepancy as line 216. Do you want `len(json.dumps(...))` instead of `len(str(...))`? Up to you

stream_alert/rule_processor/handler.py
austinbyers: The two are identical (for ASCII character sets)

stream_alert/rule_processor/handler.py
jacknagz: Good point, i'll make that change

PR #332, opened 2017-10-02
fixing the osquery:differential schema

ashmere:
confirmed working with 2.7.0 & 2.8.0 osquery outputs

### DIFF
--- a/conf/logs.json
+++ b/conf/logs.json
@@ -983,9 +983,11 @@
   },
   "osquery:differential": {
     "schema": {
-      "action": "string",
       "calendarTime": "string",
-      "columns": {},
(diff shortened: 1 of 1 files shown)


jacknagz:
hey @ashmere can you provide example logs?

mime-frame:
As highlighted in Slack, it appears you're using the `batch` format: https://github.com/facebook/osquery/blob/master/docs/wiki/deployment/logging.md#batch-format

Our conf currently only supports `snapshot` and `event` formats: https://github.com/facebook/osquery/blob/master/docs/wiki/deployment/logging.md#event-format

The correct fix is a new log type, I'll put up a PR for this shortly

PR #335, opened 2017-10-02
[cli] adding suppression for firehose log messages

ryandeivert:
to @jacknagz 

### Change
* Adding additional log output suppression for `Successfully sent * messages to Firehose*` during tests to list of suppressed messages.
* Updating the filter function to use `fnmatch` to make it a little more adaptable.

### DIFF
--- a/stream_alert_cli/logger.py
+++ b/stream_alert_cli/logger.py
@@ -13,6 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+from fnmatch import fnmatch
 import logging
 import logging.handlers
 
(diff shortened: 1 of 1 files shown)


stream_alert_cli/logger.py
```diff
@@ -29,13 +30,17 @@ class SuppressNoise(logging.Filter):
...
 
     def filter(self, record):
-        suppress_starts_with = (
-            'Starting download from S3',
-            'Completed download in'
+        suppressed_messages = (
+            'Starting download from S3*',
+            'Completed download in*',
+            '*triggered an alert on log type*',
+            'Successfully sent * messages to Firehose*'
```
mime-frame: Won't we need this for Kinesis streams as well?

mime-frame: Aka we might need this to be `Successfully sent * messages to *`

jacknagz: @mime-frame nope, Lambda only sends to Firehose

ryandeivert: @jacknagz could answer that - I don't know of any log message for Kinesis right now?

stream_alert_cli/logger.py
```diff
@@ -29,13 +30,17 @@ class SuppressNoise(logging.Filter):
...
         )
+
+        message = record.getMessage()
+
         return (
-            not record.getMessage().startswith(suppress_starts_with) and
-            'triggered an alert on log type' not in record.getMessage()
+            not any(fnmatch(message, suppression) for suppression in suppressed_messages)
```
jacknagz: why not use `in_set`? 😉

ryandeivert: I'd rather not import a helper designed for rules into the CLI

jacknagz: but why?

ryandeivert: Because if we ever do decide to start trying to use pip, we're going to have a real hard time packaging all of this if there are weird dependencies like that. and also, it's one line

jacknagz: giphy-7

ryandeivert: lamo

PR #336, opened 2017-10-02
[apps] migrating requests logic to base class

ryandeivert:
to @austinbyers 
size: small

### Changes
* To make the app class interface easier to subclass/utilize, I've moved the functionality
  for performing GET requests against an endpoint to within the base app class.
* Updating tests to account for the migrated code.
* Adding debug statement to log when a request occurs.

### DIFF
--- a/app_integrations/apps/duo.py
+++ b/app_integrations/apps/duo.py
@@ -116,6 +116,8 @@ class DuoApp(AppIntegration):
 
         # Make the request to the api, resulting in a bool or dict
         response = self._make_request(full_url, headers=headers, params=params)
+        if not response:
+            return False
 
         # Duo stores the list of logs in the 'response' key of the response
         logs = response['response']
--- a/app_integrations/apps/duo.py
+++ b/app_integrations/apps/duo.py
@@ -20,8 +20,6 @@ import hmac
 import re
 import urllib
 
-import requests
-
 from app_integrations import LOGGER
 from app_integrations.apps.app_base import app, AppIntegration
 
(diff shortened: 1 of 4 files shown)


tests/unit/app_integrations/test_apps/test_app_base.py
```diff
@@ -212,3 +212,18 @@ def test_gather_running(self, finalize_mock):
...
         self._app.gather()
         finalize_mock.assert_not_called()
+
+    @patch('requests.get')
+    def test_make_request_bad_response(self, requests_mock):
+        """App Integration - Make Request, Bad Response"""
+        requests_mock.return_value = Mock(
+            status_code=404,
+            json=Mock(side_effect=[{'message': 'something went wrong'}])
+        )
```
austinbyers: Nice! I know this is copied, but I didn't know you could use a raw Mock like this

app_integrations/apps/duo.py
```diff
@@ -116,13 +114,11 @@ def _get_duo_logs(self, hostname, full_url):
         if not headers:
             return False
 
-        # Perform the request and get the list of logs
-        response = requests.get(full_url, headers=headers, params=params)
-
-        if not self._check_http_response(response):
-            return False
+        # Make the request to the api, resulting in a bool or dict
+        response = self._make_request(full_url, headers=headers, params=params)
```
austinbyers: If `response` is False, you'll get a `KeyError` on the next line

ryandeivert: Nice good catch. Making that fix

ryandeivert:
@austinbyers  added a commit and rebased

mime-frame:
image

PR #337, opened 2017-10-02
streamalert - threat intel - duo

mime-frame:
to: @ryandeivert 

**Changes**

This adds the fields from the DUO schemas that map to CEF fields that are compared against threat intelligence.

**Testing**

```

StreamAlertCLI [INFO]: (21/21) Successful Tests
StreamAlertCLI [INFO]: (45/45) Alert Tests Passed
```

**Things to reason about as the reviewer**

1) Can/should I just specifically reference `duo:authentication` vs. `duo`, since `duo:administrator` doesn't contain either of the fields that map to threat intel?
2) What's the best way to test changes to this conf file? I'm not sure what I did for testing is good enough.

### DIFF
--- a/conf/types.json
+++ b/conf/types.json
@@ -210,5 +210,13 @@
       "srcuser",
       "dstuser"
     ]
+  },
+  "duo": {
+    "sourceAddress": [
+      "ip"
+    ],
+    "userName": [
+      "username"
+    ]
   }
 }
--- a/conf/types.json
+++ b/conf/types.json
@@ -120,6 +120,14 @@
       "invokedBy"
     ]
   },
+  "duo": {
+    "sourceAddress": [
+      "ip"
+    ],
+    "userName": [
+      "username"
+    ]
+  },
   "ghe": {
     "destinationAddress": [
       "remote_address"
@@ -209,14 +217,6 @@
     "userName": [
       "srcuser",
       "dstuser"
-    ]
-  },
-  "duo": {
-    "sourceAddress": [
-      "ip"
-    ],
-    "userName": [
-      "username"
     ]
   }
 }


ryandeivert:
@mime-frame to both of your questions: I'm not sure without taking more time to understand the threat intel code. I should have some free time later today to research this

mime-frame:
@ryandeivert awesome, thanks :)

ryandeivert:
@mime-frame  I know we're trying to keep garbage rules out of the public repo, but could you create a local test rule that uses `datatypes=['sourceAddress']` and no `logs` that will run against the current test events in `duo_fraud.json` or `duo_anonymous_ip_failure.json` for this?

I can't think of another way to ensure this works at the moment.

ryandeivert:
@mime-frame also to your first question I think you should probably use the `:authentication` suffix with these, but again, not positive

mime-frame:
Note: I'm having @chunyong-lin take this PR over when she returns

chunyong-lin:
I am going to work on this PR today. 
@mime-frame to answer your questions
> 1. Can/should I just specifically reference duo:authentication vs. duo, since duo:administrator doesn't contain either of the fields that map to threat intel?

In current implementation, normalized types are categorized by `log_source` (carbonblack, cloudtrail, ghe, osquery, duo etc). So we should reference `duo` as top key in `conf/types.json`. One advantage is we don't need to define 2 sets of normalized types for logs from `duo`. Also we defined 21 schemas for [carbonblack logs](https://github.com/airbnb/streamalert/blob/master/conf/logs.json#L2). Instead of defining 21 sets of normalized types, we only need to define 1 set of normalized types. 

> 2. What's the best way to test changes to this conf file? I'm not sure what I did for testing is good enough.

As @ryandeivert pointed out, we need local testing rule to test changes to this conf file.

chunyong-lin:
@mime-frame or @ryandeivert PTAL.
Testing against duo normalized types has been passed by adding a local testing rule and testing logs. 

A testing rule sample is 
```
from helpers.base import is_ioc

rule = StreamRules.rule


@rule(datatypes=['sourceAddress'],
      outputs=['aws-s3:sample-bucket'])
def test_detect_ioc(rec): # pylint: disable=unused-variable
    return is_ioc(rec)
```

Rule test results
```
test_detect_ioc
       [Pass]  [trigger=1]                   rule      (stream_alert_app): DUO authentication is from an IOC IP address and trigger an alert
       [Pass]                                alert     (aws-s3): sending alert to 'sample-bucket'
       [Pass]  [trigger=0]                   rule      (stream_alert_app): Duo authentication is not from IOC ip address and should not trigger an alert
```

mime-frame:
@chunyong-lin thanks :) can you approve the PR?

Issue #333, opened 2017-10-02
Bug: Normalized field is not found in a nested structure


javuto:
## Current Behavior

Writing a new rule that uses the data normalization capabilities, and in particular the `command` datatype. For carbonblack logs, the fields that are used (as defined in `types.json`) are in the same depth but for osquery logs those fields are usually within a nested structure and current tests do not match logs that should alert based on data normalization. For example, here is a carbonblack log:
```
{
    "data": {
    "cb_server": "cbserver",
    "command_line": "malicious_command",
    ...
}
```
System works as expected and `malicious_command` will be detected using `datatypes=['command']` in the rule. On the other hand here is a osquery sample log:
```
{
    "data": {
    "hostIdentifier": "osquery-host",
    "columns": {
        "command": "malicious_command",
        ...
    }
    ...
}    
```
Because the `command` field is nested within `columns`, the data does not seem to be matched properly.

## Desired Results

Data within nested structures will be extracted and used to match current and new rules. Both examples of logs should match a rule looking for `malicious_command` and at the moment only carbonblack logs do it.

chunyong-lin:
Hi @javuto Unit [test case](https://github.com/airbnb/streamalert/blob/master/tests/unit/stream_alert_rule_processor/test_rules_engine.py#L606-L677) should cover this case.

PR #339, opened 2017-10-02
[streamalert][bug] Fix to allow nested structured data and normalization

javuto:
to @ryandeivert 
cc @mime-frame 

This code fixes the problem of having nested data in normalized fields and not be detected by the rules as datatype. Resolves: https://github.com/airbnb/streamalert/issues/333

## Change

The method `match_types` now sets the value of `record['normalized_types']` instead of returning it. Given that the method was being invoked recursively, the impacted field was not updated correctly and the nested normalized fields that we were interested on, they were not set. With this fix the existing rules work as expected while other new rules can be implemented and the normalization can be used with nested fields.

## Testing

Tested for existing rules using `python manage.py lambda test --processor all`
Added coverage for this in the existing test `tests/unit/stream_alert_rule_processor/test_rules_engine.py` and checked with the following command:
```
...
Rules Engine - Match normalized types against record ... ok
Rules Engine - Recursively walk though all nested keys and update ... ok
...
```

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -167,8 +167,8 @@ class StreamRules(object):
             datatypes (list): defined in rule options, normalized_types users
                 interested in.
 
-        Returns:
-            (dict): A dict of normalized_types with original key names
+        Sets:
+            It sets the dict of normalized_types with original key names
 
         Example 1:
             datatypes=['defined_type1', 'defined_type2', 'not_defined_type']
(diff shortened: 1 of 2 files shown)


mime-frame:
Update your PR description with a Testing section with how you tested your changes @javuto :)

mime-frame:
Awesome, thanks @javuto ! @ryandeivert - you're up :)

ryandeivert:
@javuto please add the line `resolves: #333` below the to & cc lines so the issue will be closed when this PR is merged

javuto:
@chunyong-lin is taking care of this

PR #345, opened 2017-10-04
[lambda] add backoff to firehose PutRecordBatch requests

jacknagz:
to: @ryandeivert 
size: small

## Changes 

* Add `backoff` as the list of dependencies for the `rule_processor`
* Copies the backoff handlers into shared to be used by the Athena function (soon)
* Wraps the firehose request helper with a backoff decorator

## Testing

Deployed the rule processor in a test account, verified logs are sending via Firehose


mime-frame:
image

PR #346, opened 2017-10-04
[schemas] update cloudwatch and cb schemas

jacknagz:
to: @mime-frame 
size: tiny

* Cloudwatch event account IDs should be loaded as strings, not integers (for Athena)
* Add an optional key to one of the CB schemas

### DIFF
--- a/conf/logs.json
+++ b/conf/logs.json
@@ -531,6 +531,7 @@
     "schema": {
       "alert_severity": "float",
       "alert_type": "string",
+      "assigned_to": "string",
       "cb_server": "string",
       "computer_name": "string",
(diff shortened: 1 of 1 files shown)


mime-frame:
> Cloudwatch event account IDs should be loaded as strings, not integers (for Athena)

They are always integers though right? I'm curious to hear more about why Athena requires this (is it a result of how we defined the schema or?)

Either way, LGTM

jacknagz:
@mime-frame my thought was this: If we aren't going to do math on the field, it should be a string.  Athena would also require it to be a BIGINT iirc

PR #349, opened 2017-10-04
[bug][config] removing kinesis outputs from default cluster config

ryandeivert:
to @jacknagz 
size: small

### Change
* Removing the `outputs` from the default cluster config since it breaks new deploys and is tech debt left over from a TF refactor.


ryandeivert:
NM just tested this again and it's working as expected!

PR #350, opened 2017-10-05
[lambda] add a second layer of backoff to firehose

jacknagz:
to: @ryandeivert 
size: medium

## Changes

Currently, Firehose will not throw an exception if any of the records fail in the [PutRecordBatch](http://boto3.readthedocs.io/en/latest/reference/services/firehose.html#Firehose.Client.put_record_batch) request.

It will however return details in the response of the number of failed records.  This change allows for clearer diagnosis of Firehose operations.

I also modified the Firehose logger messages for clarity/consistency/usefulness

## Testing

All tests performed in a StreamAlert cluster in a staging AWS account

### DIFF
--- a/stream_alert/shared/backoff_handlers.py
+++ b/stream_alert/shared/backoff_handlers.py
@@ -25,10 +25,10 @@ def backoff_handler(details):
             target function currently executing, kwargs, args, value,
             and wait time.
     """
-    LOGGER.debug('[Backoff]: Trying again in %f seconds after %d tries calling %s',
-                 details['wait'],
-                 details['tries'],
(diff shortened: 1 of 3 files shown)


stream_alert/shared/backoff_handlers.py
```diff
@@ -40,8 +40,8 @@ def success_handler(details):
             and wait time.
     """
     LOGGER.debug('[Backoff]: Completed after %d tries calling %s',
-                 details['tries'],
-                 details['target'].__name__)
+                details['tries'],
```
ryandeivert: looks like a space got removed here

PR #355, opened 2017-10-06
[streamalert][app][onelogin] Adding new integration app for OneLogin

javuto:
to @ryandeivert
cc @mime-frame 

## Change

Adding a new integration application for OneLogin. It will allow the collection of OneLogin events and process them using StreamAlert. Also the schema for the OneLogin events format is added here.
It handles the generation of tokens, using API `client_secret/client_id` pair and pagination of requests to the API, when the number of events returned (using `since`) is over the limit (50). More information here: [https://developers.onelogin.com/api-docs/1/events/get-events](https://developers.onelogin.com/api-docs/1/events/get-events)

## Testing

Unit tests are added for the class:
```
...
OneLoginApp - Gather Events Entry Point ... ok
OneLoginApp - Generate Headers, ... ok
OneLoginApp - Get OneLogin Events, Bad Response ... ok
OneLoginApp - Get OneLogin Events, No Headers ... ok
OneLoginApp - Get Paginated Events ... ok
OneLoginApp - Get OneLogin Paginated Events, Bad Response ... ok
OneLoginApp - Required Auth Info ... ok
OneLoginApp - Sleep Seconds ... ok
OneLoginApp - Verify Events Endpoint ... ok
OneLoginApp - Verify Token Endpoint ... ok
OneLoginApp - Verify Events Type ... ok
...
```

FYI: Given that this PR only adds the new application, CI will fail. I will update it accordingly with a new rule so it will address that.

### DIFF
--- a/app_integrations/apps/onelogin.py
+++ b/app_integrations/apps/onelogin.py
@@ -1,0 +1,193 @@
+"""
+Copyright 2017-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import re
+
+from datetime import datetime
+import requests
+
+from app_integrations import LOGGER
+from app_integrations.apps.app_base import app, AppIntegration
+
+
+class OneLoginApp(AppIntegration):
+    """OneLogin StreamAlert App"""
+    _ONELOGIN_EVENTS_URL = 'https://api.us.onelogin.com/api/1/events'
+    _ONELOGIN_TOKEN_URL = 'https://api.us.onelogin.com/auth/oauth2/v2/token'
+    # OneLogin API returns 50 events per page
+    _MAX_EVENTS_LIMIT = 50
+
+    # Define our authorization headers variable
+    def __init__(self, config):
+        super(OneLoginApp, self).__init__(config)
+        self._auth_headers = None
+        self._next_page_url = None
+
+    @classmethod
+    def _type(cls):
+        return 'events'
+
+    @classmethod
+    def _endpoint(cls):
+        """Class method to return the OneLogin events endpoint
+
+        Returns:
+            str: Path of the events endpoint to query
+        """
+        return cls._ONELOGIN_EVENTS_URL
+
+    @classmethod
+    def service(cls):
+        return 'onelogin'
+
+    def _generate_headers(self, token_url, client_secret, client_id):
+        """Each request will request a new token to call the resources APIs.
+
+        More details to be found here:
+            https://developers.onelogin.com/api-docs/1/oauth20-tokens/generate-tokens-2
+
+        Returns:
+            str: Bearer token to be used to call the OneLogin resource APIs
+        """
+        authorization = 'client_id: %s, client_secret: %s' % (client_id, client_secret)
+        headers_token = {'Authorization': authorization,
+                         'Content-Type': 'application/json'}
+
+        response = requests.post(token_url,
+                                 json={'grant_type':'client_credentials'},
+                                 headers=headers_token)
+
+        if not self._check_http_response(response):
+            return False
+
+        bearer = 'bearer:' % (response.json()['access_token'])
+        self._auth_headers = {'Authorization': bearer}
+
+    def _gather_logs(self):
+        """Gather the authentication log events."""
+
+        if not self._auth_headers:
+            self._auth_headers = self._generate_headers(self._ONELOGIN_TOKEN_URL,
+                                                        self._config['auth']['client_secret'],
+                                                        self._config['auth']['client_id'])
+
+        self._next_page_url = self._ONELOGIN_EVENTS_URL
+        return self._get_onelogin_events()
+
+    def _get_onelogin_events(self):
+        """Get all events from the endpoint for this timeframe
+
+        Returns:
+            [
+                {
+                    'id': <int:id>,
+                    'created_at': <str:created_at>,
+                    'account_id': <int:account_id>,
+                    'user_id': <int:user_id>,
+                    'event_type_id': <int:event_type_id>,
+                    'notes': <str:notes>,
+                    'ipaddr': <str:ipaddr>,
+                    'actor_user_id': <int:actor_user_id>,
+                    'assuming_acting_user_id': null,
+                    'role_id': <int:role_id>,
+                    'app_id': <int:app_id>,
+                    'group_id': <int:group_id>,
+                    'otp_device_id': <int:otp_device_id>,
+                    'policy_id': <int:policy_id>,
+                    'actor_system': <str:actor_system>,
+                    'custom_message': <str:custom_message>,
+                    'role_name': <str:role_name>,
+                    'app_name': <str:app_name>,
+                    'group_name': <str:group_name>,
+                    'actor_user_name': <str:actor_user_name>,
+                    'user_name': <str:user_name>,
+                    'policy_name': <str:policy_name>,
+                    'otp_device_name': <str:otp_device_name>,
+                    'operation_name': <str:operation_name>,
+                    'directory_sync_run_id': <int:directory_sync_run_id>,
+                    'directory_id': <int:directory_id>,
+                    'resolution': <str:resolution>,
+                    'client_id': <int:client_id>,
+                    'resource_type_id': <int:resource_type_id>,
+                    'error_description': <str:error_description>
+                }
+            ]
+        """
+        # OneLogin API expects the ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ
+        formatted_date = datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
+        params = {'since': formatted_date}
+
+        # Make sure we have authentication headers
+        if not self._auth_headers:
+            return False
+
+        events = self._get_onelogin_paginated_events(params)
+
+        while self._more_to_poll and self._next_page_url:
+            LOGGER.debug('More events to retrieve for \'%s\': %s', self.type(), self._more_to_poll)
+            pagination = self._get_onelogin_paginated_events(None)
+
+            # Add the events to our results, this is equivalent to using events.extend()
+            events += pagination
+
+        # Return the list of logs to the caller so they can be send to the batcher
+        return events
+
+    def _get_onelogin_paginated_events(self, params):
+        """Get events from the API pagination url
+
+        Returns:
+            The same as the method _get_onelogin_events()
+        """
+        response = requests.get(self._next_page_url, headers=self._auth_headers, params=params)
+        if not self._check_http_response(response):
+            return False
+
+        # Extract events from response
+        events = response.json()['data']
+
+        # Do we have a pagination link to follow?
+        self._more_to_poll = len(events) >= self._MAX_EVENTS_LIMIT
+
+        # If we are already retrieved all the events, then set the value to prevent more requests
+        self._next_page_url = response.json()['pagination']['next_link']
+
+        return events
+
+    def required_auth_info(self):
+        return {
+            'client_secret':
+                {
+                    'description': ('the client secret for the OneLogin API. This '
+                                    'should a string of 57 alphanumeric characters'),
+                    'format': re.compile(r'^[a-zA-Z0-9]{57}$')
+                },
+            'client_id':
+                {
+                    'description': ('the client id for the OneLogin API. This '
+                                    'should a string of 57 alphanumeric characters'),
+                    'format': re.compile(r'^[a-zA-Z0-9]{57}$')
+                }
+            }
+
+    def _sleep_seconds(self):
+        """Return the number of seconds this polling function should sleep for
+        between requests to avoid failed requests. OneLogin tokens allows for 5000 requests
+        every hour, so returning 0 for now.
+
+        Returns:
+            int: Number of seconds that this function shoud sleep for between requests
+        """
+        return 0
--- a/tests/unit/app_integrations/test_apps/test_onelogin.py
+++ b/tests/unit/app_integrations/test_apps/test_onelogin.py
@@ -1,0 +1,211 @@
+"""
+Copyright 2017-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+# pylint: disable=abstract-class-instantiated,protected-access,no-self-use
+from mock import Mock, patch
+
+from nose.tools import assert_equal, assert_false, assert_items_equal
+
+from app_integrations.apps.onelogin import OneLoginApp
+from app_integrations.config import AppConfig
+
+from tests.unit.app_integrations.test_helpers import (
+    get_valid_config_dict,
+    MockSSMClient
+)
+
+
+@patch.object(OneLoginApp, 'type', Mock(return_value='type'))
+@patch.object(OneLoginApp, '_endpoint', Mock(return_value='endpoint'))
+@patch.object(AppConfig, 'SSM_CLIENT', MockSSMClient())
+class TestOneLoginApp(object):
+    """Test class for the OneLoginApp"""
+
+    def __init__(self):
+        self._app = None
+        self._sample_event = {
+            'id': 123,
+            'created_at': '2017-10-05T18:11:32Z',
+            'account_id': 1234,
+            'user_id': 321,
+            'event_type_id': 4321,
+            'notes': 'Notes',
+            'ipaddr': '0.0.0.0',
+            'actor_user_id': 987,
+            'assuming_acting_user_id': 654,
+            'role_id': 456,
+            'app_id': 123456,
+            'group_id': 98765,
+            'otp_device_id': 11111,
+            'policy_id': 22222,
+            'actor_system': 'System',
+            'custom_message': 'Message',
+            'role_name': 'Role',
+            'app_name': 'App Name',
+            'group_name': 'Group Name',
+            'actor_user_name': '',
+            'user_name': 'username',
+            'policy_name': 'Policy Name',
+            'otp_device_name': 'OTP Device Name',
+            'operation_name': 'Operation Name',
+            'directory_sync_run_id': 7777,
+            'directory_id': 6666,
+            'resolution': 'Resolved',
+            'client_id': 11223344,
+            'resource_type_id': 44332211,
+            'error_description': 'ERROR ERROR'
+        }
+
+    # Remove all abstractmethods so we can instantiate DuoApp for testing
+    # Also patch some abstractproperty attributes
+    @patch.object(OneLoginApp, '__abstractmethods__', frozenset())
+    def setup(self):
+        """Setup before each method"""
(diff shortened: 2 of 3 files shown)


mime-frame:
```
42.56s$ ./tests/scripts/pylint.sh
************* Module app_integrations.apps.onelogin
W: 22, 0: Unused app imported from app_integrations.apps.app_base (unused-import)
```

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,193 @@
...
+                {
+                    'description': ('the client id for the OneLogin API. This '
+                                    'should a string of 57 alphanumeric characters'),
+                    'format': re.compile(r'^[a-zA-Z0-9]{57}$')
+                }
+            }
+
+    def _sleep_seconds(self):
+        """Return the number of seconds this polling function should sleep for
+        between requests to avoid failed requests. OneLogin tokens allows for 5000 requests
```
mime-frame: Just to confirm: 5000/hour = ~83/min = ~1.3/sec - there's no way we'll accidentally exceed this?

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,193 @@
...
+        every hour, so returning 0 for now.
+
+        Returns:
+            int: Number of seconds that this function shoud sleep for between requests
```
mime-frame: should

tests/unit/app_integrations/test_apps/test_onelogin.py
```diff
@@ -0,0 +1,179 @@
...
+@patch.object(OneLoginApp, 'type', Mock(return_value='type'))
+@patch.object(OneLoginApp, '_endpoint', Mock(return_value='endpoint'))
+@patch.object(AppConfig, 'SSM_CLIENT', MockSSMClient())
+class TestOneLoginApp(object):
+    """Test class for the OneLoginApp"""
+
+    def __init__(self):
+        self._app = None
+
+    # Remove all abstractmethods so we can instantiate DuoApp for testing
```
mime-frame: OneLoginApp

app_integrations/apps/onelogin.py
mime-frame: Spoke offline, Javier will intentionally hit the limit locally so he can appropriately catch the exception and throw a useful error if it occurs

mime-frame:
Since @ryandeivert is out, @austinbyers and @jacknagz can you take a look?

app_integrations/apps/onelogin.py
javuto: I need to do some testing in the actual API to see if we can just generate a new token once we exceed the maximum number of requests per hour (500) per token. Although I doubt they would allow such a simple workaround to their rate limit. In any case I can add the code to have a safe sleep of 2 seconds per request.

javuto:
Just added a new rule, `onelogin_events_assumed_role` and tested it:
```
$ python manage.py lambda test --processor rule --rules onelogin_events_assumed_role

onelogin_events_assumed_role
       [Pass]  [trigger=1]                   rule      (stream_alert_app): OneLogin generated event when a user assumed a different role, it should alert
       [Pass]  [trigger=0]                   rule      (stream_alert_app): OneLogin generated event when a user do any other action, it should not alert



```
Schemas also validated:
```
$ python manage.py validate-schemas
...
onelogin_events_assumed_role
       [Pass]  [log='onelogin:events']       validation  (stream_alert_app): OneLogin generated event when a user assumed a different role, it should alert
       [Pass]  [log='onelogin:events']       validation  (stream_alert_app): OneLogin generated event when a user do any other action, it should not alert
...
StreamAlertCLI [INFO]: (27/27) Successful Tests
```

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,192 @@
...
+            https://developers.onelogin.com/api-docs/1/oauth20-tokens/generate-tokens-2
+
+        Returns:
+            str: Bearer token to be used to call the OneLogin resource APIs
+        """
+        authorization = 'client_id: %s, client_secret: %s' % (client_id, client_secret)
+        headers_token = {'Authorization': authorization,
+                         'Content-Type': 'application/json'}
+
+        response = requests.post(token_url,
```
ryandeivert: I anticipate potentially needing `post` abilities in other future apps as well. Could we move this functionality to the `AppIntegration` base class in `app_base.py` with a method name of something like `_make_post_request` and change the `_make_request` method to `_make_get_request`? By doing this, the new method could just return False or the JSON from the response (similar to how `_make_request` works now).

If this is too much right now, this can happen at a later date - just a thought!

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,192 @@
...
+        if not self._check_http_response(response):
+            return False
+
+        bearer = 'bearer:%s' % (response.json()['access_token'])
+        self._auth_headers = {'Authorization': bearer}
+
+    def _gather_logs(self):
+        """Gather the authentication log events."""
+
+        if not self._auth_headers:
```
ryandeivert: ++ nice!

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,192 @@
...
+                    'resolution': <str:resolution>,
+                    'client_id': <int:client_id>,
+                    'resource_type_id': <int:resource_type_id>,
+                    'error_description': <str:error_description>
+                }
+            ]
+        """
+        # OneLogin API expects the ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ
+        formatted_date = datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
+        params = {'since': formatted_date}
```
ryandeivert: So this is going to actually only get events since the current time. This will likely result in events for only a few seconds or less (the time between setting this date and actually querying their API for the events).

You'll want to use the `self._last_timestamp` for this param, but we'll have to reason about this more since you need a formatted time and not a unix/integer timestamp

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,192 @@
...
+
+        # Make sure we have authentication headers
+        if not self._auth_headers:
+            return False
+
+        events = self._get_onelogin_paginated_events(params)
+
+        while self._more_to_poll and self._next_page_url:
```
ryandeivert: I actually think we should avoid doing the pagination directly within the subclass and just let the `while` within the base classes gather handle this. We can talk more about the requirements for this.

The main thought process here is that this while loop could loop for a good amount of time, and the subclassed apps don't have a concept of when the lambda function is getting close to timing out. By letting the looping happen on the base class, where there is some checks in place to look for nearing the time out, we can operate with a little more safety.

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,192 @@
...
+        # Do we have a pagination link to follow?
+        self._more_to_poll = len(events) >= self._MAX_EVENTS_LIMIT
+
+        # If we are already retrieved all the events, then set the value to prevent more requests
+        self._next_page_url = response.json()['pagination']['next_link']
+
+        return events
+
+    def required_auth_info(self):
+        return {
```
ryandeivert: As we spoke about offline, please add the localization (us/eu) to these so they can be user-provided.

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,192 @@
...
+            'client_secret':
+                {
+                    'description': ('the client secret for the OneLogin API. This '
```
ryandeivert: Nice work with these!

app_integrations/apps/onelogin.py
ryandeivert: Agreed that additional testing should be done to best determine how this should be handled.

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,192 @@
...
+        # Return the list of logs to the caller so they can be send to the batcher
+        return events
+
+    def _get_onelogin_paginated_events(self, params):
+        """Get events from the API pagination url
+
+        Returns:
+            The same as the method _get_onelogin_events()
+        """
+        response = requests.get(self._next_page_url, headers=self._auth_headers, params=params)
```
ryandeivert: Can the base class' `_make_request` be used in this instance, since it will do the response check and automatically return the json response value.

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,192 @@
...
+import re
+import requests
+
+from app_integrations import LOGGER
+from app_integrations.apps.app_base import AppIntegration
+
+
+class OneLoginApp(AppIntegration):
+    """OneLogin StreamAlert App"""
+    _ONELOGIN_EVENTS_URL = 'https://api.us.onelogin.com/api/1/events'
```
ryandeivert: As we spoke about offline, switch to using a url that can be formatted with a region/localization (us/eu) for this link and the token link.

app_integrations/apps/onelogin.py
javuto: Should I also refactor the duo application to use the new `_make_get_requests` in this PR?

ryandeivert: Yes please! :)

app_integrations/apps/app_base.py
```diff
@@ -222,6 +222,24 @@ def _make_request(self, full_url, headers, params):
...
         return response.json()
 
+    def _make_post_request(self, full_url, headers, json):
+        """Method for returning the json loaded response for this POST request
+
+        Returns:
+            bool or dict: False if the was an error performing the request,
+                or a dictionary loaded from the json response
+        """
+        LOGGER.debug('Making request for service \'%s\' on poll #%d',
```
ryandeivert: Can you add `post` to this logger message somewhere, ie: `Making post request for...`

app_integrations/apps/app_base.py
```diff
@@ -222,6 +222,24 @@ def _make_request(self, full_url, headers, params):
 
         return response.json()
 
+    def _make_post_request(self, full_url, headers, json):
```
ryandeivert: Can we choose a different name from `json` for the variable, since that conflicts with the `json` package?

app_integrations/apps/duo.py
```diff
@@ -115,7 +115,7 @@ def _get_duo_logs(self, hostname, full_url):
             return False
 
         # Make the request to the api, resulting in a bool or dict
-        response = self._make_request(full_url, headers=headers, params=params)
+        response = self._make_get_request(full_url, headers=headers, params=params)
```
ryandeivert: Thanks for this!

app_integrations/apps/onelogin.py
```diff
@@ -124,59 +128,50 @@ def _get_onelogin_events(self):
...
+        # Return the list of logs to the caller so they can be send to the batcher
         return events
 
     def required_auth_info(self):
         return {
+            'region':
+                {
+                    'description': ('the region for the OneLogin API. This should be a '
+                                    'string of 2 letters, lowercase or uppercase'),
+                    'format': re.compile(r'^[a-zA-Z]{2}$')
```
ryandeivert: can we restrict this to `eu` or `us` since those are the only regions OneLogin supports?

app_integrations/apps/onelogin.py
```diff
@@ -124,59 +128,50 @@ def _get_onelogin_events(self):
...
 
-            # Add the events to our results, this is equivalent to using events.extend()
-            events += pagination
+        LOGGER.debug('Events to retrieve for \'%s\': %s', self.type(), self._more_to_poll)
+        response = self._make_get_request(self._next_page_url, self._auth_headers, params)
 
-        # Return the list of logs to the caller so they can be send to the batcher
-        return events
+        if not response:
+            events = False
```
ryandeivert: can we just `return False` here instead? the line below (151) could result in a KeyError

mime-frame:
From CI:

```
StreamAlertCLI [ERROR]: (2/27) Failures
StreamAlertCLI [ERROR]: (1/2) Detected old format for test event in file 'onelogin_events_assumed_role.json'. Please visit https://streamalert.io/rule-testing.html for information on the new format and update your test events accordingly.
StreamAlertCLI [ERROR]: (2/2) Detected old format for test event in file 'onelogin_events_assumed_role.json'. Please visit https://streamalert.io/rule-testing.html for information on the new format and update your test events accordingly.
```

^ You need to update your tests as @ryandeivert merged his changes. You'll need to specify `logs` and `trigger_rules`, here's an example: https://github.com/airbnb/streamalert/pull/372/files#diff-0e3d1b3c864f17c2c1e876b811c33590

Full details are here: https://streamalert.io/rule-testing.html

javuto:
@ryandeivert I have added a significant update to the PR, that It will probably require another round of code review. It does include the update to the tests format that @mime-frame pointed out above, the refactoring of get/post requests to return a tuple and handling the API rate limit sleep, more info [here](https://developers.onelogin.com/api-docs/1/oauth20-tokens/get-rate-limit).

Thanks guys!

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,233 @@
...
+                formatted_date = datetime.fromtimestamp(
+                    self._last_timestamp).strftime('%Y-%m-%dT%H:%M:%SZ')
+            else:
+                # We should have already this timestamp in format ISO 8601
+                formatted_date = self._last_timestamp
+
+            params = {'since': formatted_date}
+            request_url = self._events_endpoint()
+
+        LOGGER.debug('Events to retrieve for \'%s\': %s', self.type(), self._more_to_poll)
```
ryandeivert: hey @javuto - I added a similar logging message to this in loop within the base class:

https://github.com/airbnb/streamalert/blob/f5e866d8aca544633af416bd0bda258cff82ee2d/app_integrations/apps/app_base.py#L310

You can probably safely remove this, or change it to add more context on what's happening at this point in the code.

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,233 @@
...
+        result, response = self._make_get_request(request_url, self._auth_headers, params)
+
+        if not result:
```
ryandeivert: You'll want to make sure `response` is valid here as well, since it _could_ be None.  Below you may also want to change your lookups to `reponse.get('status')`, etc in case the response is not in a format we expect (just for safety).

ryandeivert: Maybe do this:

```
if not (result or response):
   self._more_to_poll = False
   return False
elif not result and response:
   <do your logic here>
```

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,233 @@
...
+                self._set_rate_limit_sleep()
+
+            return False
+
+        # Set pagination link, if there is any
+        self._next_page_url = response['pagination']['next_link']
+        self._more_to_poll = (self._next_page_url is not None)
+
+        # Adjust the last seen event
+        self._last_timestamp = response['data'][-1]['created_at']
```
ryandeivert: Let's add a safety check here to make sure the `data` list actually has logs:

```
logs = response['data']
if not logs:
    return False

self._last_timestamp = logs[-1]['created_at']

# Return the list of logs to the caller so they can be send to the batcher
return logs
```

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,233 @@
...
+            # If we hit the rate limit, update the sleep time
+            r_status = response['status']
+            if r_status['code'] == 400 and r_status['message'] == 'rate_limit_exceeded':
+                self._set_rate_limit_sleep()
+
+            return False
+
+        # Set pagination link, if there is any
+        self._next_page_url = response['pagination']['next_link']
+        self._more_to_poll = (self._next_page_url is not None)
```
ryandeivert: Tip (not a necessary change): you can use `bool()` to perform this logic:

`self._more_to_poll = bool(self._next_page_url)`

app_integrations/apps/onelogin.py
```diff
@@ -0,0 +1,233 @@
...
+        between requests to avoid failed requests. OneLogin tokens allows for 5000 requests
+        every hour, but if the rate limit is reached, we can retrieve how long until we are clear.
+
+        More information about this here:
+            https://developers.onelogin.com/api-docs/1/oauth20-tokens/get-rate-limit
+
+        Returns:
+            int: Number of seconds that this function should sleep for between requests
+        """
+        return self._rate_limit_sleep
```
ryandeivert: ++ Nice work on this logic!

app_integrations/apps/app_base.py
```diff
@@ -204,23 +204,35 @@ def _check_http_response(self, response):
 
         return success
 
-    def _make_request(self, full_url, headers, params):
-        """Method for returning the json loaded response for this request
+    def _make_get_request(self, full_url, headers, params):
```
ryandeivert: Let's make the `params` arg be an optional that defaults to None:

`def _make_get_request(self, full_url, headers, params=None):..`

And then update your call [here](https://github.com/airbnb/streamalert/pull/355/files#diff-02782405a17cef489b01b36d62d22196R109) so it doesn't have to pass `None` explicitly

app_integrations/apps/onelogin.py
ryandeivert: ^^ This might be unnecessary - I'll let you decide :)

ryandeivert: ^^ This might be unnecessary - I'll let you decide :)

app_integrations/apps/onelogin.py
javuto: Per our offline chat, having this logging message here it is redundant with the added one in the base class. I will remove it.

app_integrations/apps/onelogin.py
javuto: I think this approach can add some unnecessary complexity to the code. I like checking first `result`, then checking `response`. It makes the code easy to read and to follow the flow.

app_integrations/apps/onelogin.py
```diff
@@ -176,26 +193,34 @@ def _get_onelogin_events(self):
...
-        self._next_page_url = response['pagination']['next_link']
-        self._more_to_poll = (self._next_page_url is not None)
+        pagination = response.get('pagination')
+        self._next_page_url = pagination['next_link']
+        self._more_to_poll = bool(self._next_page_url)
 
         # Adjust the last seen event
-        self._last_timestamp = response['data'][-1]['created_at']
+        events = response.get('data')
+        self._last_timestamp = events[-1]['created_at']
```
ryandeivert: Good changes here, except you aren't checking to see if the `events` list is empty (and `events[-1]` could cause an index out of range error if no events are returned). You can also probably drop the `get(` above and just use `events['data']` since they key will exist even if it is an empty list.

ryandeivert: What I'm worried about is this case:
```
{
    "status": {
        "error": false,
        "code": 200,
        "type": "success",
        "message": "Success"
    },
    "pagination": {
        "before_cursor": null,
        "after_cursor": "xWNjb3VudF9pZDo6OjUzNDEzLS0jI2lkOjo6OTA0MjU3NTQ2",
        "previous_link": null,
        "next_link": null
    },
    "data": []
}
```

app_integrations/config.py
```diff
@@ -266,7 +266,7 @@ def _determine_last_time(self):
             current_time = time.mktime(time.gmtime())
             LOGGER.debug('Current timestamp: %s seconds', current_time)
 
-            self.last_timestamp = current_time - interval_time
+            self.last_timestamp = int(current_time - interval_time)
```
ryandeivert: Thanks for this fix

app_integrations/apps/onelogin.py
```diff
@@ -167,7 +184,7 @@ def _get_onelogin_events(self):
             # Check the type to understand the format stored
             if isinstance(self._last_timestamp, int):
                 # OneLogin API expects the ISO 8601 format: YYYY-MM-DDTHH:MM:SSZ
-                formatted_date = datetime.fromtimestamp(
+                formatted_date = datetime.utcfromtimestamp(
```
ryandeivert: ++

tests/unit/app_integrations/test_helpers.py
```diff
@@ -96,6 +96,12 @@ def get_auth_info(cls, app_type):
                 'integration_key': 'DI1234567890ABCDEF12',
                 'secret_key': 'abcdefghijklmnopqrstuvwxyz1234567890ABCD'
             }
+        elif app_type in {'onelogin'}:
```
ryandeivert: 👏 👏  for simplification

app_integrations/apps/onelogin.py
javuto: I see, I will handle this case so it will be safe if it appears

Issue #258, opened 2017-08-11
Improvement: Rename `log_patterns` to `required_key_values`


mime-frame:
**Proposal**

Rename `log_patterns` to `required_key_values`

Why:
- It more accurately captures what it does (it's not confusing)

**Work required**
- Code refactor
- Docs refactor

ibrahimsharaf:
Hi @mime-frame, I'll start working on this one.

jacknagz:
hey @ibrahimsharaf thanks for taking this on! unfortunately we no longer want to make this change (this is slightly outdated)

PR #358, opened 2017-10-07
[config][docs][core][testing] rename `log_patterns` to `required_key_values

ibrahimsharaf:
to: @mime-frame 
size: small
resolves: https://github.com/airbnb/streamalert/issues/258


jacknagz:
hey @ibrahimsharaf we are going to leave `log_patterns` as is, that issue was a little dated

Issue #194, opened 2017-06-23
Terraform: Do Not Refresh State For Every Deploy Step


jacknagz:
## Background

During the deployment process, Terraform runs and refreshes state several times.  To shorten this process, the state refresh can be skipped after the first iteration with `-refresh=false`

ibrahimsharaf:
Hi @jacknagz, I am interested in taking this one, where can I start?

jacknagz:
hey @ibrahimsharaf, [this function](https://github.com/airbnb/streamalert/blob/master/stream_alert_cli/helpers.py#L101) essentially needs a new kwarg for `refresh_state`, where it would conditionally add `-refresh=false` to the Terraform command.  Hope that helps!

austinbyers:
Closed via #604. This will be released with v2.0.0

PR #359, opened 2017-10-09
[cli] add refresh_state bool args

ibrahimsharaf:
to: @jacknagz 
size: small
resolves: https://github.com/airbnb/streamalert/issues/194

### DIFF
--- a/stream_alert_cli/helpers.py
+++ b/stream_alert_cli/helpers.py
@@ -115,7 +115,7 @@ def tf_runner(**kwargs):
     """
     targets = kwargs.get('targets', [])
     action = kwargs.get('action', None)
-    refresh_state = kwargs.get('refresh_state', False)
+    refresh_state = kwargs.get('refresh_state', True)
     tf_action_index = 1  # The index to the terraform 'action'
 
     var_files = {'conf/lambda.json'}


stream_alert_cli/helpers.py
```diff
@@ -110,12 +110,14 @@ def tf_runner(**kwargs):
...
         targets: a list of Terraform targets
         action: 'apply' or 'destroy'
+        refresh_state: True or False
 
     Returns:
         bool: True if the terraform command was successful
     """
     targets = kwargs.get('targets', [])
     action = kwargs.get('action', None)
+    refresh_state = kwargs.get('refresh_state', False)
```
jacknagz: This should default to `True`, since we only want to disable refreshing during latter deploy steps

stream_alert_cli/helpers.py
```diff
@@ -125,6 +127,9 @@ def tf_runner(**kwargs):
     if action == 'destroy':
         tf_command.append('-destroy')
 
+    if not refresh_state:
```
jacknagz: This logic should switch per the previous comment

ibrahimsharaf: how should it switch? what would it do in case `refresh_state` is True?

ibrahimsharaf: Are you thinking of something like that?
` if refresh_state:
    tf_command.append('-refresh=true')
  else:
        tf_command.append('-refresh=false') `

jacknagz: actually, this doesn't need to change.

ibrahimsharaf: That's what I thought too :smile:

ibrahimsharaf:
@jacknagz what should the unit tests cover?

jacknagz:
@ibrahimsharaf they should describe how you verified that this feature is working.  we don't have great unit test coverage right now for this part of the library, but integration testing should suffice

ibrahimsharaf:
@jacknagz I am not really familiar with the code-base or integration testing methodology, can you give me a hand?

jacknagz:
@ibrahimsharaf https://www.streamalert.io/deployment.html to verify things are working.  again, I didn't see any implementation of the feature, so nothing is going to happen

stream_alert_cli/helpers.py
ibrahimsharaf: Fixed.

ibrahimsharaf:
> I didn't see any implementation of the feature, so nothing is going to happen

@jacknagz should I modify this line https://github.com/airbnb/streamalert/blob/master/stream_alert_cli/manage_lambda/deploy.py#L160 to `    helpers.tf_runner(targets=targets, refresh_state=False)
`?

ibrahimsharaf:
@jacknagz can you review the changes and tell me if this implements the feature? also I don't have an AWS account so I can't setup streamalert and deploy, can you take it from here if so?

austinbyers:
Thanks @ibrahimsharaf for the PR! I'm sorry we were never able to merge it, but we definitely referenced it when implementing the fix

PR #362, opened 2017-10-09
[cli] support athena table creation with envelope keys

jacknagz:
to: @ryandeivert 
size: small
resolves: N/A

## Background

Creating Athena tables via the CLI requires conversion of StreamAlert schemas to Hive DDL statements.  To accommodate schemas with envelope keys, additional changes were necessary.

## Changes

* Support StreamAlert envelope keys in Hive CREATE TABLE commands

## Testing

* Tested by running the following command in a staging AWS account:

```
python manage.py athena create-table --refresh_type add_hive_partition \ 
--bucket <my-test-prefix>.streamalert.data \
--table_name cloudwatch_flow_logs \ 
--type data
```

### DIFF
--- a/stream_alert_cli/runner.py
+++ b/stream_alert_cli/runner.py
@@ -128,7 +128,10 @@ def athena_handler(options):
                                 options.table_name)
                 return
 
-            schema = CONFIG['logs'][options.table_name.replace('_', ':')]['schema']
+            log_info = CONFIG['logs'][options.table_name.replace('_', ':', 1)]
+            schema = dict(log_info['schema'])
+            schema_statement = ''
+
             sanitized_schema = StreamAlert.sanitize_keys(schema)
 
             athena_schema = {}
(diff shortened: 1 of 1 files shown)


stream_alert_cli/runner.py
```diff
@@ -141,24 +144,62 @@ def athena_handler(options):
...
+            def add_to_athena_schema(schema, root_key=''):
+                """Helper function to add sanitized schemas to the Athena table schema"""
+                # Setup the root_key dict
+                special_key = None
+                if root_key and not athena_schema.get('root_key'):
+                    athena_schema[root_key] = {}
+
+                for key_name, key_type in schema.iteritems():
+                    # Transform the {} or [] into hashable types
+                    if key_type == {}:
```
ryandeivert: Should `special_key` be reset to `None` here? Or just instantiate within this loop instead of outside of it?

stream_alert_cli/runner.py
```diff
@@ -141,24 +144,62 @@ def athena_handler(options):
...
-                                        for key_name, key_type
-                                        in athena_schema.iteritems()])[:-1]
-            query = ('CREATE EXTERNAL TABLE {table_name} ({schema})'
-                     'PARTITIONED BY (dt string)'
-                     'ROW FORMAT SERDE \'org.openx.data.jsonserde.JsonSerDe\''
+            def add_to_athena_schema(schema, root_key=''):
+                """Helper function to add sanitized schemas to the Athena table schema"""
+                # Setup the root_key dict
+                special_key = None
+                if root_key and not athena_schema.get('root_key'):
```
ryandeivert: I think the `get('root_key')` may be inadvertently quoted here. Are you looking for the value of the var `root_key` in the athena_schema, or the string literal `'root_key'`?

stream_alert_cli/runner.py
```diff
@@ -141,24 +144,62 @@ def athena_handler(options):
...
+
+            add_to_athena_schema(sanitized_schema)
+
+            # Support envelope keys
+            configuration_options = log_info.get('configuration')
+            if configuration_options:
+                envelope_keys = configuration_options.get('envelope_keys')
+                if envelope_keys:
+                    sanitized_envelope_keys = StreamAlert.sanitize_keys(envelope_keys)
+                    # Note: this key is wrapped in backticks to be Hive compliant
```
ryandeivert: Good note - why are back-ticks required for this key but they don't seem to be necessary for other keys?

jacknagz: because of the special char in the key name

ryandeivert: Makes sense thanks for clarifying!

stream_alert_cli/runner.py
jacknagz: oh wow good catch

Issue #351, opened 2017-10-05
Bug: Normalized Types is not reset during rule processing


ryandeivert:
### Issue
The processing of rules will iterate over records within a payload, and then iterate over each rule attempting to process the rule against the record.

During this processing, we check to see if a rule utilizes the `datatypes` directive and then we do some logic to determine if this record has any keys that reflect these normalized `datatypes`. Any `normalized_types` that are discovered get cached on the `record` itself within the `normalized_types` key, seen [here](https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/rules_engine.py#L387).

However, this `normalized_types` key is never reset, and will remain populated with any normalized data values from previous rules. Future rules that are processed for this record will have previous datatypes populated, which can cause problems/result in misleading record data.

chunyong-lin:
This issue has been fixed in PR #367.

PR #367, opened 2017-10-11
[Data Normalization][Bug] Insert normalized types to a copy of record

chunyong-lin:
to: @ryandeivert @javuto 
size: small
solves: #351

## Background

Context for the change

## Changes

* If insert `normalized_types` to the original record, it will prevent other rules which have no datatypes parameters to work correctly.
* Create a local variable to store a copy of the record and pass it to alert if necessary.

## Testing

* Unit Test
  * Create unit test case to reproduce the issue.
  * Apply fix and rerun unit test is passed. 
* Rule Test
  * Create local testing rules and rules test is passed.

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -183,7 +183,7 @@ class StreamRules(object):
                     "defined_type2": [[original_key2, sub_key2], [original_key3]]
                 }
         """
-        results = dict()
+        results = None
         if not (datatypes and cls.validate_datatypes(normalized_types, datatypes)):
             return results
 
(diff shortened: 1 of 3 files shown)


stream_alert/rule_processor/rules_engine.py
```diff
@@ -380,20 +380,24 @@ def process(cls, input_payload):
...
                     continue
+
+                types_result = None
                 if rule.datatypes:
                     types_result = cls.match_types(record,
                                                    payload.normalized_types,
                                                    rule.datatypes)
-                    record['normalized_types'] = types_result
 
+                record_copy = record.copy()
```
ryandeivert: I don't love that we're copying a record in memory for every rule in the loop. Can't we just see if the the `record` has the `normalized_types` key in it and delete it if it does? For example, around line 374 above, can you just:
```
if 'normalized_types' in record:
    del record['normalized_types']
```

chunyong-lin: We can not use `del` method here because once call to delete `normalizied_types` from a record, it will also delete `normalized_types` from alert['record'] who has `normalized_types` field. 
For a trade-off, I will make a record copy only when normalized types take effect.

ryandeivert:
@chunyong-lin thanks for the thoughtful approach to this! :)

PR #368, opened 2017-10-11
[cli][tests] rule testing framework mods

### LINKED ISSUE #306: Bug/Improvement: Folder for schema tests
**Overview**

Currently, folks are putting schema test files in `tests/integration/rules`.

These causes warnings when testing locally:

`StreamAlertCLI [WARNING]: No rules configured for test file: '...'. Please add a corresponding rule for this test file in 'rules/' with the name '...' to avoid seeing this warning and any associated errors above`

**Suggested Fix**

Create a new folder in the `tests` directory that is intended to house only schema test files. Ex: `tests/integration/schemas`.

Ensure that:
1) the above warning does not appear
2) schema checks still correctly work when running `python manage.py validate-schemas`
3) the docs are updated

### LINKED ISSUE #338: Bug/Improvement: Tests: Denote relationship in JSON file vs. relying on the filename
**Context**

1) Rule tests (`/tests/integration/rules`) require that their filename matches the name of their associated rule (aka what was defined in the `def` block). 

2) Schema tests (which will soon have their own folder for unrelated reasons - https://github.com/airbnb/streamalert/issues/306), also have this filename requirement.

Note that (2) is actually causing a bug right now for folks attempting to validate a schema only, without an associated rule:

```
  File "/Users/../streamalert/stream_alert_cli/test.py", line 467, in analyze_record_delta
    rule_info = StreamRules.get_rules()[rule_name]
KeyError: 'the_name_of_the_json_file'
```

**Desired Changes**

Test files should denote what schema or rule they are testing via a key-value pair defined within the JSON file.

It's worth discussing that once this change is made, what should we encourage from a filename perspective and from a directory structure perspective (aka should the folder structure from the rules directory be copied into the `tests/integration/rules` directory)

ryandeivert:
to: @jacknagz 
size: medium
resolves #338 
resolves #306 (the requirements for this issue are met through other means in this PR, noted below)

## Background
* StreamAlert has changed significantly with the addition of new features, and the rule testing framework needs to be updated to reflect these changes.
#### Recent changes to warrant this PR
* StreamAlert now supports data normalization, which can utilize the `datatypes` directive in rule decorators instead of specifying `logs`
* Schema validation has been added to verify a new schema added to `logs.json` via a json test event without the need to add a corresponding rule.
#### Issues
* It’s not readily apparent to end users that the basename of a json test event file must match the basename of python rule file in order for the test to run.
* Schema validation must have some knowledge of which log schema a test event should match against in order to provide more context on errors within the test event.
* Rules written using the `datatypes` directive do not have any knowledge of the type of log they will trigger on. The test events need to contain this relationship to determine if the rule resulted in an alert for this test event.

## Changes
* Test events are now log type and rule aware, so we can more easily evaluate schema issues
* Validation of schemas will not prompt about no rules existing for test event & vice versa (works on #306)
* The cli command of `python manage.py lambda test` now supports both the `--test-rules` and `--test-files` flags.
  * This change is to help with testing specific test event files OR specific rules. These flags are mutually exclusive and cannot be combined.
* Simplifying how warnings are reported and consolidating some noisy ones into a single log message
* Updating how the test event files are loaded from disk to allow for nested folders within the `tests/integration/rules/ directory`. This allows for better organization of test events.
* Adding helpers for loading the test files and extracting the rules being tested from them.
* Removing source template files (`s3.json`, `kinesis.json`, etc) in favor of codified versions of them.
* Adding support for analyzing test record errors with `json_path` nested records and for analyzing errors caused by envelope keys.
* Test event files should now be in the following format (**note the removal of the `trigger` key and addition of the `trigger_rules`, `log`, and `validate_schema_only` keys**):
```
{
  "records": [
    {
      "data": {
        "key_01": "value_01",
        "key_02": "value_02"
      },
      "description": "Description of test event",
      "service": "service_name",
      "source": "source_name",
      "log": "log_name",     # This is a 1:1 relationship
      "trigger_rules": [     # This test can trigger 0+ rule(s)
        "rule_for_this_test_01",
        "rule_for_this_test_02"
      ],
      "validate_schema_only": true   # This is optional
    }
  ]
}
```
* Test event files in the old format will be detected and reported to the user, like so:
`StreamAlertCLI [ERROR]: (1/1) Detected old format for test event in file 'cb_watchlist_hit_binary.json'. Please visit https://streamalert.io/rule-testing.html for information on the new format and update your test events accordingly.`
* The optional `validate_schema_only` flag can be used in a test event to depict that a record would only be validated, and rule logic should not be applied.
  * This change does not affect what happens if there is a rule that does not have a valid test event configured. Any rule that exists without a test event will still report a warning, similar to the following, to the user indicating so:

```
StreamAlertCLI [WARNING]: No test events configured for the following rules. Please add corresponding tests for these rules in 'tests/integration/rules' to avoid seeing this warning
    rule_without_test_example
```

## Testing
* The test event files in the repo have all been updated to conform to the new format and testing is passing with:
  * `python manage.py lambda test --processor all`
  * `python manage.py validate-schemas`
  * `python manage.py validate-schemas --test-files github_user_promotion_to_site_admin`
  * `python manage.py lambda test --processor rule --test-rules cloudtrail_put_bucket_acl fake_rule`
    * The above will report a warning that `fake_rule` does not exist
  * `python manage.py lambda test --processor rule --test-files duo_fraud.json`
* Tested against the old format to ensure the user is prompted.
* Tested `json_path` record delta analysis locally (but not committing the test event file).
* Adding some unit tests for a few CLI helper functions.

## Incoming
* PR for doc updates, namely to https://streamalert.io/rule-testing.html


stream_alert_cli/test.py
```diff
@@ -76,19 +75,21 @@ def __init__(self, context, config, print_output):
...
 
-    def test_processor(self, filter_rules, validate_only=False):
+    def test_processor(self, rules_filter, files_filter, validate_only):
         """Perform integration tests for the 'rule' Lambda function
 
         Args:
-            filter_rules (list|None): Specific rule names (or None) to restrict
-                testing to. This is passed in from the CLI using the --rules option.
+            rules_filter (set): A collection of rules to filter on, passed in by the user
+            filter_rules (set): Specific rule names to restrict testing to. This is
```
jacknagz: shouldn't this be named `files_filter`

stream_alert_cli/test.py
```diff
@@ -97,89 +98,130 @@ def test_processor(self, filter_rules, validate_only=False):
...
+
+        return file_info
+
+    def _validate_test_record(self, file_name, test_event, formatted_record, print_header_line):
         """Function to validate test records and log any errors
 
         Args:
-            rule_name (str): The rule name being tested
-            test_record (dict): A single record to test
+            file_name (str): The base name of the test event file.
```
jacknagz: `print_header_line` is missing from here

manage.py
```diff
@@ -904,33 +895,189 @@ def _add_lambda_subparser(subparsers):
...
+StreamAlertCLI v{}
+Deploy Lambda functions
+
+Command:
+
+    manage.py lambda deploy            Deploy Lambda functions
+
+Required Arguments:
+
+    -p/--processor                     The name of the Lambda function to deploy.
```
jacknagz: how does this change deployment logic? was this tested?

stream_alert_cli/helpers.py
```diff
@@ -154,6 +152,107 @@ def tf_runner(**kwargs):
...
+                    }
+                },
+                'Type': 'Notification',
+                'UnsubscribeUrl': 'EXAMPLE',
+                'TopicArn': 'arn:aws:sns:EXAMPLE',
+                'Subject': 'TestInvoke'
+            }
+        }
+
+    elif service == 'stream_alert_app':
```
jacknagz: I kind of wish this was just called `lambda`, any reason why it isn't?

ryandeivert: The only reason I can give is that we control what the incoming payload looks like (where as with sns/s3/kinesis/etc it is outside of our control). I'm not overly opposed to saying lambda, but it seem more clear to use this instead of calling it lambda one place and `stream_alert_app` another place

stream_alert_cli/test.py
ryandeivert: Yeah I think so - this must have gotten mixed up. Fixing

manage.py
ryandeivert: There is no change to how these commands work and everything still functions as it did before, testing the command works fine

stream_alert_cli/helpers.py
jacknagz: oh right, that's why.

mime-frame:
image

PR #372, opened 2017-10-11
streamalert - cloudtrail - security groups - ingress

mime-frame:
to: @ryandeivert or @jacknagz 

size: small

## Background

Alerts on Security Groups that allow anyone on the interwebs to touch your EC2 instance(s).

## Changes

- Helper function called `select_key`
- Rule called `cloudtrail_security_group_ingress_anywhere`
- Tests

## Testing

```
$ python manage.py lambda test --processor rule --test-files cloudtrail_security_group_ingress_anywhere

cloudtrail_security_group_ingress_anywhere
       [Pass]  [trigger=1]                   rule      (kinesis): A Security Group that allows ingress from anywhere on the internet should create an alert.
       [Pass]  [trigger=0]                   rule      (kinesis): A Security Group that allows ingress from an RFC1918 IP should not create an alert.



```

### DIFF
--- a/helpers/base.py
+++ b/helpers/base.py
@@ -172,13 +172,17 @@ def ghe_json_message(rec):
     return message_rec
 
 def select_key(data, search_key, results):
-    """
+    """ Recursively search for a given key and return all values
     Args:
         data (dict, list)
         search_key (string)
+        results (array)
+
+    Returns:
+        (list) all values
     """
     if not hasattr(data, 'items'):
-        return
+        return []
     if search_key in data:
         results.append(data[search_key])
     for key, val in data.iteritems():
--- a/helpers/base.py
+++ b/helpers/base.py
@@ -181,14 +181,23 @@ def select_key(data, search_key, results):
     Returns:
         (list) all values
     """
-    if not isinstance(data, (dict, list)):
-        return []
-    if search_key in data:
-        results.append(data[search_key])
+    # Check for lists - this will handle lists of lists, etc
+    if isinstance(data, list):
+        for item in data:
+            select_key(item, search_key, results)
+    # Only dictionaries can have keys, so return here if this is not one
+    if not isinstance(data, dict):
+        return
+    # Iterate over all of the key/values
     for key, val in data.iteritems():
+        # Handle nested dictionaries
         if isinstance(val, dict):
-            select_key(data[key], search_key, results)
+            select_key(val, search_key, results)
+        # Handle lists within a dictonary - the safety check at the top will handle nesting
         elif isinstance(val, list):
-            for item in val:
-                select_key(item, search_key, results)
+            select_key(val, search_key, results)
+        elif key == search_key:
+            # Finally, if this key is in the dictionary, extract the value for
+            # it and append to the results list that is passed by reference
+            results.append(val)
     return results
--- a/helpers/base.py
+++ b/helpers/base.py
@@ -171,7 +171,7 @@ def ghe_json_message(rec):
 
     return message_rec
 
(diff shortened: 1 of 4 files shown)


rules/community/cloudtrail/cloudtrail_security_group_ingress_anywhere.py
```diff
@@ -0,0 +1,33 @@
...
+
+@rule(logs=['cloudwatch:events'],
+      matchers=[],
+      outputs=['aws-s3:sample-bucket',
+               'pagerduty:sample-integration',
+               'slack:sample-channel'],
+      req_subkeys={'detail': ['eventName', 'requestParameters']})
+def cloudtrail_security_group_ingress_anywhere(rec):
+    """
+    author:         @mimeframe
```
ryandeivert: You got your own name wrong 🤔 @mime-frame

helpers/base.py
```diff
@@ -170,3 +170,25 @@ def ghe_json_message(rec):
...
+    """ Recursively search for a given key and return all values
+    Args:
+        data (dict, list)
+        search_key (string)
+        results (array)
+
+    Returns:
+        (list) all values
+    """
+    if not hasattr(data, 'items'):
```
ryandeivert: What's the goal of this check? To see if `data` is a dict? Or to check if it's iterable?

ryandeivert: If it's the former, please use `if not isinstance(data, dict): ...`

mime-frame: its the latter (it could be a dict or a list)

rules/community/cloudtrail/cloudtrail_security_group_ingress_anywhere.py
mime-frame: I use my twitter handle vs. github :)

helpers/base.py
ryandeivert: Okay in any case, lists don't have the `items` attribute, so you can pass a tuple to isinstance list: `isinstance(data, (dict, list))`

mime-frame: Done, thanks for the suggestion :)

mime-frame:
Interesting, I now only see 1 test mentioned even though I have 2?

`python manage.py lambda test --processor rule --test-rules cloudtrail_security_group_ingress_anywhere`

```

cloudtrail_security_group_ingress_anywhere
       [Pass]  [trigger=1]                   rule      (kinesis): A Security Group that allows ingress from anywhere on the internet should create an alert.



StreamAlertCLI [INFO]: (1/1) Successful Tests
```

helpers/base.py
```diff
@@ -173,3 +173,25 @@ def ghe_json_message(rec):
...
+        search_key (string)
+        results (array)
+
+    Returns:
+        (list) all values
+    """
+    if not isinstance(data, (dict, list)):
+        return []
+    if search_key in data:
+        results.append(data[search_key])
```
ryandeivert: While the above check (`if search_key in data`) will work if this is a list, this line will fail because search_key is not an number and list only support integer indexing.

Maybe we can talk through what this function is going to achieve? Mainly, my concern falls on why we'd be searching lists for a keys (when lists don't have keys)

helpers/base.py
```diff
@@ -173,3 +173,25 @@ def ghe_json_message(rec):
         return
 
     return message_rec
+
+def select_key(data, search_key, results):
+    """ Recursively search for a given key and return all values
```
ryandeivert: Super nit: leading space before docstring

PR #381, opened 2017-10-12
[bug][data normalization] Improve 'datatypes' checking.

### LINKED ISSUE #365: Bug: Validation of datatypes is too strict
## Background

In current implementation it requires all types in the rule parameter `datatypes` to be defined in `conf/types.json`.
For example, following rule will return `false` even normalized type `command` is defined. It is because the `fetch_values_by_datatype` helper will return an empty list due to code to validate datatypes at [here](https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/rules_engine.py#L282-L284) .
```
"""This is sample rule to alert on any suspicious use of wget"""
import fnmatch

@rule(datatypes=['command', 'not_exist_normalized_type'])
def alert_suspecious_wget(rec):
    for result in results:
        if fnmatch(result, "wget *"):
            return true
    return false
```

## Desired Change

Return value of normalized types if defined, and ignore normalized types which are not defined in `conf/types.json`.

chunyong-lin:
to: @mime-frame or @ryandeivert 
size: small
resolves #365

## Background
Changing rule parameter `datatypes` (it is a list) to an *OR* operation from *AND* operation.
* When it is a *AND* operation, it requires all the datatypes in the list should be defined in `conf/types.json` for all the log sources expected to be analyzed. 
* When it is *OR* operation, it will not has such constraint.
Please refer to issue #365 for more details. 

## Changes
* Remove method `validate_datatypes` from Rule Processor;
* Add checking to skip if CEF datatype is not defined for a log source, but still analyze log if other CEF datatypes in `datatype` list are defined. 
* Clean up the unconecessry code.

## Testing
* Unit Test
  * Test case `test_match_types` catch this change, and updated its assertion.

* Rule Test
  * Add local rules and testing events, and rule test is passed.

### DIFF
--- a/conf/sources.json
+++ b/conf/sources.json
@@ -3,13 +3,15 @@
     "prefix_cluster1_stream_alert_kinesis": {
       "logs": [
         "cloudwatch",
-        "ghe"
+        "ghe",
+        "osquery"
       ]
     }
   },
   "s3": {
     "prefix.cluster.sample.bucket": {
       "logs": [
+        "carbonblack",
         "cloudtrail"
       ]
     }
--- a/conf/sources.json
+++ b/conf/sources.json
@@ -3,15 +3,13 @@
     "prefix_cluster1_stream_alert_kinesis": {
       "logs": [
         "cloudwatch",
-        "ghe",
-        "osquery"
+        "ghe"
       ]
     }
   },
   "s3": {
     "prefix.cluster.sample.bucket": {
       "logs": [
-        "carbonblack",
         "cloudtrail"
       ]
     }
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -209,7 +209,7 @@ class StreamRules(object):
                 cls.update(results, key, nested_results)
             else:
                 for datatype in datatypes:
-                    if normalized_types.get(datatype) and (key in normalized_types[datatype]):
+                    if datatype in normalized_types and key in normalized_types[datatype]:
                         if not datatype in results:
                             results[datatype] = [[key]]
                         else:
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -183,9 +183,8 @@ class StreamRules(object):
                     "defined_type2": [[original_key2, sub_key2], [original_key3]]
                 }
         """
-        results = None
-        if not (datatypes and cls.validate_datatypes(normalized_types, datatypes)):
-            return results
+        if not (datatypes and normalized_types):
+            return
 
         return cls.match_types_helper(record, normalized_types, datatypes)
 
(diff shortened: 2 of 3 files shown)


tests/unit/stream_alert_rule_processor/test_rules_engine.py
```diff
@@ -490,8 +490,11 @@ def match_ipaddress(rec): # pylint: disable=unused-variable
               outputs=['s3:sample_bucket'],
               datatypes=['sourceAddress', 'command'])
         def mismatch_types(rec): # pylint: disable=unused-variable
-            """Testing rule with non-existing normalized type in the record. It
-            should not trigger alert.
+            """Testing rule with non-existing normalized type in the record.
```
mime-frame: This description is really confusing :(

What are you testing?

1) a rule has 2 datatypes defined, but only 1 exists in the record?
2) a rule has 2 datatypes defined, but none exist in the record?
3) a rule has 2 datatypes defined, but both exist in the record?

With the existing description, it's not clear why we'd see a "trigger alert" if there's a record coming into a rule that does not have the specified datatype

stream_alert/rule_processor/rules_engine.py
```diff
@@ -210,7 +209,7 @@ def match_types_helper(cls, record, normalized_types, datatypes):
                 cls.update(results, key, nested_results)
             else:
                 for datatype in datatypes:
-                    if key in normalized_types[datatype]:
+                    if normalized_types.get(datatype) and (key in normalized_types[datatype]):
```
ryandeivert: Please use:
```
if datatype in normalized_types and key in normalized_types[datatype]:...
```

stream_alert/rule_processor/rules_engine.py
```diff
@@ -183,9 +183,8 @@ def match_types(cls, record, normalized_types, datatypes):
                     "defined_type2": [[original_key2, sub_key2], [original_key3]]
                 }
         """
-        results = None
-        if not (datatypes and cls.validate_datatypes(normalized_types, datatypes)):
-            return results
+        if not (datatypes and normalized_types):
```
ryandeivert: Thanks for this change!

conf/sources.json
```diff
@@ -3,13 +3,15 @@
     "prefix_cluster1_stream_alert_kinesis": {
       "logs": [
         "cloudwatch",
-        "ghe"
+        "ghe",
```
jacknagz: why is this file changing?

ryandeivert: she added test rule tests that are osquery logs

ryandeivert: ^^ What's this guy talking about? no she didn't.. :)

chunyong-lin: I will remove this changing 😄

mime-frame:
LGTM

image

PR #386, opened 2017-10-16
[bug][threat intel] IOC matching with string case insensitive.

### LINKED ISSUE #385: Bug: IOC string case may be mismatched
## Background
This bug was discovered during rule testings. It is case-sensitive while comparing record data against IOCs from IOC files at the code [here](https://github.com/airbnb/streamalert/blob/master/helpers/base.py#L134-L135). In the use case that `fileHash` in a record is in upper cases, however, the IOCs from IOC files maybe in lower cases, it will never have a match.

## Desired Change
Implement case-insensitive comparison between IOCs and related record data.

chunyong-lin:
to: @ryandeivert 
size: small
resolves #385 

## Background
This bug was discovered during rule testings. It is case-sensitive while comparing record data against IOCs from IOC files. In the use case that fileHash in a record is in upper cases, however, the IOCs from IOC files maybe in lower cases, it will never have a match. 

This PR implements IOC comparison with case-insensitive and condition that users should have known if IOCs in IOC files are lowercase or uppercase.

## Changes

* Add `lowercase_ioc` flag to helper function `is_ioc`.
* Add unit test cases. 
* Fix another bug that `rec['streamalert:ioc']` should be a list.

## Testing
* Unit testing is passed
```
./tests/scripts/unit_tests.sh
```
* Rule testing
  * Add a local testing rule, and it is passed.

### DIFF
--- a/helpers/base.py
+++ b/helpers/base.py
@@ -118,8 +118,21 @@ def fetch_values_by_datatype(rec, datatype):
 
     return results
 
-def is_ioc(rec):
-    """Detect is any data in a record matching to known IOC"""
+def is_ioc(rec, lowercase_ioc=True):
+    """Detect is any data in a record matching to known IOC
+
+    Args:
+        rec (dict): The parsed payload of any log
+        lowercase_ioc (bool): Indicate if IOCs in IOC files are in lowercase or
+            uppercase. If true, it will convert data found in the record to
+            lowercase.
+            This flag is implemented to achieve case-insensitive comparison
+            between IOCs and related data in the record.
+
+    Returns:
+        (bool): Returns True if data matching to any IOCs, otherwise returns
+            False.
+    """
     intel = StreamThreatIntel.get_intelligence()
     datatypes_ioc_mapping = StreamThreatIntel.get_config()
 
@@ -131,6 +144,8 @@ def is_ioc(rec):
             continue
         results = fetch_values_by_datatype(rec, datatype)
         for result in results:
+            if isinstance(result, str):
+                result = result.lower() if lowercase_ioc else result.upper()
             if (intel.get(datatypes_ioc_mapping[datatype])
                     and result in intel[datatypes_ioc_mapping[datatype]]):
                 if StreamThreatIntel.IOC_KEY in rec:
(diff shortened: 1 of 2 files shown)


helpers/base.py
```diff
@@ -131,6 +144,8 @@ def is_ioc(rec):
             continue
         results = fetch_values_by_datatype(rec, datatype)
         for result in results:
+            if isinstance(result, str):
+                result = result.lower() if lowercase_ioc else result.upper()
```
mime-frame: Why are we converting to uppercase?

mime-frame:
Good catch @chunyong-lin :) Asked 1 Q

helpers/base.py
chunyong-lin: In the use case that user has IOCs are uppercase in the IOC files. `lowercase_ioc` indicates if IOCs in IOC files are lowercase (`lowercase_ioc=True`) or uppercase (`lowercase_ioc=False`).

mime-frame: Thanks :) What are your thoughts on performance? e.g.
(a) I think this means we're going to perform a lower() or upper() for every string?
(b) If the file already is lowercased, will we still incur the cost?
(c) what happens if one IOC type needs to be lowercased but another doesn't?
(d) can you share details on your approach? e.g. why this vs. lowercasing all hashes and domains on load

chunyong-lin: Thanks for bringing up performance concern. I should've wrapped it up. 
> (a) I think this means we're going to perform a lower() or upper() for every string?

It will perform a lower() or upper() for every string returned by Data Normalization helper. It means only the normalized log Data will be performed a lower() or upper() method. 

> (b) If the file already is lowercased, will we still incur the cost?
> (c) what happens if one IOC type needs to be lowercased but another doesn't?
> (d) can you share details on your approach? e.g. why this vs. lowercasing all hashes and domains on load

1. The lower() method is applied to log data which would have the different format (lowercase, uppercase or mixed case) from different log sources. 
  We could add a check to see if a string is lowercased using method `islower()`. But after measuring both approaches, there is no advantage to add the additional check. Thanks to @ryandeivert 

2. IOC file can be easily formatted when we fetch and write them to a file in the script. And we should format IOC file in the script. It is costly doing in Lambda function because the user may have millions custom IOC.

mime-frame:
Thanks for sharing additional details @chunyong-lin :) paging dr. @ryandeivert for review

PR #387, opened 2017-10-17
[parsers] bug fixes for json regex key

jacknagz:
to: @chunyong-lin 
size: small

## Background

When deploying the latest `json_regex_key` feature with the JSON parser, a couple bugs were identified.

## Changes

* Clearer debug statements
* Symmetric difference for missing keys message: reveals additional fields
* Fixes a bug where False was returned when list was expected

## Testing

Locally with unit tests


jacknagz:
@chunyong-lin I'll make the change in the next PR

PR #380, opened 2017-10-12
Nested JSON Support

jacknagz:
to: @ryandeivert 
cc: @airbnb/streamalert-maintainers
size: medium

## Background

When using forwarders such as `fluentd`, `logstash`, or `rsyslog`, log data may be wrapped with additional context keys:

```
{
  "collector": "my-app-1",
  "date-collected": "Oct 12, 2017",
  "@timestamp": "1507845487",
  "data": "<0> program[pid]: {"actual": "data is here"}
}
```

In order to get to the actual payload to apply rules against, helper functions are currently needed.  

This PR adds functionality to the `json` parser to parse JSON string payloads from wrapped keys.

Usage:

Example schema per the data above:

```
  "json:regex_key_with_envelope": {
    "schema": {
      "actual": "string"                    # the schema of the nested data
    },
    "parser": "json",
    "configuration": {
      "envelope_keys": {
        "collector": "string",
        "date-collected": "string",   # optional context to include as envelope keys
        "@timestamp": "string"
      },
      "json_regex_key": "data"        # where the nested JSON exists
    }
  },
```

Optionally, you can omit envelope keys as needed.

## Changes

* Augment the JSON parser to support nested JSON parsing within payload keys
* Update unit tests

## Testing

* Local unit testing
* Deploy in staging account

PR #390, opened 2017-10-17
[parsers] json regex key bug fixes

jacknagz:
to: @ryandeivert 
cc: @airbnb/streamalert-maintainers @chunyong-lin 
size: medium

## Background

Follow up to #380 and #387.  

Deployment of JSON Regex Key matching against a larger set of data introduced more issues with parsing.  The loading of the regex matched string into `json.loads` threw errors since it was improperly caught.

## Changes

* Catch invalid matched JSON
* Add unit test coverage

## Testing

* Locally with unit tests

### DIFF
--- a/tests/unit/stream_alert_rule_processor/test_parsers.py
+++ b/tests/unit/stream_alert_rule_processor/test_parsers.py
@@ -15,12 +15,14 @@ limitations under the License.
 """
 import json
 
+from mock import patch
 from nose.tools import (
     assert_equal,
+    assert_false,
     assert_is_instance,
     assert_items_equal,
     assert_not_equal,
-    assert_false
+    assert_true
 )
 
 from stream_alert.rule_processor.config import load_config
@@ -51,7 +53,7 @@ class TestParser(object):
         """Helper to return the parser result"""
         data = kwargs['data']
         schema = kwargs['schema']
-        options = kwargs['options']
+        options = kwargs.get('options', {})
 
         parser = self.parser_class(options)
         parsed_result = parser.parse(schema, data)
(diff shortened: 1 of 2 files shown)


tests/unit/stream_alert_rule_processor/test_parsers.py
```diff
@@ -90,6 +92,127 @@ class TestJSONParser(TestParser):
...
 
+    def test_schema_fail(self):
+        """JSON Parser - Schema Match Failure"""
+        # setup
+        options = {}
+        schema = self.config['logs']['unit_test_simple_log']['schema']
+        data = json.dumps({'hey': 1})
+
+        # get parsed data
+        parsed_data = self.parser_helper(data=data, schema=schema, options=options)
```
ryandeivert: There's a few areas where `options` is instantiated in the test as an empty dict and just passed along. The `parser_helper` function does a `get` with the keyword args for `options`, so you can safely omit `options=...` from the helper call altogether. Please update here and other places below where options is not used.

tests/unit/stream_alert_rule_processor/test_parsers.py
```diff
@@ -90,6 +92,127 @@ class TestJSONParser(TestParser):
...
+            'host': 'my-host-1'
+        })
+
+        # get parsed data
+        parsed_data = self.parser_helper(data=data, schema=schema, options=options)
+        assert_false(parsed_data)
+
+    @patch('stream_alert.rule_processor.parsers.LOGGER')
+    def test_optional_keys_missing_in_schema(self, mock_logging):
+        """JSON Parser - Optional TLK Not in Schema"""
```
ryandeivert: TLK could be a bit confusing in test output, please use a human readable string here

tests/unit/stream_alert_rule_processor/test_parsers.py
```diff
@@ -90,6 +92,127 @@ class TestJSONParser(TestParser):
...
+        })
+
+        # get parsed data
+        parsed_data = self.parser_helper(data=data, schema=schema, options=options)
+
+        assert_equal(len(parsed_data), 1)
+        assert_true(parsed_data[0]['streamalert:envelope_keys']['host'] == '')
+
+    @patch('stream_alert.rule_processor.parsers.LOGGER')
+    def test_regex_key_invalid_json(self, mock_logging):
```
ryandeivert: ++ for this test. Good thinking

tests/unit/stream_alert_rule_processor/test_parsers.py
jacknagz: `parser_helper` doesn't do a `.get`, but I can make it do one

ryandeivert: sorry you're right it doesn't. you could also just pass `None` but either is fine

tests/unit/stream_alert_rule_processor/test_parsers.py
```diff
@@ -90,6 +92,124 @@ class TestJSONParser(TestParser):
...
+        })
+
+        # get parsed data
+        parsed_data = self.parser_helper(data=data, schema=schema, options=options)
+
+        assert_true(mock_logging.debug.called)
+        assert_false(parsed_data)
+
+    def test_regex_key_non_json(self):
+        """JSON Parser - Regex Key with Invalid JSON Object"""
```
ryandeivert: I just noticed this test has the same description as the one above it. Can you update one of them please?

PR #391, opened 2017-10-18
[cli][docs] adding doc updates to reflect the --processor cli arg changes

ryandeivert:
to: @chunyong-lin 
size: small
resolves N/A

## Background

Previous changes to the CLI updated the `manage.py lambda deploy` and `rollback` commands to use a list of functions passed to the `--processor` arg. This PR updates the `test` command and also updates docs to go with previous changes.

## Changes

* Updating the `manage.py lambda test` command to support multiple args.
* Updating docs to depict the usage of CLI `--processor` flag which now accepts a list of functions to test/deploy/rollback.
* The previous method of using multiple `--processor` flags is no longer supported and will result in only the *last* `--processor` flag having any affect.

## Testing

Ran tests with various options to ensure the proper behavior occurs.

### DIFF
--- a/stream_alert_cli/test.py
+++ b/stream_alert_cli/test.py
@@ -752,13 +752,13 @@ def stream_alert_test(options, config):
         log_mem_hanlder = get_log_memory_hanlder()
 
         # Check if the rule processor should be run for these tests
-        test_rules = (set(run_options.get('processor')).issubset({'rule', 'all'})
+        test_rules = (not run_options.get('processor').isdisjoint({'rule', 'all'})
                       if run_options.get('processor') else
                       run_options.get('command') == 'live-test' or
                       run_options.get('command') == 'validate-schemas')
 
         # Check if the alert processor should be run for these tests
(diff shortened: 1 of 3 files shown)


stream_alert_cli/test.py
```diff
@@ -752,13 +752,13 @@ def run_tests(options, context):
         log_mem_hanlder = get_log_memory_hanlder()
 
         # Check if the rule processor should be run for these tests
-        test_rules = (set(run_options.get('processor')).issubset({'rule', 'all'})
+        test_rules = (not run_options.get('processor').isdisjoint({'rule', 'all'})
```
chunyong-lin: Can you provide more content about why do you choose to use `isdisjoint` over `issubset`?

ryandeivert: @chunyong-lin Do you want a code comment or just some context in the PR body?

chunyong-lin: Code comment is good for me 😄  Thanks!

ryandeivert:
hi @chunyong-lin I pushed a commit with a code comment

PR #394, opened 2017-10-18
Optional Envelope Keys

jacknagz:
to: @ryandeivert 
size: smalll

## Background

https://github.com/airbnb/streamalert/pull/387 introduced a performance hit when applying default values to envelope keys implicitly.  This change adds an explicit option to the `JSON` parser to prevent this from occuring

## Changes

* Adds a parser option for optional_envelope_keys, which is needed when using json_regex_key
* Also adds `yapf`, updates AUTHORs, and docs

## Testing

Unit tested

### DIFF
--- a/stream_alert/rule_processor/parsers.py
+++ b/stream_alert/rule_processor/parsers.py
@@ -219,15 +219,18 @@ class JSONParser(ParserBase):
             return [json_payload]
 
         envelope_schema = self.options.get('envelope_keys')
-        # If envelope_keys are declared, and this payload does not have every key specified
-        # in these envelope_keys, then it's safe to skip trying to extract records using json_path
-        if envelope_schema:
+        optional_envelope_keys = self.options.get('optional_envelope_keys')
+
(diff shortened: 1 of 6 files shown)


stream_alert/rule_processor/parsers.py
```diff
@@ -219,16 +219,19 @@ def _parse_records(self, schema, json_payload):
             return [json_payload]
 
         envelope_schema = self.options.get('envelope_keys')
-        # If envelope_keys are declared, and this payload does not have every key specified
-        # in these envelope_keys, then it's safe to skip trying to extract records using json_path
-        if envelope_schema:
+        optional_envelope_keys = self.options.get('optional_envelope_keys')
+
+        if envelope_schema and optional_envelope_keys:
```
ryandeivert: Small nit - can you just add a code comment explaining the two different behaviors?

ryandeivert:
Thanks for adding the yapf stuff! Also for putting my in authors 🤣

PR #400, opened 2017-10-19
[cli][test] fix for rule tests that should not trigger alerts

### LINKED ISSUE #397: Bug: Integration Test records always evaluate as `True`
to: @ryandeivert 

## Background

Integration tests mark test records as passed, even when the record contains a value that should evaluate as false.  This occurs when a test record contains an empty entry in the field: `trigger_rules:[]`

For instance.  Take the following simple example
```
def foobar(rec):
    if rec['name'] == 'foobar':
        return True
```

This record evaluates as TRUE because the NAME field value (foobar) satisfies the rule:
```
{
  "records": [
    {
      "data": {
        "action": ...",
        "calendarTime": "...",
        "decorations": {
          "envIdentifier": "..."
        },
        "epoch": "...",
        "hostIdentifier": "...",
        "log_type": "...",
        "name": "foobar",
        "snapshot": [
          {
            "name": "...",
            "label": "...",
            "path": "..."
          }
        ],
        "unixTime": "..."
      },
      "description": "If name matches foobar, then an alert should not be created.",
      "log": "...",
      "service": "..",
      "source": "...",
      "trigger_rules": []
    }
```

This record also evaluates as TRUE, even though the NAME field value (security) does not satisfy the rule:
```
{
  "records": [
    {
      "data": {
        "action": ...",
        "calendarTime": "...",
        "decorations": {
          "envIdentifier": "..."
        },
        "epoch": "...",
        "hostIdentifier": "...",
        "log_type": "...",
        "name": "security",
        "snapshot": [
          {
            "name": "...",
            "label": "...",
            "path": "..."
          }
        ],
        "unixTime": "..."
      },
      "description": "If name matches foobar, then an alert should not be created.",
      "log": "...",
      "service": "..",
      "source": "...",
      "trigger_rules": []
    }
```

## Desired Change

All Integration tests should validate if the records contains the appropriate values to satisfy every condition of the rule.

ryandeivert:
to: @jacknagz or @chunyong-lin 
cc: @airbnb/streamalert-maintainers, @fusionrace 
size: small
resolves #397 

## Background

See #397 

## Changes

* Fixing bug for test events that should not trigger rules were not flagging tests as failures.
  * This was due to not checking if the `trigger_rules` list had any rules defined before filtering alerts based on those rules.
* Adding more context to the output to explain various failures that can occur
* This will likely cause some other test events to fail, since it's more strict about defining rules that should trigger.

## Testing

* Updated some test events that were triggering more rules than defined in the event's `trigger_rules` list to avoid multiple alerts per test event.
  * In the future, test events that can trigger more than one rule can be placed in a single test file and have multiple rules listed within them (and the test file can be named with something more generic than the actual rule name).

### DIFF
--- a/stream_alert_cli/test.py
+++ b/stream_alert_cli/test.py
@@ -244,17 +244,17 @@ class RuleProcessorTester(object):
         alerts, all_records_matched_schema = self.test_rule(event)
 
         # Get a list of any rules that triggerer but are not defined in the 'trigger_rules'
-        bad_alerts = []
+        unexpected_alerts = []
 
         # we only want alerts for the specific rule being tested (if trigger_rules are defined)
         if test_event['trigger_rules']:
-            bad_alerts = [alert for alert in alerts
-                          if alert['rule_name'] not in test_event['trigger_rules']]
+            unexpected_alerts = [alert for alert in alerts
+                                 if alert['rule_name'] not in test_event['trigger_rules']]
 
             alerts = [alert for alert in alerts
                       if alert['rule_name'] in test_event['trigger_rules']]
 
-        alerted_properly = (len(alerts) == expected_alert_count) and not bad_alerts
+        alerted_properly = (len(alerts) == expected_alert_count) and not unexpected_alerts
         current_test_passed = alerted_properly and all_records_matched_schema
 
         self.all_tests_passed = current_test_passed and self.all_tests_passed
@@ -286,11 +286,12 @@ class RuleProcessorTester(object):
                 context = 'is triggering the following rules but should not trigger at all: {}'
                 trigger_rules = ' ,'.join('\'{}\''.format(alert['rule_name']) for alert in alerts)
                 message = '{} {}'.format(message, context.format(trigger_rules))
-            elif bad_alerts:
+            elif unexpected_alerts:
                 # If there was a failure due to alerts triggering for other rules outside
                 # of the rules defined in the trigger_rules list for the event
                 context = 'is triggering the following rules but should not be: {}'
-                bad_rules = ' ,'.join('\'{}\''.format(alert['rule_name']) for alert in bad_alerts)
+                bad_rules = ' ,'.join(
+                    '\'{}\''.format(alert['rule_name']) for alert in unexpected_alerts)
                 message = '{} {}'.format(message, context.format(bad_rules))
             elif expected_alert_count != len(alerts):
                 # If there was a failure due to alerts NOT triggering for 1+ rules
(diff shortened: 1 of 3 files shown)


stream_alert_cli/test.py
```diff
@@ -270,7 +278,31 @@ def _run_rule_tests(self, file_name, test_event, formatted_record, print_header_
...
+            if alerts and not test_event['trigger_rules']:
+                # If there was a failure due to alerts triggering for a test event
+                # that does not have any trigger_rules configured
+                context = 'is triggering the following rules but should not be: {}'
+                trigger_rules = ' ,'.join('\'{}\''.format(alert['rule_name']) for alert in alerts)
+                message = '{} {}'.format(message, context.format(trigger_rules))
+            elif bad_alerts:
+                # If there was a failure due to alerts triggering for other rules outside
+                # of the rules defined in the trigger_rules list for the event
+                context = 'is triggering the following rules but should not be: {}'
```
chunyong-lin: Would prefer to differentiate the failure here is different from above by having a different context.

ryandeivert: Just squashed - this should be different than the other one now.

stream_alert_cli/test.py
```diff
@@ -243,10 +243,18 @@ def _run_rule_tests(self, file_name, test_event, formatted_record, print_header_
         # Run tests on the formatted record
         alerts, all_records_matched_schema = self.test_rule(event)
 
-        # we only want alerts for the specific rule being tested
-        alerts = [alert for alert in alerts if alert['rule_name'] in test_event['trigger_rules']]
+        # Get a list of any rules that triggerer but are not defined in the 'trigger_rules'
+        bad_alerts = []
```
jacknagz: `bad` a slight misnomer, can  you use a different descriptor?

stream_alert_cli/test.py
```diff
@@ -243,10 +243,18 @@ def _run_rule_tests(self, file_name, test_event, formatted_record, print_header_
...
-        alerted_properly = (len(alerts) == expected_alert_count)
+        # we only want alerts for the specific rule being tested (if trigger_rules are defined)
+        if test_event['trigger_rules']:
+            bad_alerts = [alert for alert in alerts
+                          if alert['rule_name'] not in test_event['trigger_rules']]
+
+            alerts = [alert for alert in alerts
+                      if alert['rule_name'] in test_event['trigger_rules']]
+
+        alerted_properly = (len(alerts) == expected_alert_count) and not bad_alerts
```
jacknagz: `unexpected_alerts = `?

ryandeivert: updated

PR #401, opened 2017-10-19
[testing] Remove dependency on filesystem for test_config unit tests

### LINKED ISSUE #395: test_config unit tests fail for different cluster config files
## Background

Three unit tests in `test_config.py` fail when the cluster config file `prod.json` doesn't exist.

In my setup, the production cluster is named `production.json`. The contents of the cluster config are supposed to be mocked in `test_config.py`, but `CLIConfig()` will list the actual contents of the `conf/clusters` directory to retrieve the cluster filenames.

Subsequently `CLIConfig` attempts to open `conf/clusters/production.json` and doesn't get the mocked content (or mocked `prod` cluster name) leading to KeyErrors.

```
======================================================================
ERROR: CLI - Adding CloudWatch metric alarm, cluster
  File "/Users/jeremy/.local/share/virtualenvs/streamalert-iKoON-og/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/jeremy/.local/share/virtualenvs/streamalert-iKoON-og/lib/python2.7/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/Users/jeremy/projects/streamalert/tests/unit/stream_alert_cli/test_config.py", line 98, in test_cluster_alarm_creation
    config.add_metric_alarm(alarm_info)
  File "/Users/jeremy/projects/streamalert/stream_alert_cli/config.py", line 366, in add_metric_alarm
    self._add_metric_alarm_per_cluster(alarm_info, metric_function)
  File "/Users/jeremy/projects/streamalert/stream_alert_cli/config.py", line 205, in _add_metric_alarm_per_cluster
    function_config = (self.config['clusters'][cluster]['modules']
KeyError: 'prod'

======================================================================
ERROR: CLI - Metric toggling
  File "/Users/jeremy/.local/share/virtualenvs/streamalert-iKoON-og/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/jeremy/.local/share/virtualenvs/streamalert-iKoON-og/lib/python2.7/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/Users/jeremy/projects/streamalert/tests/unit/stream_alert_cli/test_config.py", line 62, in test_toggle_metric
    config.toggle_metrics(True, ['prod'], ['alert_processor'])
  File "/Users/jeremy/projects/streamalert/stream_alert_cli/config.py", line 161, in toggle_metrics
    [function]['enable_metrics'] = enabled
KeyError: 'prod'

======================================================================
FAIL: CLI - Aggregate alarm check
  File "/Users/jeremy/.local/share/virtualenvs/streamalert-iKoON-og/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/jeremy/projects/streamalert/tests/unit/stream_alert_cli/test_config.py", line 77, in test_cluster_alarm_exists
    assert_true(result)
AssertionError: False is not true

Ran 353 tests in 68.910s

FAILED (errors=2, failures=1)
```

## Desired Change

Create a mock filesystem using pyfakefs instead of only mocking out `__builtin__.open()`, so that `os.path.listdir()` will return `conf/clusters/prod.json`.

I'm happy to submit a PR for this.

stoggi:
size: small
resolves #395

## Background

CLIConfig uses the filesystem to find the filenames of cluster configuration files. In the config unit tests, these filesystem operations need to be mocked out because cluster configuration files are user configured. Currently `__builtin__.open()` is mocked out, but we also need to mock out `os.path.listdir()`. Instead of mocking individual functions, __pyfakefs__ is used to create a mock filesystem.

## Changes

* Added new python dependency pyfakefs to create a mock filesystem
* Update tests in test_config.py to use the mocked filesystem instead of mock_open

## Testing

* Remove `prod.json` cluster config and added `abc.json` cluster config
* Step through CLIConfig.load() to make sure os.listdir() returns `prod.json`
* Run `nosetests -v tests/unit`
* Run `./tests/scripts/pylint.sh`

### DIFF
--- a/tests/unit/stream_alert_cli/test_config.py
+++ b/tests/unit/stream_alert_cli/test_config.py
@@ -14,67 +14,73 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 # pylint: disable=protected-access
-from contextlib import nested
 import json
 
 from mock import Mock, patch
 from nose.tools import assert_equal, assert_true
+from pyfakefs import fake_filesystem_unittest
 
 from stream_alert_cli.config import CLIConfig
-from tests.unit.helpers.base import basic_streamalert_config, mock_open
+from tests.unit.helpers.base import basic_streamalert_config
 
 
 class TestCLIConfig(object):
     """Test class for CLIConfig"""
+    # pylint: disable=no-self-use
 
     def __init__(self):
-        self.mocked_opens = None
+        self.fs_patcher = None
 
     def setup(self):
         """Setup before each method"""
         config_data = basic_streamalert_config()
-        self.mocked_opens = [
-            mock_open('conf/global.json', json.dumps(config_data['global'])),
-            mock_open('conf/lambda.json', json.dumps(config_data['lambda'])),
-            mock_open('conf/clusters/prod.json', json.dumps(config_data['clusters']['prod']))
-        ]
+
+        self.fs_patcher = fake_filesystem_unittest.Patcher()
+        self.fs_patcher.setUp()
+
+        self.fs_patcher.fs.CreateFile('/conf/global.json',
+                                      contents=json.dumps(config_data['global']))
+        self.fs_patcher.fs.CreateFile('/conf/lambda.json',
+                                      contents=json.dumps(config_data['lambda']))
+        self.fs_patcher.fs.CreateFile('/conf/clusters/prod.json',
+                                      contents=json.dumps(config_data['clusters']['prod']))
+
+    def teardown(self):
+        """Teardown after each method"""
+        self.fs_patcher.tearDown()
 
     def test_load_config(self):
         """CLI - Load config"""
-        with nested(*self.mocked_opens):
-            config = CLIConfig()
-            assert_equal(config['global']['account']['prefix'], 'unit-testing')
+        config = CLIConfig()
+        assert_equal(config['global']['account']['prefix'], 'unit-testing')
 
     @patch('logging.Logger.error')
     @patch('stream_alert_cli.config.CLIConfig.write')
     def test_toggle_metric(self, write_mock, log_mock):
         """CLI - Metric toggling"""
-        with nested(*self.mocked_opens):
-            config = CLIConfig()
+        config = CLIConfig()
 
-            config.toggle_metrics(True, [], ['athena_partition_refresh'])
-            write_mock.assert_called()
+        config.toggle_metrics(True, [], ['athena_partition_refresh'])
+        write_mock.assert_called()
 
-            del config.config['lambda']['athena_partition_refresh_config']
-            config.toggle_metrics(True, [], ['athena_partition_refresh'])
-            log_mock.assert_called_with('No Athena configuration found; please initialize first.')
+        del config.config['lambda']['athena_partition_refresh_config']
+        config.toggle_metrics(True, [], ['athena_partition_refresh'])
+        log_mock.assert_called_with('No Athena configuration found; please initialize first.')
 
-            config.toggle_metrics(True, ['prod'], ['alert_processor'])
-            write_mock.assert_called()
+        config.toggle_metrics(True, ['prod'], ['alert_processor'])
+        write_mock.assert_called()
 
     def test_aggregate_alarm_exists(self):
         """CLI - Aggregate alarm check"""
-        with nested(*self.mocked_opens):
-            config = CLIConfig()
-            result = config._alarm_exists('Aggregate Unit Testing Failed Parses Alarm')
-            assert_true(result)
+        config = CLIConfig()
+        result = config._alarm_exists('Aggregate Unit Testing Failed Parses Alarm')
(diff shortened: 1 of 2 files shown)


tests/unit/stream_alert_cli/test_config.py
```diff
@@ -14,67 +14,73 @@
...
+from pyfakefs import fake_filesystem_unittest
 
 from stream_alert_cli.config import CLIConfig
-from tests.unit.helpers.base import basic_streamalert_config, mock_open
+from tests.unit.helpers.base import basic_streamalert_config
 
 
 class TestCLIConfig(object):
     """Test class for CLIConfig"""
+    # pylint: disable=no-self-use
```
ryandeivert: We typically follow the format of one `Test...` class per file, so please add this pylint message to the top of the file alongside the `pylint: disable=protected-access` one that is currently there

tests/unit/stream_alert_cli/test_config.py
```diff
@@ -14,67 +14,73 @@
...
+        config = CLIConfig()
+        assert_equal(config['global']['account']['prefix'], 'unit-testing')
 
     @patch('logging.Logger.error')
     @patch('stream_alert_cli.config.CLIConfig.write')
     def test_toggle_metric(self, write_mock, log_mock):
         """CLI - Metric toggling"""
-        with nested(*self.mocked_opens):
-            config = CLIConfig()
+        config = CLIConfig()
```
ryandeivert: Since this is used in most (all?) test methods, can you add this as a instance property in the `__inti__` and `setup` to be used throughout the class tests?

tests/unit/stream_alert_cli/test_config.py
```diff
@@ -14,67 +14,73 @@
...
     def setup(self):
         """Setup before each method"""
         config_data = basic_streamalert_config()
-        self.mocked_opens = [
-            mock_open('conf/global.json', json.dumps(config_data['global'])),
-            mock_open('conf/lambda.json', json.dumps(config_data['lambda'])),
-            mock_open('conf/clusters/prod.json', json.dumps(config_data['clusters']['prod']))
-        ]
+
+        self.fs_patcher = fake_filesystem_unittest.Patcher()
```
ryandeivert: Question: Could `fake_filesystem_unittest.TestCase` be subclassed here to get all of this functionality for free?

stoggi: You can, but it will convert the test case into a `unittest` test case, as `fake_filesystem_unittest.TestCase` inherits from `unittest.TestClass` (https://github.com/jmcgeheeiv/pyfakefs/wiki/Automatically-find-and-patch-file-functions-and-modules#unit-test-class).

Which `nose` can run, but we will have to change the setup method from `setup(self)` to `setUp(self)`. It may end up being more confusing.

tests/unit/stream_alert_cli/test_config.py
stoggi: Hm, good point, but in this case CLIConfig() is the unit under test. Would it make more sense to leave it in each test case?

ryandeivert: That's fine - the `setup` method gets newly called with each method that runs anyway (effectively doing the same thing you're doing here).

tests/unit/stream_alert_cli/test_config.py
ryandeivert: Good point - let's leave it!

tests/unit/stream_alert_cli/test_config.py
```diff
@@ -14,6 +14,7 @@
 limitations under the License.
 """
 # pylint: disable=protected-access
+# pylint: disable=no-self-use
```
ryandeivert: sorry one last nit - these can be on one line separated by commas :)

`pylint: disable=protected-access,no-self-use`

ryandeivert:
Thanks for the cleanup! One tiny nit and then it'll be good to merge :)

tests/unit/stream_alert_cli/test_config.py
```diff
@@ -14,67 +14,73 @@
 limitations under the License.
 """
 # pylint: disable=protected-access
-from contextlib import nested
+# pylint: disable=no-self-use
```
ryandeivert: sorry one last nit - these can be on one line separated by commas :)

`pylint: disable=protected-access,no-self-use`

stoggi: Ah, yep, sure thing. didn't realise you can do this.

stoggi: Actually, since I'm now using self in the methods I can just remove it.

mime-frame:
image

PR #402, opened 2017-10-19
[cli][test] reporting misconfigured source in test event

ryandeivert:
to: @jacknagz 
size: small
resolves N/A

## Background

Test events that do not have their source properly defined (or, more concisely, if the log type is missing from the `source.json`) would silently fail test.

## Changes

* Checking if the source is properly configured before trying to perform schema validation.

## Testing

* Changed the local `sources.json` to omit a log type and verified the following output occurs (along with failing tests):

`
StreamAlertCLI [ERROR]: (2/2) Failures
`
`
StreamAlertCLI [ERROR]: (1/2) Invalid test event in file '<filename>.json' with description 'Description would be here'. The 'sources.json' file does not include the log type 'cloudtrail' in the list of logs for this service & entitiy ('kinesis:<entity-here>').
`

### DIFF
--- a/stream_alert_cli/test.py
+++ b/stream_alert_cli/test.py
@@ -274,9 +274,7 @@ class RuleProcessorTester(object):
 
         # Add the status of the rule to messages list
         if not all_records_matched_schema:
-            message = self.analyze_record_delta(file_name, test_event)
-            if message:
-                self.status_messages.append(StatusMessage(StatusMessage.FAILURE, message))
+            self.analyze_record_delta(file_name, test_event)
         elif not alerted_properly:
             message = ('Test failure: [{}.json] Test event with description '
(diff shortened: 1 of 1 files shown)


stream_alert_cli/test.py
```diff
@@ -458,6 +456,29 @@ def test_rule(self, record):
 
         return alerts, all_records_matched_schema
 
+    def check_log_in_source(self, base_message, test_event):
```
jacknagz: how about `check_log_declared_in_sources`

stream_alert_cli/test.py
```diff
@@ -458,6 +456,29 @@ def test_rule(self, record):
...
+        """Do a simply check to see if this log type is defined in the sources for the service
```
jacknagz: `A simple check...`

PR #404, opened 2017-10-20
[rules] Detect right-to-left character in commands or files

javuto:
to: @mime-frame @ryandeivert 
size: small

## Background

Malicious files can be disguised by using a encoding trick that uses the unicode character U+202E, also known as right-to-left-override (RLO). The trick hides a potentially malicious extension and makes them appear harmless.

[https://krebsonsecurity.com/2011/09/right-to-left-override-aids-email-attacks](https://krebsonsecurity.com/2011/09/right-to-left-override-aids-email-attacks)

## Changes

* Adding `osquery` logs to kinesis placeholder.
* Adding `carbonblack` logs to s3 placeholder.
* Adding rule `right_to_left_character` using data normalization and tests.

## Testing

Verification of schemas:
```
$ python manage.py validate-schemas
``` 
Testing of rule:
```
python manage.py lambda test --processor rule --test-file right_to_left_character
```

### DIFF
--- a/conf/sources.json
+++ b/conf/sources.json
@@ -3,14 +3,16 @@
     "prefix_cluster1_stream_alert_kinesis": {
       "logs": [
         "cloudwatch",
-        "ghe"
+        "ghe",
+        "osquery"
       ]
     }
   },
   "s3": {
     "prefix.cluster.sample.bucket": {
       "logs": [
-        "cloudtrail"
+        "cloudtrail",
+        "carbonblack"
       ]
     }
   },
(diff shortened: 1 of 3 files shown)


conf/sources.json
```diff
@@ -3,14 +3,16 @@
...
+        "ghe",
+        "osquery"
       ]
     }
   },
   "s3": {
     "prefix.cluster.sample.bucket": {
       "logs": [
-        "cloudtrail"
+        "cloudtrail",
```
ryandeivert: I think cloudtrail would more appropriately go under `kinesis` above. Can you update here and within your test events?

PR #409, opened 2017-10-20
Athena Logger, JSON Regex Key Bug Fixes, and More

jacknagz:
to: @chunyong-lin 
size: small

## Background

Various bugs identified from development.

## Changes

See commits

## Testing

Deployed in testing AWS account

### DIFF
--- a/stream_alert/athena_partition_refresh/main.py
+++ b/stream_alert/athena_partition_refresh/main.py
@@ -257,11 +257,9 @@ class StreamAlertAthenaClient(object):
         if query_success and query_resp['ResultSet']['Rows']:
             return True
 
-        LOGGER.info('The streamalert table \'%s\' does not exist. '
-                    'For alert buckets, create it with the following command: \n'
-                    '$ python manage.py athena create-table '
-                    '--type alerts --bucket s3.bucket.id',
-                    table_name)
+        LOGGER.info('The streamalert table \'%s\' does not exist.', table_name)
+        LOGGER.info('For help with creating tables: '
+                    '$ python manage.py athena create-table --help')
         return False
 
     def repair_hive_table(self, unique_buckets):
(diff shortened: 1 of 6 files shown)


stream_alert/athena_partition_refresh/main.py
```diff
@@ -257,11 +257,9 @@ def check_table_exists(self, table_name):
         if query_success and query_resp['ResultSet']['Rows']:
             return True
 
-        LOGGER.info('The streamalert table \'%s\' does not exist. '
-                    'For alert buckets, create it with the following command: \n'
-                    '$ python manage.py athena create-table '
-                    '--type alerts --bucket s3.bucket.id',
-                    table_name)
+        LOGGER.info('The streamalert table \'%s\' does not exist.', table_name)
+        LOGGER.info('Create it with the following command: '
```
chunyong-lin: The following command is to show `create-table` helper. The statement is misleading here.

stream_alert_cli/runner.py
```diff
@@ -150,6 +150,10 @@ def add_to_athena_schema(schema, root_key=''):
                     athena_schema[root_key] = {}
 
                 for key_name, key_type in schema.iteritems():
+                    # When using special characters in the beginning or end
+                    # of a column name, they have to be wrapped in backticks
+                    key_name = '`{}`'.format(key_name)
```
chunyong-lin: I am concerned to use backticks here. The `key_name` is getting from the schema which is defined in the log file. I don't think it is safe to use backticks here.

jacknagz: @chunyong-lin see the comment above the line which explains why

stream_alert/athena_partition_refresh/main.py
jacknagz: yep! I did that on purpose, so users could figure out the settings they need

ryandeivert: Since the command you point to doesn't actually _create_ the table, like the log statement indicates, maybe change the statement to something like `'For help with creating tables, use the following command...'`

PR #411, opened 2017-10-23
[apps] app successive invocations when nearing timeout

ryandeivert:
to: @jacknagz 
size: medium
resolves N/A

## Background

The apps function has built in logic to detect if the execution is close to timing out, and will save state/exit cleanly if that is the case. However, if there were more logs to collect for the interval in which this lambda executed, they would not be collected until the next scheduled invocation occurs.
 For instance, if a functions executes every hour, but only has enough time to process 1/2 hour worth of log data before timing out, that remaining 1/2 hour of log data would not be collected until the next scheduled invocation of the function. This is typically not an issue, and the function will catch up eventually. However, if there is a considerable change in log throughput and (ie - a company's scale doubles), this could result in an undesired effect of always being behind in log processing.

## Changes

* Adding support for perpetually invoking an apps function if there are more logs to collect and the function is nearing time out. It will now invoke itself with a custom `event` object indicating there are more logs to be collected.
* The new execution will be started, and the current execution will not be marked as complete. This is to allow it to continue collecting logs without having other scheduled lambda invocations occurring at the same time, which would potentially cause duplicative data or collisions with 
reading/writing the config.
  * This makes collecting a lot of logs easier to handle and will allow an apps function to completely collect logs for an interval, instead of relying on a subsequent scheduled invocation to collect the leftover logs.
* The `AppConfig` object now contains the passed in `event` dict to be read from during execution. The event can only have one key right now, and the is `invocation_type`.
* Bumping the buffer multiplier to account for the additional time needed to invoke a lambda function at the end of execution.
*  Adding helper function to config for reporting the remaining seconds in the lambda execution.

## Testing

Updating unit test to verify a new invocation occurs during the `finalize` step if there are still more logs to poll.

### DIFF
--- a/app_integrations/apps/app_base.py
+++ b/app_integrations/apps/app_base.py
@@ -68,6 +68,12 @@ def get_app(config, init=True):
 class AppIntegration(object):
     """Base class for all app integrations to be implemented for various services"""
     __metaclass__ = ABCMeta
+    # This _POLL_BUFFER_MULTIPLIER is a multiplier that will be used, along with the time it
+    # took to perform an API request and forward logs, to determine if there is enough
+    # time remaining in the execution of this function to perform another request.
+    # The buffer is also to account for any finalization that must occur, like config
+    # saving to parameter store and spawning a new Lambda invocation if there are more
+    # logs to poll for this interval
     _POLL_BUFFER_MULTIPLIER = 1.5
 
     def __init__(self, config):
@@ -170,7 +176,7 @@ class AppIntegration(object):
     def _initialize(self):
         """Method for performing any startup steps, like setting state to running"""
         # Perform another safety check to make sure this is not being invoked already
-        # If this is a successive invocation to a previous execution, allow it to continue
+        # If this is an invocation spawned from a previous execution, allow it to continue
         if not self._config.is_successive_invocation and self._config.is_running:
             LOGGER.error('App already running for service \'%s\'.', self.type())
             return False
@@ -219,7 +225,7 @@ class AppIntegration(object):
         self._config.mark_success()
 
     def _invoke_successive_gather(self):
-        """Method for invoking a successive app function to handle more logs
+        """Invoke a successive app function to handle more logs
 
         This is useful when there were more logs to collect than could be accomplished
         in this execution. Instead of marking the config with 'success' and waiting
@@ -357,7 +363,8 @@ class AppIntegration(object):
 
         LOGGER.info('Gather process for \'%s\' executed in %f seconds.', self.type(), exec_time)
 
-        # Add a 50% buffer to the time it too to account for some unforeseen delay
+        # Add a 50% buffer to the time it took to account for some unforeseen delay and to give
+        # this function enough time to spawn a new invocation if there are more logs to poll
         # Cast this back to float so general arithemtic works
         return float(exec_time * Decimal(self._POLL_BUFFER_MULTIPLIER))
 
--- a/app_integrations/apps/app_base.py
+++ b/app_integrations/apps/app_base.py
@@ -15,9 +15,12 @@ limitations under the License.
 """
 from abc import ABCMeta, abstractmethod, abstractproperty
 from decimal import Decimal
+import json
 import time
 from timeit import timeit
 
+import boto3
+from botocore.exceptions import ClientError
 import requests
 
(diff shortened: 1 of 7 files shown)


app_integrations/apps/app_base.py
```diff
@@ -65,7 +68,7 @@ def get_app(config, init=True):
 class AppIntegration(object):
     """Base class for all app integrations to be implemented for various services"""
     __metaclass__ = ABCMeta
-    _POLL_BUFFER_MULTIPLIER = 1.2
+    _POLL_BUFFER_MULTIPLIER = 1.5
```
jacknagz: I'd add a comment explaining what this is for

app_integrations/apps/app_base.py
```diff
@@ -167,12 +170,16 @@ def _sleep(self):
     def _initialize(self):
         """Method for performing any startup steps, like setting state to running"""
         # Perform another safety check to make sure this is not being invoked already
-        if self._config.is_running:
+        # If this is a successive invocation to a previous execution, allow it to continue
```
jacknagz: This comment is slightly confusing, should it just be `This prevents multiple Lambda functions from running simultaneously`?

app_integrations/apps/app_base.py
```diff
@@ -184,21 +191,64 @@ def _initialize(self):
...
+        # mark the config as 'success' (keeping the state as 'running' prevents
+        # scheduled function invocations from running alongside this invocation).
+        if self._more_to_poll:
+            self._invoke_successive_gather()
+            return
+
+        self._config.mark_success()
+
+    def _invoke_successive_gather(self):
+        """Method for invoking a successive app function to handle more logs
```
jacknagz: I don't think you need to mention it's a method, that's a little too `meta`

app_integrations/config.py
```diff
@@ -47,9 +47,14 @@ class States(object):
...
 
-    def __init__(self, config, **kwargs):
+    class Events(object):
+        """Events object to encapsulate various acceptable events"""
+        SUCCESSIVE_INVOKE = 'successive_invoke'
+
+    def __init__(self, config, event=None, **kwargs):
         dict.__init__(self, config, **kwargs)
         self._validate_config()
+        self._event = event or {}
```
jacknagz: why not just make the default `{}`?

ryandeivert: Good thought, but: 
image

app_integrations/apps/app_base.py
ryandeivert: Weird nit but I'll change it

app_integrations/config.py
jacknagz: why is that dangerous?

ryandeivert: See here:
http://pylint-messages.wikidot.com/messages:w0102

> To a Python novice, this may seem like a reasonable way to default to an empty array for the extras param. What really happens is that this "default" array gets created as a persistent object, and every invocation of my_method that doesn't specify an extras param will be using that same list object—any changes to it will persist and be carried to every other invocation!

jacknagz: oh right, I forgot Python did that

app_integrations/apps/app_base.py
ryandeivert: This is explained when it's accessed here (where the context matters more than when it's defined): https://github.com/airbnb/streamalert/blob/3cf8b22dbc6b42a0b10db96a6a66d346c98995dc/app_integrations/apps/app_base.py#L310-L312

I  can add an extra comment here to be doubly sure it's seen

ryandeivert:
@jacknagz what do you mean by re-use of `successive invocation` exactly? Re-use from where 🤔

ryandeivert:
also @jacknagz implementing a step function is a much larger refactor that we could consider more in the future, but is honestly not super likely to happen. We (Austin and I) discussed using step functions for this when adding the initial feature and there was no huge benefit to it.

Since this operates within a single function, it doesn't get us additional error handling we don't already get and would be most useful if we had multiple lambdas invoking in chain here. The biggest benefit would be the ability raise custom exceptions and fork what we do based off of that. But even then, in this particular case we'd have to do a hacky thing to raise an exception to indicate there are more logs to poll, which would then cause the lambda to get invoked again. Let me know if you'd like to discuss more offline.

PR #416, opened 2017-10-24
Spiper enrichments

### LINKED ISSUE #305: Feature: Custom Alert text based on configuration and fields
When alerts are written to Slack, they provide a title, a description of the alert (as extracted from the rule's source code comment), and the event that caused the alert.  I would like the ability to format the text of the alert.  For example, for the sample alerts, there is one for cloudtrail_critical_apis.  I would like to high-light in the alert text, who did it (.userIdentity.arn) and what they did (.eventName) to more easily understand what happened instead of expanding the json and looking through it for the information that most interests me.  Is this possible?

0xdabbad00:
size: small
resolves #305 

## Background

Adds "enrichments" which are similar to "matchers", but instead of working on rules, they work on alerts.  This has only been added to Slack outputs currently.  This allows you to:
- Modify the alert header and pretext of the alert, so for example, you don't have to have the rule's docstring in the alert that fires.
- Enhance the alert description to do things like include the AWS account ID in the description and give a human name for that account (ex. "Prod" instead of "111111111111").
- Drop the alert.  For example, if you set an output to two slack channels ("security" and "prod") and you want all alerts to go to "security", but additionally you want alerts relevant to your prod account to go to your prod channel as well.

The main concept of what this does exhibits itself in `enrichments/sample.py`

## Testing

I tested this on our local repo and then copied the pages to this fork of upstream.  I'm not sure what all testing to do or how to add unit tests for this.


enrichments/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Raise this to stop an alert from firing from an enrichment
```
ryandeivert: For clarity and readability sake, can you make this comment the actual docstring for the `class` (and drop the `pass`), like:
```
class DropAlertException(Exception):
    """Raise this to stop an alert from firing from an enrichment"""
```

stream_alert/alert_processor/main.py
```diff
@@ -79,6 +81,22 @@ def run(alert, region, function_name, config):
         LOGGER.error('Invalid alert format:\n%s', json.dumps(alert, indent=2))
         return
 
+    # Load enrichments
+    modules_to_import = set()
+    # walk the enrichments directory to dymanically import
+    for folder in (['enrichments']):
```
ryandeivert: This outer for loop is unnecessary since there is only one folder to walk

0xdabbad00:
The failed Travis CI check is for "./tests/scripts/pylint.sh".  The error is `E:485,38: Instance of 'SlackOutput' has no '_StreamOutputBase__enrichments' member (no-member)` due to this https://github.com/airbnb/streamalert/pull/416/files#diff-d0c7b9320a87a742c03e437e7397bc3cR485

I had used `pdb` to discover that `_StreamOutputBase__enrichments` variable, so I believe it does exist, but I don't know a different way to reference it to make this happy.

ryandeivert:
hey @0xdabbad00 - that error is because the `__enrichments` class variable gets 'mangled' by the interpreter to avoid collisions with subclasses. This is due to the double preceding underscores (and can be read more about [here](https://docs.python.org/2.7/tutorial/classes.html#tut-private)). If you notice the error, it's a subclass (`SlackOutput`) that is unable to access that member property.

One workaround for this would be to have the variable be `__enrichments__` (with two trailing underscores).

However, the preferred workaround would be to use a class method on the `StreamOutputBase` class that returns to the `__enrichments` dictionary you're trying to access, like so:
```
class StreamOutputBase(object):
    __enrichments = {}

    @classmethod
    def get_enrichments(cls):
        return cls.__enrichments
```

0xdabbad00:
I should mention in all this, that what I've done works for my needs, however it feels unclean in the sense that it results in code like this in the enrichments:
```
kwargs['alert']['header_text'] = "*{}*".format(kwargs['alert']['rule_name'])
```
which means I'm just reusing and setting variables in the original `kwargs` that was passed to the `dispatch` call instead of creating any sort of interface for the modifications.  For example, it might be cleaner to have some functions like `get_rule_name()` and `set_header_text(str)`, such that you'd use:
```
set_header_text("*{}*".format(get_rule_name()))
```

0xdabbad00:
I need guidance on how to add tests to improve that coverage score.

ryandeivert:
Hey Scott - I’m grabbing food now but happy to assist after lunch.

ryandeivert:
hey @0xdabbad00 - it looks like you've updated previous unit tests to pass but not added any new ones for your feature(s). Take a look in `tests/unit/stream_alert_alert_processor/test_output_base.py` and `tests/unit/stream_alert_alert_processor/test_output.py` for examples of other tests for those files. Is there any specific test you're struggling with?

stream_alert/alert_processor/main.py
```diff
@@ -79,6 +81,22 @@ def run(alert, region, function_name, config):
         LOGGER.error('Invalid alert format:\n%s', json.dumps(alert, indent=2))
         return
 
+    # Load enrichments
+    modules_to_import = set()
+    # walk the enrichments directory to dymanically import
+    for root, _, files in os.walk('enrichments'):
+        filtered_files = [enrichment_file for enrichment_file in files if not (
+            enrichment_file.startswith(('.', '__init__')) or
```
ryandeivert: This is an odd indentation - it should be aligned with `[enrichment_file... ` in the preceding line.

Tip: The `yapf --in-place <path/to/file.py>` will help with file formatting.

0xdabbad00:
@ryandeivert Anything else I need to do for this?

jacknagz:
Hey @0xdabbad00, we need some time to think about how we want this feature to be implemented to fit current/future use cases.  If you can define a set of requirements you need? I think that can really help with how we review the design here.

0xdabbad00:
Requirements:
- Ability to change any of the text in output messages. Including the full doc string as is done currently not working for our interests.
- Ability to include items from the record into the alert and to enhance that data.  We might want our alerts to just include "Root login to AWS account Prod (111111111111)".  This would mean no description, no record listing, just that line of text.  That line of text would extract out the relevant account ID, and would enhance it by doing a lookup that gives the human readable name of that account.
- (Optional) Ability to drop alerts.  We are currently using my new functionality to drop alerts to make our alerts routable, such that we specify two outputs for our rules (a "security" and a "prod" channel), and when an alert fires that is not relevant to the Prod account, but is destined for the "prod" channel, we just drop it.  A more intelligent routing mechanism may be desired here, for example, you might have a lookup table of account IDs, and arrays of Slack channel names those alerts should go to.  Currently the only other way of implementing that functionality is to copy/paste all your alerts into a bunch of files with the only differences between those files being account ID checks and the output Slack channels.
- Ability to easily add this functionality to different rules.  For example, we have a single "minimal" enrichment that I use across all of my alerts that instead of the doc string for the rule in the pretext, it simply provides a link to the rule in our git repo.
- (Optional) Ability to chain enrichments.  It might be useful for some use cases to include multiple enrichments for a rule if they adding generic functionality.  My code does support that, however currently I specify one enrichment for each rule type I have (ex. cloudtrail vs flow_logs), and then I call a "base" enrichment function directly from there to perform some of the general things I want done and to give me access to some general functions (such as an account ID -> Name mapping).

jacknagz:
Thanks @0xdabbad00, we'll follow up soon

0xdabbad00:
Any updates?

ryandeivert:
closing due to staleness - we're happy to revisit this if desired

0xdabbad00:
This was such a good improvement in concept (maybe not in my implementation of it though). @ryandeivert Did the concept of "enhancements" ever make it into alerts as has been mentioned in https://github.com/airbnb/streamalert/issues/305#issuecomment-328568088 ?

ryandeivert:
hey @0xdabbad00 - we are actually in the process of working on a solution for this now. admittedly, I lost your previously filed issue (#305) and filed a new one as #870. I updated the description in there mentioning they are related. Please comment there with any new requests.

PR #417, opened 2017-10-25
[apps] fix for improperly marking app config as failing

ryandeivert:
to: @javuto 
size: small
resolves N/A

## Background

Discovered that a partial execution of log collection would mark the config state as 'failing' upon completion, even though there was no failure. This was due to the state toggling naively marking the state to 'failing' if it was still set to 'running' at the time of exit.

## Changes

* Adding a new state of `'partial'` that will get set when there are more logs to poll and a new apps function is invoked.
* The new 'partial' state must now exist in conjunction with the 'successive_invocation' event type in order for an invocation to continue.

## Testing

* Adding unit tests to verify that the 'partial' flag is working as expected, and additional unit tests.


javuto:
LGTM 🎸

PR #422, opened 2017-10-25
 getting started minor update

securityclippy:
cc: <optional-cc-to-specific-users>
size: small|medium|large
resolves #<related-issue-goes-here>

## Background

Just a small documentation update

## Changes

* how to solve psutil installation problems with python-dev
* ...

## Testing

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -20,7 +20,7 @@ Dependencies
 .. code-block:: bash
 
   $ pip install virtualenv
-  $ virtualenv venv
+  $ virtualenv -p python2.7 venv
   $ source venv/bin/activate
 
 5. Install dependencies:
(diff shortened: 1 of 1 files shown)


ryandeivert:
hey @securityclippy I looked in the slack to try to see if you mentioned the problem you experienced with `psutil` but came up short. Do you have steps to reproduce the problem you had?

securityclippy:
@ryandeivert sure.  Any fresh install of ubuntu 16.04 should produce the issue.  I'm sure there are probably others as well, but basically all you need to do is follow the ```getting started``` through to the ```pip install -r  requirements.txt```.  During the requirements installation,  the wheel build for psutil will fail.  The above  listed fix will get your past that  point and able to continue on.

securityclippy:
PS.  I added the ```virtualenv -p python2.7 venv``` too here, apparently, because I'm bad at breaking up issues.  That was just a clarification for explicit use of python2.7 rather than implicit, as some distro's are starting to ship with python3 as the default.

jacknagz:
@securityclippy Thanks for the change, can you rebase this branch off latest master?

ryandeivert:
@securityclippy thanks for the clarification! I also support the addition of `-p python2.7` 👍

securityclippy:
@jacknagz @ryandeivert ack, rebase incoming shortly

PR #423, opened 2017-10-25
[apps] fix for potential but unlikely race condition

ryandeivert:
to: @javuto 
size: small
resolves N/A

## Background

Identified a very unlikely but possible race condition that could occur. In short, a 'chained' invocation of a app function could be started before the config in Parameter Store was properly updated with the `partial` state value. This could then cause the new invocation of the lambda to fail if it read the state as `running` before it was written to `partial`

## Changes

* Writing the `partial` state config to parameter store before invoking the new lambda function.

## Testing

* Unit tests passing.


ryandeivert:
FYI the coverage drop is due to the total number of lines decreasing by 1 🤣

PR #425, opened 2017-10-27
[tf][kinesis] Disabling shard level metrics by default

javuto:
to: @ryandeivert @jacknagz 
size: small

## Background

StreamAlert was enabling by default [enhanced monitoring](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_EnableEnhancedMonitoring.html) to surface infrastructure metrics at a granular level. This could ended up costly since those metrics are charged and they should be enabled when needed, instead of adding a unnecessary cost and not being utilized properly. 

## Changes

* Shard level metrics are now empty by default, with the necessary documentation to enable them.

## Testing

Changes were tested running `python manage.py terraform init`, `python manage.py terraform build` and `python manage.py terraform destroy` to make sure all the infrastructure is created and destroyed properly.

### DIFF
--- a/docs/source/metrics.rst
+++ b/docs/source/metrics.rst
@@ -1,17 +1,25 @@
 Metrics
 =======
 
-StreamAlert enables `Enhanced Monitoring`_ to surface infrastructure metrics at a granular level.
+StreamAlert allows to enable `Enhanced Monitoring`_ to surface infrastructure metrics at a granular level.
 
 .. _Enhanced Monitoring: https://docs.aws.amazon.com/kinesis/latest/APIReference/API_EnableEnhancedMonitoring.html
 
-Go to AWS Console -> CloudWatch -> Metrics -> Kinesis to view them.
+When enabled, access them by going to AWS Console -> CloudWatch -> Metrics -> Kinesis.
 
-Example metrics:
+Enhanced metrics can be enabled in ``conf/global.json`` as ``shard_level_metrics``, for example:
 
-* IncomingBytes
-* IncomingRecords
-* Error/Exceptions
+.. code-block:: bash
+
+  "shard_level_metrics": [
+    "IncomingBytes",
+    "IncomingRecords",
+    "OutgoingBytes",
+    "OutgoingRecords",
+    "WriteProvisionedThroughputExceeded",
+    "ReadProvisionedThroughputExceeded",
+    "IteratorAgeMilliseconds",
+  ]
 
 These metrics can be viewed at the shard-level or the stream-level (cluster/environment).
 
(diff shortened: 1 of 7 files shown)


stream_alert_cli/terraform/kinesis_streams.py
```diff
@@ -29,6 +29,7 @@ def generate_kinesis_streams(cluster_name, cluster_dict, config):
     """
     prefix = config['global']['account']['prefix']
     config_modules = config['clusters'][cluster_name]['modules']
+    shard_level_metrics = config['global']['infrastructure']['monitoring']['shard_level_metrics']
```
jacknagz: what happens if this doesn't exist in the config?

docs/source/metrics.rst
```diff
@@ -1,17 +1,25 @@
...
+.. code-block:: bash
+
+  "shard_level_metrics": [
+    "IncomingBytes",
+    "IncomingRecords",
+    "OutgoingBytes",
+    "OutgoingRecords",
+    "WriteProvisionedThroughputExceeded",
+    "ReadProvisionedThroughputExceeded",
+    "IteratorAgeMilliseconds",
```
jacknagz: The trailing comma isn't valid JSON

javuto:
Since `shard_level_metrics` is an optional argument, when you pass an empty list it does not add any metric into the stream. I tested it with an empty list and with the values outlined in the example and terraform behaved as expected in both scenarios. Thanks for the review!

stream_alert_cli/terraform/kinesis_streams.py
javuto: `KeyError` exception I guess. I should default it to an empty list, if the value isn't there.

stream_alert_cli/terraform/kinesis_streams.py
```diff
@@ -30,6 +30,12 @@ def generate_kinesis_streams(cluster_name, cluster_dict, config):
     prefix = config['global']['account']['prefix']
     config_modules = config['clusters'][cluster_name]['modules']
 
+    infrastructure_config = config['global'].get('infrastructure')
+    shard_level_metrics = []
+    if infrastructure_config and 'monitoring' in infrastructure_config:
+        if 'shard_level_metrics' in infrastructure_config['monitoring']:
```
jacknagz: Check out the [get](https://docs.python.org/2/library/stdtypes.html#dict.get) method on a dictionary

stream_alert_cli/terraform/kinesis_streams.py
```diff
@@ -30,6 +30,11 @@ def generate_kinesis_streams(cluster_name, cluster_dict, config):
     prefix = config['global']['account']['prefix']
     config_modules = config['clusters'][cluster_name]['modules']
 
+    shard_level_metrics = config['global']['infrastructure']['monitoring'].get(
+        'shard_level_metrics',
+        list()
```
jacknagz: one last change: you can just use `[]`, this is a pattern in the repo 😄

PR #426, opened 2017-10-27
[apps] adding G Suite app for all activity audit reports

### LINKED ISSUE #348: App: Create StreamAlert App for GSuite Admin Reports
### Background
With the completion of, #278 we have the ability to add new log forwarding apps to StreamAlert. The apps execute on a schedule, sending logs to the StreamAlert rule processor for processing.

### Request
Create an App to support **all** GSuite Admin Reports:
* https://developers.google.com/admin-sdk/reports/v1/quickstart/python
* https://developers.google.com/admin-sdk/reports/v1/guides/manage-audit-admin
* https://developers.google.com/admin-sdk/reports/v1/guides/manage-audit-login
* https://developers.google.com/admin-sdk/reports/v1/guides/manage-audit-tokens
...

ryandeivert:
to: @javuto and @austinbyers 
size: large
resolves #348 

## Background

See #348 

tl;dr - Consume all GSuite logs:

https://developers.google.com/admin-sdk/reports/v1/guides/manage-audit-admin
https://developers.google.com/admin-sdk/reports/v1/guides/manage-audit-login
https://developers.google.com/admin-sdk/reports/v1/guides/manage-audit-tokens
https://developers.google.com/admin-sdk/reports/v1/guides/manage-audit-drive
https://developers.google.com/admin-sdk/reports/v1/appendix/activity/saml
https://developers.google.com/admin-sdk/reports/v1/appendix/activity/rules
https://developers.google.com/admin-sdk/reports/v1/appendix/activity/mobile
https://developers.google.com/admin-sdk/reports/v1/appendix/activity/gplus
...

## Changes

* Adding initial implementation of G Suite StreamAlertApp
* This app utilizes the `google-api-python-client` module, which has been added to the `requirements.txt` (along with other dependencies) and to the 3rd party libraries that get packaged with the Apps lambda function.
* All `GSuite` apps must use a Google [service account key](https://developers.google.com/identity/protocols/OAuth2ServiceAccount) which must be configured by an admin, along with access to the `Admin SDK` api.
  * The service account key type must be in the Google-recommended JSON format (not p12):
image
  * Attempting to create a GSuite application with a p12 keyfile will fail and inform the user that a JSON keyfile must be provided.

### Other Changes
* Adding an `auth` property helper to the `AppConfig` so the `auth` subdictionary within the config dict can now be accessed via 'AppConfig.auth' instead of requiring apps to do a dictionary lookup of `'auth'` any time authentication info needs to be retrieved.
  * Also updating all unit tests to go with the above change.
* Making the `AppIntegration.required_auth_info` method a classmethod to the app does not need to be instantiated in the CLI to access this method.
  * Also making corresponding change to the cli's `runner.py` and `config.py` so the apps do not get instantiated when prompting the user for required authentication information.
* Adding support for complex input validation for items that are provided by the user.
  * A function can now be passed to the `user_input` cli helper that can be used to properly validate input. Previously only a simple `x is not in y` or regex comparison could be used.

### Bug Fixes
* Marking the `AppIntegration._more_to_poll` to `False` if there was nothing returned from an app's gather function. This avoids potentially bad looping in the subclasses.
* Fixing potential issue where casting auth values to a string could potentially cast dict values. This is meant to only case `unicode` to str, so checking for unicode first.
* Fixing bug in `validate-schemas` command that caused the `--test-files` flag to not work as expected

## Testing

* Added unit tests for the `GSuiteReportsApp`s.
* Also added the schema for `gsuite:reports` and added a test event for validation purposes only to test the schema.
* All tests passing.

### DIFF
--- a/app_integrations/apps/gsuite.py
+++ b/app_integrations/apps/gsuite.py
@@ -180,6 +180,15 @@ class GSuiteAdminReports(GSuiteReportsApp):
 
 
 @app
+class GSuiteCalendarReports(GSuiteReportsApp):
+    """G Suite Calendar Activity Report app integration"""
+
+    @classmethod
+    def _type(cls):
+        return 'calendar'
+
+
+@app
 class GSuiteDriveReports(GSuiteReportsApp):
     """G Suite Drive Activity Report app integration"""
 
@@ -189,12 +198,57 @@ class GSuiteDriveReports(GSuiteReportsApp):
 
 
 @app
+class GSuiteGroupsReports(GSuiteReportsApp):
+    """G Suite Groups Activity Report app integration"""
+
+    @classmethod
+    def _type(cls):
+        return 'groups'
+
+
+@app
+class GSuiteGPlusReports(GSuiteReportsApp):
+    """G Suite Google Plus Activity Report app integration"""
+
+    @classmethod
+    def _type(cls):
+        return 'gplus'
+
+
+@app
 class GSuiteLoginReports(GSuiteReportsApp):
     """G Suite Login Activity Report app integration"""
 
     @classmethod
     def _type(cls):
         return 'login'
+
+
+@app
+class GSuiteMobileReports(GSuiteReportsApp):
+    """G Suite Mobile Activity Report app integration"""
+
+    @classmethod
+    def _type(cls):
+        return 'mobile'
+
+
+@app
+class GSuiteRulesReports(GSuiteReportsApp):
+    """G Suite Rules Activity Report app integration"""
+
+    @classmethod
+    def _type(cls):
+        return 'rules'
+
+
+@app
+class GSuiteSAMLReports(GSuiteReportsApp):
+    """G Suite SAML Activity Report app integration"""
+
+    @classmethod
+    def _type(cls):
+        return 'saml'
 
 
 @app
--- a/app_integrations/apps/gsuite.py
+++ b/app_integrations/apps/gsuite.py
@@ -1,0 +1,206 @@
+"""
+Copyright 2017-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import json
+
+from apiclient import discovery, errors
+from oauth2client.service_account import ServiceAccountCredentials
+
+from app_integrations import LOGGER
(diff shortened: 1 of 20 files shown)


mime-frame:
@ryandeivert can you add support for https://developers.google.com/admin-sdk/reports/v1/guides/manage-audit-mobile? It appears to be the last remaining one per my Google foo: `inurl:developers.google.com inurl:manage-audit`

ryandeivert:
@mime-frame your google foo is excellent :) however, there are a couple of others we aren't doing:

https://developers.google.com/admin-sdk/reports/v1/reference/activities/list

`calendar`
`groups`
`gplus`
`mobile`
`rules`

mime-frame:
@ryandeivert - yup, it's worth supporting all of the above - you think its worth doing this in a 2nd PR or this one? There's also a SAML one too it seems: https://developers.google.com/admin-sdk/reports/v1/appendix/activity/saml

ryandeivert:
Supporting them is a very simple change tbh so it doesn't really matter to me either way. Really annoying that googles docs say one thing one place and other thing somewhere else 🤔

mime-frame:
> Supporting them is a very simple change tbh so it doesn't really matter to me either way.

Cool, lets get it into this PR then :)

ryandeivert:
@mime-frame done

app_integrations/apps/gsuite.py
```diff
@@ -0,0 +1,260 @@
...
+                Otherwise, return a list of activies for this application type.
+        """
+        if not self._create_service():
+            return False
+
+        activities_list = self._activities_service.list(
+            userKey='all',
+            applicationName=self._type(),
+            startTime=self._last_timestamp,
+            pageToken=self._next_page_token if self._next_page_token else None
```
austinbyers: Is there some reason that `self._next_page_token` could be an empty string? If so, you could do simply:

```python
pageToken = self._next_page_token or None
```

ryandeivert: The case of it potentially being an empty string is exactly why I added this, so good catch :). I don't know for sure if passing `''` to `pageToken` would have a different effect than passing `None` so I just wanted to guard against it.

I added this because I'm not sure what the default value (if any) is within the [response](https://developers.google.com/admin-sdk/reports/v1/reference/activities/list) for `nextPageToken` that gets returned. It's unclear if 1) the key exists at all if there is no value, and 2) what the value would be if it _does_ exist but is empty (is it an empty string or a json null)?

The code where this is set is here, using a `.get`:
https://github.com/airbnb/streamalert/blob/14480665dc214185f6c69986abf8092ddb3badb8/app_integrations/apps/gsuite.py#L121

ryandeivert:
@austinbyers excellent question - the answer is yes, the API tokens/secrets are sent to SSM as a `SecureString`:

https://github.com/airbnb/streamalert/blob/14480665dc214185f6c69986abf8092ddb3badb8/stream_alert_cli/apps.py#L41-L49

app_integrations/apps/gsuite.py
ryandeivert: Maybe a second set of eyes on the docs would help unearth something I'm not seeing ;)

PR #430, opened 2017-10-27
[docs][setup] use aws profiles instead of exporting creds directly

jacknagz:
to: @austinbyers 
size: small

## Background

Provide instructions on how to use named AWS profiles for account credential setup.

## Changes

* Default to using AWS profiles

## Testing

Ran the `test_the_docs.sh` script locally

### DIFF
--- a/docs/source/account.rst
+++ b/docs/source/account.rst
@@ -86,12 +86,18 @@ Next, create the user:
 * Attach the previously created ``streamalert`` policy
 * Click: ``Next: Review``, and then ``Create user``
 
-Copy the Access Key ID and Secret Access Key and export them to your environment variables:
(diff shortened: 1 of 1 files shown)


docs/source/account.rst
```diff
@@ -86,12 +86,20 @@ Next, create the user:
...
+  Default region name [None]: REGION GOES HERE
+  Default output format [None]: json
+
+Finally, export the following environment variables to activate the profile:
+
+.. code-block:: bash
+  
+  $ export AWS_PROFILE=streamalert
 
 .. note:: Remember to save your credentials in a safe place!
```
austinbyers: With your new instructions, credentials are now saved in `.aws/credentials`, so we probably don't need this note anymore

PR #433, opened 2017-10-28
Prefix streamalert to normalized types and remove duplicated IOC info

### LINKED ISSUE #413: Normalized Type Missing streamalert Prefix
## Background

For fields added to the parsed record during rule processing, such as `envelope_keys` and `normalized_types`, they should always be prefixed with `streamalert:` to avoid collision with valid data fields.  `normalized_types` does not follow this convention.

## Desired Change

Add a constant for the `streamalert:normalized_types` field, similar to how `streamalert:envelope_keys` works.

https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/rules_engine.py#L372

chunyong-lin:
to: @jacknagz or @ryandeivert 
size: small
resolves #413 

## Background
1. Data Normalization: For fields added to the parsed record during rule processing, such as `envelope_keys` and `normalized_types`, they should always be prefixed with `streamalert:` to avoid collision with valid data fields.  `normalized_types` does not follow this convention.

2. Threat Intel: Remove IOC info duplication. In some records, different keys may contain the same value. After the record being normalized, there may have duplicated IOC info inserted into the record. 
  * Example, a coming record 
```
{
  "local_ip": "1.2.3.4",
  "ipv4": "1.2.3.4"
}
```
  * Keys `local_ip` and `ipv4` will be normalized to [sourceAddress](https://github.com/airbnb/streamalert/blob/master/conf/types.json#L42-L44). Assume IP `1.2.3.4` is malicious, record will have duplicated IOC info after applying Threat Intel feature. 
```
  "local_ip": "1.2.3.4",
  "ipv4": "1.2.3.4",
  "streamalert:ioc": [
    {
      "type": "ip",
      "value": "1.2.3.4"
    },
   {
      "type": "ip",
      "value": "1.2.3.4"
    }
  ]
```
## Changes
1. Data Normalization: Prefix normalization key with `streamalert:`.
2. Threat Intel: Remove dumplicaiton IOC info by changing type of `streamalert:ioc` to a dictionary from list and add logic to check if the IOC value exists or not. 

## Testing
```
...

./tests/scripts/unit_tests.sh
Ran 392 tests in 5.502s
OK
```

### DIFF
--- a/helpers/base.py
+++ b/helpers/base.py
@@ -20,6 +20,7 @@ import time
 from netaddr import IPAddress, IPNetwork
 from netaddr.core import AddrFormatError
 
+from stream_alert.rule_processor.rules_engine import NORMALIZATION_KEY
 from stream_alert.rule_processor.threat_intel import StreamThreatIntel
 
 logging.basicConfig()
@@ -101,13 +102,13 @@ def fetch_values_by_datatype(rec, datatype):
         (list) The values of normalized types
     """
     results = []
-    if not rec.get('normalized_types'):
+    if not rec.get(NORMALIZATION_KEY):
         return results
 
-    if not datatype in rec['normalized_types']:
+    if not datatype in rec[NORMALIZATION_KEY]:
         return results
 
-    for original_keys in rec['normalized_types'][datatype]:
+    for original_keys in rec[NORMALIZATION_KEY][datatype]:
         result = rec
         if isinstance(original_keys, list):
             for original_key in original_keys:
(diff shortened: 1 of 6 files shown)


helpers/base.py
```diff
@@ -147,23 +147,47 @@ def is_ioc(rec, lowercase_ioc=True):
...
+                ],
+                "domain" : [
+                    "evil1.com",
+                    "evil2.com"
+                ]
+              }
+    """
+    if StreamThreatIntel.IOC_KEY in rec:
+        if (ioc_type in rec[StreamThreatIntel.IOC_KEY] and
+                not ioc_value in rec[StreamThreatIntel.IOC_KEY][ioc_type]):
```
jacknagz: what about `ioc_value not in ...`

chunyong-lin:
@ryandeivert or @jacknagz PTAL, the latest commit is to move Data Normalization constant in `shared` package which is shared with Rule and Alert processors. Thanks!

Issue #23, opened 2017-02-06
Feature: Outputs: JIRA


mime-frame:
Support JIRA as an output

armtash:
I'm going to take a stab at this.

PR #437, opened 2017-10-31
[outputs] jira output

armtash:
size:medium
resolves: #23 

## Background
Introduction of an additional output for Jira.  Alerts can now generate Jira issues.

## Changes
* Jira output option added which supports multiple Jira apps (core, service desk, software)
* Jira output can be configured to create individual issues per StreamAlert alert, or to aggregate into a single issue based on rule name
* Aggregation behavior is identical to how the Phantom output aggregates alerts today
* Alert content appears in the "description" field of a Jira issue, while appended alerts are added as comments
* Unused "get_comments" method added to assist with more advanced aggregation in the future
* output_base.py modified to prevent noisy warning messages related to unverified requests from flooding the logs.  Performed with urllib3.disable_warnings()
* Both outputs.py and test_outputs.py now exceed 1000 lines, so pylint "too-many-lines" temporarily disabled
* Modified one of the mocked attributes in a Phantom unit tests as is doesn't seem possible to use "side_effect" on an object attribute.  The fix looks kind of hacky, but the [docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.PropertyMock) mention this method.
* Modified _check_http_response() to use a more general error message, and not call response.json() in favor of response.content
## Testing
* Live outputs with aggregation enabled and disabled
* Integration/Unit tests passing

### DIFF
--- a/stream_alert/alert_processor/output_base.py
+++ b/stream_alert/alert_processor/output_base.py
@@ -19,12 +19,14 @@ import json
 import os
 import tempfile
 import requests
+import urllib3
 
 import boto3
 from botocore.exceptions import ClientError
 
 from stream_alert.alert_processor import LOGGER
 
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 OutputProperty = namedtuple('OutputProperty',
                             'description, value, input_restrictions, mask_input, cred_requirement')
 OutputProperty.__new__.__defaults__ = ('', '', {' ', ':'}, False, False)
@@ -224,11 +226,9 @@ class StreamOutputBase(object):
         """
         success = response is not None and (200 <= response.status_code <= 299)
         if not success:
-            resp_json = response.json()
-            LOGGER.error('Encountered an error while sending to %s: %s\n%s',
+            LOGGER.error('Encountered an error while sending to %s:\n%s',
                          cls.__service__,
-                         resp_json.get('message'),
-                         resp_json.get('errors'))
+                         response.content)
         return success
 
     @classmethod
--- a/stream_alert/alert_processor/outputs.py
+++ b/stream_alert/alert_processor/outputs.py
@@ -38,6 +38,8 @@ from stream_alert.shared.backoff_handlers import (
 # All included subclasses are designated using the '@output' class decorator
 # The keys are the name of the service and the value is the class itself
 # {cls.__service__: <cls>}
+
+# pylint: disable=too-many-lines
 STREAM_OUTPUTS = {}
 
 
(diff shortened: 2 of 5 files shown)


armtash:
Addressing the failed unit tests.

ryandeivert:
hey @armtash  - thanks for the contribution! I hadn't realized you were refactoring the `urllib` --> `requests` stuff for this and @javuto already had that in the works.

Would you mind waiting on this for his open PR (#440) to merge? I know you may go through rebase hell, but he's added a handful of other tests that may be beneficial as well. It also looks like most of your non-requests changes are encapsulated in the one new output class, so it might not be so awful.

armtash:
@ryandeivert it was my fault for not mentioning it.  I'll clean this up once @javuto's changes are implemented

ryandeivert:
hey @armtash PR #440 has merged - if you'd like to update your branch to match now that'd be awesome! also if there is anything in @javuto 's changes that you'd like to adapt, feel free to make changes & call them out in the PR body

armtash:
Hey @ryandeivert and @javuto.  I've updated the PR, however I have run into an issue I couldn't figure out with the AlertProcessorTester.  I'm finding that when a Jira output is configured, my changes to the AlertProcessorTester cause the Phantom tests to fail, along with the Jira tests.

What I noticed was that the mock variables were actually getting swapped between the two outputs.  For example, if you set the following for Jira:
`post_mock.return_value.json.return_value = {'id':9000}`
and the following for Phantom:
`post_mock.return_value.json.return_value = {'id':1000}` 
you'll notice that the values actually get reversed.  ` {'id':1000}` gets passed to Jira, and ` {'id':9000}` gets passed to phantom.  I suspect this has something to do with the mock objects being shared between the different services in AlertProcessorTester.

Once this issue gets sorted out, I can also add the unit tests for Jira.

ryandeivert:
hey @armtash thanks for the update. I'll checkout your branch and see if I can diagnose

ryandeivert:
hey @armtash when you get a chance to add some tests, that would be excellent!

armtash:
hey @ryandeivert, sorry for the delay!  I just added the unit tests and updated the PR notes.

ryandeivert:
hey @armtash ! totally didn't see this update from 3 days ago. thanks for adding the tests, I'll take another look at this tomorrow

armtash:
Hey @ryandeivert, no problem! I simplified the error checking and the unit tests today, so it all worked out.

stream_alert/alert_processor/output_base.py
```diff
@@ -19,12 +19,14 @@
 import os
 import tempfile
 import requests
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
```
ryandeivert: Please just import urllib3 directly instead of accessing through the requests package. Example:

```
import urllib3
...
urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
Unless I'm missing something and the method you're using is the only way this will work? Can you also add a note as to why this is necessary?

stream_alert/alert_processor/outputs.py
```diff
@@ -37,6 +37,8 @@
 # All included subclasses are designated using the '@output' class decorator
 # The keys are the name of the service and the value is the class itself
 # {cls.__service__: <cls>}
+
+# pylint: disable=too-many-lines
```
ryandeivert: Note to other reviewers: I asked @armtash to add this for now with the intent that I would break this out after this PR merges.

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,272 @@ def dispatch(self, **kwargs):
                                  Payload=alert_string)
 
         return self._log_status(resp)
+
+@output
+class JiraOutput(StreamOutputBase):
+    """JiraOutput handles all alert dispatching for Jira"""
+    __service__ = 'jira'
+
+    LOGIN = 'rest/auth/1/session'
```
ryandeivert: Can you maybe add a suffix to these so their usage throughout the class is more clear? Maybe `LOGIN_ENDPOINT`, etc

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,272 @@ def dispatch(self, **kwargs):
...
+        Args:
+            jql (str): The JQL query
+            max_results (int): Maximum number of results to return
+            validate_query (bool): Whether to validate the JQL query or not
+            fields (list): List of fields to return for each issue
+        Returns:
+            list: list of issues matching JQL query
+        """
+        search_url = os.path.join(self.base_url, self.SEARCH)
+        headers = {'Content-Type': 'application/json', 'cookie' : self.auth_cookie}
```
ryandeivert: Remove the extra space after `'cookie'` here. I noticed a few other instances like this in the file as well. The following command is what we've been using to help with small things like this:

`yapf --in-place /path/to/file.py`

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,272 @@ def dispatch(self, **kwargs):
...
+        """Add a comment to an existing issue
+
+        Args:
+            issue_id (str): The existing issue ID or key
+            comment (str): The body of the comment
+
+        Returns:
+            int: ID of the created comment or False if unsuccessful
+        """
+
```
ryandeivert: Remove the extra line here

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,272 @@ def dispatch(self, **kwargs):
...
+        comment_url = os.path.join(self.base_url, self.COMMENT.format(issue_id))
+        comment_body = {'body' : comment}
+        headers = {'Content-Type': 'application/json', 'cookie' : self.auth_cookie}
```
ryandeivert: Suggestion: there's a good amount of duplication here with `headers = ..` code. I'd suggest consolidating a bit with something along the lines of:

```
@output
class JiraOutput(StreamOutputBase):
    DEFAULT_HEADERS = {'Content-Type': 'application/json'}

    @classmethod
    def _get_default_headers(cls):
        """Class method to be used when you do not need the 'cookie' param"""
        return cls.DEFAULT_HEADERS.copy()

    def _get_headers(self):
        """Instance method to be used when you want to combine the cookie with the default headers"""
        return dict(self._get_default_headers(), **{'cookie': self.auth_cookie})
```
This also has the benefit of the headers being a class constant with less chance of a typo bringing things down.

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,272 @@ def dispatch(self, **kwargs):
...
+
+        Returns:
+            int: ID of the found issue or False if existing issue does not exist
+        """
+        jql = 'summary ~ "{}" and project="{}"'.format(issue_summary, project_key)
+        resp = self._search_jira(jql, fields=['id', 'summary'], max_results=1)
+        jira_id = False
+
+        try:
+            jira_id = resp[0]['id']
```
ryandeivert: Could this ever result in a `KeyError` as well (with `'id'`)? Genuinely curious (I don't have much jira experience)

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,272 @@ def dispatch(self, **kwargs):
...
+
+        Args:
+            issue_summary (str): The Jira issue summary
+            project_key (str): The Jira project to search
+
+        Returns:
+            int: ID of the found issue or False if existing issue does not exist
+        """
+        jql = 'summary ~ "{}" and project="{}"'.format(issue_summary, project_key)
+        resp = self._search_jira(jql, fields=['id', 'summary'], max_results=1)
```
ryandeivert: Though this is technically not an issue, I'd suggest calling the `_search_jira` function with the named args in the same order as their declared in the function declaration:  max_results --> validate_query --> fields (or re-order the params in the function declaration if you this order better :) )

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,272 @@ def dispatch(self, **kwargs):
...
+            project_key (str): The Jira project key which issues will be associated with
+            issue_type (str): The type of issue being created
+            description (str): The body of text which describes the issue
+
+        Returns:
+            int: ID of the created issue or False if unsuccessful
+        """
+        issue_url = os.path.join(self.base_url, self.ISSUE)
+        headers = {'Content-Type': 'application/json', 'cookie' : self.auth_cookie}
+        issue_body = {'fields':{}}
```
ryandeivert: It would be okay, and probably a little more readable, to create this object all in one go, like:
```
issue_body = {
    'fields': {
        'project': {
            'key': project_key
        },
        'summary': issue_name,
        'description': description,
        'issuetype': {
            'name': issue_type
        }
    }
}
```

stream_alert/alert_processor/output_base.py
armtash: Will do.  The change is related to the requests library generating warning messages for every request that is made with the certificate verification disabled (verify=False).  The code above is to suppress those warnings, as the streamalert logs were full of these warnings.  A single alert to Phantom or Jira could generate 3 of these warnings per alert, so it can get pretty noisy.

Unfortunately, it looks like this is the only way to suppress the errors as we are using an older version of requests:
 [v2.13.0](https://github.com/airbnb/streamalert/blob/master/requirements.txt#L71).  With the most recent version of requests (v2.18.4), it seems like what you suggested would actually work.  I am referencing this SO [post](https://stackoverflow.com/a/28002687).

stream_alert/alert_processor/outputs.py
armtash: I originally did define the dict like this but pylint was throwing a fit.  I must have screwed up the indentation or something as it seems to work just fine now.

stream_alert/alert_processor/outputs.py
armtash: If a list of dicts is passed back, it would be fairly likely for the 'id' to be present, but this should definitely be modified to "except (IndexError, KeyError)" just in case.  Nice catch! (hehe)

stream_alert/alert_processor/output_base.py
ryandeivert: Thanks for the clarification - in that case, let's just bump our `requests` min requirement to that version and switch to the aforementioned way of disabling the warnings.

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,272 @@ def dispatch(self, **kwargs):
...
+            ('project_key',
+             OutputProperty(description='the Jira project key',
+                            mask_input=False,
+                            cred_requirement=True)),
+            ('issue_type',
+             OutputProperty(description='the Jira issue type',
+                            mask_input=False,
+                            cred_requirement=True)),
+            ('aggregate',
+             OutputProperty(description='the Jira aggregation behavior to aggregate '
```
armtash: @ryandeivert I was meaning to ask you about this.  The question is asked to the user in a very awkward way, due to "Please provide" being prefixed to these descriptions.  I essentially want to ask "Do you want to aggregate alerts?".  In its current form, the user is asked "Please provide the aggregation behavior to aggregate alerts by rule name (yes/no)", which is quite the mouthful and I'm not entirely sure if it makes any sense.  Let me know if you have any suggestions.

ryandeivert: lol yeah that is a bit corny. we may have to just go with it for now and address it in the next alert processor refactor

ryandeivert: It'll probably be best to just have the full prompt message here (and change the `OutputProperty`'s `description` attribute to something better, like `prompt`)

ryandeivert: Just to clarify -  wasn't implying you should make that change :) I'll handle it in the future

stream_alert/alert_processor/outputs.py
armtash: After running this on outputs.py, it looks like there are a lot of changes made to almost every output.  Do you want me to add these changes?

armtash:
Hey @ryandeivert, I think I addressed everything you found.  Let me know if you spot anything else!

ryandeivert:
I think that would be fine - we want to start using yapf as our standard anyway. Thanks for asking!

ryandeivert:
Awesome, I’ll take a peek in the morning - thanks!

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,296 @@ def dispatch(self, **kwargs):
...
+        # Create new issue if aggregation is disabled or existing issue not found.
+        if not jira_id:
+            jira_id = self._create_issue(issue_summary,
+                                         creds['project_key'],
+                                         creds['issue_type'],
+                                         alert_body)
+
+        if jira_id or comment_id:
+            success = True
+            LOGGER.debug('sending alert to Jira %s', jira_id)
```
ryandeivert: This logger wouldn't tell us if there was a comment added, only if an issue was created right? I'd suggest moving  the logger calls to closer to where the action happens (ie: after `comment_id = self....` above and add two others after the `_get_existing_issue` and `_create_issue` call. It'll help us track the flow in cloudwatch logs as well.

Then you can simply return this at the end:
`return self._log_status(jira_id or comment_id)`

Since `_log_status` will log and error similar to what you're logging here

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,296 @@ def dispatch(self, **kwargs):
...
+    DEFAULT_HEADERS = {'Content-Type': 'application/json'}
+    LOGIN_ENDPOINT = 'rest/auth/1/session'
+    SEARCH_ENDPOINT = 'rest/api/2/search'
+    ISSUE_ENDPOINT = 'rest/api/2/issue'
+    COMMENT_ENDPOINT = 'rest/api/2/issue/{}/comment'
+
+    def __init__(self, *args, **kwargs):
+        StreamOutputBase.__init__(self, *args, **kwargs)
+        self.base_url = None
+        self.auth_cookie = None
```
ryandeivert: Maybe make these instance properties protected? ie - `self._ auth_cookie`, since they should never be accessed from outside an instance of this class.

stream_alert/alert_processor/outputs.py
```diff
@@ -803,3 +805,296 @@ def dispatch(self, **kwargs):
...
+                }
+            }
+        }
+
+        resp = self._post_request(issue_url,
+                                  data=issue_body,
+                                  headers=self._get_headers(),
+                                  verify=False)
+
+        try:
```
ryandeivert: While I'm a fan of your pythonic approach here (using try/except), I'm realizing we should _always_ expect the response (if there is one) to be json. With that in mind, this could be consolidated to:

```
resp = self._post_...
if not self._check_http_response(resp):    # If this returns False, `_check_http_response` will log the error
    return False

response = resp.json()

return response.get('id', False)
```

armtash: I can simplify this a bit more.  For all Jira responses, if a successful status code is passed back, I think you can make the assumption that valid JSON is also passed back, though I tend to be a little paranoid around exactly what is returned from the api calls. 

This is what the method looked like before consolidating the error conditions, when I was leveraging both _check_http_response(), as well as the try/catch to validate that the json content contained the expected data.

        resp = self._post_request(issue_url, data=issue_body, headers=headers, verify=False)
        if not self._check_http_response(resp):
            LOGGER.error('An error occurred during Jira issue creation %s', resp.content)
            return False

        try:
            issue_id = resp_dict.json()['id']
        except (ValueError, KeyError):
            LOGGER.error('An error occurred when parsing Jira issue creation %s', resp.content)
            return False

        return issue_id

One thing I just noticed when looking into this is that _check_http_response() might need to be tweaked, as it will most likely raise an exception at this [line](https://github.com/airbnb/streamalert/blob/master/stream_alert/alert_processor/output_base.py#L227).  If a non 2XX response is passed back, I don't think there is any guarantee that valid json is passed, and assuming that valid json is passed back, I think it is likely that [this](https://github.com/airbnb/streamalert/blob/master/stream_alert/alert_processor/output_base.py#L230) will raise a KeyError exception, as 'messages' and 'errors' are not guaranteed to be present.

This is what _check_http_respons() looked like back when I attempted the urllib -> requests transition, when I had a check to validate json:  

    @staticmethod
    def _check_http_response(resp, validate_json=False):
        if validate_json:
            try:
                resp.json()
            except ValueError:
                return False
        return resp and (200 <= resp.status_code <= 299)


Getting back to your original comments, let me know if you want me to modify all of the methods to remove the try/except, and if you want me to tweak _check_http_response().

ryandeivert: I was thinking the same thing re `_check_http_response` potentially causing an exception. I'd be good with you adding a try/except in that function similar to your example that does something like:
```
try:
    content = resp.json()
except ValueError:
    content = resp.content

# Then we could still log the content, even if it's not json
...
```
To your point, though, it wouldn't raise a KeyError because it uses `get`on the response.

Hopefully to answer you question - tweak away (_check_http_response, etc)! If you can make the change one place and benefit from it 5x places, I'm all for it.

armtash: Ah my mistake.  Whenever I use "get" its been with a default value, so I just assumed it would also raise a KeyError exception if the key was missing, rather than returning None.

I'll update _check_http_response and clean up the try/excepts.

armtash: Just out of curiosity, is it necessary to make the distinction between these two:

    LOGGER.error('Encountered an error while sending to %s: %s\n%s',
                  cls.__service__,
                  content.get('message'),
                  content.get('errors'))

and 

    LOGGER.error('Encountered an error while sending to %s:\n%s',
                  cls.__service__,
                  resp.content)

The later would include the "message" and "error" values in resp.content should they be present, along with everything else passed back with the response.  My preference would be to do away with the former unless you have any objections.

ryandeivert: Sure - that's fine, I suppose you're right that they'd all be present in the content. Ship it!

armtash:
@ryandeivert, I think I took care of your last comments.  I've updated the PR summary as well.

ryandeivert:
hey @armtash thanks - taking a look this morning. I'm having jack take a peak as well to make sure I'm not overlooking anything :)

requirements.txt
```diff
@@ -74,6 +74,7 @@ pyOpenSSL==17.3.0
 python-dateutil==2.6.1
 pytz==2017.3
 PyYAML==3.12
+requests==2.18.4
```
jacknagz: this file shouldn't be directly edited, add this pin to the `requirements-top-level.txt`

stream_alert/alert_processor/output_base.py
```diff
@@ -215,20 +217,16 @@ def _post_request(cls, url, data=None, headers=None, verify=True):
     @classmethod
     def _check_http_response(cls, response):
         """Method for checking for a valid HTTP response code
```
jacknagz: why squash the newlines?

stream_alert/alert_processor/outputs.py
```diff
@@ -1054,3 +1056,285 @@ def dispatch(self, **kwargs):
...
+
+    def _search_jira(self, jql, fields=None, max_results=100, validate_query=True):
+        """Search Jira for issues using a JQL query
+
+        Args:
+            jql (str): The JQL query
+            fields (list): List of fields to return for each issue
+            max_results (int): Maximum number of results to return
+            validate_query (bool): Whether to validate the JQL query or not
+        Returns:
```
jacknagz: add a space before `Returns`

stream_alert/alert_processor/outputs.py
```diff
@@ -1054,3 +1056,285 @@ def dispatch(self, **kwargs):
...
+                rule_name (str): Name of the triggered rule
+                alert (dict): Alert relevant to the triggered rule
+        """
+        creds = self._load_creds(kwargs['descriptor'])
+        if not creds:
+            return self._log_status(False)
+
+        issue_id = None
+        comment_id = None
+        issue_summary = 'StreamAlert {}'.format(kwargs['rule_name'])
```
jacknagz: We should consider moving strings like this into class definitions to easily change/modify in the future cc @ryandeivert

stream_alert/alert_processor/outputs.py
ryandeivert: you mean newline right?

requirements.txt
armtash: gotcha.  it looks like you already updated the requests requirement a few days ago so this is redundant anyways.

armtash:
@jacknagz, took care of the formatting issues you found, and removed the changes to requirements.txt as they were unnecessary.

PR #440, opened 2017-11-02
[output] Replace urllib2 with requests

### LINKED ISSUE #158: Switch to using `requests` library instead of `urllib2`
For alert dispatching, we currently use `urllib2` (see: [here](https://github.com/airbnb/streamalert/blob/master/stream_alert/alert_processor/output_base.py) ) which works for our purposes. However, we should consider migrating to using the `requests` package going forward to save ourselves some headache.

http://docs.python-requests.org/en/latest/index.html

javuto:
to: @ryandeivert 
size: medium
resolves #158 

## Background

Outputs in StreamAlert were using `urllib2` to process HTTP requests. For certain things (parsing output, parameter encoding) is simpler to use `requests` and we are already using it in app integrations.

## Changes

* Replacing `urllib2` by `requests` in the `output_base.py`, to handle `GET`, `POST` checking for successful traffic.
* Added all the relevant code to the dispatchers for PagerDuty, Phantom and Slack to use the new methods using `requests`.
* Updated unit tests.
* Added `requests` as third party library in the `alert_processor_config`.

## Testing

Unit tests, linter and StreamAlert processor:
```
...
TOTAL                                            2219     46    98%
Ran 397 tests in 8.224s

OK

$ ./tests/scripts/pylint.sh

$ python manage.py lambda test --processor all
...
```

### DIFF
--- a/stream_alert/alert_processor/output_base.py
+++ b/stream_alert/alert_processor/output_base.py
@@ -184,9 +184,9 @@ class StreamOutputBase(object):
 
         Args:
             url (str): Endpoint for this request
-            headers (dict): Dictionary containing request-specific header parameters
             params (dict): Payload to send with this request
-            verify (bool): Whether or not SSL should be used for this request
+            headers (dict): Dictionary containing request-specific header parameters
+            verify (bool): Whether or not the server's SSL certificate should be verified
         Returns:
             dict: Contains the http response object
         """
@@ -198,16 +198,15 @@ class StreamOutputBase(object):
 
         Args:
             url (str): Endpoint for this request
+            data (dict): Payload to send with this request
             headers (dict): Dictionary containing request-specific header parameters
-            params (dict): Payload to send with this request
-            verify (bool): Whether or not SSL should be used for this request
+            verify (bool): Whether or not the server's SSL certificate should be verified
         Returns:
             dict: Contains the http response object
         """
         return requests.post(url, headers=headers, json=data, verify=verify)
 
-    @staticmethod
-    def _check_http_response(response):
+    def _check_http_response(self, response):
         """Method for checking for a valid HTTP response code
 
         Args:
@@ -216,7 +215,14 @@ class StreamOutputBase(object):
         Returns:
             bool: Indicator of whether or not this request was successful
         """
-        return response is not None and (200 <= response.status_code <= 299)
+        success = response is not None and (200 <= response.status_code <= 299)
+        if not success:
+            resp_json = response.json()
+            LOGGER.error('Encountered an error while sending to %s: %s\n%s',
+                         self.__service__,
+                         resp_json.get('message'),
+                         resp_json.get('errors'))
+        return success
 
     @classmethod
     def _get_default_properties(cls):
--- a/stream_alert/alert_processor/output_base.py
+++ b/stream_alert/alert_processor/output_base.py
@@ -17,9 +17,8 @@ from abc import ABCMeta, abstractmethod
 from collections import namedtuple
 import json
 import os
-import ssl
 import tempfile
-import urllib2
+import requests
 
 import boto3
 from botocore.exceptions import ClientError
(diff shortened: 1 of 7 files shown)


stream_alert/alert_processor/output_base.py
```diff
@@ -180,39 +179,44 @@ def _log_status(self, success):
...
     @staticmethod
-    def _check_http_response(resp):
-        return resp and (200 <= resp.getcode() <= 299)
+    def _post_request(url, data=None, headers=None, verify=True):
+        """Method to return the json loaded response for this POST request
+
+        Args:
+            url (str): Endpoint for this request
+            headers (dict): Dictionary containing request-specific header parameters
+            params (dict): Payload to send with this request
```
ryandeivert: These args don't match the function declaration, please update

stream_alert/alert_processor/outputs.py
```diff
@@ -101,24 +100,24 @@ def dispatch(self, **kwargs):
...
             'description': message,
             'details': details,
             'client': 'StreamAlert'
-        })
+        }
+        headers = None
+        verify = True
 
-        resp = self._request_helper(creds['url'], values_json)
+        resp = self._post_request(creds['url'], data, headers, verify)
```
ryandeivert: Suggestion: don't instantiate the `headers` and `verify` vars above just to pass them to this function - just pass them directly since they aren't used anywhere else in this functions scope.

stream_alert/alert_processor/outputs.py
```diff
@@ -188,27 +187,26 @@ def dispatch(self, **kwargs):
...
             'payload': payload,
             'event_action': 'trigger',
             'client': 'StreamAlert'
-        })
+        }
+        headers = None
+        verify = True
 
-        resp = self._request_helper(creds['url'], values_json)
+        resp = self._post_request(creds['url'], data, headers, verify)
```
ryandeivert: Same comment as above.

stream_alert/alert_processor/outputs.py
```diff
@@ -188,27 +187,26 @@ def dispatch(self, **kwargs):
...
         success = self._check_http_response(resp)
 
-        if not success:
-            response_value = json.load(resp)
-            error_message = response_value['error']['message']
-            detailed_errors = response_value['error']['errors']
+        if not success and resp:
```
ryandeivert: Can we have the `_check_http_response` method handle this error logging instead of doing it in a bunch of disparate places? You should be able to do something like this from the base class so the message is still context specific:

```
LOGGER.error('Encountered an error while sending to %s: %s\n%s',
    self.__service__,
    response['message'],
    response['errors'])
```

stream_alert/alert_processor/outputs.py
```diff
@@ -258,19 +256,17 @@ def _check_container_exists(self, rule_name, container_url, headers):
...
+            '_filter_name': rule_name,
+            'page_size': 1
+        }
+        resp = self._get_request(container_url, params, headers, False)
         if not self._check_http_response(resp):
             return False
 
         try:
-            resp_dict = json.loads(resp.read())
+            response = json.loads(resp.text)
```
ryandeivert: This is no longer necessary with requests (I don't think) and you can just use `resp.json()` without the try/except

stream_alert/alert_processor/outputs.py
```diff
@@ -303,20 +299,19 @@ def _setup_container(self, rule_name, rule_description, base_url, headers):
...
-        container_string = json.dumps(ph_container)
-        resp = self._request_helper(container_url, container_string, headers, False)
+        resp = self._post_request(container_url, ph_container, headers, False)
 
         if not self._check_http_response(resp):
             return False
 
         try:
-            resp_dict = json.loads(resp.read())
+            response = json.loads(resp.text)
```
ryandeivert: Same comment as above

stream_alert/alert_processor/outputs.py
```diff
@@ -553,12 +547,14 @@ def dispatch(self, **kwargs):
...
-        resp = self._request_helper(creds['url'], slack_message)
+        resp = self._get_request(creds['url'], slack_message)
         success = self._check_http_response(resp)
 
-        if not success:
-            LOGGER.error('Encountered an error while sending to Slack: %s',
-                         resp.read())
+        if not success and resp:
+            response = resp.json()
+            LOGGER.error('Encountered an error while sending to Slack: %s\n%s',
```
ryandeivert: Same comment as before, having this handled in the base class' `_check_http_response` method.

PR #443, opened 2017-11-03
[output] Using timeout in requests methods

javuto:
to: @ryandeivert 
size: small

## Background

Requests is replacing urllib2 in the outputs.

## Changes

* Having `requests` to use a default timeout (for both `connection` and `read` calls) of 3.05 seconds, as recommended by the [requests docs](http://docs.python-requests.org/en/master/user/advanced/#timeouts).

## Testing

```
...
TOTAL                                            2219     46    98%
Ran 397 tests in 8.224s

OK

$ ./tests/scripts/pylint.sh

$ python manage.py lambda test --processor all
...
```

### DIFF
--- a/stream_alert/alert_processor/output_base.py
+++ b/stream_alert/alert_processor/output_base.py
@@ -55,6 +55,10 @@ class StreamOutputBase(object):
     __metaclass__ = ABCMeta
     __service__ = NotImplemented
 
+    # _DEFAULT_REQUEST_TIMEOUT indicates how long the requests library will wait before timing
+    # out for both get and post requests. This applies to both connection and read timeouts
+    _DEFAULT_REQUEST_TIMEOUT = 3.05
+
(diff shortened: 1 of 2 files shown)


stream_alert/alert_processor/output_base.py
```diff
@@ -55,6 +55,10 @@ class StreamOutputBase(object):
     __metaclass__ = ABCMeta
     __service__ = NotImplemented
 
+    # _DEFAULT_REQUEST_TIMEOUT indicates long the requests library will wait before timing
```
ryandeivert: This should be `indicates how long` not `indicates long` 🤣

ryandeivert: @javuto before merging can you also update the comment where this was copied from?

javuto:
u0iluqv

PR #457, opened 2017-11-08
[outputs] Firehose Output and Terraform Cleanup

jacknagz:
to: @ryandeivert (Lambda) @chunyong-lin (Infra)
size: large

## Background

Alerts are generally sent in low volume, but there may be scenarios where this is not the case.  When using S3 as an alert output, the Alert Processor can only handle so many requests per second before timing out. 

To fix this issue, alerts can now land in S3 to be searchable in Athena via a new Firehose Delivery Stream.  This enables the Alert Processor to asynchronously send data to S3 in high volume.

## Lambda Changes

* Write a new `KinesisFirehose` output plugin.
* Adjust the Athena Partition Refresh to determine the S3 key regex dynamically.
  * This allows for backwards compatibility between the S3 and Firehose output plugins, since the resultant S3 key is determined based on the service.
  * S3: `alerts/dt=YYYY-MM-DD-HH/<uuid>.json`
  * Firehose: `alerts/YYYY/MM/DD/HH/<uuid>.json`

## Infra Changes

* Create a new Firehose Delivery Stream for alert delivery.
  * Used a buffer interval of 300 (5 minutes) to deliver alerts to S3 quicker
  * Includes Error Logging too:
screen shot 2017-11-07 at 4 26 21 pm
* Create IAM Role and Policies for the new Firehose Stream
* Move CloudWatch Metrics code into a separate file in the main `stream_alert` module (cc @ryandeivert)
* Cleanup the Rule and Alert processor Role Policy naming.  They were previously unnecessarily verbose.
* Cleanup the Kinesis Events policy.  It was previously overly permissive with the ability to read from any Kinesis Stream.
* Cleanup the IAM policy names for Athena, similar to the previous changes above.
* Add permissions to account for Glue+Athena users.

## Testing

* Linted
* Unit Tested
* Deployed in AWS test account and verified end-to-end

### DIFF
--- a/stream_alert/alert_processor/outputs.py
+++ b/stream_alert/alert_processor/outputs.py
@@ -21,10 +21,17 @@ import json
 import os
 import uuid
 
+import backoff
+from botocore.exceptions import ClientError
 import boto3
 
 from stream_alert.alert_processor import LOGGER
 from stream_alert.alert_processor.output_base import OutputProperty, StreamOutputBase
+from stream_alert.shared.backoff_handlers import (
+    backoff_handler,
+    success_handler,
+    giveup_handler
+)
 
 # STREAM_OUTPUTS will contain each subclass of the StreamOutputBase
 # All included subclasses are designated using the '@output' class decorator
@@ -531,6 +538,7 @@ class AWSOutput(StreamOutputBase):
     """Subclass to be inherited from for all AWS service outputs"""
     def format_output_config(self, service_config, values):
         """Format the output configuration for this AWS service to be written to disk
+
         AWS services are stored as a dictionary within the config instead of a list so
         we have access to the AWS value (arn/bucket name/etc) for Terraform
 
@@ -553,6 +561,88 @@ class AWSOutput(StreamOutputBase):
     def dispatch(self, **kwargs):
         """Placeholder for implementation in the subclasses"""
         pass
+
+
+@output
+class KinesisFirehoseOutput(AWSOutput):
+    """High throughput Alert delivery to AWS S3"""
+    MAX_RECORD_SIZE = 1000 * 1000
+    MAX_BACKOFF_ATTEMPTS = 3
+
+    __service__ = 'aws-firehose'
+    __aws_client__ = None
+
+    def get_user_defined_properties(self):
+        """Properties asssigned by the user when configuring a new Firehose output
+
+        Every output should return a dict that contains a 'descriptor' with a description of the
+        integration being configured.
+
+        Returns:
+            OrderedDict: Contains various OutputProperty items
+        """
+        return OrderedDict([
+            ('descriptor',
+             OutputProperty(
+                 description='a short and unique descriptor for this Firehose Delivery Stream')),
+            ('aws_value',
+             OutputProperty(description='the Firehose Delivery Stream name'))
+        ])
+
+    def dispatch(self, **kwargs):
+        """Send alert to a Kinesis Firehose Delivery Stream
+
+        Keyword Args:
+            descriptor (str): Service descriptor (ie: slack channel, pd integration)
+            rule_name (str): Name of the triggered rule
+            alert (dict): Alert relevant to the triggered rule
+
+        Returns:
+            bool: Indicates a successful or failed dispatch of the alert
+        """
+        @backoff.on_exception(backoff.fibo,
+                              ClientError,
+                              max_tries=self.MAX_BACKOFF_ATTEMPTS,
+                              jitter=backoff.full_jitter,
+                              on_backoff=backoff_handler,
+                              on_success=success_handler,
+                              on_giveup=giveup_handler)
+        def _firehose_request_wrapper(json_alert, delivery_stream):
+            """Make the PutRecord request to Kinesis Firehose with backoff
+
+            Args:
+                json_alert (str): The JSON dumped alert body
+                delivery_stream (str): The Firehose Delivery Stream to send to
+
+            Returns:
+                dict: Firehose response in the format below
+                    {'RecordId': 'string'}
+            """
+            return self.__aws_client__.put_record(DeliveryStreamName=delivery_stream,
+                                                  Record={'Data': json_alert})
+
+        if self.__aws_client__ is None:
+            self.__aws_client__ = boto3.client('firehose', region_name=self.region)
+
+        json_alert = json.dumps(kwargs['alert'], separators=(',', ':')) + '\n'
+        if len(json_alert) > self.MAX_RECORD_SIZE:
+            LOGGER.error('Alert too large to send to Firehose: \n%s...', json_alert[0:1000])
+            return False
+
+        delivery_stream = self.config[self.__service__][kwargs['descriptor']]
+        LOGGER.info('Sending alert [%s] to aws-firehose:%s',
+                    kwargs['rule_name'],
+                    delivery_stream)
+
+        resp = _firehose_request_wrapper(json_alert, delivery_stream)
+
+        if resp.get('RecordId'):
+            LOGGER.info('Alert [%s] successfully sent to aws-firehose:%s with RecordId:%s',
+                        kwargs['rule_name'],
+                        delivery_stream,
+                        resp['RecordId'])
+
+        return self._log_status(resp)
 
 
 @output
--- a/stream_alert/athena_partition_refresh/main.py
+++ b/stream_alert/athena_partition_refresh/main.py
@@ -323,15 +323,15 @@ class StreamAlertAthenaClient(object):
                              athena_table)
                 continue
 
-            # Gather all of the partitions to add per bucket
-            s3_key_regex = self.STREAMALERTS_REGEX if athena_table == 'alerts' \
-                else self.FIREHOSE_REGEX
             # Iterate over each key
             for key in keys:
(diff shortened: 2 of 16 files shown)


stream_alert/alert_processor/outputs.py
```diff
@@ -14,17 +14,24 @@
...
-import cgi
 from collections import OrderedDict
 from datetime import datetime
+import cgi
 import json
 import os
 import uuid
 
+from botocore.exceptions import ClientError
+import backoff
```
ryandeivert: This needs added to the `alert_processor_config` in `conf/lambda.json`

jacknagz: oh right, it is in my internal config, i'll push it to the default

stream_alert/alert_processor/outputs.py
```diff
@@ -14,17 +14,24 @@
 limitations under the License.
 """
 from abc import abstractmethod
-import cgi
 from collections import OrderedDict
 from datetime import datetime
+import cgi
```
ryandeivert: Please move this to it's previous location, per our conversation with @austinbyers  a few weeks ago.

The convention we now follow post his pylint refactor was that imports will be alphabetical by top level package (regardless of import/from grouping). This change is now inconsistent with all other import orders throughout the code base.

stream_alert/alert_processor/outputs.py
ryandeivert: `backoff` should come before `botocore` here. see previous comment.

stream_alert/alert_processor/outputs.py
```diff
@@ -556,6 +564,88 @@ def dispatch(self, **kwargs):
...
+
+            Args:
+                json_alert (str): The JSON dumped alert body
+                delivery_stream (str): The Firehose Delivery Stream to send to
+
+            Returns:
+                dict: Firehose response in the format below
+                    {'RecordId': 'string'}
+            """
+            return self.__aws_client__.put_record(DeliveryStreamName=delivery_stream,
```
ryandeivert: Should this be in try/except that catches `ClientError` and returns something falsey?

stream_alert/alert_processor/outputs.py
```diff
@@ -556,6 +564,88 @@ def dispatch(self, **kwargs):
...
+        if len(json_alert) > self.MAX_RECORD_SIZE:
+            LOGGER.error('Alert too large to send to Firehose: \n%s...', json_alert[0:1000])
+            return False
+
+        delivery_stream = self.config[self.__service__][kwargs['descriptor']]
+        LOGGER.info('Sending alert [%s] to aws-firehose:%s',
+                    kwargs['rule_name'],
+                    delivery_stream)
+
+        resp = _firehose_request_wrapper(json_alert, delivery_stream)
```
ryandeivert: if `_firehose_request_wrapper` is updated to catch `ClientError` above, also check this response here before doing the `resp.get()` below.

stream_alert/alert_processor/outputs.py
jacknagz: nope - backoff handles it

stream_alert/alert_processor/outputs.py
```diff
@@ -556,6 +564,88 @@ def dispatch(self, **kwargs):
...
+                dict: Firehose response in the format below
+                    {'RecordId': 'string'}
+            """
+            return self.__aws_client__.put_record(DeliveryStreamName=delivery_stream,
+                                                  Record={'Data': json_alert})
+
+        if self.__aws_client__ is None:
+            self.__aws_client__ = boto3.client('firehose', region_name=self.region)
+
+        json_alert = json.dumps(kwargs['alert'], separators=(",", ":")) + '\n'
```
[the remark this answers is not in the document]
jacknagz: 😆 copypasta

stream_alert/alert_processor/outputs.py
jacknagz: by using `backoff.on_exception` you want the wrapped function to raise, or else backoff won't happen

ryandeivert: oh right I forgot that. so what will this return if there there are 3 ClientErrors in a row (ie max attempts are reached)? none? the 4th clienterror? 🤔

jacknagz: it raises, shouldn't [the handler](https://github.com/airbnb/streamalert/blob/master/stream_alert/alert_processor/main.py#L112-L120) catch it?

ryandeivert: yeah it will - just wanted to make sure that was expected, sgtm!

terraform/modules/tf_stream_alert/iam.tf
```diff
@@ -18,9 +18,17 @@ data "aws_iam_policy_document" "lambda_assume_role_policy" {
...
+  name = "CloudwatchWriteLogs"
+  role = "${aws_iam_role.streamalert_rule_processor_role.id}"
+
+  policy = "${data.aws_iam_policy_document.alert_processor_cloudwatch.json}"
+}
+
 // IAM Role Policy: Allow the Rule Processor to invoke the Alert Processor
 resource "aws_iam_role_policy" "streamalert_rule_processor_lambda" {
-  name = "${var.prefix}_${var.cluster}_streamalert_rule_processor_invoke_alert_proc"
+  name = "LambdaInvokeAlertProcessor"
```
chunyong-lin: The `name` of role policy should be unique. If it is not unique, this policy will be shared with multiple IAM roles when there are multiple clusters in the environment. When the user destroy one cluster for some reason, this policy might be deleted, although other IAM roles from other clusters are still referring to this policy.
I tested it by creating two simple `aws_iam_role_policy` with same `name`. Policies can be created successfully, however, `terraform destroy` throws error says `NoSuchEntity` when it tries to destroy second iam role policy.

chunyong-lin: Same comment is applied to following `aws_iam_role_policy` changes in this tf file.

terraform/modules/tf_stream_alert_kinesis_events/main.tf
```diff
@@ -1,33 +1,41 @@
...
-  "Version": "2012-10-17",
-  "Statement": [
-    {
-      "Effect": "Allow",
-      "Action": [
-        "kinesis:DescribeStream",
-        "kinesis:GetRecords",
-        "kinesis:GetShardIterator",
-        "kinesis:ListStreams",
-        "logs:CreateLogGroup",
```
chunyong-lin: Want to confirm that `logs:*` actions are not required for kinesis read? I see these actions are removed from `aws_iam_policy_document`.

jacknagz: Yeah it has nothing to do with Kinesis, it was just technical debt

terraform/modules/tf_stream_alert/iam.tf
jacknagz: @chunyong-lin can you explain your testing a little more? Did you create two role policies in the same role with the same name? or different roles? they are namespaced per role, so it shouldn't conflict

jacknagz: this did just lead me to finding another bug though in the naming of the Firehose role

stream_alert/alert_processor/outputs.py
ryandeivert: Ignore this since the expectation is that an exception would be raised if an error occurs

terraform/modules/tf_stream_alert/iam.tf
chunyong-lin: I created two role policies in the same role with the same name. 
Example:
```
resource "aws_iam_role_policy" "foo" {
  name   = "role_policy_name_baz"
  role   = "${aws_iam_role.pan.id}"
  policy = "${data.aws_iam_policy_document.foo.json}"
}

resource "aws_iam_role_policy" "bar" {
  name   = "role_policy_name_baz"
  role   = "${aws_iam_role.pan.id}"
  policy = "${data.aws_iam_policy_document.bar.json}"
}
```
Creation was successful, but deletion complaint `NoSuchEntity` when it tries to destroy second policy.

stream_alert/athena_partition_refresh/main.py
```diff
@@ -323,15 +323,15 @@ def add_hive_partition(self, s3_buckets_and_keys):
...
                 continue
 
-            # Gather all of the partitions to add per bucket
-            s3_key_regex = self.STREAMALERTS_REGEX if athena_table == 'alerts' \
-                else self.FIREHOSE_REGEX
             # Iterate over each key
             for key in keys:
-                match = s3_key_regex.search(key)
+                for pattern in (self.FIREHOSE_REGEX, self.STREAMALERTS_REGEX):
+                    match = pattern.search(key)
```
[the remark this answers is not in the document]
jacknagz: @ryandeivert I didn't think it was an issue since the loop always runs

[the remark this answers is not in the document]
jacknagz: covreport says it's covered as well

stream_alert/alert_processor/outputs.py
jacknagz: I personally like grouping by `from` and `import` but ok

PR #436, opened 2017-10-31
[output] Adding support for PagerDuty Events API v2

javuto:
to: @ryandeivert 
cc: @airbnb/streamalert-maintainers
size: small

## Background

The API v2 is available since June 2016 with no new additions added to the Events API v1 since July 2016. It will be deprecated in February 2018.

## Changes

* Without breaking current code, this PR adds a new output to handle the PagerDuty Events API v2. This can be enabled gradually for rules or at once, but in any case the code is ready.

## Testing

Added unit tests for the new class `PagerDutyOutputV2`. Ran unit tests, linter and rules checks:
```
$ ./tests/scripts/unit_tests.sh
...
TOTAL                                            2231     63    97%
----------------------------------------------------------------------
Ran 396 tests in 7.514s

OK

$ ./tests/scripts/pylint.sh

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

$ python manage.py lambda test --processor all
...
StreamAlertCLI [INFO]: (55/55) Successful Tests
StreamAlertCLI [INFO]: (90/90) Alert Tests Passed
StreamAlertCLI [INFO]: Completed
```

PR #458, opened 2017-11-08
[output] Adding pagerdutyv2 to choices in manage.py

javuto:
to: @ryandeivert 
size: small

## Background

PagerDuty Events API v2 [was added](https://github.com/airbnb/streamalert/pull/436).

## Changes

* Added `pagerdutyv2` to the list of choices in `manage.py` so it can be added from the CLI.

## Testing

```
...
TOTAL                                            2332     45    98%
Ran 420 tests in 7.407s

OK

$ ./tests/scripts/pylint.sh

...
StreamAlertCLI [INFO]: (96/96) Alert Tests Passed

```

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -102,7 +102,7 @@ Examples:
     # Output service options
     output_parser.add_argument(
         '--service',
-        choices=['aws-lambda', 'aws-s3', 'pagerduty', 'phantom', 'slack'],
+        choices=['aws-lambda', 'aws-s3', 'pagerduty', 'pagerdutyv2', 'phantom', 'slack'],
         required=True,
         help=ARGPARSE_SUPPRESS
     )
--- a/manage.py
+++ b/manage.py
@@ -102,7 +102,7 @@ Examples:
     # Output service options
     output_parser.add_argument(
         '--service',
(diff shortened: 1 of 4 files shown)


jacknagz:
@javuto would you be open to renaming the output to `pagerduty-v2`? it's slightly easier to read

manage.py
```diff
@@ -102,7 +102,7 @@ def _add_output_subparser(subparsers):
     # Output service options
     output_parser.add_argument(
         '--service',
-        choices=['aws-lambda', 'aws-s3', 'pagerduty', 'phantom', 'slack'],
+        choices=['aws-lambda', 'aws-s3', 'pagerduty', 'pagerduty-v2', 'phantom', 'slack'],
```
ryandeivert: I don't think your implementation of this uses a hyphen does it? Shouldn't this just be `pagerdutyv2`. See: [here](https://github.com/airbnb/streamalert/blob/master/stream_alert/alert_processor/outputs.py#L126)

javuto: Not a typo, but trying to change that to use the hyphen, and also see @jacknagz comment above.

PR #465, opened 2017-11-10
[test] make the region used for tests configurable

0xdabbad00:
size: small
resolves N/A

## Background

We use `us-west-2` for our deployment.  We had a lambda configured as an output.  The tests failed because their mocking has been hard-coded to use `us-east-1`.  This change allows those tests to run by using the region that the deployment has been configured with.

## Changes

* Changes hard-coded `us-east-1` string to the region used by the deployment configuration.

## Testing

Change you region to `us-west-2`, configure a lambda output, run `python manage.py lambda test --processor all` and note your tests fail with `ClientError: An error occurred (404) when calling the Invoke operation:` until you apply this change.


stream_alert_cli/test.py
```diff
@@ -607,6 +607,7 @@ def __init__(self, config, context):
         self.kms_alias = 'alias/stream_alert_secrets_test'
         self.secrets_bucket = 'test.streamalert.secrets'
         self.outputs_config = load_outputs_config()
+        self.config = config
```
ryandeivert: suggestion: don't cache the whole config and instead just get the region from it here once (also avoids having to do the lookup a bunch throughout code below).

ryandeivert:
hey @0xdabbad00 - can you resolve the conflicts with this branch and I'll merge it today?

PR #467, opened 2017-11-10
[output] Adding new output for PagerDuty Incidents

javuto:
to: @ryandeivert @jacknagz 
size: medium

## Background

Migrating to the PagerDuty REST API helps to simplify the code and utilize features only present in v2. You can read up on more advantages of the migration to the new API [here](https://v2.developer.pagerduty.com/docs/migrating-to-api-v2).
After adding support for the PagerDuty Events API v2, next step is to be able to create Incidents using the API.

## Changes

* Adding support for PagerDuty Incidents API v2 (REST API) as output. [API reference for creating incidents.](https://v2.developer.pagerduty.com/v2/page/api-reference#!/Incidents/post_incidents).
* Added tests for new output.
* Adding context field as decorator in the alert, as `RuleAttributes` field. This allows specific rules to add some context to the alert record to be used in places like the outputs.
* Added documentation for new field for rule `context`.

## Testing

```
...
TOTAL                                            2388     81    97%
Ran 425 tests in 8.940s

OK

$ ./tests/scripts/pylint.sh

...
StreamAlertCLI [INFO]: (96/96) Alert Tests Passed

```

### DIFF
--- a/docs/source/rules.rst
+++ b/docs/source/rules.rst
@@ -153,6 +153,23 @@ Examples:
         req_subkeys={'columns':['port', 'protocol']})
         ...
 
+context
+~~~~~~~~~~~
+
+``context`` is an optional field to pass extra instructions to the alert processor on how to route the alert. It can be particulary helpful to pass data to an output.
+
+Example:
+
+.. code-block:: python
+
+  # Context provided to the pagerduty-incident output
+  # with instructions to assign the incident to a user.
+
+  @rule(logs=['osquery:differential'],
+        outputs=['pagerduty', 'aws-s3'],
+        context={'pagerduty-incident':{'assigned_user': 'valid_user'}})
+        ...
+
 
 Helpers
 -------
--- a/stream_alert/alert_processor/outputs.py
+++ b/stream_alert/alert_processor/outputs.py
@@ -64,9 +64,7 @@ class PagerDutyOutput(StreamOutputBase):
         Returns:
             dict: Contains various default items for this output (ie: url)
         """
-        return {
-            'url': 'https://events.pagerduty.com/generic/2010-04-15/create_event.json'
-        }
+        return {'url': 'https://events.pagerduty.com/generic/2010-04-15/create_event.json'}
 
     def get_user_defined_properties(self):
         """Get properties that must be asssigned by the user when configuring a new PagerDuty
@@ -133,9 +131,7 @@ class PagerDutyOutputV2(StreamOutputBase):
         Returns:
             dict: Contains various default items for this output (ie: url)
         """
-        return {
-            'url': 'https://events.pagerduty.com/v2/enqueue'
-        }
+        return {'url': 'https://events.pagerduty.com/v2/enqueue'}
 
     def get_user_defined_properties(self):
         """Get properties that must be asssigned by the user when configuring a new PagerDuty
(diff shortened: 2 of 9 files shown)


docs/source/rules.rst
```diff
@@ -133,6 +133,14 @@ req_subkeys
 
 This feature should be avoided, but it is useful if you defined a loose schema to trade flexibility for safety; see `Schemas <conf-schemas.html#json-example-osquery>`_.
 
+context
+~~~~~~~~~~~
+
+``context`` is an optional argument which defines an extra field of information to pass on inside
```
jacknagz: Can probably be simplified to something like: `context is an optional field to pass extra instructions to the alert processor on how to route the alert.`

stream_alert/alert_processor/outputs.py
```diff
@@ -13,6 +13,7 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+# pylint: disable=too-many-lines
```
jacknagz: hah we need to start splitting these out

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +200,171 @@ def dispatch(self, **kwargs):
...
+
+    @classmethod
+    def _get_default_properties(cls):
+        """Get the standard url used for PagerDuty Incidents API v2. This value the same for
+        everyone, so is hard-coded here and does not need to be configured by the user
+
+        Returns:
+            dict: Contains various default items for this output (ie: url)
+        """
+        return {
```
jacknagz: nit: can this be a single line?

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +200,171 @@ def dispatch(self, **kwargs):
...
+
+        response = resp.json()
+
+        # If there are results, get the first occurence from the list
+        return response and response.get(target_key)[0]['id']
+
+    def dispatch(self, **kwargs):
+        """Send incident to Pagerduty Incidents API v2
+
+        Args:
```
jacknagz: you can just declare `Keyword Args` instead of `Args`

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +200,171 @@ def dispatch(self, **kwargs):
...
+                descriptor (str): Service descriptor (ie: slack channel, pd integration)
+                rule_name (str): Name of the triggered rule
+                alert (dict): Alert relevant to the triggered rule
+        """
+        creds = self._load_creds(kwargs['descriptor'])
+        if not creds:
+            return self._log_status(False)
+
+        # Extracting context data to assign the incident
+        rule_context = kwargs['alert']['context'][self.__service__]
```
jacknagz: what if `context` isn't in the alert? does it always need to be defined for this output?

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +200,171 @@ def dispatch(self, **kwargs):
...
+            else:
+                user_to_assign = False
+
+        if not user_to_assign and rule_context.get('assigned_policy'):
+            policy_to_assign = rule_context.get('assigned_policy')
+        else:
+            policy_to_assign = creds['escalation_policy']
+
+        policies_url = os.path.join(creds['api'], self.POLICIES_ENDPOINT)
+        policy_id = self._check_exists_get_id(policy_to_assign,
```
jacknagz: nit: use consistent line breaks

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +200,171 @@ def dispatch(self, **kwargs):
...
+        }
+
+        # Start preparing the incident JSON blob to be sent to the API
+        incident_title = 'StreamAlert Incident - Rule triggered: {}'.format(kwargs['rule_name'])
+        incident_body = {
+            'type': '',
+            'details': ''
+        }
+        # We need to get the service id from the API
+        services_url = os.path.join(creds['api'], self.SERVICES_ENDPOINT)
```
jacknagz: these two lines seem to be repeated in this function, can this be extracted into a helper?

stream_alert/alert_processor/outputs.py
jacknagz: or rather, what if it's an empty map?

tests/unit/stream_alert_alert_processor/test_outputs.py
```diff
@@ -268,6 +268,144 @@ def test_dispatch_bad_descriptor(self, log_error_mock):
 
         log_error_mock.assert_called_with('Failed to send alert to %s', self.__service)
 
+class TestPagerDutyIncidentOutput(object):
```
jacknagz: Add a test case for an empty context arg

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +200,171 @@ def dispatch(self, **kwargs):
...
+        resp = self._get_request(target_url, params, headers, False)
+        if not self._check_http_response(resp):
+            return False
+
+        response = resp.json()
+
+        # If there are results, get the first occurence from the list
+        return response and response.get(target_key)[0]['id']
+
+    def dispatch(self, **kwargs):
```
jacknagz: Please add an expected schema for the `context` either here or in the class docstring

stream_alert/alert_processor/outputs.py
ryandeivert: ^^ @jacknagz I told him to do this for now. I also told @armtash to do it for his JIRA output stuff. I will break this out after these merge

stream_alert/alert_processor/outputs.py
ryandeivert: to jack's point, the default for this will be an empty dict, given this change: https://github.com/airbnb/streamalert/pull/467/files#diff-30cc622fccfff0217d84af2ff5bcd7ffR74

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +200,171 @@ def dispatch(self, **kwargs):
...
+            if user_id:
+                assigned_key = 'assignments'
+                assigned_value = [{
+                    'assignee' : {
+                        'id': '',
+                        'type': 'user_reference'}
+                }]
+            # If the user retrieval did not succeed, default to policies
+            else:
+                user_to_assign = False
```
ryandeivert: Remove this `else` and add the default value to the `get(..)` above:

`user_to_assign = rule_context.get('assigned_user', False)`

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +200,171 @@ def dispatch(self, **kwargs):
...
+                                              policies_url, headers, self.POLICIES_ENDPOINT)
+        assigned_key = 'escalation_policy'
+        assigned_value = {
+            'id': policy_id,
+            'type': 'escalation_policy_reference'
+        }
+
+        # Start preparing the incident JSON blob to be sent to the API
+        incident_title = 'StreamAlert Incident - Rule triggered: {}'.format(kwargs['rule_name'])
+        incident_body = {
```
ryandeivert: are the `type` and `details` keys within this required? It looks like these values never get set, and this is just passed as the `body in the incident payload

If these are required, can you add a comment saying so and why?

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +200,171 @@ def dispatch(self, **kwargs):
...
+        }
+        # We need to get the service id from the API
+        services_url = os.path.join(creds['api'], self.SERVICES_ENDPOINT)
+        service_id = self._check_exists_get_id(creds['service_key'],
+                                               services_url, headers, self.SERVICES_ENDPOINT)
+        incident_service = {
+            'id': service_id,
+            'type': 'service_reference'
+        }
+        incident_priority = {
```
ryandeivert: I'd be a fan of just moving this dict and the above dict right into the structure below for the `incident`.

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +200,171 @@ def dispatch(self, **kwargs):
...
+                        'type': 'user_reference'}
+                }]
+            # If the user retrieval did not succeed, default to policies
+            else:
+                user_to_assign = False
+
+        if not user_to_assign and rule_context.get('assigned_policy'):
+            policy_to_assign = rule_context.get('assigned_policy')
+        else:
+            policy_to_assign = creds['escalation_policy']
```
ryandeivert: move this above the else so it's set by default and then just set it, like so:
```
policy_to_assign = creds['escalation_policy']
if not user_to_assign and rule_context.get('assigned_policy'):
    policy_to_assign = rule_context.get('assigned_policy')
```

one less line :)

stream_alert/alert_processor/outputs.py
ryandeivert: this dispatch function is huge - can you move some of it to functions to make it more readable and make testing easier

stream_alert/alert_processor/outputs.py
javuto: Even it is defaulted to an empty dict, I will add a test for that because the output should just take the default escalation policy if no context is passed.

stream_alert/alert_processor/outputs.py
javuto: They are required, I just forgot about them :)

stream_alert/alert_processor/outputs.py
javuto: I was digging into the API docs and it turns out that this is just optional, so maybe something else to provide via `context`.

tests/unit/stream_alert_alert_processor/helpers.py
```diff
@@ -72,7 +72,18 @@ def get_random_alert(key_count, rule_name, omit_rule_desc=False):
...
 
-def get_alert(index=0):
+def get_alert(index=0, context=None):
+    """This function generates a sample alert for testing purposes
+
+    Args:
+        index (int): test_index value (0 by default)
+        context(dict): context dictionary (empty by default)
+    """
+    if not context:
```
ryandeivert: this can be simplified to `context = context or {}`

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +196,245 @@ def dispatch(self, **kwargs):
...
+                rule_name (str): Name of the triggered rule
+                alert (dict): Alert relevant to the triggered rule
+                alert['context'] (dict): Provides user or escalation policy
+        """
+        creds = self._load_creds(kwargs['descriptor'])
+        if not creds:
+            return self._log_status(False)
+
+        # Preparing headers for API calls
+        headers = {
```
ryandeivert: suggestion: just set `self.headers` right here (or `self._headers` once the instance properties are made protected)

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +196,245 @@ def dispatch(self, **kwargs):
...
+    """PagerDutyIncidentOutput handles all alert dispatching for PagerDuty Incidents API v2"""
+    __service__ = 'pagerduty-incident'
+    INCIDENTS_ENDPOINT = 'incidents'
+    USERS_ENDPOINT = 'users'
+    POLICIES_ENDPOINT = 'escalation_policies'
+    SERVICES_ENDPOINT = 'services'
+
+    def __init__(self, *args, **kwargs):
+        StreamOutputBase.__init__(self, *args, **kwargs)
+        self.base_url = None
```
ryandeivert: Maybe make these instance properties protected? ie - `self._headers`, etc..  since they should never be accessed from outside an instance of this class.

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +196,245 @@ def dispatch(self, **kwargs):
...
+        }
+        resp = self._get_request(url, params, self.headers, False)
+        if not self._check_http_response(resp):
+            return False
+
+        response = resp.json()
+
+        # If there are results, get the first occurence from the list
+        target = response.get(target_key, False)
+        if response and target:
```
ryandeivert: Suggestion: returning early if there is ever the opportunity to do so, instead of having to do compound `if` checks throughout a function helps with decrease confusion and increase readability. For instance, this could be make more readable via:

```
response = resp.json()
if not response:
    return False           # Maybe even add a LOGGER.error(...) here (?)

# If there are results, get the first occurence from the list
target = response.get(target_key, False)
if not target:
    return False           # Maybe add a LOGGER.error(...) here too (?)

return target[0]['id']
```

Theoretically the last if/return can be consolidated into just:
`return response['target_key'][0]['id'] if 'target_key' in response else False`


Just food for thought :)

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +196,245 @@ def dispatch(self, **kwargs):
...
+        policies_url = self._get_endpoint(self.base_url, self.POLICIES_ENDPOINT)
+
+        return self._item_verify(policies_url, policy, self.POLICIES_ENDPOINT,
+                                 'escalation_policy_reference')
+
+    def _service_verify(self, service):
+        """Method to verify the existance of a service with the API
+
+        Args:
+            api_url (str): Base URL of the API
```
ryandeivert: I think this args list needs updated to only include `service`

stream_alert/alert_processor/outputs.py
```diff
@@ -199,6 +196,245 @@ def dispatch(self, **kwargs):
...
+        }
+
+        # Cache base_url and headers
+        self.base_url = creds['api']
+        self.headers = headers
+
+        # Extracting context data to assign the incident
+        rule_context = kwargs['alert'].get('context', {})
+        if rule_context:
+            rule_context = rule_context[self.__service__]
```
ryandeivert: Is there a chance that this service won't exist within the `rule_context` dict? It looks like this is performed without a `get` so is it safe to assume there will always be a value in it for this service? (I don't think this is the case so just making sure I'm not missing something?)

stream_alert/alert_processor/outputs.py
javuto: I like replacing the last if/else with that super pythonic statement

stream_alert/alert_processor/outputs.py
ryandeivert: @javuto still seeing `api_url` in this list but this isn't an arg in the function declaration 🤔

stream_alert/alert_processor/outputs.py
javuto: I presumed that the service will be there, but it is a good idea to make sure and add some checks. Nice catch!

ryandeivert: This is the context that is sent from the rule processor, correct? Cause I'm thinking that will only rarely get set in some rules, no?

javuto: Yeah that would be the approach. The `context` element will be there, even if it is just an empty dict but the `__service__` could be missing. Better do the check.

stream_alert/alert_processor/outputs.py
```diff
@@ -392,7 +390,7 @@ def dispatch(self, **kwargs):
         # Extracting context data to assign the incident
         rule_context = kwargs['alert'].get('context', {})
         if rule_context:
-            rule_context = rule_context[self.__service__]
+            rule_context = rule_context.get(self.__service__, {})
```
ryandeivert: Could we break out the logic below into a function that returns a tuple of the `user_to_assign` and the `assigned_value`? just trying to understand this flow better

stream_alert/alert_processor/outputs.py
```diff
@@ -364,6 +365,32 @@ def _item_verify(self, item_url, item_str, item_key, item_type):
...
+
+        # If provided, verify the user and get the id from API
+        if user_to_assign:
+            user_assignee = self._user_verify(user_to_assign)
+            # User is verified, return tuple
+            if user_assignee:
+                return 'assignments', [{'assignee': user_assignee}]
+
+        # If escalation policy was not provided, use default one
+        policy_to_assign = context.get('assigned_policy') or self._escalation_policy
```
ryandeivert: with a lookup like this, you can use a default value with the get like `context.get('assigned_policy', self._escalation_policy)` instead of doing the `... or ...`

PR #468, opened 2017-11-11
Global Alert Firehose

jacknagz:
to: @chunyong-lin 
size: medium
follow up: #457 

## Background

The Firehose to deliver alerts to S3 should not be cluster based, it should be global.  See the tagged PR above for more context.

## Changes

* Introduce a new `tf_stream_alert_globals` module to house global configuration sets.  The idea with this module is to be a placeholder for any other global resources we'll need in the future.  We can eventually move resources into it as well.
* Update the CLI to allow for adding new `aws-firehose` outputs.

## Testing

Deployed and verified end-to-end in testing account

### DIFF
--- a/terraform/modules/tf_stream_alert/iam.tf
+++ b/terraform/modules/tf_stream_alert/iam.tf
@@ -50,7 +50,7 @@ data "aws_iam_policy_document" "rule_processor_invoke_alert_proc" {
   }
 }
 
-// IAM Role Policy: Allow the Rule Processor to put records on Firehose
+// IAM Role Policy: Allow the Rule Processor to put data on Firehose
 resource "aws_iam_role_policy" "streamalert_rule_processor_firehose" {
   name = "FirehoseWriteData"
   role = "${aws_iam_role.streamalert_rule_processor_role.id}"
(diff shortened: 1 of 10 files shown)


terraform/modules/tf_stream_alert/iam.tf
```diff
@@ -234,7 +234,7 @@ data "aws_iam_policy_document" "alert_processor_firehose" {
     ]
 
     resources = [
-      "${aws_kinesis_firehose_delivery_stream.stream_alerts.arn}",
+      "arn:aws:firehose:${var.region}:${var.account_id}:deliverystream/${var.prefix}_streamalert_alert_delivery",
```
chunyong-lin: Any reason to use full arn of firehose delivery stream, instead of `${aws_kinesis_firehose_delivery_stream.stream_alerts.arn}`?

jacknagz:
@chunyong-lin we are moving towards using `yapf` to format all files in the repo, those changes were not made by hand.

terraform/modules/tf_stream_alert/iam.tf
jacknagz: That resource isn't in this module, so that interpolation isn't possible.  I'd have to pass it in as a variable to this module instead, which I can do!

chunyong-lin: You are right. Thanks for the explanation.

jacknagz: I'm actually going to leave it, since we do the same thing for Athena data streams

PR #476, opened 2017-11-14
rule - cloudtrail_put_object_acl_public

mime-frame:
size: small

## Background

Sync latest version of rule that looks for use of PutObjectAcl that creates public objects.

## Changes

* Works for logs generated via CLI or API

## Testing

```
```


mime-frame:
@jacknagz I cool feature we could ship in the future is to make a helper that abstracts CTW from CWE so that way you can just write one rule

PR #480, opened 2017-11-14
[apps] adding simple read me for box dependencies

ryandeivert:
to: @javuto 
size: small
resolves N/A

## Background

The `boxsdk` requirement needs to be natively compiled on an EC2 instance for the time being. This is simple walkthrough on how to accomplish that if the dependencies need updated in the future (before this stop gap is removed).

## Changes

* Adding README.rst to demonstrate how to natively compile the boxsdk package on an EC2 instance.

## Testing

N/A

### DIFF
--- a/app_integrations/apps/README.rst
+++ b/app_integrations/apps/README.rst
@@ -1,0 +1,63 @@
+How to update boxsdk dependencies
+#################################
+
+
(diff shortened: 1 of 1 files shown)


ryandeivert:
@javuto  lol @ 🍞  😋

PR #484, opened 2017-11-15
[requirements] update requirements and add script for updating

### LINKED ISSUE #439: Add "requirements_top_level.txt"
## Background

The `requirements.txt` file now lists 79 dependencies, many of which are out of date (for example, boto3). This has two downsides:

- Upgrading the requirements list is not trivial because an upgrade may add or remove dependencies as well.
- It's hard to see, at a glance, what high-level packages are actually used (boto3, sphinx, etc)

## Desired Change

Add a separate `requirements_top_level.txt` which stores only the top-level dependencies. Then, updating `requirements.txt` is much easier:

```bash
pip install -r requirements_top_level.txt --upgrade
pip freeze > requirements.txt
```

BinaryAlert uses this model to great success, and it was inspired by [this blog post](http://www.jfabre.com/how-not-to-to-pip-requirements).

jacknagz:
to: @ryandeivert or @austinbyers 
size: small
resolves #439

## Background

As @austinbyers pointed out in the relevant issue, many deps were out of date and it was unclear which were top level.

## Changes

* Add a top level requirements 
* Add a script to manage and update reqs
* Update the pylint script so I know it has started

## Testing

Ran all the testing scripts locally


austinbyers:
Thanks for making the change! A few comments:

1. I would recommend specifying the version of every native package you use in `requirements_top_level.txt`. For example, specify which version of `cryptography` you are using so that the virtualenv is consistent with the natively compiled package we provide.
2. `virtualenv` need not be in the requirements. You need virtualenv installed _before_ you `pip install` the requirements
3. I'm a bit confused by the `requirements.txt` - why is there a comment halfway down? It should just be the output of `pip freeze`

I can make these changes on Friday if they are agreeable

jacknagz:
> I would recommend specifying the version of every native package you use in requirements_top_level.txt. For example, specify which version of cryptography you are using so that the virtualenv is consistent with the natively compiled package we provide.

Definitely, feel free to make this change.

> virtualenv need not be in the requirements. You need virtualenv installed before you pip install the requirements

That's fair, feel free to make this change too.

> I'm a bit confused by the requirements.txt - why is there a comment halfway down? Why not just make it the output of pip freeze?

This happens automatically via the script that was checked in (inspired from the link you provided in the issue)

PR #488, opened 2017-11-16
[cli][tests] adding mocking for firehose output

ryandeivert:
to: @javuto 
size: small
resolves N/A

## Background

Rule tests that were run with `python manage.py lambda test -p all` would fail for any rules that were using firehose outputs with an error similar to:
```
ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the PutRecord operation: Stream airbnb_streamalert_alert_delivery under account 123456789012 not found.
```

## Changes

* Adding mocking support for AWS firehose output

## Testing

Added a firehose output for rules and ran tests successfully.


javuto:
Thanks for fixing this!

PR #491, opened 2017-11-18
[Threat Intel] Threat Intel Downloader implementation

chunyong-lin:
to: @jacknagz (terraform module, streamalert CLI) @ryandeivert (Lambda function code in `stream_alert/threat_intel_downloader/`)
size: large
(Please don't review commits individually, I fixed many thing during testing stage 😄 )

## Background
**Threat Intel Downloader** is a Lambda function triggered by CloudWatch Event periodically to retrieve IOCs ([Indicators of Compromise](https://en.wikipedia.org/wiki/Indicator_of_compromise)) from threat feeds which support RESTful API calls, and it stores IOCs in DynamoDB table where StreamAlert rule processor will refer to while doing threat detection. 

### Supported features
* Retrieve IOCs from threat feeds which support RESTful API calls. And update IOCs periodically, once a day by default. It is scheduled by CloudWatch Event. 
* Store retrieved IOCs in DynamoDB table.
* API credentials are encrypted and store in AWS parameter store. 

### Deploy Threat Intel Downloader
* It is required to initialized StreamAlert before enabling Threat Intel Downloader. Please following [StreamAlert documentation](https://streamalert.readthedocs.io/deployment.html) to initialize StreamAlert. 
* After StreamAlert initialization, you can enable Threat Intel Downloader feature using following command
```
python manage.py threat_intel_downloader enable --interval 'rate(1 day)' --timeout 120 --memory 128
``` 
  * This command will do two things
    1. Generate settings for Threat Intel Downloader lambda function and write them into `conf/lambda.json` file.
    2. Ask you to provide API key and username to fetch IOCs from threat feed which supports RESTful API calls. And store them in AWS Parameter Store with encryption.
```
Please supply API key to retrieve IOCs via API calls. This should be a string of 40 alphanumeric characters.: 0123456789abcdefghijklmnopqrstuvwxyzabcd

Please supply API username to retrieve IOCs via API calls. This should be an email address.: ti-testing@foo.bar
```
  * Deploy Threat Intel Downloader to your infrastructure with following one line command. When the command finishes, you have Threat Intel Downloader Lambda function running once a day and retrieve IOCs from threat feed. 
```
python manage.py lambda deploy --processor threat_intel_downloader
```

### Lambda function
* A new Lambda function `Threat Intel Downloader` which is running independently from Rule and Alert processors.
* This lambda function is triggered by CloudWatch Event initially and it will invoke itself if there are more IOCs need to be retrieved when lambda function approaching to timeout. 
* This lambda function will send filtered IOCs to DynamoDB table. StreamAlert rule processors will refer to this table when it is doing threat detection (this feature is coming soon).

### Terraform module
* Add a `Threat Intel Downloader` module to create
  * IAM role
  * Lambda function
  * DynamoDB table
  * CloudWatch Event
  * CloudWatch Log Group

### StreamAlert CLI
* Add new CLI code to support enabling Threat Intel Downloader feature via command line.
* Add new CLI code to generate Threat Intel Downloader configuration and terraform settings.

## Testing
* Unit test was passed
```
./tests/scripts/unit_tests.sh
```
* Tested in AWS test account and it was passed.


terraform/modules/tf_threat_intel_downloader/main.tf
```diff
@@ -0,0 +1,76 @@
...
+    target_arn = "arn:aws:sns:${var.region}:${var.account_id}:${var.monitoring_sns_topic}"
+  }
+
+  tags {
+    Name = "ThreatIntel"
+  }
+}
+
+// Lambda Alias: Threat Intel Downloader Production
+resource "aws_lambda_alias" "threat_intel_downloader_production" {
```
jacknagz: you can just name this resource `production`

terraform/modules/tf_threat_intel_downloader/main.tf
```diff
@@ -0,0 +1,76 @@
...
+  name             = "production"
+  description      = "Production Threat Intel Dowwnloader Alias"
+  function_name    = "${aws_lambda_function.threat_intel_downloader.arn}"
+  function_version = "${var.current_version}"
+}
+
+// Lambda Permission: Allow Cloudwatch Scheduled Events to invoke Lambda
+resource "aws_lambda_permission" "allow_cloudwatch_events_invocation" {
+  statement_id  = "CloudwatchEventsInvokeAthenaRefresh"
```
jacknagz: update this SID

terraform/modules/tf_threat_intel_downloader/main.tf
```diff
@@ -0,0 +1,76 @@
...
+  qualifier     = "production"
+
+  # explicit dependency
+  depends_on = ["aws_lambda_alias.threat_intel_downloader_production"]
+}
+
+// Cloudwatch Event Rule: Invoke the threat_intel_downloader once a day
+resource "aws_cloudwatch_event_rule" "invoke_threat_intel_downloader" {
+  name        = "invoke_threat_intel_downloader"
+  description = "Invoke the Threat Intel Downloader Lambda function once a day"
```
jacknagz: couldn't it be arbitrarily invoked? maybe just make the description more generic - `Invoke the Threat Intel Downloader Lambda function periodically`

terraform/modules/tf_threat_intel_downloader/main.tf
```diff
@@ -0,0 +1,76 @@
...
+  arn  = "${aws_lambda_function.threat_intel_downloader.arn}:production"
+
+  # explicit dependency
+  depends_on = ["aws_lambda_alias.threat_intel_downloader_production"]
+}
+
+// Log Retention Policy: lambda function
+resource "aws_cloudwatch_log_group" "threat_intel_downloader" {
+  name              = "/aws/lambda/${var.prefix}_streamalert_threat_intel_downloader"
+  retention_in_days = 60
```
jacknagz: make this a variable

terraform/modules/tf_threat_intel_downloader/main.tf
```diff
@@ -0,0 +1,76 @@
...
+}
+
+# TODO: check if `aws_cloudwatch_log_metric_filter` needed.
```
jacknagz: leftover comments?

terraform/modules/tf_threat_intel_downloader/variables.tf
```diff
@@ -0,0 +1,72 @@
+variable "region" {}
+
+variable "account_id" {}
+
+variable "lambda_handler" {
+  type    = "string"
+  default = "main.handler"
```
jacknagz: this seems invalid considering how we package lambda functions, I'd just nix it

terraform/modules/tf_threat_intel_downloader/dynamodb.tf
```diff
@@ -0,0 +1,26 @@
...
+    type = "S"
+  }
+
+  ttl {
+    attribute_name = "expiration_date"
+    enabled        = true
+  }
+
+  tags {
+    Name = "ThreatIntel"
```
jacknagz: should this be StreamAlert so it's picked up by billing?

stream_alert_cli/threat_intel_downloader.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # Save these to the parameter store
+    param_name = 'threat_intel_downloader_api_creds'
+    saved = save_parameter(region, param_name, creds_dict, description, overwrite)
+    if saved:
+        LOGGER_CLI.info('Threat Intel Downloader credentials were successfully saved to parameter store.')
+    else:
+        LOGGER_CLI.error('Threat Intel Downloader credentials were not saved to parameter store.')
+
+    return saved
```
jacknagz: is anything storing this value?

stream_alert/threat_intel_downloader/threat_stream.py
```diff
@@ -0,0 +1,260 @@
...
+                - First object is a list of intelligence.
+                - Second object is a string of next token to retrieve more IOCs.
+                - Third object is bool to indicated if retrieve more IOCs from
+                    threat feed.
+                    Return False if next token is empty or threshold of number
+                    of IOCs is reached.
+        """
+        retry, retry_num, continue_invoke = True, self._MAX_RETRY, False
+        intelligence = list()
+        while retry:
```
ryandeivert: I think a good amount of this complex logic can be handled using the `backoff` library via a few stacked decorators. Maybe @jacknagz would be willing to offer some pointers here?

stream_alert/threat_intel_downloader/threat_stream.py
```diff
@@ -0,0 +1,260 @@
...
+        try:
+            ssm = boto3.client('ssm', self.region)
+            response = ssm.get_parameters(
+                Names=[self._PARAMETER_NAME], WithDecryption=True
+            )
+            for cred in response['Parameters']:
+                if cred['Name'] == self._PARAMETER_NAME:
+                    try:
+                        decoded_creds = json.loads(cred['Value'])
+                        self.api_user = decoded_creds['api_user']
```
ryandeivert: again - only put the items that may result in a `ValueError` in the try block. (move the `self.api_user = decoded_creds['api_user']` and `self.api_key = decoded_creds['api_key']` outside of this)

manage.py
```diff
@@ -1157,6 +1156,101 @@ def _add_athena_subparser(subparsers):
...
+                              See the link in the Resources section below for more information.
+        --table_name       The DynamoDB table name where stores IOCs. (required)
+        --table_rcu        The DynamoDB table Read Capacity Unit. (optional)
+        --table_wcu        The DynamoDB table Write Capacity Unit. (optional)
+
+    manage.py threat_intel_downloader update-auth   Update API credentials to parameter store.
+
+Examples:
+
+    manage.py threat_intel_downloader enable \\
```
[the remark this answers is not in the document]
chunyong-lin: I want to make enabling Threat Intel Downloader feature as simple as I can. One commandline is easier for to use. And this command is just for generating lambda function settings and writing settings to lambda.json file.

stream_alert_cli/terraform/threat_intel_downloader.py
```diff
@@ -0,0 +1,57 @@
...
+def generate_threat_intel_downloader(config):
+    """Generate Threat Intel Downloader Terrafrom
+
+    Args:
+        config (dict): The loaded config from the 'conf/' directory
+
+    Returns:
+        dict: Athena dict to be marshalled to JSON
+    """
+    # Use the monitoring topic as a dead letter queue
```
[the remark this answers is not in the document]
chunyong-lin: I adopted this pattern from StreamAlert App. Using Dead Letter Queue, we can indicate to Lambda that unprocessed events should be sent to SNS topic in our case. Then it will trigger CloudWatch alarm.

stream_alert_cli/threat_intel_downloader.py
```diff
@@ -0,0 +1,101 @@
...
+from stream_alert_cli.logger import LOGGER_CLI
+
+def threat_intel_downloader_handler(options, config):
+    """Configure Threat Intel Downloader from command line
+
+    Args:
+        options (namedtuple): The parsed args passed from the CLI
+        config (CLIConfig): Loaded StreamAlert CLI
+    """
+    def _validate_options(options):
```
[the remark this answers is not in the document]
chunyong-lin: Yes, it is enforced at the parser level. This method is imported and used in `stream_alert_cli/runner.py` where doing command parsing.

terraform/modules/tf_threat_intel_downloader/dynamodb.tf
chunyong-lin: Hmmm, I will change it to `StreamAlert`. The reason I chose `ThreatIntel` as tag is because we may want to see how much does this standalone lambda function (Threat Intel Downloader) cost. Actually, we can easily get sense since this lambda function mainly use DynamoDB and it is the only feature using DynamoDB in StreamAlert.

manage.py
```diff
@@ -1157,6 +1156,114 @@ def _add_athena_subparser(subparsers):
...
+    ti_downloader_parser.set_defaults(command='threat_intel_downloader')
+
+    ti_downloader_parser.add_argument(
+        'subcommand',
+        choices=['enable',
+                 'update-auth'],
+        help=ARGPARSE_SUPPRESS
+    )
+
+    ti_downloader_parser.add_argument(
```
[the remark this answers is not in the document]
chunyong-lin: Thanks for the suggestion. The helper functions are very useful!

manage.py
```diff
@@ -1157,6 +1156,114 @@ def _add_athena_subparser(subparsers):
...
+        default=['expiration_ts', 'itype', 'source', 'type', 'value']
+    )
+
+    ti_downloader_parser.add_argument(
+        '--ioc_filters',
+        help=ARGPARSE_SUPPRESS,
+        default=['crowdstrike', '@airbnb.com']
+    )
+
+    ti_downloader_parser.add_argument(
```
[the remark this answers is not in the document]
chunyong-lin: Unfortunately, the name of IOC types will be different from different threat feeds. We can not fix IOC types.

stream_alert_cli/config.py
```diff
@@ -125,6 +125,8 @@ def set_prefix(self, prefix):
             'rule_processor_config']['source_bucket'].replace('PREFIX_GOES_HERE', prefix)
         self.config['lambda']['stream_alert_apps_config']['source_bucket'] = self.config['lambda'][
             'stream_alert_apps_config']['source_bucket'].replace('PREFIX_GOES_HERE', prefix)
+        self.config['lambda']['threat_intel_downloader_config']['source_bucket'] = self.config['lambda'][
```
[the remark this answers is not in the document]
ryandeivert: @jacknagz that replace only happens if it exists, so what's wrong with it being here?

stream_alert_cli/terraform/threat_intel_downloader.py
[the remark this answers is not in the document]
ryandeivert: @jacknagz what do you mean by 'how are they handled?' The DLQ SNS topic will receive the messages from the lambda if there is a failure, and will do whatever actions are linked to that sns topic.

terraform/modules/tf_threat_intel_downloader/dynamodb.tf
ryandeivert: @chunyong-lin you can use multiple tags here if you'd like to keep threatintel as well

manage.py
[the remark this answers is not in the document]
ryandeivert: If we move to this type of pattern, we may as well have a base config that has _all_ possible lambda configs in it and just have an enable/disable flag in them (so we wouldn't be 'adding' a default config to it if it's **enabled** at some point and just would be disabled by default)

stream_alert/threat_intel_downloader/threat_stream.py
chunyong-lin: Thanks! I looked `backoff` library last night, it is very useful. I will include it in my next commit.

stream_alert/threat_intel_downloader/threat_stream.py
```diff
@@ -0,0 +1,260 @@
...
+                        {
+                            'value': 'malicious_domain.com',
+                            'itype': 'c2_domain',
+                            'source': 'crowdstrike',
+                            'type': 'domain',
+                            'expiration_ts': 1513658718,
+                        }
+                    ]
+        """
+        def _epoch_time(time_str):
```
ryandeivert: making this is `@staticmethod` instead of a nested function would make testing this much easier. I'd avoid nested functions if possible. Currently there is no way to reference this function from outside of its containing function, so testing it properly becoming inherently difficult.

stream_alert/threat_intel_downloader/threat_stream.py
```diff
@@ -0,0 +1,260 @@
...
+                time_str (str): expiration time in string format
+                    Example: '2017-12-19T04:45:18.412Z'
+
+            Returns:
+                (int): Epoch time. If no expiration time presented, return to
+                    default value which is current time + 90 days.
+            """
+
+            if not time_str:
+                return int((datetime.now() + timedelta(90) - datetime(1970, 1, 1)).total_seconds())
```
ryandeivert: I think the preferred way for the latter part of this (`datetime(1970, 1, 1)`) would be `datetime.utcfromtimestamp(0)`.

Can you also add the `days=90` keyword arg to the `timedelta(90)` call, since timedelta can also take `seconds=`, `minutes=` - it just helps with readability without having to reference the docstring above.

stream_alert/threat_intel_downloader/threat_stream.py
```diff
@@ -0,0 +1,260 @@
...
+
+            try:
+                utc_time = datetime.strptime(time_str, "%Y-%m-%dT%H:%M:%S.%fZ")
+                return int((utc_time - datetime(1970, 1, 1)).total_seconds())
+            except ValueError:
+                LOGGER.error('Can not convert expiration date \'%s\' to epoch time',
```
ryandeivert: `Can not` --> `Cannot`

stream_alert/threat_intel_downloader/main.py
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+def load_config():
+    """Load the Threat Intel Downloader configuration from conf/lambda.json file
+
+    Returns:
+        (dict): Configuration for Threat Intel Downloader
+
+    Raises:
+        ConfigError: For invalid or missing configuration files.
```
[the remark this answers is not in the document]
chunyong-lin: Good catch! Thanks!

stream_alert/threat_intel_downloader/threat_stream.py
```diff
@@ -0,0 +1,260 @@
...
+                - First object is a list of intelligence.
+                - Second object is a string of next token to retrieve more IOCs.
+                - Third object is bool to indicated if retrieve more IOCs from
+                    threat feed.
+        """
+        if not event:
+            return None, None, False
+
+        next_url = None
+        if 'next_url' not in event:
```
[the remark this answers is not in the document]
chunyong-lin: Your 🚀  
Your understanding is correct. The `event` may contain `next_url` key but with empty value. It is just special handling in following lines.

chunyong-lin:
@jacknagz @ryandeivert PTAL
I have addressed your comments. Thank you so much, they are really helpful 😄 
Mainly changes are,
* Combine methods to generate lambda settings for `Athena` and `Threat Intel Downloader`
* Use existing `MockSSMClient` class during testing
* Promote helpers for validating arguments to be globally accessed
* Use `backoff` library to handle `requests` retries

I noticed, the coverage is dropped by 1.2%. I will add more test cases.

stream_alert_cli/helpers.py
```diff
@@ -30,10 +30,56 @@
...
 
+from app_integrations.config import AWS_RATE_RE
 from stream_alert_cli.logger import LOGGER_CLI
 from stream_alert_cli.terraform._common import enabled_firehose_logs
 
 
+class ParseError(Exception):
+    """Custom expection rasied when argument parser has error"""
+
+def validate_scheduled_interval(val):
```
ryandeivert: As discussed offline, can we actually do away with the helpers and just duplicate this code (sigh) in the arg parsers? I don't love that this will actually raise an exception and print the traceback, instead of acting like the normal argparse error that prints the usage and the error (without printing a traceback). I'm sorry :(

tests/unit/app_integrations/test_helpers.py
```diff
@@ -24,8 +24,9 @@
 class MockSSMClient(object):
     """Helper mock class to act as the ssm boto3 client"""
 
-    def __init__(self, suppress_params=False, app_type=''):
-        self._parameters = dict()
+    def __init__(self, suppress_params=False, app_type='', parameters=None):
+        #self._parameters = dict()
+        self._parameters = parameters if parameters else dict()
```
ryandeivert: you can just do `self._parameters = parameters or dict()`

stream_alert/threat_intel_downloader/threat_stream.py
```diff
@@ -0,0 +1,273 @@
...
+                    continue_invoke = True
+            elif https_req.status_code == 401:
+                raise ThreatStreamRequestsError('Response status code 401, unauthorized.')
+            elif https_req.status_code == 500:
+                raise ThreatStreamRequestsError('Response status code 500, retry now.')
+            else:
+                raise ThreatStreamRequestsError('Unknown status code {}, '
+                                                'do not retry.'.format(https_req.status_code)
+                                               )
+        except (requests.exceptions.Timeout,
```
ryandeivert: hey @chunyong-lin, I think when using `backoff`, you shouldn't wrap the code in a try/except block like this. You should be able to get rid of this and achieve what you're trying to do.

stream_alert/threat_intel_downloader/threat_stream.py
```diff
@@ -0,0 +1,273 @@
...
+            days (int): default expiration days which 90 days from now
+
+        Returns:
+            (int): Epoch time. If no expiration time presented, return to
+                default value which is current time + 90 days.
+        """
+
+        if not time_str:
+            return int((datetime.now()
+                        + timedelta(days)
```
ryandeivert: this is interesting formatting. pylint doesn't complain about this 🤔 ?

I'd suggest putting the `+` and `-` on the preceding line as such:
```
return int((datetime.now() +
            timedelta(days) -
            datetime.utcfromtimestamp(0)).total_seconds())
```

stream_alert_cli/helpers.py
chunyong-lin: :( Yeah, I will make change. I am agree that it is ugly to see the traceback when there is argument parse error. So we have to make a compromise that to use clean build-in argument error handling (have to duplicate code) over custom exception which will print out traceback.

stream_alert/threat_intel_downloader/threat_stream.py
chunyong-lin: [PEP8](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) recommends to break lines before operants, but actually before and after operants are both okay if we have style consistently.

stream_alert/threat_intel_downloader/threat_stream.py
chunyong-lin: Ha, good catch! I will remove `try..except`.

stream_alert/threat_intel_downloader/threat_stream.py
ryandeivert: Weird 🤔

terraform/modules/tf_threat_intel_downloader/iam.tf
```diff
@@ -0,0 +1,144 @@
...
+data "aws_iam_policy_document" "invoke_lambda_function" {
+  statement {
+    effect = "Allow"
+
+    actions = [
+      "lambda:InvokeFunction",
+    ]
+
+    resources = [
+      "arn:aws:lambda:${var.region}:${var.account_id}:function:${var.prefix}_streamalert_threat_intel_downloader",
```
jacknagz: I would feed this in via variables vs interpolation

terraform/modules/tf_threat_intel_downloader/iam.tf
```diff
@@ -0,0 +1,144 @@
...
+
+    actions = [
+      "dynamodb:BatchGetItem",
+      "dynamodb:GetItem",
+      "dynamodb:BatchWriteItem",
+      "dynamodb:PutItem",
+    ]
+
+    resources = [
+      "arn:aws:dynamodb:${var.region}:${var.account_id}:table/${var.prefix}_streamalert_threat_intel_downloader",
```
jacknagz: can this be interpolated?

terraform/modules/tf_threat_intel_downloader/iam.tf
```diff
@@ -0,0 +1,144 @@
...
+    ]
+  }
+}
+
+// IAM role policy: Allow lambda function to read from parameter store
+resource "aws_iam_role_policy" "read_api_creds_from_ssm" {
+  name   = "read_ssm"
```
jacknagz: I'd make this camel case, like the others:

`SSMGetThreatIntelParms`

terraform/modules/tf_threat_intel_downloader/iam.tf
jacknagz: you can use outputs from the main `tf_stream_alert` module

stream_alert/threat_intel_downloader/threat_stream.py
```diff
@@ -0,0 +1,273 @@
...
+        if not (self.api_user and self.api_key):
+            LOGGER.error('API Creds Error')
+            raise ThreatStreamCredsError('API Creds Error')
+
+    exceptions_to_backoff = (requests.exceptions.Timeout,
+                             requests.exceptions.ConnectionError,
+                             requests.exceptions.ChunkedEncodingError,
+                             ThreatStreamRequestsError)
+    @backoff.on_exception(backoff.constant,
+                          exceptions_to_backoff,
```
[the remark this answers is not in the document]
chunyong-lin: I thought about using `expo`, but I don't want to add too much timeout. I can change to `expo` if constant doesn't work well. Using constant with 3 times retries, the max timeout will be 3 seconds. If use `expo`, it will take `1+2+4 = 7` seconds.

stream_alert_cli/config.py
```diff
@@ -431,6 +438,55 @@ def add_app_integration(self, app_info):
...
+
+        Args:
+            ti_downloader_info (dict): Settings for Threat Intel Downloader Lambda
+                function, generated from commandline
+                "manage.py threat_intel_downloader enable"
+
+        Returns:
+            (bool): Return True if writing settings of Lambda function successfully.
+        """
+        def _generate_default_ti_downloader_configure():
```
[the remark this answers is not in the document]
chunyong-lin: It is weird that I use function here 😄  Don't know what I was doing 😂 
Will change to  variable.

terraform/modules/tf_threat_intel_downloader/iam.tf
chunyong-lin: I am following the same style to interpolate resources across streamalert terraform modules, example in [streamalert iam tf file](https://github.com/airbnb/streamalert/blob/master/terraform/modules/tf_stream_alert/iam.tf#L48)

terraform/modules/tf_threat_intel_downloader/iam.tf
chunyong-lin: Can you have more content? Do you mean interpolated in different way?

chunyong-lin:
@jacknagz PTAL. I addressed some of your comments, thanks! Also, I may need more more context about two of your comments. Please see above.

stream_alert/threat_intel_downloader/threat_stream.py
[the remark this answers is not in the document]
jacknagz: expo is actually not linear like that, but sounds good!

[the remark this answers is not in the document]
jacknagz: there's jitter/randomization involved

PR #493, opened 2017-11-20
[output] PagerDuty incident requires "From:" header

javuto:
to: @ryandeivert @jacknagz 
size: small

## Background

New output to use PagerDuty Incidents was added [here](https://github.com/airbnb/streamalert/pull/467). The REST API requires the header `From: validpagerdutyuser@email.com` to create a new incident.

## Changes

* The header `From: ` is added to authentication, and the user will be added when creating new output.
* Modified method `_check_exists` to optionally return just a boolean.
* All tests have been modified to accommodate this change.

## Testing

```
...
TOTAL                                                2516     76    97%
Ran 462 tests in 7.015s

OK

$ ./tests/scripts/pylint.sh

...
```

### DIFF
--- a/stream_alert/alert_processor/outputs.py
+++ b/stream_alert/alert_processor/outputs.py
@@ -249,7 +249,11 @@ class PagerDutyIncidentOutput(StreamOutputBase):
                             mask_input=True,
                             cred_requirement=True)),
             ('escalation_policy',
-             OutputProperty(description='the name of the default escalation policy'))
+             OutputProperty(description='the name of the default escalation policy')),
+            ('email_from',
+             OutputProperty(description='valid user email from the PagerDuty '
+                                        'account linked to the token',
+                            cred_requirement=True))
         ])
 
     @staticmethod
@@ -265,7 +269,7 @@ class PagerDutyIncidentOutput(StreamOutputBase):
         """
         return os.path.join(base_url, endpoint)
 
-    def _check_exists_get_id(self, filter_str, url, target_key):
+    def _check_exists(self, filter_str, url, target_key, get_id=True):
         """Generic method to run a search in the PagerDuty REST API and return the id
         of the first occurence from the results.
 
@@ -273,10 +277,11 @@ class PagerDutyIncidentOutput(StreamOutputBase):
             filter_str (str): The query filter to search for in the API
             url (str): The url to send the requests to in the API
             target_key (str): The key to extract in the returned results
+            get_id (boolean): Whether to generate a dict with result and reference
 
         Returns:
             str: ID of the targeted element that matches the provided filter or
-                 False if a matching element does not exists.
+                 True/False whether a matching element exists or not.
         """
         params = {
             'query': '{}'.format(filter_str)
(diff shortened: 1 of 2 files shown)


stream_alert/alert_processor/outputs.py
```diff
@@ -340,32 +344,34 @@ def _service_verify(self, service):
...
             item_type (str): Type of item reference to be returned
+            get_id (boolean): Whether to generate a dict with result and reference
 
         Returns:
             dict: JSON object be used in the API call, containing the item id
-                  and the item reference, or False if it fails
+                  and the item reference, True if it just exists or False if it fails
         """
-        item_id = self._check_exists_get_id(item_str, item_url, item_key)
+        item_url = self._get_endpoint(self._base_url, item_key)
```
ryandeivert: good change, putting this in one place!

stream_alert/alert_processor/outputs.py
```diff
@@ -340,32 +344,34 @@ def _service_verify(self, service):
...
         if not item_id:
             LOGGER.info('%s not found in %s, %s', item_str, item_key, self.__service__)
             return False
 
-        return {
-            'id': item_id,
-            'type': item_type
-        }
+        if get_id:
+            return {
```
ryandeivert: put this dict on one line:

`return {id': item_id, 'type': item_type}`

stream_alert/alert_processor/outputs.py
```diff
@@ -407,14 +413,22 @@ def dispatch(self, **kwargs):
...
 
-        # Cache base_url
-        self._base_url = creds['api']
+        # Get user email to be added as From header and verify
+        user_email = creds['email_from']
+        if not self._user_verify(user_email, False):
+            return self._log_status(False)
+
+        # Add From to the headers after verifying
+        self._headers['From'] = user_email
```
ryandeivert: is this header value really title case (`From`)?

stream_alert/alert_processor/outputs.py
```diff
@@ -291,21 +296,23 @@ def _check_exists_get_id(self, filter_str, url, target_key):
...
 
         # If there are results, get the first occurence from the list
-        return response[target_key][0]['id'] if target_key in response else False
+        if get_id:
+            return response[target_key][0]['id'] if target_key in response else False
+
+        return True
 
-    def _user_verify(self, user):
+    def _user_verify(self, user, get_id=True):
```
ryandeivert: It looks like the only time this function is called, the default value for `get_id` is overridden with `False`... am I missing something? When would `get_id` ever be `True` here?

stream_alert/alert_processor/outputs.py
javuto: It seems so: https://v2.developer.pagerduty.com/docs/incident-creation-api

stream_alert/alert_processor/outputs.py
javuto: In `_incident_assignment` it gets called and the default `True` value is used

stream_alert/alert_processor/outputs.py
ryandeivert: weird 🤣

PR #497, opened 2017-11-27
S3 Events Prefix/Suffix + Athena Bug Fixes

jacknagz:
to: @chunyong-lin 
size: medium

## Background

When configuring S3 event notifications on CloudTrail buckets, having the ability to filter based on a S3 key prefix allows us to route processing to certain StreamAlert clusters.  This is especially useful when multiple accounts send CloudTrail into the same S3 Bucket.

## Changes

* Support `filter_prefix` and `filter_suffix` Terraform args to the s3 event notification resource
* Fix a couple bugs in Athena table creation

## Testing

* Verified in a test AWS account

### DIFF
--- a/stream_alert_cli/athena/handler.py
+++ b/stream_alert_cli/athena/handler.py
@@ -29,7 +29,7 @@ SCHEMA_TYPE_MAPPING = {
     dict: 'map<string, string>',
     list: 'array<string>'
 }
-
+MAX_QUERY_LENGTH = 262144
 
 def create_database(athena_client):
     """Create the 'streamalert' Athena database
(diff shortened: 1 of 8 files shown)


stream_alert_cli/athena/handler.py
```diff
@@ -94,6 +94,13 @@ def rebuild_partitions(athena_client, options, config):
         new_partitions_statement = athena_helpers.partition_statement(
             unique_partitions, options.bucket, options.table_name)
 
+        # Make sure our new alter table statement is within the query API limits
+        if len(new_partitions_statement) > MAX_QUERY_LENGTH:
+            LOGGER_CLI.error('Partition statement too large, writing to local file')
+            with open('partitions_{}.txt'.format(options.table_name), 'w') as partition_file:
```
chunyong-lin: Do you think about log the `new_partitions_statement` to CloudWatch, instead of writing to a local file?

stream_alert_cli/athena/handler.py
```diff
@@ -249,6 +256,7 @@ def create_table(athena_client, options, config):
             'CREATE EXTERNAL TABLE {table_name} ({schema}) '
             'PARTITIONED BY (dt string) '
             'ROW FORMAT SERDE \'org.openx.data.jsonserde.JsonSerDe\' '
+            'WITH SERDEPROPERTIES ( \'ignore.malformed.json\' = \'true\') '
```
chunyong-lin: Minor nit: additional space before `\'ignore.malformed.json\' = \'true\'`, but no space after it.

stream_alert_cli/athena/handler.py
jacknagz: this doesn't run in Lambda

chunyong-lin: Oh, I see. How about logging to LOGGER_CLI?

PR #498, opened 2017-11-27
[output] Adding support for assign priority to incidents

javuto:
to: @ryandeivert @jacknagz 
size: small

## Background

New output to use PagerDuty Incidents was added [here](https://github.com/airbnb/streamalert/pull/467). Incidents will be created with a low priority by default, unless specified.

## Changes

* Added support to specify a valid priority name in the context.
* Added testing code.

## Testing

```
...
TOTAL                                                  2573     73    97%
Ran 470 tests in 8.373s

OK

$ ./tests/scripts/pylint.sh

...
```

### DIFF
--- a/stream_alert/alert_processor/outputs/pagerduty.py
+++ b/stream_alert/alert_processor/outputs/pagerduty.py
@@ -176,6 +176,7 @@ class PagerDutyIncidentOutput(OutputDispatcher):
     USERS_ENDPOINT = 'users'
     POLICIES_ENDPOINT = 'escalation_policies'
     SERVICES_ENDPOINT = 'services'
+    PRIORITIES_ENDPOINT = 'priorities'
 
     def __init__(self, *args, **kwargs):
         OutputDispatcher.__init__(self, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -401,6 +437,14 @@ def dispatch(self, **kwargs):
         if rule_context:
             rule_context = rule_context.get(self.__service__, {})
 
+        # Use the priority provided in the context, use it or the incident will be low priority
+        incident_priority = {}
+        _priority = rule_context.get('incident_priority', False)
+        if _priority:
+            verified_priority = self._priority_verify(_priority)
```
ryandeivert: Suggestion: pass the rule_context to the `_priority_verify` function and have it do all of this, and instead of having `_priority_verify` return `False` have it return the default value of `{}` or `dict()`..

So your call would just look something like:
`incident_priority = self._priority_verify(rule_context)`

And everywhere above that returns false would instead return `dict()`

stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -401,6 +441,11 @@ def dispatch(self, **kwargs):
         if rule_context:
             rule_context = rule_context.get(self.__service__, {})
 
+        # Use the priority provided in the context, use it or the incident will be low priority
+        incident_priority = {}
+        if rule_context:
```
ryandeivert: one last suggestion - don't even create the `incident_priority` here or do the `if rule_context` check here.. do this all in the `_priority_verify` and return `dict()` immediately if the rule_context == None

Issue #363, opened 2017-10-09
Improvement: Output Backoff


jacknagz:
## Background

Currently, if any API calls to specified outputs fail, the Lambda function will retry 3 times before giving up.  The issue is that if an output is being rate limited, there is no backoff logic to ensure the alert makes it through. 

## Desired Change

Implement backoff in alerting output logic.

jacknagz:
@javuto I believe we have resolved this, feel free to close (and tagging the relevant PRs if you wouldn't mind) 😄

PR #499, opened 2017-11-30
[output] Backoff support for all PagerDuty related outputs

javuto:
to: @ryandeivert @jacknagz 
size: small

## Background

There were some issues with the PagerDuty multiple APIs that was making StreamAlert to lose alerts becase the API access was being rate limited. With these changes the requests sent to the API will retry when a problem is detected, that way we can effectively mitigate the encountered rate limits, due to high volume alerts. 

## Changes

* Added changes to the base output class to add retry methods for both GET and POST requests. This way the support for backoff/retry can be progressively added in each single output without breaking existing code.
* Added specific support in PagerDuty outputs to use the backoff/retry code.
* Modified existing tests to support new approach.

## Testing

```
...
TOTAL                                                  2617     75    97%
Ran 470 tests in 13.873s

OK

$ ./tests/scripts/pylint.sh

...
```

### DIFF
--- a/stream_alert/alert_processor/outputs/output_base.py
+++ b/stream_alert/alert_processor/outputs/output_base.py
@@ -21,10 +21,16 @@ import tempfile
 import requests
 import urllib3
 
+import backoff
 import boto3
 from botocore.exceptions import ClientError
 
 from stream_alert.alert_processor import LOGGER
+from stream_alert.shared.backoff_handlers import (
+    backoff_handler,
+    success_handler,
+    giveup_handler
+)
 
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 OutputProperty = namedtuple('OutputProperty',
@@ -35,6 +41,23 @@ OutputProperty.__new__.__defaults__ = ('', '', {' ', ':'}, False, False)
 class OutputRequestFailure(Exception):
     """OutputRequestFailure handles any HTTP failures"""
 
+
+def retry_on_exception(exceptions):
+    """Decorator function to attempt retry based on passed exceptions"""
+    print exceptions
+    def real_decorator(func):
+        """Actual decorator to retry on exceptions"""
+        @backoff.on_exception(backoff.fibo,
+                              exceptions, # This is a tuple with exceptions
+                              max_tries=OutputDispatcher.MAX_RETRY_ATTEMPTS,
+                              jitter=backoff.full_jitter,
+                              on_backoff=backoff_handler,
+                              on_success=success_handler,
+                              on_giveup=giveup_handler)
+        def wrapper(*args, **kwargs):
+            func(*args, **kwargs)
+        return wrapper
+    return real_decorator
 
 class StreamAlertOutput(object):
     """Class to be used as a decorator to register all OutputDispatcher subclasses"""
@@ -109,6 +132,9 @@ class OutputDispatcher(object):
     """
     __metaclass__ = ABCMeta
     __service__ = NotImplemented
+
+    # How many times it will attempt to retry something failing using backoff
+    MAX_RETRY_ATTEMPTS = 3
 
     # _DEFAULT_REQUEST_TIMEOUT indicates how long the requests library will wait before timing
     # out for both get and post requests. This applies to both connection and read timeouts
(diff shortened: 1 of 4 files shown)


tests/unit/stream_alert_alert_processor/test_outputs/test_pagerduty.py
```diff
@@ -599,8 +600,6 @@ def test_dispatch_success_bad_policy(self, get_mock, post_mock, log_mock):
                                               rule_name='rule_name',
                                               alert=get_alert(context=ctx)))
 
-        log_mock.assert_called_with('Successfully sent alert to %s', self.SERVICE)
```
ryandeivert: I think you should be able to leave this test still - can you confirm?

tests/unit/stream_alert_alert_processor/test_outputs/test_pagerduty.py
```diff
@@ -574,12 +576,11 @@ def test_dispatch_failure_bad_everything(self, get_mock, post_mock, log_mock):
                                                rule_name='rule_name',
                                                alert=get_alert()))
 
-        log_mock.assert_called_with('Failed to send alert to %s', self.SERVICE)
```
ryandeivert: Same with this one

tests/unit/stream_alert_alert_processor/test_outputs/test_pagerduty.py
```diff
@@ -251,6 +250,8 @@ def test_policy_verify_success_default(self, get_mock):
         assert_equal(policy_verified['id'], 'good_policy_id')
         assert_equal(policy_verified['type'], 'escalation_policy_reference')
 
+    @patch('stream_alert.alert_processor.outputs.output_base.OutputDispatcher.MAX_RETRY_ATTEMPTS',
```
ryandeivert: Can you instead patch this on the entire `TestPagerDutyIncidentOutput` instead of individual methods? :)

tests/unit/stream_alert_alert_processor/test_outputs/test_pagerduty.py
javuto: Yes, just confirmed it. Thanks!

stream_alert/alert_processor/outputs/output_base.py
```diff
@@ -252,6 +295,32 @@ def _get_request(cls, url, params=None, headers=None, verify=True):
...
+            headers (dict): Dictionary containing request-specific header parameters
+            verify (bool): Whether or not the server's SSL certificate should be verified
+        Returns:
+            dict: Contains the http response object
+        Raises:
+            OutputRequestFailure
+        """
+        @retry_on_exception(cls._catch_exceptions())
+        def do_get_request():
+            """Decorator to perform the request with retry/backoff"""
```
ryandeivert: Can you update this docstring? This technically isn't a `decorator` and is a decorated nested function. Same with the other similar function below.

stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -259,13 +264,12 @@ def _check_exists(self, filter_str, url, target_key, get_id=True):
...
-        resp = self._get_request(url, params, self._headers, False)
-
-        if not self._check_http_response(resp):
-            return False
-
-        response = resp.json()
-        if not response:
+        try:
+            resp = self._get_request_retry(url, params, self._headers, False)
+            response = resp.json()
```
ryandeivert: can we do this outside of the try/except block? and just have the block be:
```
try:
    resp = self._get_request_retry(url, params, self._headers, False)
except OutputRequestFailure:
    return False
```
I think there a few different places that could be changed to move this outside of the block

tests/unit/stream_alert_alert_processor/test_outputs/test_pagerduty.py
```diff
@@ -84,7 +84,6 @@ def test_dispatch_bad_descriptor(self, log_mock):
 
         log_mock.assert_called_with('Failed to send alert to %s', self.SERVICE)
 
-
```
ryandeivert: where dis line go? 🤔

javuto:
Implements this https://github.com/airbnb/streamalert/issues/363

PR #501, opened 2017-12-03
[apps] utilizing delegation email for gsuite app

ryandeivert:
to: @javuto 
size: small
resolves N/A

## Background

During testing of the GSuite streamalert app, it was found that the proper permissions were not being used by the app, resulting in the following error:
`HttpError: <HttpError 401 when requesting https://www.googleapis.com/admin/reports/v1/activity/users/all/applications/admin?alt=json returned "Access denied. You are not authorized to read activity records.">`

## Changes

* The GSuite app now uses a service account email to perform [domain-wide auth delegation](https://developers.google.com/admin-sdk/directory/v1/guides/delegation#delegate_domain-wide_authority_to_your_service_account).
  * This email address must be provided during app configuration and will be save the parameter store along with the keyfile's data.
* Fixing bug where the incorrect last timestamp would be saved to the state.
  * The Google Admin SDK Reports API returns all logs in reverse chronological order (contrary to every other API I've seen or we support). Therefore, the _first_ log in the list of logs returned will have the most recent date and the _last_ log returned (after all pagination, etc) will be the oldest log.
  * The ideal fix for this would make sure ALL logs were retrieved before saving the timestamp, since this can now result in missed data if there are any errors. However, that's a significant refactor and we can limit the impact of this by setting a high AWS Lambda timeout and a low CloudWatch Scheduled Events interval.
* Adding additional debug log statements

### Other Changes
* Adding a debug logger statement to the box app so it's more clear when requests are happening.

## Testing

* Unit tests passing (`tests/scripts/unit_tests.sh`)
* Tested in scripted form locally.

### DIFF
--- a/app_integrations/apps/gsuite.py
+++ b/app_integrations/apps/gsuite.py
@@ -173,7 +173,7 @@ class GSuiteReportsApp(AppIntegration):
             'delegation_email':
                 {
                     'description': 'the service account user email to delegate access to',
-                    'format': re.compile(r'^[A-Za-z0-9-_.]+@[A-Za-z0-9-.]+\.[A-Za-z]{2,}$')
+                    'format': re.compile(r'^[A-Za-z0-9-_.+]+@[A-Za-z0-9-.]+\.[A-Za-z]{2,}$')
                 }
             }
 
--- a/app_integrations/apps/gsuite.py
+++ b/app_integrations/apps/gsuite.py
@@ -14,8 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 import json
+import re
 
-from apiclient import discovery, errors
+import apiclient
 from oauth2client.service_account import ServiceAccountCredentials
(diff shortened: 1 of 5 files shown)


app_integrations/apps/gsuite.py
```diff
@@ -156,6 +169,11 @@ def keyfile_validator(keyfile):
                     'description': ('the path on disk to the JSON formatted Google '
                                     'service account private key file'),
                     'format': keyfile_validator
+                },
+            'delegation_email':
+                {
+                    'description': 'the service account user email to delegate access to',
+                    'format': re.compile(r'^[A-Za-z0-9-_.]+@[A-Za-z0-9-.]+\.[A-Za-z]{2,}$')
```
javuto: Maybe change this to allow also `+`, so emails like `foo+bar@foobar.com` works. It should be just adding the character to the first grouping.

PR #507, opened 2017-12-05
Threat Intel Beta

chunyong-lin:
to: @jacknagz @ryandeivert 
size: large
resolves N/A

## Background
* The implementation of the [Threat Intelligence (Alpha)](https://github.com/airbnb/streamalert/pull/322) has a limitation that it requires to bump rule processor memory in each cluster if the Threat Intelligence is enabled. Because it reads all IOCs (Indicators of Compromise) into memory. When users have millions of IOCs, it will consume more than 1G memory easily. 
* We identified the solution and split it into two parts to solve this problem. For **Part1**, we introduced [Threat Intel Downloader](https://github.com/airbnb/streamalert/pull/491), a separate lambda function to fetch IOCs from threat feed that supports RESTful API, and store IOCs in a DynamoDB table. So it doesn't require to read all IOCs into memory in Rule Processor.
* This PR is **Part2**, implement new threat detection logic in StreamAlert Rule Processor. 
  * To enable Threat Intel, add or update threat intel setting in `conf/global.json`
  ```
  "threat_intel": {
      "enabled": true,
      "dynamodb_table": "PREFIX_GOES_HERE_streamalert_threat_intel_downloader"
    }
  ```
  * Rule Processor queries DynamoDB IOC table when it is doing threat detection, instead of looking up from memory. 
  * If there are any malicious IOC detected in the record, IOC information will be inserted to the record. 
  * Thus, user is able to write a rule alert on any suspicious IOCs by checking if `streamalert:ioc` key inserted to the record or not.

## Changes
* Improve the way to config Data Normalization and Threat Intel setting.
  * More Threat Intel Enabled flag into `conf/global.json`.
  * Combine normalized CEF types with IOCs in `conf/types.json`.
* Create a new `StreamIoc` class to store IOC information extract from records which have been normalized.
* Add new methods in `StreamThreatIntel` class.
* Remove Threat Intel Alpha implementation and related test cases.
* Grant rule processor IAM role to have permission to read DynamoDB IOC table in Terraform code.

## Testing
* Unit test
```
./tests/scripts/unit_tests.sh
...
TOTAL                                                    2935     45    98%
Ran 492 tests in 10.721s

OK
```
* Integration test
```
...
```
* AWS test
  * Testing in AWS testing account was passed.


conf/global.json
```diff
@@ -15,5 +15,9 @@
     "tfstate_bucket": "PREFIX_GOES_HERE.streamalert.terraform.state",
     "tfstate_s3_key": "stream_alert_state/terraform.tfstate",
     "tfvars": "terraform.tfvars"
+  },
```
jacknagz: We should leave optional conf flags out of the default config

jacknagz: They should be enabled via the CLI, which would then add them into this file

stream_alert/rule_processor/threat_intel.py
```diff
@@ -17,15 +17,372 @@
...
+        if self.IOC_KEY in rec:
+            if (ioc_type in rec[self.IOC_KEY] and
+                    ioc_value not in rec[self.IOC_KEY][ioc_type]):
+                rec[self.IOC_KEY][ioc_type].append(ioc_value)
+            else:
+                rec[self.IOC_KEY][ioc_type] = [ioc_value]
+        else:
+            rec.update({self.IOC_KEY: {ioc_type: [ioc_value]}})
+
+    def _extract_values(self, record):
```
jacknagz: Can this name be more descriptive?

stream_alert/rule_processor/threat_intel.py
```diff
@@ -17,15 +17,372 @@
...
     IOC_KEY = 'streamalert:ioc'
 
     __intelligence = {}
     __config = {}
 
+    # Class variable stores Data Normalization types mapping.
+    __normalized_types = {}
+
+    # Class variable stores mapping between CEF normalized types and IOC types
+    __normalized_types2ioc_types = {}
```
jacknagz: Could it be `__normalized_ioc_types_mapping` maybe?

stream_alert/rule_processor/threat_intel.py
```diff
@@ -17,15 +17,372 @@
...
+# Build boto3 DynamoDB resource at import time so it can be cached between invocations.
+DYNAMODB = boto3.client('dynamodb')
+
+class StreamIoc(object):
+    """Class to store IOC info"""
+    def __init__(self, **kwargs):
+        """Initialize StreamIoc instance and store useful information
+
+        Keyword arguments:
+            - value (str): IOC value
```
jacknagz: nit: you don't need the `-`in front of each arg 😄

stream_alert/rule_processor/threat_intel.py
```diff
@@ -17,15 +17,372 @@
...
+        Returns:
+            (list): List of subset of StreamIoc instances
+        """
+        result = []
+        end = len(ioc_collections)
+        for index in range(0, end, MAX_QUERY_CNT):
+            result.append(ioc_collections[index:min(index+MAX_QUERY_CNT, end)])
+        return result
+
+    exceptions_to_backoff = (ClientError)
```
jacknagz: this could be a constant in the top of the file

stream_alert/rule_processor/threat_intel.py
```diff
@@ -17,15 +17,372 @@
...
+    @backoff.on_exception(backoff.expo,
```
jacknagz: 💯

stream_alert/rule_processor/threat_intel.py
```diff
@@ -17,15 +17,372 @@
...
+                          exceptions_to_backoff,
+                          max_tries=3,
+                          on_backoff=backoff_handler,
+                          on_success=success_handler,
+                          on_giveup=giveup_handler)
+    def _query(self, values):
```
jacknagz: I would simplify this method to only have the `batch_get_item` request, and handle unprocessed keys by feeding them back into this function as input.  The reason is that if for whatever reason **either** `batch_get_item` request fails, the entire request will retry with all `values`, which could be time consuming

stream_alert/rule_processor/rules_engine.py
```diff
@@ -374,24 +376,46 @@ def process(cls, input_payload):
...
+            if ioc_records:
+                for ioc_record in ioc_records:
+                    for rule in rules:
+                        cls.rule_analysis(ioc_record, rule, payload, alerts)
 
         return alerts
+
+    @classmethod
+    def rule_analysis(cls, record, rule, payload, alerts):
+        """Static method to analyze rule
```
jacknagz: nit: I think it's ok to leave out `static method` from here and just describe the main use case of the function 😄

stream_alert/rule_processor/threat_intel.py
```diff
@@ -94,6 +94,13 @@ def associated_record(self, rec):
         """Associate the original record to an IOC instance"""
         self._associated_record = rec
 
+
+def exceptions_to_giveup(err):
```
jacknagz: oh wow that's awesome, I had no idea you could do that

stream_alert/rule_processor/threat_intel.py
chunyong-lin: Unfortunately, I can not make it as a constant. Pylint complaint
```
Starting pylint script
************* Module stream_alert.rule_processor.threat_intel
C: 36, 0: Invalid class name "EXCEPTIONS_TO_BACKOFF" (invalid-name)
```
After digging pylint's error, looks pylint is able to recognize `ClientError` is a class, so `EXCEPTIONS_TO_BACKOFF` will be a class. The [regular expression](http://pylint-messages.wikidot.com/messages:c0103) of class name is different from constant. 
```
Constant: (([A-Z_][A-Z0-9_]*)|(__.*__))$
Class: [A-Z_][a-zA-Z0-9]+$
```
Translate to human readable version is that Class variable is started with Capital letter or underscore, but no underscore in the middle. 
I don't think it is necessary to make change to this part.

stream_alert/rule_processor/threat_intel.py
```diff
@@ -13,90 +13,389 @@
...
+
+    exceptions_to_backoff = (ClientError)
+    @backoff.on_exception(backoff.expo,
+                          exceptions_to_backoff,
+                          max_tries=3,
+                          giveup=exceptions_to_giveup,
+                          on_backoff=backoff_handler,
+                          on_success=success_handler,
+                          on_giveup=giveup_handler)
+    def _query(self, values):
```
jacknagz: I would simplify this method to only have the batch_get_item request, and handle unprocessed keys by feeding them back into this function as input. The reason is that if for whatever reason either batch_get_item request fails, the entire request will retry with all values, which could be time consuming

stream_alert/rule_processor/rules_engine.py
```diff
@@ -374,24 +376,52 @@ def process(cls, input_payload):
...
-                        'outputs': rule.outputs,
-                        'source_service': payload.service(),
-                        'source_entity': payload.entity,
-                        'context': rule.context}
-                    alerts.append(alert)
+                cls.rule_analysis(record_copy, rule, payload, alerts)
+
+        # Apply the Threat Intelligence against normalized records
+        if StreamThreatIntel.enabled():
+            threat_intel = StreamThreatIntel()
```
ryandeivert: Since the `boto3` dynamo client is created in the init of this `StreamThreatIntel` class, consider moving this outside of the `get_rules` function that could be called N number of times where N = the number of logs processed in one run.

For instance, this could be created and cached in the `__init__` of this `StreamRules` class instead, as:

```
def __init__(self):
    self._threat_intel = StreamThreatIntel() if StreamThreadIntel.enabled() else None
```

stream_alert/rule_processor/threat_intel.py
```diff
@@ -13,90 +13,389 @@
...
+        """Get the value of an IOC instance"""
+        return self._value
+
+    @value.setter
+    def value(self, value):
+        """Set the value of an IOC instance"""
+        self._value = value
+
+    @property
+    def ioc_type(self):
```
ryandeivert: Not that I'm not proud of you for the pythonic use of `@property` 😄 .. the getter/setter approach is most useful if you must transform the data being set/got before setting/returning it (ie: transforming a string, adding 100 to a base integer, etc). In this case, you're actually just adding unnecessary complexity. You could just have the instance properties declared as public, avoid all of the extra code, and have the same affect:

```
def __init__(self, **kwargs):
    self.value = kwargs.get('value', None)
    self.ioc_type = kwargs.get('ioc_type', None)
    self.associated_record = kwargs.get('rec', None)
    self.is_ioc = kwargs.get('is_ioc', False)
```

stream_alert/rule_processor/threat_intel.py
```diff
@@ -13,90 +13,389 @@
...
+        end = len(ioc_collections)
+        for index in range(0, end, MAX_QUERY_CNT):
+            result.append(ioc_collections[index:min(index+MAX_QUERY_CNT, end)])
+        return result
+
+    exceptions_to_backoff = (ClientError)
+    @backoff.on_exception(backoff.expo,
+                          exceptions_to_backoff,
+                          max_tries=3,
+                          giveup=exceptions_to_giveup,
```
ryandeivert: you could use a `lambda` here like `lambda err: err.response['Error']['Code'] == 'AccessDeniedException'`

stream_alert/rule_processor/handler.py
```diff
@@ -92,7 +92,9 @@ def __init__(self, context, enable_alert_processor=True):
 
         # Firehose client initialization
         self.firehose_client = None
-        StreamThreatIntel.load_intelligence(self.config)
+
+        # Load Threat Intel config
+        StreamThreatIntel.load_config(self.config)
```
ryandeivert: Hey @chunyong-lin let's create an instance of the `StreamRules` class that gets cached on the `StreamAlert` class as an instance property. And, within the `StreamRules` initializer, you load an instance of this `StreamThreatIntel` class from the config.

stream_alert/rule_processor/threat_intel.py
```diff
@@ -13,90 +13,389 @@
...
+                    if isinstance(original_keys, list):
+                        for original_key in original_keys:
+                            value = value[original_key]
+                    ioc_values.add(value)
+        return [StreamIoc(value=value, ioc_type=ioc_type, rec=record) for value in ioc_values]
 
     @classmethod
-    def read_compressed_files(cls, intel_dir, delimiter=','):
-        """Read intelligence into memory
+    def load_config(cls, config):
```
ryandeivert: Can you make this method return an instance of the `StreamThreatIntel` class (or `None` if it's not enabled). This would allow you to pass the region, table, etc to the `__init__` instead of dynamically swapping out the class's protected properties (`__enabled`, `__region`, `__table`) in this method. Doing this is bad practice and should be avoided.

stream_alert/rule_processor/threat_intel.py
ryandeivert: @chunyong-lin this error is because you're accidentally just making `exceptions_to_backoff` reference `ClientError` when what you're really tryin to do is create a tuple with one item of `ClientError` within it. To accomplish this, add a trailing comma to this like so:

`exceptions_to_backoff = (ClientError,)`

or as a constant: `EXCEPTIONS_TO_BACKOFF = (ClientError,)`

stream_alert/rule_processor/threat_intel.py
chunyong-lin: @jacknagz  Ideally, we should have a while loop to process unprocessed keys. But I am worry about the lambda timeout in Rule Processor. That's why I only re-query against `UnprocessedKeys` once now. We definitely can improve this in the near future. 
What do you think?

chunyong-lin:
@jacknagz @ryandeivert PTAL, 🙇‍♀️

manage.py
```diff
@@ -1332,11 +1374,19 @@ def build_parser():
...
-    manage.py output           Configure new StreamAlert outputs
-    manage.py lambda           Deploy, test, and rollback StreamAlert AWS Lambda functions
-    manage.py live-test        Send alerts to configured outputs
-    manage.py configure        Configure StreamAlert settings
+    manage.py app                        Create, list, or update a StreamAlert app integration function
+    manage.py athena                     Configure Athena for StreamAlert
+    manage.py configure                  Configure StreamAlert settings
+    manage.py create-alarm               Add a CloudWatch alarm for predefined metrics
+    manage.py kinesis                    Configure Kinesis for StreamAlert
+    manage.py lambda                     Deploy, test, and rollback StreamAlert AWS Lambda functions
```
ryandeivert: Thanks for updating this block :) This was super outdated!

chunyong-lin:
@jacknagz @ryandeivert This PR is ready for another reviews. Thank you for all the feedback so far. They are very helpful.

tests/unit/stream_alert_rule_processor/test_threat_intel.py
```diff
@@ -271,25 +289,27 @@ def test_segment(self):
...
+    @patch('boto3.client')
+    def test_query_with_exception(self, mock_client):
         """Threat Intel - Test DynamoDB query method with exception"""
-        mock_dynamodb.return_value = MockDynamoDBClient.response(exception=True)
+        mock_client('dynamodb').batch_get_item.return_value = \
+            MockDynamoDBClient.response(exception=True)
 
         self.threat_intel._query(['1.1.1.2'])
-        assert_equal(mock_dynamodb.call_count, 3)
+        assert_equal(mock_client.call_count, 3)
```
ryandeivert: hey @chunyong-lin  my only comment is related to this test - I actually don't think this assert is ever occurring and this test will not be hit. Since the `threat_intel._query(['1.1.1.2'])` call is what raises the exception, this line will never be touched. You can verify this by adding a `print ...` before this assertion.

chunyong-lin: Hi @ryandeivert You are absolutely right! No, this assertion doesn't take effect 😞 
Thank you for flag this out!

stream_alert/rule_processor/threat_intel.py
```diff
@@ -13,90 +13,348 @@
...
+        """
+        result = []
+        end = len(ioc_collections)
+        for index in range(0, end, MAX_QUERY_CNT):
+            result.append(ioc_collections[index:min(index+MAX_QUERY_CNT, end)])
+        return result
+
+    @backoff.on_exception(backoff.expo,
+                          EXCEPTIONS_TO_BACKOFF,
+                          max_tries=3,
```
ryandeivert: hey @chunyong-lin I was just thinking, this `max_tries` value should probably be stored as a constant somewhere (on the class is fine). This will help us greatly with patching this value in testing, as seen here: https://github.com/airbnb/streamalert/pull/514/files#diff-aaa5916e38545f10cd7b0e9f57953818R33

stream_alert/rule_processor/rules_engine.py
```diff
@@ -346,52 +349,113 @@ def process(cls, input_payload):
...
             return alerts
 
+        # store normalized records for future process in Threat Intel
+        normalized_records = []
+
         for record in payload.records:
             for rule in rules:
                 # subkey check
-                has_sub_keys = cls.process_subkeys(record, payload.type, rule)
+                has_sub_keys = StreamRules.process_subkeys(record, payload.type, rule)
```
jacknagz: why call `StreamRules` vs `self`? there's some inconsistency in the subsequent blocks below

stream_alert/rule_processor/rules_engine.py
```diff
@@ -346,52 +349,113 @@ def process(cls, input_payload):
...
+                'outputs': rule.outputs,
+                'source_service': payload.service(),
+                'source_entity': payload.entity,
+                'context': rule.context}
+
+            alerts.append(alert)
+
+    @staticmethod
+    def check_alerts_duplication(record, rule, alerts):
+        """Class method to check if the record has been added to alerts list already.
```
jacknagz: typo - it's a static method, but you can actually just omit that altogether

stream_alert/rule_processor/threat_intel.py
jacknagz: Well, you technically wouldn't query it once, because backoff is applying to both requests.   My original comment still applies in this situation

chunyong-lin: `backoff` will only be applied when there is `ClientError` exception. When the response includes `UnprocessedKeys`, it doesn't mean it returns error code. I would love to talk more offline.

stream_alert/rule_processor/rules_engine.py
chunyong-lin: Good catch! Thanks!

stream_alert/rule_processor/threat_intel.py
```diff
@@ -246,12 +246,22 @@ def _process_ioc(self, ioc_collections):
...
+        for subset in self._segment(ioc_collections):
             query_values = []
             for ioc in subset:
                 if ioc.value not in query_values:
                     query_values.append(ioc.value)
-            query_result = self._query(query_values)
+
+            query_result = []
+            has_unprocessed_key = True
+            while has_unprocessed_key:
```
jacknagz: We probably want to give a limit here so it doesn't run forever 😄 

As an example, we can also optionally stack decorators: https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/handler.py#L273-L301

In your experience, did you ever observe a subset of keys being unprocessed? or is this feature to be proactive?

If anything, you could just feed it back into `query` once, and then backoff on predicate or on exception .  We can chat more tomorrow if that's still unclear

PR #508, opened 2017-12-05
[tf] Remove sort key from DynamoDB table.

chunyong-lin:
to: @jacknagz 
size: small
resolves N/A

## Background
Remove sort key from DynamoDB table. It is not necessary to have sort key in DynamoDB table, since the primary key (hash key) will also be identical from each item in the table. The primary key in the DynamoDB table is the value of IOC. 

## Changes

* Remove `range_key` from terraform module, and it translates to `sort key` in DynamoDB table. 

## Testing
* Tested it in AWS testing account, and it was passed.


chunyong-lin:
@jacknagz Can you stamp on this PR when you get time? Thanks!

PR #511, opened 2017-12-06
[tf][cli] vpc flow log support clusters & cross-account delivery

jacknagz:
to: @austinbyers 
size: medium

## Background

The VPC flow log Terraform module currently only creates a subscription filter within a single account.  In order to support cross account subscription filters, some IAM permissions needed to change.

## Changes

* Refactor to use IAM policy doc data resources
* Add an explicit cloudwatch log destination resource which allows us to grant cross account access
* Add an option in the `flow_logs` stream alert cluster module which accepts a list of `cross_account_ids`

## Testing

Deployed and verified by using two test accounts

### DIFF
--- a/terraform/modules/tf_stream_alert_flow_logs/iam.tf
+++ b/terraform/modules/tf_stream_alert_flow_logs/iam.tf
@@ -1,98 +1,126 @@
 // Allow flow logs to write to CloudWatch
 // http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/flow-logs.html#flow-logs-iam
-
+// IAM Role: Clustered VPC Flow Log
 resource "aws_iam_role" "flow_log_role" {
-  name = "flow_log_role"
+  name = "stream_alert_${var.cluster}_flow_log_role"
 
-  assume_role_policy = <<EOF
-{
-  "Version": "2012-10-17",
-  "Statement": [
-    {
-      "Sid": "",
-      "Effect": "Allow",
-      "Principal": {
-        "Service": "vpc-flow-logs.amazonaws.com"
-      },
-      "Action": "sts:AssumeRole"
-    }
-  ]
-} 
-EOF
+  assume_role_policy = "${data.aws_iam_policy_document.flow_log_assume_role_policy.json}"
 }
 
+// IAM Policy Doc: AssumeRole for VPC Flow Logs
+data "aws_iam_policy_document" "flow_log_assume_role_policy" {
+  statement {
+    effect  = "Allow"
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "Service"
+      identifiers = ["vpc-flow-logs.amazonaws.com"]
+    }
+  }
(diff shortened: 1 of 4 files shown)


terraform/modules/tf_stream_alert_flow_logs/iam.tf
```diff
@@ -1,98 +1,126 @@
...
+      "logs:CreateLogStream",
+      "logs:PutLogEvents",
+      "logs:DescribeLogGroups",
+      "logs:DescribeLogStreams",
+    ]
+
+    resources = [
+      "*",
+    ]
+  }
```
austinbyers: Nit: Add `effect: Allow`. It is the default, but let's put it to be explicit and consistent

terraform/modules/tf_stream_alert_flow_logs/iam.tf
```diff
@@ -1,98 +1,126 @@
...
+
+// IAM Policy Doc: CloudWatch Put Events
+data "aws_iam_policy_document" "flow_log_put_cloudwatch_logs" {
+  statement {
+    actions = [
+      "logs:CreateLogGroup",
+      "logs:CreateLogStream",
+      "logs:PutLogEvents",
+      "logs:DescribeLogGroups",
+      "logs:DescribeLogStreams",
```
austinbyers: Nit: alphabetical action ordering (here and elsewhere)

terraform/modules/tf_stream_alert_flow_logs/iam.tf
```diff
@@ -1,98 +1,126 @@
...
+    actions = [
+      "kinesis:PutRecord*",
+      "kinesis:DescribeStream",
+      "kinesis:ListStreams",
+    ]
+
+    resources = [
+      "${var.destination_stream_arn}",
+    ]
+  }
```
austinbyers: Add the `effect: Allow` and alphabetical action ordering, as above

terraform/modules/tf_stream_alert_flow_logs/iam.tf
```diff
@@ -1,98 +1,126 @@
...
+}
+
+// IAM Policy Doc: Allow Cross Account Flow Logs
+data "aws_iam_policy_document" "cross_account_subscription_filter" {
+  count = "${length(var.cross_account_ids) > 0 ? 1 : 0}"
+
+  statement {
+    effect = "Allow"
```
austinbyers: Can we add a statement ID `SID` here which is very clear about what it does? E.g. `sid = "AllowCrossAccountLogSubscription" or something

jacknagz: There's a comment on the resource, is that still needed?

austinbyers: The `SID` is attached to the actual policy, which I think can be searched / show up in alerts, etc. I don't have a strong preference either way

PR #514, opened 2017-12-07
[output] Increasing backoff time between retries

javuto:
to: @ryandeivert @jacknagz 
size: small

## Background

Backoff was added to handle retries for outputs when something goes wrong. The current code uses `backoff.fibo` (Fibonacci), as method to define how long to wait between attempts. Also the maximum number of attempts was 3. For PagerDuty it was not making much of a difference since the requests were still rate limited, as the wait was not that long.

By moving to `backoff.expo` (Exponential), we will increase more the wait from attempt to the next one and also the maximum number to attempts will be 5.

Here is a comparison between the two approaches, with a quick script:
```
$ python compare.py
Exponential
1
2
4
8
16

Fibonacci
1
1
2
3
5
```

## Changes

* Increased the timeout for default clusters.
* Changed the backoff behavior from `backoff.fibo` to backoff.expo` to increase the wait between attempts.
* Increased the maximum number of retries.
* Patched maximum number of retries for all outputs to avoid unnecessary delays in tests.

## Testing

```
...
TOTAL                                                    2803     99    96%
Ran 494 tests in 9.749s

OK

$ ./tests/scripts/pylint.sh

...
```

### DIFF
--- a/conf/clusters/prod.json
+++ b/conf/clusters/prod.json
@@ -25,7 +25,7 @@
         "current_version": "$LATEST",
         "log_level": "info",
         "memory": 128,
-        "timeout": 10
+        "timeout": 60
       },
       "rule_processor": {
         "current_version": "$LATEST",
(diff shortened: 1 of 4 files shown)


javuto:
Related to https://github.com/airbnb/streamalert/issues/363

PR #515, opened 2017-12-07
[schema] updating  with opt keys

ryandeivert:
to: @chunyong-lin 
size: small
resolves N/A

## Background

Observed failed parse due to some optional top level keys in the `carbonblack:watchlist.hit.binary` schema.

## Changes

* Adding some optional keys to this schema.

## Testing

* Tested on locally on test event that is not committed here.

### DIFF
--- a/conf/logs.json
+++ b/conf/logs.json
@@ -537,6 +537,18 @@
   },
   "carbonblack:watchlist.hit.binary": {
     "schema": {
+      "alliance_data_srsthreat": [],
+      "alliance_data_srstrust": [],
+      "alliance_data_virustotal": [],
+      "alliance_link_srstrust": "string",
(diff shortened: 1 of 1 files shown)


conf/logs.json
```diff
@@ -580,7 +584,16 @@
         "type": "string",
         "watchlist_id": "integer",
         "watchlist_name": "string"
-      }
+      },
+      "optional_top_level_keys": [
+        "alliance_data_srsthreat",
```
mime-frame: id add all of the other alliance_ keys that exist in other ones too - i imagine theyll pop up

ryandeivert: good point @mime-frame I've added all options

ryandeivert:
@chunyong-lin PTAL when you have a second

PR #516, opened 2017-12-07
[data-normalization] adding gsuite ipAddress to data normalization/threat intel

ryandeivert:
to: @mime-frame 
size: small
resolves N/A

## Background

G Suite Admin SDK Report logs have an `ipAddress` field that maps to the source address of where an event took place.

## Changes

* Adding the `sourceAddress` --> `ipAddress` mapping to the `types.json` file so we can get threat intelligence alerting on gsuite logs.

### DIFF
--- a/conf/types.json
+++ b/conf/types.json
@@ -144,6 +144,11 @@
       "user"
     ]
   },
+  "gsuite": {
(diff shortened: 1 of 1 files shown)


mime-frame:
LGTM

PR #520, opened 2017-12-11
Deploy by Cluster, Batch Size Param, and More

### LINKED ISSUE #434: CLI Bug: Deploy continues if S3 upload fails
## Background

While deploying, if your source bucket is incorrectly set, Terraform will still attempt to run even though the new deployment package did not make it to S3.

## Desired Change

Break out of the deployment process and log an error if the S3 upload fails

jacknagz:
to: @ryandeivert 
size: medium
resolves #434 

## Background

When building new clusters, or deploying in the event of an infrastructure based issue, having the ability to deploy by cluster greatly decreases the time necessary to make changes.  This PR focuses on refactoring to make this possible, and also adds features for supporting batch size, and more.

## Changes

* Add the ability to deploy and publish versions by cluster
* Support the `all` option in the deployment process for optional Lambdas
* Do not deploy if S3 package upload fails
* `yapf` on `stream_alert_cli/config.py`
* Update unit tests accordingly

## Testing

* Deploy the rule processor
* Deploy all (enabled) processors
* Deploy by cluster
* Terraform build specific cluster

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1002,6 +1002,8 @@ Available Options:
     --target                                   The Terraform module name to apply.
                                                Valid options: stream_alert, kinesis, kinesis_events,
                                                cloudtrail, monitoring, and s3_events.
+    --cluster                                  The StreamAlert cluster(s) to apply to.
+
 Examples:
 
     manage.py terraform init
@@ -1038,6 +1040,11 @@ Examples:
             'athena', 'cloudwatch_monitoring', 'cloudtrail', 'flow_logs', 'kinesis',
             'kinesis_events', 'stream_alert', 's3_events', 'threat_intel_downloader'
         ],
+        help=ARGPARSE_SUPPRESS,
+        nargs='+')
+
+    tf_parser.add_argument(
+        '--cluster',
         help=ARGPARSE_SUPPRESS,
         nargs='+')
 
--- a/manage.py
+++ b/manage.py
@@ -976,6 +976,11 @@ def _add_default_lambda_args(lambda_parser):
         action=UniqueSetAction,
         required=True)
 
+    lambda_parser.add_argument(
+        '--cluster',
+        help=ARGPARSE_SUPPRESS,
+        nargs='+')
+
     # Allow verbose output for the CLI with the --debug option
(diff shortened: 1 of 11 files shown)


manage.py
```diff
@@ -1041,6 +1043,11 @@ def _add_terraform_subparser(subparsers):
         help=ARGPARSE_SUPPRESS,
         nargs='+')
 
+    tf_parser.add_argument(
+        '--cluster',
+        help=ARGPARSE_SUPPRESS,
+        nargs='+')
```
ryandeivert: Recommendation: use the `action=UniqueSetAction` keyword arg to ensure this is always a set of unique clusters with no duplicates. I know you cast this to a `set` down the line (when you do `clusters = set(options.cluster or config.clusters())`), but using this action will avoid requiring that the caller knows this could contain dupes. Setting the `default=set()` is also a good idea.

stream_alert_cli/terraform/handler.py
```diff
@@ -151,40 +146,43 @@ def terraform_handler(options, config):
...
+        'threat_intel_downloader': 'module.threat_intel_downloader'
+    }
+    clusters = set(options.cluster or config.clusters())
+
+    if options.target:
+        tf_runner_targets.update({
+            'module.{}_{}'.format(target, cluster)
+            for cluster in clusters for target in options.target
+        })
+        for name, _ in custom_module_mapping.iteritems():
```
ryandeivert: if you don't need to use the value of in this for loop, don't use `.iteritems` and instead just do `for name in custom_module_mapping:...`

manage.py
```diff
@@ -1041,6 +1043,11 @@ def _add_terraform_subparser(subparsers):
         help=ARGPARSE_SUPPRESS,
         nargs='+')
 
+    tf_parser.add_argument(
+        '--cluster',
```
ryandeivert: consider making this flag plural (`--clusters`) since it can take 1 or more args

manage.py
```diff
@@ -976,6 +976,11 @@ def _add_default_lambda_args(lambda_parser):
         action=UniqueSetAction,
         required=True)
 
+    lambda_parser.add_argument(
+        '--cluster',
```
ryandeivert: see same comments related to tf_parser

PR #525, opened 2017-12-12
Assign Kinesis Users to a Group, and Make User Creation Optional

jacknagz:
to: @austinbyers 
size: small

## Background

It's general best practice to not directly assign user policies, and instead assign them to groups.  This change applies this best practice, and also creates the ability to optionally create the user.  In some instances, data is delivered to Lambda in other ways (CloudWatch events, S3 notifications, etc)

## Changes

* Create a clustered stream_alert group
* Assign Kinesis users to this group
* Assign the Kinesis WriteOnly policy to this group
* Add a variable to optionally create users/groups

## Testing

* Local CI testing
* Deployed the change on an existing cluster, no users or access keys were deleted
* Deployed the change with `"create_user": false`

### DIFF
--- a/docs/source/kinesis.rst
+++ b/docs/source/kinesis.rst
@@ -49,12 +49,13 @@ Kinesis Streams settings for each cluster:
 Options
 ~~~~~~~
 
-=============  =========  ===========
-Key            Required   Description
--------------  ---------  -----------
-``retention``  ``Yes``    The data record retention period of your stream.
-``shards``     ``Yes``    A shard provides a fixed unit of capacity to your stream.
-=============  =========  ===========
+===============    =========  ===========
+Key                Required   Description
+---------------    ---------  -----------
+``retention``      ``Yes``    The data record retention period of your stream.
+``shards``         ``Yes``    A shard provides a fixed unit of capacity to your stream.
+``create_user``    ``No``     A boolean to enable/disable user creation for the Kinesis stream.  This is true by default.
+===============    =========  ===========
 
 Scaling
 ~~~~~~~
--- a/terraform/modules/tf_stream_alert_kinesis_streams/iam.tf
+++ b/terraform/modules/tf_stream_alert_kinesis_streams/iam.tf
@@ -1,0 +1,57 @@
+// IAM User: stream_alert user for systems to send data to the stream
+resource "aws_iam_user" "stream_alert" {
+  count = "${var.create_user ? 1 : 0}"
+  name  = "${var.prefix}_${var.cluster_name}_stream_alert_user"
+  path  = "/streamalert/"
+}
+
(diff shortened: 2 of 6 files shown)


terraform/modules/tf_stream_alert_kinesis_streams/iam.tf
```diff
@@ -0,0 +1,57 @@
...
+
+  group = "${aws_iam_group.stream_alert.name}"
+}
+
+// IAM Group Policy: Allow users in the group to PutRecords to Kinesis
+resource "aws_iam_group_policy" "stream_alert_kinesis_put_records" {
+  count = "${var.create_user ? 1 : 0}"
+  name  = "KinesisPutRecords"
+  group = "${aws_iam_group.stream_alert.id}"
+
```
austinbyers: I don't think there should be a blank line here, did you `terraform fmt`?

terraform/modules/tf_stream_alert_kinesis_streams/iam.tf
```diff
@@ -0,0 +1,57 @@
...
+// IAM Policy Doc: allow the stream_alert user to write to the generated Kinesis Stream
+data "aws_iam_policy_document" "stream_alert_writeonly" {
+  count = "${var.create_user ? 1 : 0}"
+
+  statement {
+    actions = [
+      "kinesis:PutRecord*",
+      "kinesis:DescribeStream",
+      "kinesis:ListStreams",
```
austinbyers: Nit: alphabetical order

docs/source/kinesis.rst
```diff
@@ -49,12 +49,13 @@ Kinesis Streams settings for each cluster:
...
--------------  ---------  -----------
-``retention``  ``Yes``    The data record retention period of your stream.
-``shards``     ``Yes``    A shard provides a fixed unit of capacity to your stream.
-=============  =========  ===========
+===============    =========  ===========
+Key                Required   Description
+---------------    ---------  -----------
+``retention``      ``Yes``    The data record retention period of your stream.
+``shards``         ``Yes``    A shard provides a fixed unit of capacity to your stream.
+``create_user``    ``No``     A boolean to enable/disable user creation for the Kinesis stream.
```
austinbyers: Can you also mention that the default is `create_user=True`?

terraform/modules/tf_stream_alert_kinesis_streams/iam.tf
jacknagz: I guess the Atom plugin didn't catch this one, I think it was originally a docstring and that's why

docs/source/kinesis.rst
jacknagz: Sure, good call

PR #526, opened 2017-12-12
Rule context can be modified within rule

javuto:
to: @ryandeivert @jacknagz 
size: small

## Background

Context was added to rules to be able to extract information from the rule record and be utilized after (for example, outputs). This code makes that alteration possible and adds test for it.

## Changes

* If context is used, it will be passed as argument of the rule.
* Test to verify the added code.

## Testing

```
...
TOTAL                                                    2951     94    97%
Ran 501 tests in 8.074s

OK

$ ./tests/scripts/pylint.sh

...
```

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -284,7 +284,10 @@ class StreamRules(object):
             (bool): The return function of the rule
         """
         try:
-            rule_result = rule.rule_function(record)
+            if rule.context:
+                rule_result = rule.rule_function(record, rule.context)
+            else:
+                rule_result = rule.rule_function(record)
         except Exception:  # pylint: disable=broad-except
             rule_result = False
             LOGGER.exception(
--- a/tests/unit/stream_alert_rule_processor/test_rules_engine.py
+++ b/tests/unit/stream_alert_rule_processor/test_rules_engine.py
@@ -852,3 +852,41 @@ class TestStreamRules(object):
             payload = load_and_classify_payload(toggled_config, service, entity, raw_record)
 
             assert_equal(len(new_rules_engine.process(payload)), 1)
+
+    def test_rule_modify_context(self):
+        """Rules Engine - Testing Context Modification"""
+        @rule(logs=['test_log_type_json_nested_with_data'],
+              outputs=['s3:sample_bucket'],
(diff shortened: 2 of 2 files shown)


stream_alert/rule_processor/rules_engine.py
```diff
@@ -284,7 +284,10 @@ def process_rule(record, rule):
             (bool): The return function of the rule
         """
         try:
-            rule_result = rule.rule_function(record)
+            if rule.context:
+                rule_result = rule.rule_function(record, rule.context)
+            else:
+                rule_result = rule.rule_function(record)
         except Exception:  # pylint: disable=broad-except
```
ryandeivert: @javuto I know this isn't part of your changes, but can you get rid of this `except Exception:` line and just have this be `try: ... except: ... `. This would get rid of that pylint disable message too

tests/unit/stream_alert_rule_processor/test_rules_engine.py
```diff
@@ -852,3 +852,41 @@ def match_ipaddress(_): # pylint: disable=unused-variable
...
+
+        # prepare the payloads
+        service, entity = 'kinesis', 'test_kinesis_stream'
+        raw_record = make_kinesis_raw_record(entity, kinesis_data)
+        payload = load_and_classify_payload(self.config, service, entity, raw_record)
+
+        # process payloads
+        alerts = self.rules_engine.process(payload)
+
+        print alerts
```
ryandeivert: Rouge print statement should be removed

tests/unit/stream_alert_rule_processor/test_rules_engine.py
```diff
@@ -852,3 +852,41 @@ def match_ipaddress(_): # pylint: disable=unused-variable
             payload = load_and_classify_payload(toggled_config, service, entity, raw_record)
 
             assert_equal(len(new_rules_engine.process(payload)), 1)
+
+    def test_rule_modify_context(self):
+        """Rules Engine - Testing Context Modification"""
+        @rule(logs=['test_log_type_json_nested_with_data'],
+              outputs=['s3:sample_bucket'],
+              context={'assigned_user': 'not_set', 'assigned_policy': 'not_set2'})
+        def modify_context_test(rec, context): # pylint: disable=unused-variable
```
ryandeivert: you shouldn't need this `# pylint: disable=unused-variable` do you?

PR #543, opened 2017-12-15
Threat intel error handling

### LINKED ISSUE #539: [Threat Intel] Errors handling
## Background
Threat Intel Beta enables Rule Processor(s) to query a DynamoDB table where stores IOC data. In the implementation it doesn't have exception handling. 

### Description
This issue is open to track progress. We should handle/log exceptions
* boto3 ClientError
* DynamoDB ProvisionedThroughPutExceededException

chunyong-lin:
to: @ryandeivert or @jacknagz 
size: small
resolves #539

## Background

Threat Intel Beta enables Rule Processor(s) to query a DynamoDB table where stores IOC data. In the implementation it doesn't have exception handling. This PR is to add exception handling for 
* ClientError
* ParamValidationError

## Changes
* Add error handling when querying dynamodb table.
* Make MockDynamoDBClient great (again). It includes raising more exceptions now.
* Enforce lowercase IOC value from first place.
* Add more test cases to catch errors.
* Remove `autoscale` setting from `threat_intel` and it is only available in `threat_intel_downloader`.  This bug was caused by careless Copy&Paste ;(

## Testing
* Unit test
```
./tests/scripts/unit_tests.sh
...
Ran 510 tests in 8.852s

OK
```
* Rule test
```
...
```
* Tested it in AWS testing account, and it was passed.

### DIFF
--- a/stream_alert/rule_processor/threat_intel.py
+++ b/stream_alert/rule_processor/threat_intel.py
@@ -16,7 +16,7 @@ limitations under the License.
 import backoff
 import boto3
 from boto3.dynamodb.types import TypeDeserializer
-from botocore.exceptions import ClientError
+from botocore.exceptions import ClientError, ParamValidationError
 
 from stream_alert.shared import NORMALIZATION_KEY
 from stream_alert.shared.backoff_handlers import (
@@ -58,10 +58,13 @@ class StreamIoc(object):
         self.is_ioc = kwargs.get('is_ioc', False)
 
 def exceptions_to_giveup(err):
-    """Function to decide if giveup backoff or not.
-    Give up backoff retries if DynamoDB IOC table doesn't exist.
-    """
-    return err.response['Error']['Code'] == 'AccessDeniedException'
+    """Function to decide if giveup backoff or not."""
+    error_code = {
+        'AccessDeniedException',
+        'ProvisionedThroughputExceededException',
+        'ResourceNotFoundException'
+    }
+    return err.response['Error']['Code'] in error_code
 
 class StreamThreatIntel(object):
     """Load intelligence from csv.gz files into a dictionary."""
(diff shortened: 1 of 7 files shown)


stream_alert/rule_processor/threat_intel.py
```diff
@@ -251,24 +254,40 @@ def _process_ioc(self, ioc_collections):
...
 
             query_result = []
 
-            result, unprocesed_keys = self._query(query_values)
-            query_result.extend(result)
+            query_error_msg = 'An error occured while quering dynamodb table. Error is: %s'
+            try:
+                result, unprocesed_keys = self._query(query_values)
+                query_result.extend(result)
+            except ClientError as err:
```
jacknagz: why handle the different errors separately? you can catch both on a single line as a tuple:

`except (Error1, Error2) as err`

chunyong-lin: It is because error message is different. `ClientError` has `response` attribute, `ParamValidationError` doesn't. It will be useful if we can log error details. So I separate them to be able to log details.

ryandeivert: Suggestion on this: use `LOGGER.exception` instead of `error` so we can see the full stack & exception value(s).

stream_alert/rule_processor/threat_intel.py
```diff
@@ -251,24 +254,40 @@ def _process_ioc(self, ioc_collections):
...
                 deserializer = self._deserialize(unprocesed_keys[self._table]['Keys'])
                 query_values = [elem[PRIMARY_KEY] for elem in deserializer]
-                result, _ = self._query(query_values)
-                query_result.extend(result)
+                query_error_msg = 'An error occured while processing unprocesed_keys. Error is: %s'
+                try:
+                    result, _ = self._query(query_values)
+                    query_result.extend(result)
+                except ClientError as err:
+                    LOGGER.error(query_error_msg, err.response)
```
ryandeivert: See other comment about using `.exception(..)`

PR #544, opened 2017-12-16
[Threat Intel] Cast IOC value to string

chunyong-lin:
to: @jacknagz or @ryandeivert 
size: small
resolves N/A

## Background

Context for the change

## Changes
* Cast IOC value to string always.
* Also use `LOGGER.error` to avoid confusion while searching crash/exception in CloudWatch.

## Testing
* Unit test
```
./tests/scripts/unit_tests.sh
...
TOTAL                                                    3001    124    96%
Ran 510 tests in 10.134s

OK
```

* Rule test
```
...
```

### DIFF
--- a/stream_alert/rule_processor/threat_intel.py
+++ b/stream_alert/rule_processor/threat_intel.py
@@ -161,7 +161,7 @@ class StreamThreatIntel(object):
                             value = value[original_key]
                     if value:
                         ioc_values.add(value)
-        return [StreamIoc(value=value.lower(), ioc_type=ioc_type, associated_record=record)
+        return [StreamIoc(value=str(value).lower(), ioc_type=ioc_type, associated_record=record)
                 for value in ioc_values]
 
     @classmethod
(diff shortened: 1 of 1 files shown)


stream_alert/rule_processor/threat_intel.py
```diff
@@ -305,7 +305,7 @@ def _segment(ioc_collections):
         """
         result = []
         end = len(ioc_collections)
-        LOGGER.info('[Threat Inel] Rule Processor queries %d IOCs', end)
+        LOGGER.debug('[Threat Inel] Rule Processor queries %d IOCs', end)
```
jacknagz: Can we move this debug statement before the caller to `_query`? I'm not sure it provides a lot of value in this function since

chunyong-lin: It will be same thing. `_process_ioc` is the caller to both `_segment` and `_query`. 
This debug info will be helpful when infrastructure is in a small scale, or does testing in AWS testing account.

PR #545, opened 2017-12-20
Fix read capacity overload issue

### LINKED ISSUE #531: Issues in Threat Intel Beta
## Background
[Threat Intel Beta PR](https://github.com/airbnb/streamalert/pull/507) is using DynamoDB table to store IOC. During deploy, we found Read Capacity Throttling issue when there are significant amount of log records processed by Rule Processor. We recommend not to use Threat Intel Beta until we fix the throttling issue. I am actively working on this issue now.

chunyong-lin:
size: medium
resolves #531

## Background
[Threat Intel Beta PR](https://github.com/airbnb/streamalert/pull/507) is using DynamoDB table to store IOC. During deploy, we found Read Capacity Throttling issue when there are significant amount of log records processed by Rule Processor. 

This issue has been identified that threat intel should be called from rule processor handler, instead of in rules engine.
This [line](https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/rules_engine.py#L395) is called by every normalized record, so it doesn't take advantage of batch_get_item from DynamoDB.

### Solution:
Collect all normalized records and call threat_detection from Rule Processor's handler.

## Changes

* Call threat intel from StreamAlert `handler`, instead of in StreamRules `process`. This change also requires to return Payload metadata (`payload.log_source`, `payload.type`, `payload.service` and `payload.entity`) of each parsed record. These Payload metadata is needed when trigger the alerts.
* Add more test cases to guard that only one DynamoDB table query been called in one invocation with multiple events coming.
* Fix a bug about handling `ioc_type` incorrectly. The `ioc_type` doesn't match to `ioc_value` if there are multiple type of IOCs in a record. Also add a test case to catch this bug. 

## Testing
* Unit test
```
./tests/scripts/unit_tests.sh
...
TOTAL                                                    3013    117    96%
Ran 513 tests in 12.082s

OK
```
* Rule test
```
...
```
* Tested in AWS testing acount

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -342,14 +342,16 @@ class StreamRules(object):
         and the rule itself to determine if a match occurs.
 
         Returns:
-            list: alerts
-
-            An alert is represented as a dictionary with the following keys:
-                rule_name: the name of the triggered rule
-                payload: the StreamPayload object
-                outputs: list of outputs to send to
+            (list): First return is a list of alerts.
+                An alert is represented as a dictionary with the following keys:
+                    rule_name: the name of the triggered rule
+                    payload: the StreamPayload object
+                    outputs: list of outputs to send to
+            (list): Second return is a list of payload instance with normalized records.
         """
         alerts = []
+        # store normalized records for future process in Threat Intel
+        normalized_records = []
         payload = copy(input_payload)
 
         rules = [rule_attrs for rule_attrs in self.__rules.values()
@@ -357,10 +359,7 @@ class StreamRules(object):
 
         if not rules:
             LOGGER.debug('No rules to process for %s', payload)
-            return alerts
-
-        # store normalized records for future process in Threat Intel
-        normalized_records = []
+            return alerts, normalized_records
 
         for record in payload.records:
             for rule in rules:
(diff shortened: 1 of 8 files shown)


stream_alert/rule_processor/rules_engine.py
```diff
@@ -342,25 +342,24 @@ def process(self, input_payload):
         and the rule itself to determine if a match occurs.
 
         Returns:
-            list: alerts
-
-            An alert is represented as a dictionary with the following keys:
-                rule_name: the name of the triggered rule
-                payload: the StreamPayload object
-                outputs: list of outputs to send to
+            (list): First return is a list of alerts.
```
ryandeivert: For the `Returns` blocks, the parenthesis should not be included around the return type. Please change the back the `list: ...`

stream_alert/rule_processor/rules_engine.py
```diff
@@ -342,25 +342,24 @@ def process(self, input_payload):
...
+                An alert is represented as a dictionary with the following keys:
+                    rule_name: the name of the triggered rule
+                    payload: the StreamPayload object
+                    outputs: list of outputs to send to
+            (list): Second return is a list of payload instance with normalized records.
```
ryandeivert: Same comment as above. Now that I'm looking at it, please specify that this returns a `tuple<list, list>` instead.

stream_alert/rule_processor/rules_engine.py
```diff
@@ -384,22 +383,42 @@ def process(self, input_payload):
                     record_copy = record.copy()
                     record_copy[NORMALIZATION_KEY] = types_result
                     if self._threat_intel:
-                        normalized_records.append(record_copy)
+                        # An copy of payload which includes payload metadata.
```
ryandeivert: Nit: `A copy...`, not `An copy...`

stream_alert/rule_processor/rules_engine.py
```diff
@@ -384,22 +383,42 @@ def process(self, input_payload):
...
+                        # The payload metadata includes log_source, type, service,
+                        # and entity. The meatdata will be returned to along with
```
ryandeivert: Typo: `The meatdata will be returned to along with...` -- > `The meatdata will be returned along with...`

stream_alert/rule_processor/rules_engine.py
```diff
@@ -384,22 +383,42 @@ def process(self, input_payload):
...
+
+        Args:
+            payload_with_normalized_records (list): A list of payload instances.
+                And it pre_parsed_record is replaced by normalized record. The
+                reason to pass a copy of payload into Threat Intelligence is because
+                alerts require to include payload metadata (payload.log_source,
+                payload.type, payload.service and payload.entity).
+
+        Returns:
+            (list): A list of alerts triggered by Threat Intelligence.
```
ryandeivert: Same comment about no parenth around return types in `Returns` block.

stream_alert/rule_processor/threat_intel.py
```diff
@@ -87,10 +87,10 @@ def threat_detection(self, records):
         records contain malicious IOC(s).
 
         Args:
-            records (list): A list of the normalized records which are dictionaries.
+            records (list): A list of payload instance with normalized records.
 
         Returns:
-            (list): A list of records including IOC information.
+            (list): A list of payload instances including IOC information.
```
ryandeivert: Again :)

PR #546, opened 2018-01-02
Firehose Table Creation Bug Fixes

jacknagz:
to: @ryandeivert 
size: medium

## Background

While creating Athena tables for CarbonBlack log types, I discovered a bug in how the CLI converts the log name format to the Athena format.  Athena tables can only have underscores in them, so all other special characters need to be converted.

## Changes

* Fix a bug when looking up schemas for log names containing special characters
* Pylint fixes

## Testing

Local test suite

### DIFF
--- a/stream_alert/rule_processor/firehose.py
+++ b/stream_alert/rule_processor/firehose.py
@@ -212,8 +212,29 @@ class StreamAlertFirehose(object):
                         record_batch_size,
                         stream_name)
 
-    @staticmethod
-    def _load_enabled_log_sources(firehose_config, log_sources):
+    def _firehose_log_name(self, log_name):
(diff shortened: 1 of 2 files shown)


stream_alert/rule_processor/firehose.py
```diff
@@ -212,8 +212,29 @@ def firehose_request_wrapper():
                         record_batch_size,
                         stream_name)
 
-    @staticmethod
-    def _load_enabled_log_sources(firehose_config, log_sources):
+    def firehose_log_name(self, log_name):
```
ryandeivert: I should have caught this before, but this could still be a `classmethod` instead of converting from a staticmethod to an instance method (since the `SPECIAL_CHAR_REGEX` is a class constant). Not a crucial change

PR #548, opened 2018-01-03
[metrics] getting a more accurate processed record count in rp handler

### LINKED ISSUE #527: Metrics: TotalRecords Improvement
### Description

Currently, the `TotalRecords` metric is based around the number of records in the event which invoked Lambda:

https://github.com/airbnb/streamalert/blob/560c0f7db88074cdeba6aaa0bc7a73745df97a9c/stream_alert/rule_processor/handler.py#L119

This is slightly misleading when records are nested, such as with VPC flow logs, or perhaps with S3 files that are pulled and parsed.

### Steps to Reproduce

Observe your Lambda function execution for nested records, and this pattern will show as metrics are logged to CloudWatch.

## Desired Change

Base the `TotalRecords` metric on the number of records parsed into the `StreamPayload` object:

https://github.com/airbnb/streamalert/blob/560c0f7db88074cdeba6aaa0bc7a73745df97a9c/stream_alert/rule_processor/handler.py#L381

ryandeivert:
to: @jacknagz 
size: small
resolves #527 

## Background

See #527. Metrics should track the records for the processed payload, not the incoming 'records' array.

## Changes

* Using a instance property to track the total records from each processed payload.
* This value is then logged at the end of processing as the total records metric.

## Testing

* Adding unit test to check for tracking the proper processed log count.

### DIFF
--- a/stream_alert/rule_processor/handler.py
+++ b/stream_alert/rule_processor/handler.py
@@ -56,6 +56,7 @@ class StreamAlert(object):
 
         self.enable_alert_processor = enable_alert_processor
         self._failed_record_count = 0
+        self._processed_record_count = 0
         self._processed_size = 0
         self._alerts = []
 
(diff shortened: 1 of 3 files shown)


mime-frame:
image

PR #549, opened 2018-01-03
[schema] cb watchlist.hit.binary vt connector optional keys

ryandeivert:
to: @mime-frame 
size: small
resolves N/A

## Background

Observed a failed parse for carbonblack watchlist.hit.binary log type due to missing keys related to virustotal connector.

## Changes

* Adding `*_virustotalconnector` keys as optional keys for this log type.
* This also syncs some other previous changes related to other optional keys.

## Testing

* Tested internally with test event.


mime-frame:
LGTM

PR #550, opened 2018-01-04
[threat intel] Add threat_intel flag in clusters' config

chunyong-lin:
to: @ryandeivert 
size: small
resolves N/A

## Background
During internal deploy, we realized it will be super helpful if we support enable/disable threat_intel per cluster based. So users can have choice to enable threat intel on interested clusters or trying to avoid running threat intel against certain clusters. 

## Changes

* This PR is to add an explicitly disable flag for threat_intel feature based on clusters. In another words, even global threat_intel flag in conf/global.json is set to true, the cluster still can disable it by adding "enable_threat_intel": false, under rule_processor settings.
* If there is no enable_threat_intel flag in clusters' conf, it is default to True and global threat intel flag will take effect.

## Testing
* Rule testing
```
...
```
* Unit testing
```
./tests/scripts/unit_tests.sh
...
TOTAL                                                    3037    119    96%
Ran 516 tests in 12.293s

OK

### DIFF
--- a/stream_alert/rule_processor/config.py
+++ b/stream_alert/rule_processor/config.py
@@ -17,6 +17,8 @@ from collections import OrderedDict
 import json
 import os
 
+CLUSTER = os.environ.get('CLUSTER', '')
+
 class ConfigError(Exception):
     """Exception class for config file errors"""
 
@@ -33,15 +35,29 @@ def load_config(conf_dir='conf/'):
     key denotes the name of the log type, and includes 'keys' used to match
     rules to log fields.
     """
-    conf_files = ('sources', 'logs', 'types', 'global')
+    conf_files = ('sources', 'logs', 'types', 'global', 'clusters')
     config = dict()
     for base_name in conf_files:
-        path = '{}.json'.format(os.path.join(conf_dir, base_name))
-        with open(path) as data:
-            try:
-                config[base_name] = json.load(data, object_pairs_hook=OrderedDict)
-            except ValueError:
-                raise ConfigError('Invalid JSON format for {}.json'.format(base_name))
+        if base_name == 'clusters':
+            # Load current cluster config into memory for threat intel
+            path = os.path.join(conf_dir, base_name)
+            config['clusters'] = {}
+            if CLUSTER:
+                cluster_conf_path = '{}.json'.format(os.path.join(path, CLUSTER))
+                with open(cluster_conf_path) as data:
+                    try:
+                        config['clusters'][CLUSTER] = json.load(data)
+                    except ValueError:
+                        raise ConfigError('Invalid JSON format for {}'.format(cluster_conf_path))
+        else:
+            path = '{}.json'.format(os.path.join(conf_dir, base_name))
+            with open(path) as data:
+                try:
+                    # we use object_pairs_hook=OrderdDict to preserve schema order
+                    # for CSV/KV log types
+                    config[base_name] = json.load(data, object_pairs_hook=OrderedDict)
+                except ValueError:
+                    raise ConfigError('Invalid JSON format for {}.json'.format(base_name))
 
     # Validate the config. This will raise an exception on any errors,
     # which bubbles up and will immediately break execution of the function
--- a/stream_alert/rule_processor/config.py
+++ b/stream_alert/rule_processor/config.py
@@ -17,7 +17,6 @@ from collections import OrderedDict
 import json
 import os
 
-CLUSTER = os.environ.get('CLUSTER', '')
 
 class ConfigError(Exception):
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: stream_alert/rule_processor/threat_intel.py [hunk 1]
```diff
@@ -179,6 +183,12 @@ def load_from_config(cls, config):
         if config.get('types'):
             cls._process_types_config(config['types'])
 
+        # Threat Intel will be disabled for the cluster if it is explicityly
```
jacknagz: explicityly -> explicitly

stream_alert/rule_processor/config.py
```diff
@@ -33,15 +35,29 @@ def load_config(conf_dir='conf/'):
...
+                with open(cluster_conf_path) as data:
+                    try:
+                        config['clusters'][CLUSTER] = json.load(data)
+                    except ValueError:
+                        raise ConfigError('Invalid JSON format for {}'.format(cluster_conf_path))
+        else:
+            path = '{}.json'.format(os.path.join(conf_dir, base_name))
+            with open(path) as data:
+                try:
+                    # we use object_pairs_hook=OrderdDict to preserve schema order
```
jacknagz: Can this condense into a single line? or is it above limits

stream_alert/rule_processor/config.py
```diff
@@ -33,15 +35,29 @@ def load_config(conf_dir='conf/'):
...
-        path = '{}.json'.format(os.path.join(conf_dir, base_name))
-        with open(path) as data:
-            try:
-                config[base_name] = json.load(data, object_pairs_hook=OrderedDict)
-            except ValueError:
-                raise ConfigError('Invalid JSON format for {}.json'.format(base_name))
+        if base_name == 'clusters':
+            # Load current cluster config into memory for threat intel
+            path = os.path.join(conf_dir, base_name)
+            config['clusters'] = {}
```
jacknagz: this could be loaded into a `cluster` key instead since we load a singular file

chunyong-lin: I just want to clarify that do you mean to have key order 
1. `config['prod']` when the current cluster name is `prod`? 
2. Or, `config['cluster']['prod']`?

stream_alert/rule_processor/config.py
chunyong-lin: it will be above limits (104 chars) if condense into a single line.

stream_alert/rule_processor/config.py
chunyong-lin: I would prefer to keep `config['clusters']` style, because it is inline with directory structure in `conf/`

jacknagz: sounds good @chunyong-lin, feel free to leave then

stream_alert/rule_processor/config.py
ryandeivert: Just so I'm clear, the requirement for the objet pairs hook is only really for the logs.json file, correct? We could (theoretically) check if that's the file being loaded and only use that keyword arg for that specific file, instead of all non-cluster files.

EDIT: probably not a change that needs to happen now, just a thought

stream_alert_cli/logger.py
```diff
@@ -34,7 +34,8 @@ def filter(self, record):
             'Starting download from S3*',
             'Completed download in*',
             '*triggered an alert on log type*',
-            '*Firehose*'
+            '*Firehose*',
+            'Got * normalized records'
```
ryandeivert: Thanks for this!

tests/unit/stream_alert_rule_processor/test_threat_intel.py
```diff
@@ -76,6 +80,10 @@ def setup(self):
         self.config = load_config('tests/unit/conf')
         self.config['global']['threat_intel']['enabled'] = True
         self.threat_intel = StreamThreatIntel.load_from_config(self.config)
+        os.environ['CLUSTER'] = ''
```
ryandeivert: Since you always use the `with patch.dict(...` context manager, there is no need to 'reset' this value upon setup here. Can you comment this out and see if the tests still pass? The same may apply for the 2 `reload` statements that follow, but I'm not 100% sure.

chunyong-lin:
@ryandeivert Thanks for the feedback and help. I have made change to change `CLUSTER` to local variable and it makes unit test easier. 

As we discussed, it is not necessary to store current cluster name in a module constant. The cluster information is only used in `load_config` and `load_from_config` methods. It doesn't have much benefit to store it as a constant. However, I had experienced difficult to write correct unit test cases when cluster name store in a constant which requires to reload modules if I change cluster name in environment variable.

PR #553, opened 2018-01-04
Athena Schema Overrides and Cleanup

jacknagz:
to: @ryandeivert 
size: small

## Background

Add a new flag to the create-table command to allow for schema overrides. There is a case where the defined field type (like float for instance) would differ in Athena/Hive.  

I realize this is not a perfect solution, since the override state isn't stored anywhere (perhaps it can be added in the global Athena or Firehose configurations in the future.

## Changes

* New `--schema_override` flag
* Cleanup

## Testing

Local test suite, deployed in test account too


manage.py
```diff
@@ -1186,6 +1186,13 @@ def _add_athena_subparser(subparsers):
 
     athena_parser.add_argument('--table_name', help=ARGPARSE_SUPPRESS)
 
+    athena_parser.add_argument(
+        '--schema_override',
```
ryandeivert: Just realized you're using an underscore as a delimiter in the flag. Does that mean it has to be used as `--schem_override` in the cli, or will `--schema-override` also work? Idk if argparse enforces it...

I only mention this because our other commands/flags are delimited with a hyphen if there are two words (ie: `--test-files`, `--alarm-description`, etc...). But now that I'm looking, it seems we've been following different patterns here 🤣 . Maybe we should choose one and standardize on it.

jacknagz: Merging now @ryandeivert, we'll need to revisit later

PR #557, opened 2018-01-05
[threat intel] apply ip filter to only check public ip

chunyong-lin:
to: @ryandeivert or @jacknagz 
size: small
resolves 

## Background
When processing network related events, we will encounter the situation that many of the events are generated in the private subnet. It is totally not necessary to run Threat Intel against these private ip addresses. This PR proposes to filter out broadcast and private IP addresses. 

## Changes

* Use IPAddress class from `netaddr` lib.
* Add unit test cases. 

## Testing
* Performance scoping
Tested with 5466 cloudwatch events with IP filter, the IOC number is dropped to 2324, by 57%. It means the RC to DDB will be dropped by 57% based on the test events I have. 
* Rule Test
```
...
```
* Unit test
```
./tests/scripts/unit_tests.sh
...
TOTAL                                                    3047    119    96%
Ran 517 tests in 11.176s

OK
```

### DIFF
--- a/stream_alert/rule_processor/threat_intel.py
+++ b/stream_alert/rule_processor/threat_intel.py
@@ -412,6 +412,8 @@ class StreamThreatIntel(object):
     def is_public_ip(ip_address):
         try:
             ip_addr = IPAddress(str(ip_address))
+            # We also need to filter multicast ip which is a private ip. For example,
+            # multicast ip '239.192.0.1', is a private ip.
             return ip_addr.is_unicast() and not ip_addr.is_private()
         except AddrFormatError:
             return False
--- a/stream_alert/rule_processor/threat_intel.py
+++ b/stream_alert/rule_processor/threat_intel.py
@@ -19,6 +19,8 @@ import backoff
 import boto3
 from boto3.dynamodb.types import TypeDeserializer
 from botocore.exceptions import ClientError, ParamValidationError
+from netaddr import IPAddress
+from netaddr.core import AddrFormatError
 
 from stream_alert.shared import NORMALIZATION_KEY
 from stream_alert.shared.backoff_handlers import (
@@ -164,6 +166,10 @@ class StreamThreatIntel(object):
                         for original_key in original_keys:
                             value = value[original_key]
                     if value:
+                        # Threat Intel will only check against public IP addresses
+                        # while processing IP IOCs.
+                        if ioc_type == 'ip' and not self.is_public_ip(value):
+                            continue
                         ioc_value_type_tuples.add((value, ioc_type))
         return [StreamIoc(value=str(value).lower(), ioc_type=ioc_type, associated_record=record)
                 for value, ioc_type in ioc_value_type_tuples]
@@ -401,3 +407,11 @@ class StreamThreatIntel(object):
     def normalized_type_mapping(cls):
         """Get normalized CEF types mapping to original keys from the records."""
         return cls.__normalized_types
+
+    @staticmethod
+    def is_public_ip(ip_address):
+        try:
+            ip_addr = IPAddress(str(ip_address))
+            return ip_addr.is_unicast() and not ip_addr.is_private()
+        except AddrFormatError:
+            return False
--- a/tests/unit/stream_alert_rule_processor/test_threat_intel.py
+++ b/tests/unit/stream_alert_rule_processor/test_threat_intel.py
@@ -247,6 +247,57 @@ class TestStreamThreatIntel(object):
             assert_equal((results[3].value, results[3].ioc_type),
                          ('abcdef0123456789abcdef0123456789', 'md5'))
 
+    def test_extract_ioc_from_record_with_private_ip(self):
+        """Threat Intel - Test extrac values from a record based on normalized keys"""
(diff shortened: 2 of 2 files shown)


tests/unit/stream_alert_rule_processor/test_threat_intel.py
```diff
@@ -247,6 +247,57 @@ def test_extract_ioc_from_record(self):
             assert_equal((results[3].value, results[3].ioc_type),
                          ('abcdef0123456789abcdef0123456789', 'md5'))
 
+    def test_extract_ioc_from_record_with_private_ip(self):
+        """Threat Intel - Test extrac values from a record based on normalized keys"""
+        records = [
```
jacknagz: I'd add a public IP example in here

stream_alert/rule_processor/threat_intel.py
```diff
@@ -164,6 +166,10 @@ def _extract_ioc_from_record(self, record):
                         for original_key in original_keys:
                             value = value[original_key]
                     if value:
+                        # Threat Intel will only check against public IP addresses
+                        # when process IP IOCs.
```
jacknagz: processing*

stream_alert/rule_processor/threat_intel.py
```diff
@@ -164,6 +166,10 @@ def _extract_ioc_from_record(self, record):
...
+                        if ioc_type == 'ip' and not self.is_public_ip(value):
```
mime-frame: why negation logic when you can just do `.is_private()`?

mime-frame: see https://viewsby.wordpress.com/2015/02/12/python-check-if-ip-is-public-or-private/

tests/unit/stream_alert_rule_processor/test_threat_intel.py
chunyong-lin: It considers `1.1.1.2` as public address 😄  So it is already covered by old test cases. 
The private subnet range falls into three classes.
1. class A: 10.0.0.0 – 10.255.255.255
2. class B: 172.16.0.0 – 172.31.255.255
3. class C: 192.168.0.0 – 192.168.255.255

stream_alert/rule_processor/threat_intel.py
mime-frame: # Private IP addresses (ex: RFC1918) will not be compared against threat intelligence

stream_alert/rule_processor/threat_intel.py
chunyong-lin: We also need to filter out non unicast ip address (multicast, broadcast).

stream_alert/rule_processor/threat_intel.py
```diff
@@ -401,3 +407,11 @@ def _deserialize(dynamodb_data):
     def normalized_type_mapping(cls):
         """Get normalized CEF types mapping to original keys from the records."""
         return cls.__normalized_types
+
+    @staticmethod
+    def is_public_ip(ip_address):
+        try:
+            ip_addr = IPAddress(str(ip_address))
+            return ip_addr.is_unicast() and not ip_addr.is_private()
```
mime-frame: add a code comment about why you're checking if its unicast

tests/unit/stream_alert_rule_processor/test_threat_intel.py
```diff
@@ -247,6 +247,57 @@ def test_extract_ioc_from_record(self):
             assert_equal((results[3].value, results[3].ioc_type),
                          ('abcdef0123456789abcdef0123456789', 'md5'))
 
+    def test_extract_ioc_from_record_with_private_ip(self):
+        """Threat Intel - Test extrac values from a record based on normalized keys"""
```
mime-frame: `extrac` = `extracting`?

tests/unit/stream_alert_rule_processor/test_threat_intel.py
mime-frame: and a malformed IP address

mime-frame: and a test log with both a public and a private ip address, ensuring that a single private entry doesn't result in the other one not being processed

stream_alert/rule_processor/threat_intel.py
mime-frame: ex: do we not care about multicast?

stream_alert/rule_processor/threat_intel.py
chunyong-lin: (netaddr doc)[https://netaddr.readthedocs.io/en/latest/tutorial_01.html#public] shows how  to check is the ip address is public or not.

stream_alert/rule_processor/threat_intel.py
chunyong-lin: Good question. I just verified that. A multicast ip is a private ip. 
```
from netaddr import IPAddress

ip = IPAddress('239.192.0.1')
ip.is_multicast()
True

ip.is_unicast()
False

ip.is_private()
True
```

PR #562, opened 2018-01-09
Add Bandit security linting; update requirements

austinbyers:
to: @jacknagz or @ryandeivert 
size: small

## Background

[Bandit](https://pypi.python.org/pypi/bandit/1.0.1) is a Python scanner which checks for common security issues in Python source code. I've already successfully [added Bandit to BinaryAlert](https://github.com/airbnb/binaryalert/pull/100), and now we add to StreamAlert.

## Changes

Fortunately, Bandit did not flag any security issues! No code changes were necessary

* Adds `bandit` to requirements
* Removes `virtualenv` from requirements (you don't need to install `virtualenv`, you are already in a venv when installing requirements)
* Updates frozen requirements set with latest versions

## Testing

* Travis CI: unit tests, linting, and now security checks


ryandeivert:
@austinbyers one comment: you mention this is enforced in travis ci... does it only take adding a `.bandit` file for travis to pick it up? Asking because I don't see any changes to the `.travis.yml` :)

.bandit
```diff
@@ -0,0 +1,10 @@
+[bandit]
```
jacknagz: can this live in the `setup.cfg` file instead?

requirements.txt
```diff
@@ -1,90 +1,85 @@
-Sphinx==1.6.5
+alabaster==0.7.10
```
jacknagz: the changes this file breaks the separation that previously existed, can you follow these two steps in order to retain that?

```
# Installing the top-level dependencies, upgrading them if necessary
pip install –r requirements-top-level.txt --upgrade 

# Updating requirements.txt
pip freeze –r requirements-top-level.txt > requirements.txt
```

ryandeivert: The main issue I foresee with this is that dev requirements are lost with this workflow

jacknagz: @ryandeivert how's that? developers just `pip install requirements.txt` as normal

ryandeivert: ah sorry I didn't realize the dev/test requirements were also in the normal requirements file. we should definitely move away from that

.bandit
austinbyers: Great idea! Yes, it works

requirements.txt
austinbyers: @jacknagz, that is exactly what I did, with the exception that I didn't order the requirements with the `-r` flag to `pip freeze`. So it was just the requirements order that changed (the files have the same meaning and contents).

I've put the order back, thanks for listing the command!

austinbyers:
@ryandeivert good catch! Yep, I forgot to add `bandit` to `.travis.yml`. It's there now!

jacknagz:
👍

PR #563, opened 2018-01-10
Add StreamAlert tag to Athena SQS queue

austinbyers:
cc: @chunyong-lin 
size: small

## Changes

* Adds the "StreamAlert" tag to SQS, since SQS tagging support was added a few months ago

## Testing

* `terraform validate`


jacknagz:
👍

PR #565, opened 2018-01-10
[threat intel] Catch ValueError and TypeError while processing IP IOCs

chunyong-lin:
to: @ryandeivert or @jacknagz 
size: small
resolves N/A

## Background

IPAddress class raise `ValueError` when the address value contains `/ `(subnet prefix), instead of raising `AddrFormatError`. (Refer to IPAddress [source code](http://netaddr.readthedocs.io/en/latest/_modules/netaddr/ip.html).)

## Changes
* This fix is to catch `ValueError`, in addition to catch `TypeError` also.

## Testing
* Rule testing
```
...
```
* Unit testing
```
./tests/scripts/unit_tests.sh
...
TOTAL                                                    3048    119    96%
Ran 518 tests in 10.170s

OK
```

### DIFF
--- a/stream_alert/rule_processor/threat_intel.py
+++ b/stream_alert/rule_processor/threat_intel.py
@@ -20,7 +20,6 @@ import boto3
 from boto3.dynamodb.types import TypeDeserializer
 from botocore.exceptions import ClientError, ParamValidationError
 from netaddr import IPAddress
-from netaddr.core import AddrFormatError
 
 from stream_alert.shared import NORMALIZATION_KEY
 from stream_alert.shared.backoff_handlers import (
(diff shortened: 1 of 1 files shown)


mime-frame:
Id suggest we dont log an error, as some data providers might leave the field blank or populate it with something dumb like 'none' or 'empty' every once in awhile

chunyong-lin:
@ryandeivert Pylint will complaint if do a generic exception (I did generic exception in the beginning), unless we disable pylint to check this line.

ryandeivert:
@chunyong-lin see my updated comment above on how to suppress the pylint errors :)

chunyong-lin:
I agree with Joren, we don't want to log these errors, it will be many logs.

ryandeivert:
@chunyong-lin can you update the PR description to reflect reality with your new change? :)

chunyong-lin:
@ryandeivert Done! Thanks for the reminder. 👍

PR #567, opened 2018-01-11
[outputs] PagerDuty Incident rich content

javuto:
to: @ryandeivert @mime-frame 
size: medium

## Background

PagerDuty REST API v2 allows to create incidents and assign them to a specific user from the PagerDuty account. That allows alerts to be able to be assigned when they are created. The problem is that the API does not allow to have any JSON object data, like the PagerDuty Events API v2. With this code, we are able to utilize the incident creation capabilities and also have a rich content by merging alerts/events created with the Events API v2.

## Changes

* Added PUT requests support to the output base class.
* Added capability to create events and merge them with the created incidents.
* Modified unit tests to test new capabilities.

## Testing

```
...
Ran 518 tests in 11.193s

OK

$ ./tests/scripts/pylint.sh

...
```

### DIFF
--- a/stream_alert/alert_processor/outputs/pagerduty.py
+++ b/stream_alert/alert_processor/outputs/pagerduty.py
@@ -24,6 +24,35 @@ from stream_alert.alert_processor.outputs.output_base import (
     StreamAlertOutput
 )
 
+def events_v2_data(routing_key, **kwargs):
+    """Helper method to generate the payload to create an event using PagerDuty Events API v2
+    Args:
+        routing_key:
+        **kwargs: consists of any combination of the following items:
+            descriptor (str): Service descriptor (ie: slack channel, pd integration)
+            rule_name (str): Name of the triggered rule
+            alert (dict): Alert relevant to the triggered rule
+    Returns:
+            dict: Contains JSON blob to be used as event
+    """
+    summary = 'StreamAlert Rule Triggered - {}'.format(kwargs['rule_name'])
+
+    details = {
+        'rule_description': kwargs['alert']['rule_description'],
+        'record': kwargs['alert']['record']
+    }
+    payload = {
+        'summary': summary,
+        'source': kwargs['alert']['log_source'],
+        'severity': 'critical',
+        'custom_details': details
+    }
+    return {
+        'routing_key': routing_key,
+        'payload': payload,
+        'event_action': 'trigger',
+        'client': 'StreamAlert'
+    }
 
 @StreamAlertOutput
 class PagerDutyOutput(OutputDispatcher):
@@ -147,24 +176,7 @@ class PagerDutyOutputV2(OutputDispatcher):
         if not creds:
             return self._log_status(False)
 
-        summary = 'StreamAlert Rule Triggered - {}'.format(kwargs['rule_name'])
-
-        details = {
-            'rule_description': kwargs['alert']['rule_description'],
-            'record': kwargs['alert']['record']
-        }
-        payload = {
-            'summary': summary,
-            'source': kwargs['alert']['log_source'],
-            'severity': 'critical',
-            'custom_details': details
-        }
-        data = {
-            'routing_key': creds['routing_key'],
-            'payload': payload,
-            'event_action': 'trigger',
-            'client': 'StreamAlert'
-        }
+        data = events_v2_data(creds['routing_key'], **kwargs)
 
         try:
             success = self._post_request_retry(creds['url'], data, None, True)
@@ -232,6 +244,9 @@ class PagerDutyIncidentOutput(OutputDispatcher):
             ('email_from',
              OutputProperty(description='valid user email from the PagerDuty '
                                         'account linked to the token',
+                            cred_requirement=True)),
+            ('integration_key',
+             OutputProperty(description='the integration key for this PagerDuty integration',
                             cred_requirement=True))
         ])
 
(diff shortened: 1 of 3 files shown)


stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -23,7 +24,41 @@
...
 )
+from stream_alert.shared.backoff_handlers import (
+    backoff_handler,
+    success_handler,
+    giveup_handler
+)
 
+def events_v2_data(routing_key, **kwargs):
+    """Helper method to generate the payload to create an event using PagerDuty Events API v2
+    Args:
```
ryandeivert: Please change this to `Keyword Args:` and omit the `**kwargs` line below

stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -248,6 +275,97 @@ def _get_endpoint(base_url, endpoint):
...
+        """
+        params = {
+            'incident_key': '{}'.format(incident_key)
+        }
+        incidents_url = self._get_endpoint(self._base_url, self.INCIDENTS_ENDPOINT)
+        response = self._generic_api_get(incidents_url, params)
+
+        incident = response.get('incidents', [])
+        if not incident:
+            raise PagerdutySearchDelay
```
ryandeivert: This should be `PagerdutySearchDelay()`

stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -248,6 +275,97 @@ def _get_endpoint(base_url, endpoint):
...
+
+        return incident[0].get('id')
+
+    def _merge_incidents(self, url, to_be_merged_id):
+        """Helper to merge incidents by id using the PagerDuty REST API v2 and
```
ryandeivert: Please add Args info in this doctstring

stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -479,8 +592,38 @@ def dispatch(self, **kwargs):
...
-            success = False
+            incident = False
 
-        return self._log_status(success)
+        if not incident:
+            LOGGER.error('Could not create main incident, %s', self.__service__)
+            return self._log_status(False)
+
+        # Extract the incident id from the incident that was just created
+        incident_id = incident.get('incident').get('id')
```
ryandeivert: If the first get returns `None` the second `get` is going to cause an `AttributeError` here. Suggestion:

`incident_id = incident.get('incident', {}).get('id')`

or omit the first get altogether if you're sure it will exist (`incident_id = incident['incident'].get('id')`)

stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -479,8 +592,38 @@ def dispatch(self, **kwargs):
...
+        # Keep that id to be merged later with the created incident
+        event_incident_id = self._get_event_incident_id(incident_key)
+
+        # Merge the incident with the event, so we can have a rich context incident
+        # assigned to a specific person, which the PagerDuty REST API v2 does not allow
+        merging_url = '{}/{}'.format(incidents_url, incident_id)
+
+        merged = self._merge_incidents(merging_url, event_incident_id)
+
+        if not merged:
```
ryandeivert: I think this if block is safe to leave out, since essentially the same thing will happen just below on:

`return self._log_status(merged)`

And there is a logger message in `_log_status` that does something similar to what you're logging here.

stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -248,6 +275,97 @@ def _get_endpoint(base_url, endpoint):
...
+
+        Args:
+            incident_key (str): Incident key that indentifies uniquely an incident
+
+        Returns:
+            str: ID of the incident after look up the incident_key
+
+        """
+        params = {
+            'incident_key': '{}'.format(incident_key)
```
ryandeivert: this `'{}'.format(...)` is unnecessary I believe(?)

stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -479,8 +592,38 @@ def dispatch(self, **kwargs):
...
+
+        # Create alert to hold all the incident details
+        event_data = events_v2_data(creds['integration_key'], **kwargs)
+        event = self._create_event(event_data)
+        if not event:
+            LOGGER.error('Could not create incident event, %s', self.__service__)
+            return self._log_status(False)
+
+        # Lookup the incident_key returned as dedup_key to get the incident id
+        incident_key = event.get('dedup_key')
```
ryandeivert: What would occur here if this `.get('dedup_key')` returns `None`? Should it be evaluated before continuing on?

javuto: This is a good point, since the API will return all open incidents if the `incident_key` is empty, and that can cause bad things because it will pick up the first element of the results to be merged.

ryandeivert: @javuto good to know! any ideas on how we can account for that?

javuto: I just pushed some code that fails the incident creation and logs it, that should be easy to spot and does not put any alerts/incidents at risk to be merged with the wrong thing.

PR #568, opened 2018-01-13
athena: added recursive schema support

javefang:
size: medium
resolves #<related-issue-goes-here>

## Background

Nested log schema is not supported when creating tables in Athena.

## Changes

* Added recursive support when converting log schema to Athena schema.

## Testing

```
15:00 $ ./tests/scripts/pylint.sh
Starting pylint script

15:01 $ ./tests/scripts/unit_tests.sh
...
CLI - Generate Athena schema: simple ... ok (0.0003s)
CLI - Generate Athena schema: special key ... ok (0.0003s)
CLI - Generate Athena schema: nested ... ok (0.0002s)
...
Ran 521 tests in 66.392s

OK
```

### DIFF
--- a/stream_alert_cli/athena/helpers.py
+++ b/stream_alert_cli/athena/helpers.py
@@ -40,10 +40,13 @@ def to_athena_schema(log_schema):
     for key_name, key_type in log_schema.iteritems():
         key_name = '`{}`'.format(key_name)
         if key_type == {}:
+            # For empty dicts
             athena_schema[key_name] = SCHEMA_TYPE_MAPPING[dict]
         elif key_type == []:
+            # For empty array
             athena_schema[key_name] = SCHEMA_TYPE_MAPPING[list]
         elif isinstance(key_type, dict):
+            # For recursion
             athena_schema[key_name] = to_athena_schema(key_type)
         else:
             athena_schema[key_name] = SCHEMA_TYPE_MAPPING[key_type]
--- a/stream_alert_cli/athena/helpers.py
+++ b/stream_alert_cli/athena/helpers.py
@@ -1,0 +1,43 @@
+"""
+Copyright 2017-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+# How to map log schema types to Athena/Hive types
+SCHEMA_TYPE_MAPPING = {
+    'string': 'string',
+    'integer': 'bigint',
+    'boolean': 'boolean',
+    'float': 'decimal(10,3)',
+    dict: 'map<string, string>',
+    list: 'array<string>'
+}
+
+def to_athena_schema(log_schema):
+    """Convert streamalert log schema to athena schema"""
+
+    athena_schema = {}
+
+    for key_name, key_type in log_schema.iteritems():
+        key_name = '`{}`'.format(key_name)
+        if key_type == {}:
+            athena_schema[key_name] = SCHEMA_TYPE_MAPPING[dict]
+        elif key_type == []:
+            athena_schema[key_name] = SCHEMA_TYPE_MAPPING[list]
+        elif isinstance(key_type, dict):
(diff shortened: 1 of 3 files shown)


jacknagz:
hey @javefang, you can run our local test suite with `./tests/scripts/unit_tests.sh`, and also with `./tests/scripts/pylint.sh` locally.  I'd also recommend adding a new test file for `stream_alert_cli.athena.handler` to verify the changes.  Finally, if you have specific test cases that you tried with Athena that would also be great.

javefang:
@jacknagz fixed linter warnings, and added unit tests

stream_alert_cli/athena/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+    'string': 'string',
+    'integer': 'bigint',
+    'boolean': 'boolean',
+    'float': 'decimal(10,3)',
+    dict: 'map<string, string>',
+    list: 'array<string>'
+}
+
+def to_athena_schema(log_schema):
+    """Convert streamalert log schema to athena schema"""
```
jacknagz: Please add additional context on the Args and Return value

javefang:
```
(venv) ✔ ~/git/streamalert [add_athena_recursive_schema_support|✔]
23:04 $ ./tests/scripts/autopep8.sh stream_alert_cli/athena/helpers.py
(venv) ✔ ~/git/streamalert [add_athena_recursive_schema_support|✔]
23:05 $ ./tests/scripts/autopep8.sh stream_alert_cli/athena/handler.py
(venv) ✔ ~/git/streamalert [add_athena_recursive_schema_support|✚ 1]
23:05 $ ./tests/scripts/autopep8.sh tests/unit/stream_alert_cli/athena/test_handler.py
```

stream_alert_cli/athena/helpers.py
```diff
@@ -0,0 +1,51 @@
...
+
+    athena_schema = {}
+
+    for key_name, key_type in log_schema.iteritems():
+        key_name = '`{}`'.format(key_name)
+        if key_type == {}:
+            athena_schema[key_name] = SCHEMA_TYPE_MAPPING[dict]
+        elif key_type == []:
+            athena_schema[key_name] = SCHEMA_TYPE_MAPPING[list]
+        elif isinstance(key_type, dict):
```
ryandeivert: hey @javefang  - One small suggestion here. It's not clear (since there are no comments) why we do the `if key_type == {}` above and then do `elif isinstance(key_type, dict)` here. I know via deduction that this is looking for any empty dict vs a non-empty dict, to do the recursion, but for someone else looking at this it may not be as readily obvious.

Suggestion for this would be to modify so we don't do any type checking with the `==` operator and instead do something like the following:

```
if isinstance(key_type, dict):
    # Check if the dict is empty
    if not key_type:
        athena_schema[key_name] = SCHEMA_TYPE_MAPPING[dict]
    else:
        # Handle recursive key/values in sub-dicts
        athena_schema[key_name] = to_athena_schema(key_type)
elif isinstance(key_type, list):
    athena_schema[key_name] = SCHEMA_TYPE_MAPPING[list]
else:
    athena_schema[key_name] = SCHEMA_TYPE_MAPPING[key_type]
```

javefang: Make sense, this made me think: is it better to remove the special keys {} and [], and replace them with string "dict" and "list" in the streamalert log schema? This way we can reserve {} for representing nested value only. I think this makes sense since if you mean to represent "list" in the schema, one should never use anything other than [] (e.g. [1, 2] should be invalid). But the code you provided will not allow that?

I understand this will be a breaking change so you might not want to include it here. But if we do so, then the code can be simplified to:

```
SCHEMA_TYPE_MAPPING = {
...
    'dict': 'map<string, stirng>',
    'list': 'array<string>'
...
}

...

if isinstance(key_type, dict):
    athena_schema[key_name] = to_athena_schema(key_type)
else:
    athena_schema[key_name] = SCHEMA_TYPE_MAPPING[key_type]
```

What do you think?

ryandeivert: @javefang yes, this is ideally where we want to get with our schema declarations, but is a much larger overhaul. Thanks for pointing this out :) I'm actually going to file an issue for this now so we don't forget about it.

javefang: Thanks, sure this will be a breaking change to many people who already have schema defined. 

In this case, I feel like checking with `log_type == {}` and `log_type == []` is probably more straightforward for the moment since it directly dictates what is in the schema? Do you still want to make the recommended change @ryandeivert ?

ryandeivert: @javefang let's leave it how you have it for now and address in the future. Can you just add a comment to the if blocks stating that one is for empty dicts and one is for recursion?

ryandeivert:
Thanks for the contribution @javefang. This PR has now been merged :)

Issue #478, opened 2017-11-14
Bug: Rule processor Firehose bug - `error: [Errno 104] Connection reset by peer`


ryandeivert:
## Background

Observed the following traceback in CloudWatch logs:
```
[Errno 104] Connection reset by peer: error
Traceback (most recent call last):
File "/var/task/stream_alert/rule_processor/main.py", line 39, in handler
StreamAlert(context).run(event)
File "/var/task/stream_alert/rule_processor/handler.py", line 172, in run
self._send_to_firehose()
File "/var/task/stream_alert/rule_processor/handler.py", line 348, in _send_to_firehose
self._firehose_request_helper(stream_name, sized_batch)
File "/var/task/stream_alert/rule_processor/handler.py", line 304, in _firehose_request_helper
resp = firehose_request_wrapper()
File "/var/task/backoff/_sync.py", line 42, in retry
ret = target(*args, **kwargs)
File "/var/task/backoff/_sync.py", line 85, in retry
ret = target(*args, **kwargs)
File "/var/task/stream_alert/rule_processor/handler.py", line 299, in firehose_request_wrapper
in record_batch])
File "/var/runtime/botocore/client.py", line 312, in _api_call
return self._make_api_call(operation_name, kwargs)
File "/var/runtime/botocore/client.py", line 592, in _make_api_call
operation_model, request_dict)
File "/var/runtime/botocore/endpoint.py", line 141, in make_request
return self._send_request(request_dict, operation_model)
File "/var/runtime/botocore/endpoint.py", line 170, in _send_request
success_response, exception):
File "/var/runtime/botocore/endpoint.py", line 249, in _needs_retry
caught_exception=caught_exception, request_dict=request_dict)
File "/var/runtime/botocore/hooks.py", line 227, in emit
return self._emit(event_name, kwargs)
File "/var/runtime/botocore/hooks.py", line 210, in _emit
response = handler(**kwargs)
File "/var/runtime/botocore/retryhandler.py", line 183, in __call__
if self._checker(attempts, response, caught_exception):
File "/var/runtime/botocore/retryhandler.py", line 251, in __call__
caught_exception)
File "/var/runtime/botocore/retryhandler.py", line 269, in _should_retry
return self._checker(attempt_number, response, caught_exception)
File "/var/runtime/botocore/retryhandler.py", line 317, in __call__
caught_exception)
File "/var/runtime/botocore/retryhandler.py", line 223, in __call__
attempt_number, caught_exception)
[...]

PR #583, opened 2018-01-25
Add Firehose Request Id and Support Custom CW Alarm Params

jacknagz:
to: @austinbyers 
size: medium
related to: #478 

## Background

Two changes in this PR, one is to add the Firehose Request Id string after each batch for better troubleshooting with #478, and one is to add knobs for the Cloudwatch Monitoring Terraform module.

## Changes

* Add RequestId in the logger for each Firehose batch request
* Make the CloudWatch Alarm module configurable
* Fix a bug in the CloudTrail module by defaulting logging to enabled
* Update tests

## Testing

Tested locally with unit tests and pylinting, deployed Lambda change onto a test account

### DIFF
--- a/stream_alert_cli/terraform/cloudtrail.py
+++ b/stream_alert_cli/terraform/cloudtrail.py
@@ -33,8 +33,10 @@ def generate_cloudtrail(cluster_name, cluster_dict, config):
     cloudtrail_module = 'cloudtrail_{}'.format(cluster_name)
 
     enabled_legacy = modules['cloudtrail'].get('enabled')
-    cloudtrail_enabled = modules['cloudtrail'].get('enable_logging')
-    kinesis_enabled = modules['cloudtrail'].get('enable_kinesis')
+
+    cloudtrail_enabled = modules['cloudtrail'].get('enable_logging', True)
+    kinesis_enabled = modules['cloudtrail'].get('enable_kinesis', True)
+
     account_ids = list(
         set([config['global']['account']['aws_account_id']] + modules['cloudtrail'].get(
             'cross_account_ids', [])))
--- a/stream_alert_cli/terraform/monitoring.py
+++ b/stream_alert_cli/terraform/monitoring.py
@@ -90,9 +90,8 @@ def generate_monitoring(cluster_name, cluster_dict, config):
     # Note: This does not strictly check for proper variable names, since there are so many.
     #       Instead, Terraform will error out if an imporper name is used.
     #       Also, every value in these settings should be a string, so cast for safety.
-    if monitoring_config.get('settings', {}):
-        for setting_name, setting_value in monitoring_config['settings'].iteritems():
-            cluster_dict['module']['cloudwatch_monitoring_{}'.format(
-                cluster_name)][setting_name] = str(setting_value)
+    for setting_name, setting_value in monitoring_config.get('settings', {}).iteritems():
+        cluster_dict['module']['cloudwatch_monitoring_{}'.format(
+            cluster_name)][setting_name] = str(setting_value)
 
     return True
--- a/stream_alert_cli/terraform/monitoring.py
+++ b/stream_alert_cli/terraform/monitoring.py
@@ -19,6 +19,19 @@ from stream_alert_cli.logger import LOGGER_CLI
 def generate_monitoring(cluster_name, cluster_dict, config):
     """Add the CloudWatch Monitoring module to the Terraform cluster dict.
 
+    Example configuration:
+
+    "cloudwatch_monitoring": {
+      "enabled": true,
+      "kinesis_alarms_enabled": true,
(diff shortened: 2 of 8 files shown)


stream_alert_cli/terraform/monitoring.py
```diff
@@ -19,6 +19,19 @@
...
+    Example configuration:
+
+    "cloudwatch_monitoring": {
+      "enabled": true,
+      "kinesis_alarms_enabled": true,
+      "lambda_alarms_enabled": true,
+      "settings": {
+        "lambda_invocation_error_period": "600",
+        "kinesis_iterator_age_error_period": "600",
+        "kinesis_write_throughput_exceeded_threshold": "100"
```
austinbyers: To be consistent with the rest of the configuration (kinesis `batch_size`, lambda `current_version`, etc), these should be numbers instead of strings. Terraform works fine with numbers, too.

stream_alert_cli/terraform/monitoring.py
```diff
@@ -69,4 +86,13 @@ def generate_monitoring(cluster_name, cluster_dict, config):
             'kinesis_alarms_enabled': True
         })
 
+    # Add support for custom settings for tweaking alarm thresholds, eval periods, and periods
+    # Note: This does not strictly check for proper variable names, since there are so many.
+    #       Instead, Terraform will error out if an imporper name is used.
+    #       Also, every value in these settings should be a string, so cast for safety.
+    if monitoring_config.get('settings', {}):
+        for setting_name, setting_value in monitoring_config['settings'].iteritems():
```
austinbyers: You don't need the top-level `if` statement: `.iteritems()` will not return anything for an empty dict.

```python
for setting_name, setting_value in monitoring_config.get('settings', {}).iteritems():
```

tests/unit/stream_alert_cli/terraform/test_monitoring.py
```diff
@@ -44,6 +44,39 @@ def test_generate_cloudwatch_monitoring():
         cluster_dict['module']['cloudwatch_monitoring_test'],
         expected_cloudwatch_tf)
 
+def test_generate_cloudwatch_monitoring_with_settings():
+    """CLI - Terraform Generate Cloudwatch Monitoring with Custom Settings"""
+    cluster_dict = _common.infinitedict()
+    result = monitoring.generate_monitoring('advanced', cluster_dict, CONFIG)
+
+    # Test a the default SNS topic option
```
austinbyers: Nit: grammar

stream_alert_cli/terraform/cloudtrail.py
```diff
@@ -33,8 +33,10 @@ def generate_cloudtrail(cluster_name, cluster_dict, config):
     cloudtrail_module = 'cloudtrail_{}'.format(cluster_name)
 
     enabled_legacy = modules['cloudtrail'].get('enabled')
-    cloudtrail_enabled = modules['cloudtrail'].get('enable_logging')
-    kinesis_enabled = modules['cloudtrail'].get('enable_kinesis')
+
+    cloudtrail_enabled = modules['cloudtrail'].get('enable_logging', True)
+    kinesis_enabled = modules['cloudtrail'].get('enable_kinesis', True)
```
austinbyers: Does this mean we are changing the default values for these configuration options? If so, we probably need to update the documentation

terraform/modules/tf_stream_alert_monitoring/variables.tf
```diff
@@ -19,3 +19,69 @@ variable "lambda_functions" {
   type    = "list"
   default = []
 }
+
+// Lambda Invocation Error Alarm Settings
+
+variable "lambda_invocation_error_threshold" {
+  default = "0"
```
austinbyers: These can (and should, IMO) all be numbers

stream_alert_cli/terraform/cloudtrail.py
jacknagz: No, they should be enabled by default, it's a bug

stream_alert_cli/terraform/monitoring.py
jacknagz: the documentation uses strings for all the alarms, I was just following their convention: https://www.terraform.io/docs/providers/aws/r/cloudwatch_metric_alarm.html

terraform/modules/tf_stream_alert_monitoring/variables.tf
jacknagz: Just being consistent with their examples: https://www.terraform.io/docs/providers/aws/r/cloudwatch_metric_alarm.html

stream_alert_cli/terraform/cloudtrail.py
jacknagz: If we don't put a default, Python will marshal None to JSON, which throws Terraform errors

PR #584, opened 2018-01-26
bumping version from 1.5.0 -> 1.6.0

ryandeivert:
to: @jacknagz or @chunyong-lin 
size: small
resolves N/A

## Changes

* Version bump from `1.5.0` --> `1.6.0`

### DIFF
--- a/stream_alert_cli/__init__.py
+++ b/stream_alert_cli/__init__.py
@@ -1,2 +1,0 @@
-"""Define the version of the CLI."""
-__version__ = '1.4.0'
--- a/manage.py
+++ b/manage.py
@@ -28,8 +28,9 @@ from argparse import Action, ArgumentParser, RawTextHelpFormatter, SUPPRESS as A
 import os
 import string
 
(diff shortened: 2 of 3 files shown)


stream_alert_cli/__init__.py
```diff
@@ -1,2 +1,2 @@
 """Define the version of the CLI."""
-__version__ = '1.4.0'
```
jacknagz: should this just point to the main lib version?

ryandeivert: I wasn't sure if there was a reason the cli was versioned separately. should we remove this?

jacknagz: yeah let's just keep them consistent for now.

austinbyers:
If you want to add a version flag @ryandeivert , it's very easy: https://github.com/airbnb/binaryalert/pull/84

ryandeivert:
@austinbyers done!! thanks :)

PR #592, opened 2018-02-10
[athena] changes to make athena part of a default deployment

ryandeivert:
cc: @airbnb/streamalert-maintainers, @jacknagz 
size: medium
resolves #N/A

## Background

* In order to support some upcoming features, StreamAlert must always have alerts sent to S3 to allow for Athena querying. Therefore, the Athena partition refresh function is no longer optional and should be created upon StreamAlert initialization.

## Changes

* Updating cli to handle deploying athena on init
  * This removes some now unnecessary `manage.py` athena subcommands like `init`, `create-db`, etc.
  * Athena database creation is now handled via Terraform.
* The athena AWS resources are now name-spaced to the user-specified `prefix` as to not conflict if multiple deploys exist in one AWS account.
* Updating athena terraform generate code and updating default local config for athena.
* Three additional options related to the athena function are now configurable by the user (within the `athena_partition_refresh_config` in `lambda.json`:
  * `database_name`: the name of the Athena database to use. (default is: `<prefix>_streamalert`)
  * `results_bucket `: the S3 bucket to use for Athena query results and metastore storage (default is: `<prefix>.streamalert.athena-results`)
  * `queue_name`: the name of the sqs queue to use for bucket notifications (default is: `<prefix>_streamalert_athena_data_bucket_notifications`)

### Other Changes

* Preliminary doc updates to go with changes. Another wave will follow.
* Syncing carbon black logs schema updates

## Testing

* Updating some unit tests for changes.
* Deployed in test account end-to-end and tested to make sure alerts searchable in S3.


stream_alert/rule_processor/firehose.py
```diff
@@ -49,7 +49,10 @@ class StreamAlertFirehose(object):
     def __init__(self, region, firehose_config, log_sources):
         self._firehose_client = boto3.client('firehose', region_name=region)
         # Expand enabled logs into specific subtypes
-        self._enabled_logs = self._load_enabled_log_sources(firehose_config, log_sources)
+
+        self._enabled_logs = []
+        if firehose_config:
```
jacknagz: what prompted this change? the code cannot run without the loaded config, so it may as well just return or raise

stream_alert_cli/manage_lambda/deploy.py
```diff
@@ -149,5 +156,14 @@ def deploy(options, config):
...
         return
 
+    if 'athena' in processors:
+        # create the streamalerts table since terraform does not support this
+        # See: https://github.com/terraform-providers/terraform-provider-aws/issues/1486
+        alerts_bucket = '{}.streamalerts'.format(config['global']['account']['prefix'])
+
+        athena_opts = namedtuple('AthenaOptions', ['bucket', 'refresh_type'])
+        opts = athena_opts(alerts_bucket, 'add_hive_partition')
+        athena_create_table(opts, 'alerts', config)
```
jacknagz: shouldn't this just happen on initialization?

stream_alert_cli/terraform/athena.py
```diff
@@ -33,8 +33,25 @@ def generate_athena(config):
     for refresh_type in athena_config['refresh_type']:
         data_buckets.update(set(athena_config['refresh_type'][refresh_type]))
 
+    prefix = config['global']['account']['prefix']
+    database = athena_config.get('database_name', '').strip()
```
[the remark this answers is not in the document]
ryandeivert: there was a very specific reason I did this, but I'm failing to recall what that was right now. I think I had a check in here that would remove any whitespace, since spaces are invalid in database names. I'll review and make changes

terraform/modules/tf_stream_alert_athena/main.tf
```diff
@@ -22,6 +22,32 @@ resource "aws_lambda_function" "athena_partition_refresh" {
...
+  }
+
+  logging {
+    target_bucket = "${var.s3_logging_bucket}"
+    target_prefix = "${var.results_bucket}/"
+  }
+}
+
+// Athena Database: streamalert
+resource "aws_athena_database" "streamalert" {
```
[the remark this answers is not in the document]
ryandeivert: I'm not sure but I'm assuming they would get an error and have to choose a different name (part of the reason we're allowing users to specify their own DB names). But also, we discussed early on that these big refactors are going to be `breaking` with no backwards compatibility...?

stream_alert_cli/manage_lambda/deploy.py
ryandeivert: Yes & no - so the thought with putting it here was that the `create_table` command will first check if the table already exists and do nothing if it does. Doing it here vs the `init` workflow ensures the table gets created if it wasn't previously.. but now, as I'm typing this out, I'm not sure this is the best approach.

stream_alert/rule_processor/firehose.py
ryandeivert: We can talk about this in person but this was added because the default global config does not contain a `firehose` config.. yet this class is very tightly coupled to the table creation. Would be better alternative be to have a default `firehose` config that has an empty list of `enabled_logs`?

stream_alert_cli/athena/handler.py
```diff
@@ -41,25 +41,7 @@
...
-    create_db_success, create_db_result = athena_client.run_athena_query(
-        query='CREATE DATABASE streamalert')
-
-    if create_db_success and create_db_result['ResultSet'].get('Rows'):
-        LOGGER_CLI.info('streamalert database successfully created!')
-        LOGGER_CLI.info('results: %s', create_db_result['ResultSet']['Rows'])
-
-
-def rebuild_partitions(athena_client, options, config):
+def rebuild_partitions(options, config):
```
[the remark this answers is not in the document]
ryandeivert: No, I think I need to make another change for that to work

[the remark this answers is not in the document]
jacknagz: The check for `options.type` should stay in there then

terraform/modules/tf_stream_alert_athena/main.tf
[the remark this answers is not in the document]
jacknagz: I'd test it and understand what will happen, because any existing Athena+SA user could potentially have issues.  There is a pretty easy Terraform workaround, and it's to import the resource.  You can proactively build this functionality into the CLI as well

stream_alert_cli/manage_lambda/deploy.py
jacknagz: Yeah, it seems a little overkill to do this on every deploy

stream_alert_cli/athena/handler.py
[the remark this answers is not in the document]
ryandeivert: I'll add that back but as an optional flag that defaults to data

docs/source/athena-arch.rst
```diff
@@ -1,16 +1,12 @@
-Athena Deployment
-=================
+Athena Architecture
+===================
 
 Overview
 --------
 
-After the initial `Athena Setup <athena-setup.html>`_, a new Lambda function must be deployed.
+The Athena Partition Refresh function exists to periodically refresh Athena tables, enabling the searchability of data.
```
austinbyers: I don't think searchability is a proper word. In any case, this sentence is kinda awkward. Maybe something along the lines of "The Athena Refresh function periodically refreshes your Athena tables, giving you the ability to run SQL-style queries over historical log and alert data"

docs/source/athena-arch.rst
```diff
@@ -90,20 +83,18 @@ Key                                  Required  Default                Descriptio
...
-After configuring the above settings, deploy the Lambda function:
+If any of the settings above are changed from the initialized defaults, the Lambda function will need to be deployed in order for them to take effect:
 
 .. code-block:: bash
 
   $ python manage.py lambda deploy --processor athena
 
-This will create all of the underlying infrastructure to automatically refresh Athena tables.
-
 Going forward, if the deploy flag ``--processor all`` is used, it will redeploy this function along with the ``rule_processor`` and ``alert_processor``.
```
austinbyers: "Going forward" no longer makes sense

terraform/modules/tf_stream_alert_athena/main.tf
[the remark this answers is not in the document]
ryandeivert: do we do any terraform importing currently?

docs/source/athena-setup.rst
```diff
@@ -34,33 +34,13 @@ Concepts
 Getting Started
 ---------------
 
-To get started with Athena, run the following commands:
+By default, Athena will be enabled for searching alerts.
```
austinbyers: "By default" implies there is an option to disable

docs/source/athena-user-guide.rst
```diff
@@ -11,7 +11,7 @@ Concepts
 Querying Data
 -------------
 
-After completion of the `Athena Setup <athena-setup.html>`_ and `Athena Deploy <athena-deploy.html>`_, alerts generated from StreamAlert are now searchable in the Athena UI.
+Adding an output for rules that sends alerts generated by StreamAlert to the ``alerts`` S3 bucket will allow them to be searchable in the Athena UI.
```
austinbyers: Isn't this output going to be automatically added to all rules?

terraform/modules/tf_stream_alert_athena/main.tf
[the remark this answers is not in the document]
jacknagz: Not automatically, but it should be pretty trivial to discover if the `streamalert` database exists and then import it in so it's manageable

stream_alert/athena_partition_refresh/main.py
```diff
@@ -139,14 +137,21 @@ def __init__(self, config, **kwargs):
...
+        athena_config = self.config['lambda']['athena_partition_refresh_config']
+
+        # GEt the S3 bucket to store Athena query results
+        results_bucket = athena_config.get('results_bucket', '').strip()
+        if results_bucket == '':
+            self.athena_results_bucket = 's3://{}.streamalert.athena-results'.format(self.prefix)
+        elif results_bucket[:5] != 's3://':
+            self.athena_results_bucket = 's3://{}'.format(results_bucket)
+        else:
+            self.athena_results_bucket = results_bucket
```
austinbyers: Instead of `if ... != ..., else` it's usually easier to read the positive case first (also you can use `startswith`):

```python
if not results_bucket:
    ...
elif results_bucket.startswith('s3://'):
    self.athena_results_bucket = results_bucket
else:
    self.athena_results_bucket = 's3://{}'.format(results_bucket)
```

terraform/modules/tf_stream_alert_athena/main.tf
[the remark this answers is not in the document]
ryandeivert: I'm feeling like this may be a rabbit hole but will flag it to investigate

stream_alert/rule_processor/firehose.py
jacknagz: that's a reasonable workaround

austinbyers:
Quick question @ryandeivert - you're trying to merge this into master? I thought we wanted to create a development branch to merge into?

docs/source/athena-user-guide.rst
ryandeivert: Yes good call out - but that change will happen in the next round of changes (pt 2 of this) so I just updated to doc to reflect reality for now.

ryandeivert:
hey @austinbyers you're right, ideally we should get a dev workflow in place for these breaking changes. would you be open to setting up a meeting to discuss the best approach for this?

stream_alert/athena_partition_refresh/main.py
ryandeivert: good suggestions!

docs/source/athena-arch.rst
ryandeivert: the internets say otherwise :) but you're right this could be improved - I'll make a change for it!

austinbyers: Haha I was just judging by the spellcheck as I was typing it. But it does seem to be in many dictionaries

manage.py
```diff
@@ -1151,18 +1151,11 @@ def _add_athena_subparser(subparsers):
...
     manage.py athena create-table         Create an Athena table
 
 Examples:
 
-    manage.py athena create-db
-
-    manage.py athena create-table --type alerts --bucket s3.bucket.name --refresh_type add_hive_partition
-
-    manage.py athena create-table --type data --bucket s3.bucket.name --refresh_type add_hive_partition --table_name my_athena_table
+    manage.py athena create-table --bucket s3.bucket.name --refresh_type add_hive_partition --table_name my_athena_table
```
chunyong-lin: Should we support `interval` setting in the CLI?

stream_alert/athena_partition_refresh/main.py
```diff
@@ -155,6 +160,16 @@ def __init__(self, config, **kwargs):
             results_key_prefix,
             datetime.now().strftime('%Y/%m/%d'))
 
+    @property
+    def sa_database(self):
+        """Return the name of the streamalert database. This can be overridden in the config"""
+        database = self.config['lambda']['athena_partition_refresh_config'].get('database_name', '')
+        database = database.replace(' ', '') # strip any spaces which are invalid database names
+        if database == '':
```
chunyong-lin: Use  `if not database:` will cover all empty value cases.

stream_alert/athena_partition_refresh/main.py
```diff
@@ -418,13 +434,24 @@ def __init__(self, config):
 
         self.setup()
 
+    @property
+    def queue_name(self):
+        """Return the name of the sqs queue to use. This can be overridden in the config"""
+        queue = self.config['lambda']['athena_partition_refresh_config'].get('queue_name', '')
+        queue = queue.replace(' ', '') # strip any spaces which are invalid queue names
+        if queue == '':
```
chunyong-lin: `if not queue:`

stream_alert/athena_partition_refresh/main.py
```diff
@@ -605,10 +632,13 @@ def handler(*_):
...
     # Check that the 'streamalert' database exists before running queries
     if not stream_alert_athena.check_database_exists():
-        raise AthenaPartitionRefreshError('The \'streamalert\' database does not exist')
+        raise AthenaPartitionRefreshError(
+            'The \'{}\' database does not exist'.format(stream_alert_athena.sa_database)
+        )
 
     if not stream_alert_athena.add_hive_partition(s3_buckets_and_keys):
-        raise AthenaPartitionRefreshError('Adding partition(s) has failed')
+        LOGGER.error('Failed to add hive partition(s)')
```
chunyong-lin: Why don't you want to raise the exception here? Any concern doing that?

stream_alert/athena_partition_refresh/main.py
ryandeivert: This code will probably change but great suggestion!

stream_alert/athena_partition_refresh/main.py
ryandeivert: So this was changed because the `add_hive_partition` can return `False` even if there wasn't real error that requires an `exception`ional response. Errors that should raise an exception now raise directly from within `add_hive_partition`

ryandeivert:
closing this since #599 is open to address this

PR #599, opened 2018-02-14
[athena] changes to make athena part of a default deployment

ryandeivert:
cc: @airbnb/streamalert-maintainers, @jacknagz 
size: large
resolves #N/A

## Background

* In order to support some upcoming features, StreamAlert must always have alerts sent to S3 to allow for Athena querying. Therefore, the Athena partition refresh function is no longer optional and should be created upon StreamAlert initialization.

## Changes

* Updating cli to handle deploying athena on init
  * This removes some now unnecessary `manage.py` athena subcommands like `init`, `create-db`, etc.
  * Athena database creation is now handled via Terraform.
* The athena AWS resources are now name-spaced to the user-specified `prefix` as to not conflict if multiple deploys exist in one AWS account.
* Updating athena terraform generate code and updating default local config for athena.
* Three additional options related to the athena function are now configurable by the user (within the `athena_partition_refresh_config` in `lambda.json`:
  * `database_name`: the name of the Athena database to use. (default is: `<prefix>_streamalert`)
  * `results_bucket `: the S3 bucket to use for Athena query results and metastore storage (default is: `<prefix>.streamalert.athena-results`)
  * `queue_name`: the name of the sqs queue to use for bucket notifications (default is: `<prefix>_streamalert_athena_data_bucket_notifications`)

### Other Changes

* Preliminary doc updates to go with changes. Another wave will follow.
* Syncing carbon black logs schema updates

## Testing

* Updating some unit tests for changes.
* Deployed in test account end-to-end and tested to make sure alerts searchable in S3.


stream_alert/athena_partition_refresh/main.py
```diff
@@ -139,14 +137,21 @@ def __init__(self, config, **kwargs):
...
-        self.athena_results_bucket = 's3://aws-athena-query-results-{}-{}'.format(
-            self.config['global']['account']['aws_account_id'],
-            self.config['global']['account']['region'])
+        athena_config = self.config['lambda']['athena_partition_refresh_config']
+
+        # GEt the S3 bucket to store Athena query results
+        results_bucket = athena_config.get('results_bucket', '').strip()
+        if results_bucket == '':
+            self.athena_results_bucket = 's3://{}.streamalert.athena-results'.format(self.prefix)
+        elif results_bucket[:5] != 's3://':
```
austinbyers: `results_bucket.startswith('s3://')`

stream_alert_cli/athena/handler.py
```diff
@@ -41,25 +41,7 @@
...
-    create_db_success, create_db_result = athena_client.run_athena_query(
-        query='CREATE DATABASE streamalert')
-
-    if create_db_success and create_db_result['ResultSet'].get('Rows'):
-        LOGGER_CLI.info('streamalert database successfully created!')
-        LOGGER_CLI.info('results: %s', create_db_result['ResultSet']['Rows'])
-
-
-def rebuild_partitions(athena_client, options, config):
+def rebuild_partitions(options, config):
```
jacknagz: This method is still needs the constraint on `options.type == data`, since it won't work for the `alerts` table

stream_alert_cli/athena/handler.py
```diff
@@ -199,17 +191,15 @@ def _construct_create_table_statement(schema, table_name, bucket):
         bucket=bucket)
 
 
-def create_table(athena_client, options, config):
+def create_table(options, table_type, config):
```
jacknagz: I might be missing something, but is there an added benefit of passing in `table_type` as an arg if it's in `options.type`?

stream_alert_cli/manage_lambda/deploy.py
```diff
@@ -149,5 +156,14 @@ def deploy(options, config):
...
         return
 
+    if 'athena' in processors:
+        # create the streamalerts table since terraform does not support this
+        # See: https://github.com/terraform-providers/terraform-provider-aws/issues/1486
+        alerts_bucket = '{}.streamalerts'.format(config['global']['account']['prefix'])
+
+        athena_opts = namedtuple('AthenaOptions', ['bucket', 'refresh_type'])
+        opts = athena_opts(alerts_bucket, 'add_hive_partition')
+        athena_create_table(opts, 'alerts', config)
```
jacknagz: I was under the impression we were revisiting table creation on every deploy per the last PR convo, is that not the case?

stream_alert_cli/config.py
```diff
@@ -116,26 +70,29 @@ def set_prefix(self, prefix):
             LOGGER_CLI.error('Prefix cannot contain underscores')
             return
 
+        tf_state_bucket = '{}.streamalert.terraform.state'.format(prefix)
```
jacknagz: (not related to this line)

Any reason to delete the `generate_athena` method? It could still be useful for current users who don't have an Athena config

stream_alert_cli/athena/handler.py
ryandeivert: the idea was this would allow for a more generic function with a better interface by accepting a string for the table_type.

the only property of `options` that was accessed within this function was `type`, which means if any caller wanted to use this function it essentially has to create a namedtuple first that would have a `type` property corresponding to the table type. it's fine to keep as is, just makes using it more difficult

ryandeivert:
@jacknagz & @austinbyers I should clarify that this PR is a direct mirror of #592 and your feedback there was **not** addressed yet. This PR was simply to go against the release branch instead of master. Your comments lead me to believe you thought I had addressed previous PR feedback in this PR, which is not the case. Sorry for the misunderstanding!!

ryandeivert:
hey @austinbyers and @jacknagz PTAL. I've updated the branch and addressed the feedback from previous comments

stream_alert_cli/config.py
austinbyers: We don't auto-generate any other part of the config (e.g. Lambda template code for the alert processor or alert merger), and I think we should be consistent. So either we should auto-generate all of it or none of it, and IMO it would be over-engineering to auto-generate every part of the config an upgrading user might not have (because they'll likely need to change the config anyway). With documentation, users can just update their config manually. So my vote is to get rid of the `generate_athena` method

#simplify 💣

stream_alert_cli/terraform/handler.py
```diff
@@ -65,96 +65,79 @@ def terraform_handler(options, config):
...
-        # Destroy all of the infrastructure
-        if not tf_runner(action='destroy', auto_approve=True):
-            return
+    # Use a named tuple to match the 'processor' attribute in the argparse options
+    deploy_opts = namedtuple('DeployOptions', ['processor', 'clusters'])
 
-        # Remove old Terraform files
-        terraform_clean(config)
+    LOGGER_CLI.info('Deploying Lambda Functions')
+    # deploy both lambda functions
```
austinbyers: We don't really need this comment

tests/unit/stream_alert_athena_partition_refresh/test_main.py
```diff
@@ -89,15 +89,10 @@
...
-            "enabled": True,
-            "refresh_type": {
-                "repair_hive_table": {
-                    "unit-testing.streamalerts": "alerts"
-                },
-                "add_hive_partition": {
-                    "unit-testing.streamalerts": "alerts",
-                    "unit-testing.streamalert.data": "data"
-                }
+            'enabled': True,
```
austinbyers: This is no longer used in the config, let's also remove it from the test

manage.py
```diff
@@ -1166,18 +1166,10 @@ def _add_athena_subparser(subparsers):
...
-    manage.py athena create-table         Create an Athena table
-
-Examples:
-
-    manage.py athena create-db
-
-    manage.py athena create-table --type alerts --bucket s3.bucket.name --refresh_type add_hive_partition
-
-    manage.py athena create-table --type data --bucket s3.bucket.name --refresh_type add_hive_partition --table_name my_athena_table
+    manage.py athena init                     Initialize the Athena base config (for legacy support)
```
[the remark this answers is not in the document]
ryandeivert: thanks @austinbyers, that was my impression as well

stream_alert_cli/config.py
ryandeivert: @austinbyers and @jacknagz maybe we need to discuss this because I'm not sure what the best approach is. my initial impression was to nix it as well (as seen in my commit history). let's chat today

tests/unit/stream_alert_athena_partition_refresh/test_main.py
ryandeivert: yes good catch - I'll remove :)

stream_alert_cli/terraform/athena.py
```diff
@@ -29,23 +29,36 @@ def generate_athena(config):
...
-    data_buckets = set()
-    for refresh_type in athena_config['refresh_type']:
-        data_buckets.update(set(athena_config['refresh_type'][refresh_type]))
+    data_buckets = list(set(athena_config['buckets']))
+
+    prefix = config['global']['account']['prefix']
+    database = athena_config.get('database_name', '{}_streamalert'.format(prefix))
+
+    results_bucket_name = athena_config.get('results_bucket', '').strip()
+    if results_bucket_name == '':
```
chunyong-lin: You may just use `if not results_bucket_name:`.  The empty string is treated as false anyway.

stream_alert_cli/terraform/athena.py
```diff
@@ -29,23 +29,36 @@ def generate_athena(config):
...
+        results_bucket_name = '{}.streamalert.athena-results'.format(prefix)
+
+    queue_name = athena_config.get('queue_name', '').strip()
+    if queue_name == '':
```
chunyong-lin: Same as above, optional to use `if not queue_name:`

manage.py
```diff
@@ -1189,32 +1181,204 @@ def _add_athena_subparser(subparsers):
...
+        usage=athena_create_table_usage,
+        formatter_class=RawTextHelpFormatter,
+        help=ARGPARSE_SUPPRESS)
+
+    athena_create_table_parser.set_defaults(subcommand='create-table')
+
+    _add_default_athena_args(athena_create_table_parser)
+
+    # Validate the provided schema-override options
+    def _validate_override(val):
```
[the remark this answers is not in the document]
ryandeivert: hey @chunyong-lin - `val` will actually not be a set, but will be each individual item _within_ the set.

Good suggestion though - I can add an additional check to make sure the right about of equals-separated-values is provided

manage.py
```diff
@@ -1189,32 +1181,204 @@ def _add_athena_subparser(subparsers):
...
+
+    athena_drop_all_parser.set_defaults(subcommand='drop-all-tables')
+
+    athena_drop_all_parser.add_argument(
+        '--debug',
+        action='store_true',
+        help=ARGPARSE_SUPPRESS)
+
+
+def _add_default_athena_args(athena_parser):
```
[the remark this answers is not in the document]
ryandeivert: I can rename but chose this since the arguments are the 'defaults' related to the actual argparsers that we're constructing here (not the 'defaults' the user will have to provide). Also, you'll notice that not ever argument that is added within the function is in fact 'required' - see the 'debug' param.

stream_alert_cli/terraform/athena.py
ryandeivert: Oh good catch on this - I missed updating this one :)

manage.py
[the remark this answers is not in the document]
ryandeivert: Made the change!

stream_alert/athena_partition_refresh/main.py
```diff
@@ -403,7 +383,7 @@ class StreamAlertSQSClient(object):
         received_messages: A list of receieved SQS messages
         processed_messages: A list of processed SQS messages
     """
-    QUEUENAME = 'streamalert_athena_data_bucket_notifications'
+    DEFAULT_QUEUE_NAME = '{}_streamalert_athena_data_bucket_notifications'
```
jacknagz: This is a pretty long Queue name, can we shorten it? The max length is 80 chars.

stream_alert_cli/athena/handler.py
```diff
@@ -309,22 +297,23 @@ def athena_handler(options, config):
         options (namedtuple): The parsed args passed from the CLI
         config (CLIConfig): Loaded StreamAlert CLI
     """
-    athena_client = StreamAlertAthenaClient(config, results_key_prefix='stream_alert_cli')
```
jacknagz: What's the benefit of removing the instantiation here and doing it in each subcommand's method?

conf/lambda.json
```diff
@@ -23,6 +23,18 @@
       "subnet_ids": []
     }
   },
+  "athena_partition_refresh_config": {
```
jacknagz: is this missing the `results_bucket`?

jacknagz: also `database_name`?

ryandeivert: those options are only needed if the user wants to override the defaults that we use. I can add the defaults here but have some concerns. Mainly, I worry that our config is growing unnecessarily large with superfluous options that the vast majority of users wouldn't need to worry about. What are your thoughts? Have it here or omit it and just document the options config settings well?

stream_alert_cli/athena/handler.py
ryandeivert: This allows any caller using these functions to not have to worry about created an athena client that needs to be passed it.

Example here:
https://github.com/airbnb/streamalert/blob/8d653cb2e423d00fcc640b780fa3675c0f05821f/stream_alert_cli/terraform/handler.py#L121

Also, not every subcommand needs this (ie: `init`)

stream_alert/athena_partition_refresh/main.py
ryandeivert: Yeah we can - I was just mimicking what our queue name is internally

stream_alert_cli/athena/handler.py
```diff
@@ -69,27 +51,26 @@ def rebuild_partitions(athena_client, options, config):
       - Re-create partitions
 
     Args:
-        athena_client (boto3.client): Instantiated CLI AthenaClient
-        options (namedtuple): The parsed args passed from the CLI
+        table (str): The name of the table being rebuilt
+        bucket (str): The s3 bucket to be used as the location for Athena data
+        table_type (str): The type of table being refreshed
+            Types of 'data' and 'alert' are accepted, but only 'data' is implemented
```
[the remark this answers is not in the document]
ryandeivert: Good catch! But this is just for 'rebuilding' part, not simply supporting the alerts

PR #603, opened 2018-02-21
Consolidate to a single alert processor (instead of clustered)

austinbyers:
to: @jacknagz and @ryandeivert 
size: large

## Background

In order to support alert merging, we need to consolidate to a single alert processor instead of one for every cluster. This will also simplify some of StreamAlert's architecture.

## Changes

* Removes alert processor from `stream_alert` Terraform module
* In its place, the alert processor is defined by an instance of the `lambda` module (#596) and a new `alert_processor_iam` module for its permissions.
* Moves alert processor config from `conf/clusters/*` to `conf/lambda.json`
* Change rule processors to point to the single global alert processor (specified as an environment variable)
* Update all affected CLI commands: Lambda deploy and rollback, Terraform generation, etc
* Adds new unit tests for Lambda rollback and alert processor Terraform generation

### Setup and Single Cluster
1. Clone the branch (do not change any files) and setup AWS test account
1. `./manage.py configure prefix global-alert`
1. `./manage.py configure aws_account_id ...`
1. `./manage.py terraform init`
1. Manually create an SNS topic for rule input
1. Manually create an S3 bucket, Lambda function, and Slack webhook for alert outputs
1. `./manage.py output new --service aws-lambda`
1. `./manage.py output new --service aws-s3`
1. `./manage.py output new --service slack`
1. Add a rule with the new outputs
1. Update `conf/prod.json` with the SNS topic input and `conf/lambda.json` with the outputs
1. `./manage.py terraform build` (permissions for Lambda/S3/SNS)
1. `./manage.py lambda deploy -p all`
1. Publish to the SNS topic - alerts landed in Lambda, S3 and Slack

### Multi-Cluster
1. Add 2 more clusters with different SNS inputs (update `conf/clusters` and `conf/sources.json`)
1. `./manage.py terraform build` (had to apply twice for SNS input permissions to take)
1. `./manage.py lambda deploy -p all`
1. Publish to each SNS topic - rule processors from all 3 clusters invoke the same alert processor

### Advanced Alert Processor Configuration
1. Disable metric alarms, move behind VPC, and change other alert processor settings (`conf/lambda.json`)
    * You need to configure Internet access for the VPC - see [AWS Docs](https://docs.aws.amazon.com/lambda/latest/dg/vpc.html) and [terraform-aws-modules/terraform-aws-vpc](https://github.com/terraform-aws-modules/terraform-aws-vpc)
1. `./manage.py terraform build` (destroys Alert Processor and recreates in VPC)
1. `./manage.py lambda deploy -p alert`
1. Publish to SNS topics - alert processor invoked
    * So far, I haven't succeeded in getting the alert processor to connect to resources outside of the VPC in my test account
1. Enable custom metrics - `./manage.py metrics -e -f alert`
    * The alert processor doesn't actually implement custom metrics, but this preserves the existing CLI functionality
1. `./manage.py lambda deploy -p alert`

### Rollback and Destroy
1. `./manage.py lambda rollback -p alert`
1. `./manage.py lambda rollback -p rule`
1. `./manage.py lambda rollback -p rule --cluster prod`
1. `./manage.py terraform destroy`


stream_alert/rule_processor/sink.py
```diff
@@ -31,10 +32,8 @@ def __init__(self, env):
             env (dict): loaded dictionary containing environment information
         """
         self.env = env
-        self.client_lambda = boto3.client('lambda',
-                                          region_name=self.env['lambda_region'])
-        self.function = self.env['lambda_function_name'].replace(
-            '_streamalert_rule_processor', '_streamalert_alert_processor')
+        self.client_lambda = boto3.client('lambda', region_name=self.env['lambda_region'])
+        self.function = os.environ['ALERT_PROCESSOR']
```
ryandeivert: ++

stream_alert_cli/manage_lambda/rollback.py
```diff
@@ -14,9 +14,86 @@
...
+    tf_targets = []
+
+    for cluster in clusters:
+        apps_config = config['clusters'][cluster]['modules'].get('stream_alert_apps', {})
+        for lambda_name, lambda_config in apps_config.iteritems():
+            clustered_name = '{}_{}'.format(lambda_name, cluster)
+            if _try_decrement_version(lambda_config, clustered_name):
+                tf_targets.append('module.{}'.format(clustered_name))
+
+    print('Apps Targets: ', tf_targets)
```
ryandeivert: was this a missed debugging line?

stream_alert_cli/test.py
```diff
@@ -907,6 +907,9 @@ def run_tests(options, context):
             options (namedtuple): CLI options (debug, processor, etc)
             context (namedtuple): A constructed aws context object
         """
+        os.environ['ALERT_PROCESSOR'] = '{}_{}'.format(
```
ryandeivert: Consider adding a comment here so it's clear why this is necessary

stream_alert_cli/manage_lambda/rollback.py
austinbyers: Yep! Good catch, thanks

stream_alert_cli/test.py
austinbyers: Good idea, thanks

terraform/modules/tf_alert_processor_iam/main.tf
```diff
@@ -0,0 +1,105 @@
+// Permissions specific to the alert processor: decrypting secrets, sending alerts to outputs
+
+locals {
+  firehose_arn_prefix = "arn:aws:firehose:${var.region}:${var.account_id}"
+  lambda_arn_prefix   = "arn:aws:lambda:${var.region}:${var.account_id}:function"
+}
+
+resource "aws_iam_role_policy" "base" {
+  name   = "DecryptSecretsAndUseDefaultOutputs"
```
jacknagz: can we split `DecryptSecrets` and `DefaultOutputs` into two policies? I tend to follow the single responsibility model for policies in order to easily understand what an IAM role can do (via the UI especially)

terraform/modules/tf_stream_alert/metrics.tf
```diff
@@ -13,21 +13,6 @@ resource "aws_cloudwatch_log_metric_filter" "rule_processor_cw_metric_filters" {
   }
 }
 
-// CloudWatch metric filters for the alert processor
```
jacknagz: Mabye I'm missing something, but is this in the new set of modules?

terraform/modules/tf_alert_processor_iam/main.tf
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+// Allow the Alert Processor to invoke the configured output Lambda functions
+resource "aws_iam_role_policy" "invoke_lambda_outputs" {
+  count = "${length(var.output_lambda_functions)}"
+  name  = "LambdaInvoke_${element(var.output_lambda_functions, count.index)}"
+  role  = "${var.role_id}"
+
+  policy = <<EOF
```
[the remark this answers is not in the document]
austinbyers: No, I tried :(

You can create multiple `data` resources with `count` as you can for other resources, but then you can't access them using the same array/splat syntax. This may be fixed in TF 0.11, I'm not sure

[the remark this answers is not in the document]
austinbyers: Initially, I didn't think that was possible - I tried. But I found a different syntax which seems to do the trick

terraform/modules/tf_stream_alert/metrics.tf
austinbyers: It is not - there are no custom metrics in the alert processor, so this resource is never used. I'm removing it for now - we can always re-add it if we ever do put custom metrics in the alert processor

austinbyers:
PTAL @jacknagz @ryandeivert - your comments have been addressed

PR #604, opened 2018-02-21
Speed up deploys with -auto-approve

### LINKED ISSUE #194: Terraform: Do Not Refresh State For Every Deploy Step

austinbyers:
to: @jacknagz 
size: small
resolves #194  (supersedes #359)

## Background

Two separate issues unnecessarily slow down StreamAlert deploys:

- Most Terraform operations are implemented as a `plan` followed by an `apply` after user confirmation. This requires Terraform to parse the config and walk its graph twice. The built-in `-auto-approve=false` accomplishes the same plan + user confirmation, but faster (because Terraform is only invoked once)
- There are some cases (e.g. `lambda deploy`) where we unnecessarily refresh between Terraform commands

## Changes

* `tf_runner` accepts 4 keyword arguments: `action`, `refresh`, `auto_approve`, and `targets`
* `-auto-approve=false` is used instead of requesting user confirmation ourselves. Besides speed, this has another benefit: the user need not approve empty changes when Terraform is up to date
* Applying the Lambda alias update after a Lambda deploy now uses `refresh=False` and `auto_approve=True` 
* The final Terraform build during `./manage.py terraform init` uses `refresh=False`

## Testing
For each command, I tested with the old and new deploy process with 3 clusters with the default configurations.

* `./manage.py terraform init` - 190 seconds down to 145 seconds (23.7% faster)
* `./manage.py lambda deploy` - 57 seconds down to 37 seconds (35% faster)
* `./manage.py terraform destroy` - unchanged

### DIFF
--- a/stream_alert_cli/helpers.py
+++ b/stream_alert_cli/helpers.py
@@ -129,7 +129,7 @@ def tf_runner(action='apply', refresh=True, auto_approve=False, targets=None):
         LOGGER_CLI.info('Destroying infrastructure')
         tf_command.append('-force={}'.format(str(auto_approve).lower()))
     else:
-        LOGGER_CLI.info('{} changes'.format('Applying' if auto_approve else 'Planning'))
+        LOGGER_CLI.info('%s changes', 'Applying' if auto_approve else 'Planning')
         tf_command.append('-auto-approve={}'.format(str(auto_approve).lower()))
 
     if targets:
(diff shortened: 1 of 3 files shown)


jacknagz:
@austinbyers 

```
W:132,24: Use % formatting in logging functions and pass the % parameters as arguments (logging-format-interpolation)
```

ryandeivert:
thanks for this!! @austinbyers

PR #612, opened 2018-02-27
Fork alerts to a Dynamo table

austinbyers:
to: @jacknagz and @ryandeivert 
size: medium

## Background

In order to support alert merging (and more reliable alert delivery), we need to start keeping state about the alerts. We will do this with a Dynamo table that sits in between the Rule Processor and Alert Processor.

## Changes
This PR creates the new alerts table and forks all alerts to the table *in addition to the alert processor*, so we can test the alerts table in production. We will sever the rule processor - alert processor connection in a subsequent pull request after we verify that writing to the table is working as expected.

* Fixes a bug in the `tf_alert_processor_iam` module when the output lists are empty
* Replaces the `LogsAndMetrics` custom policy in the `tf_lambda` module with the managed `AWSLambdaBasicExecutionRole` policy
* Adds the alerts table to the `tf_globals` module, with capacity configurable in `conf/global.json`
* Changes and renames `sink.py` in the rule processor to write alerts in batches to Dynamo after sending them to Lambda.

## Testing

Deployed in a test account, alerts are sent to both the alert processor and Dynamo as expected. The Dynamo entry looks like this:

screen shot 2018-02-28 at 2 58 31 pm

### DIFF
--- a/stream_alert/rule_processor/alert_forward.py
+++ b/stream_alert/rule_processor/alert_forward.py
@@ -29,6 +29,7 @@ from stream_alert.rule_processor import LOGGER
 class AlertForwarder(object):
     """Sends alerts to the Alert Processor and the alerts Dynamo table."""
     # TODO: Do not send to Alert Processor after Alert Merger is implemented
+    BACKOFF_MAX_RETRIES = 6
 
     def __init__(self, env):
         """Initialize the Forwarder with the boto3 clients and resource names.
@@ -41,6 +42,9 @@ class AlertForwarder(object):
         self.client_lambda = boto3.client('lambda', region_name=self.env['lambda_region'])
         self.function = os.environ['ALERT_PROCESSOR']
         self.table = os.environ['ALERT_TABLE']
+
+        # Keep track of unprocessed items when retrying batch_write_item()
+        self.unprocessed_items = None
 
     def _send_to_lambda(self, alerts):
         """DEPRECATED: Invoke Alert Processor directly
@@ -104,7 +108,7 @@ class AlertForwarder(object):
             (dict) The constructed request for batch_write_item, containing <= 25 alerts.
                 Maps table name to a list of requests.
         """
-        for i in xrange(0, len(alerts), batch_size):
+        for i in range(0, len(alerts), batch_size):
             batch = alerts[i:i+batch_size]
             yield {
                 self.table: [
@@ -120,7 +124,9 @@ class AlertForwarder(object):
                                 'RuleDescription': {'S': alert['rule_description']},
                                 'Outputs': {'SS': alert['outputs']},
                                 # Compact JSON encoding (no extra spaces)
-                                'Record': {'S': json.dumps(alert['record'], separators=(',', ':'))}
+                                'Record': {'S': json.dumps(alert['record'], separators=(',', ':'))},
+                                # TODO: Remove TTL after alert merger is implemented
+                                'TTL': {'N': str(int(time.time()) + 7200)}  # 2 hour TTL
                             }
                         }
                     }
@@ -128,45 +134,54 @@ class AlertForwarder(object):
                 ]
             }
 
-    @backoff.on_exception(backoff.expo, ClientError, max_tries=5, jitter=backoff.full_jitter,
-                          on_backoff=backoff_handlers.backoff_handler,
-                          on_success=backoff_handlers.success_handler,
-                          on_giveup=backoff_handlers.giveup_handler)
-    def _batch_write(self, request_items, max_attempts=5):
+    def _batch_write(self):
         """Write a batch of alerts to Dynamo, retrying with exponential backoff for failed items.
 
-        Args:
-            request_items (dict): The generated RequestItems dict from _alert_batches().
-            max_attempts (int): Maximum number of times to retry UnprocessedItems.
-
         Returns:
-            (bool) True if the batch write was eventually successful, False otherwise.
+            (bool) True if *all* items were written successfully, False otherwise.
         """
-        for attempt in xrange(1, max_attempts + 1):
-            response = self.client_dynamo.batch_write_item(RequestItems=request_items)
+        @backoff.on_predicate(backoff.expo,
+                              max_tries=self.BACKOFF_MAX_RETRIES, jitter=backoff.full_jitter,
+                              on_backoff=backoff_handlers.backoff_handler,
+                              on_success=backoff_handlers.success_handler,
+                              on_giveup=backoff_handlers.giveup_handler)
+        @backoff.on_exception(backoff.expo, ClientError,
+                              max_tries=self.BACKOFF_MAX_RETRIES, jitter=backoff.full_jitter,
+                              on_backoff=backoff_handlers.backoff_handler,
+                              on_success=backoff_handlers.success_handler,
+                              on_giveup=backoff_handlers.giveup_handler)
+        def decorated_batch_write(cls):
+            """batch_write_item with the unprocessed_items from the AlertForwarder instance.
 
-            # If Dynamo experiences an internal error, unprocessed items are listed in the response.
-            # AWS recommends retrying unprocessed items in a loop with exponential backoff.
-            request_items = response['UnprocessedItems']
-            if request_items:
-                LOGGER.warn(
-                    'Batch write failed: %d alerts were not written (attempt %d/%d)',
-                    len(request_items[self.table]), attempt, max_attempts)
-                # Simple exponential backoff: Sleep 0.5, 1, 2, 4 and 8 seconds.
-                time.sleep(0.25 * 2 ** attempt)
-                attempt += 1
-            else:
-                return True
+            There are 2 different errors to handle here:
+                (1) If Dynamo is unresponsive, a boto ClientError will be raised.
+                (2) The batch_write_item operation can fail halfway through, in which case the
+                    unprocessed items are returned in the response. In this case, unprocessed items
+                    are stored in the class instance, and we return False.
+                    The backoff.on_predicate will automatically retry with any Falsey value, and
+                    batch_write will run again, but only with the remaining unprocessed items.
 
-        return False
+            Args:
+                cls (AlertForwarder): Instance of the AlertForwarder
+
+            Returns:
+                (bool) True if the batch write succeeded, False if there were UnprocessedItems.
+            """
+            response = cls.client_dynamo.batch_write_item(RequestItems=cls.unprocessed_items)
+            cls.unprocessed_items = response['UnprocessedItems']
+            return len(cls.unprocessed_items) == 0
+
+        return decorated_batch_write(self)
 
     def _send_to_dynamo(self, alerts):
         """Write alerts in batches to Dynamo."""
         for batch_num, batch in enumerate(self._alert_batches(alerts), start=1):
-            LOGGER.info('Sending batch #%d to Dynamo with %d alert(s)',
+            LOGGER.info('Sending batch %d to Dynamo with %d alert(s)',
                         batch_num, len(batch[self.table]))
-            if not self._batch_write(batch):
-                LOGGER.error('Unable to save alert batch %s', json.dumps(batch))
+            self.unprocessed_items = batch
+            if not self._batch_write():
+                LOGGER.error('Unable to save alert batch; unprocessed items remain: %s',
+                             json.dumps(self.unprocessed_items))
 
     def send_alerts(self, alerts):
         """Send alerts to the Alert Processor and to the alerts Dynamo table.
@@ -176,7 +191,8 @@ class AlertForwarder(object):
         """
         self._send_to_lambda(alerts)
 
-        # While we are testing this, exceptions should be logged but not raise errors.
+        # For now, don't blow up the rule processor if there is a problem sending to Dynamo.
+        # TODO: Remove/refine broad exception handling once tested.
         try:
             self._send_to_dynamo(alerts)
         except Exception:  # pylint: disable=broad-except
--- a/stream_alert/rule_processor/alert_forward.py
+++ b/stream_alert/rule_processor/alert_forward.py
@@ -26,11 +26,12 @@ from stream_alert.shared import backoff_handlers
 from stream_alert.rule_processor import LOGGER
 
 
-class StreamSink(object):
-    """StreamSink class is used for sending actual alerts to the alert processor"""
+class AlertForwarder(object):
+    """Sends alerts to the Alert Processor and the alerts Dynamo table."""
+    # TODO: Do not send to Alert Processor after Alert Merger is implemented
 
     def __init__(self, env):
-        """StreamSink initializer
+        """Initialize the Forwarder with the boto3 clients and resource names.
 
         Args:
             env (dict): loaded dictionary containing environment information
@@ -41,8 +42,8 @@ class StreamSink(object):
         self.function = os.environ['ALERT_PROCESSOR']
         self.table = os.environ['ALERT_TABLE']
 
-    def _sink_lambda(self, alerts):
-        """Original sink: Invoke alert processor directly
+    def _send_to_lambda(self, alerts):
+        """DEPRECATED: Invoke Alert Processor directly
 
         Sends a message to the alert processor with the following JSON format:
             {
@@ -159,7 +160,7 @@ class StreamSink(object):
 
         return False
 
-    def _sink_dynamo(self, alerts):
+    def _send_to_dynamo(self, alerts):
         """Write alerts in batches to Dynamo."""
         for batch_num, batch in enumerate(self._alert_batches(alerts), start=1):
             LOGGER.info('Sending batch #%d to Dynamo with %d alert(s)',
@@ -167,16 +168,16 @@ class StreamSink(object):
             if not self._batch_write(batch):
                 LOGGER.error('Unable to save alert batch %s', json.dumps(batch))
 
-    def sink(self, alerts):
-        """Sink triggered alerts from the StreamRules engine.
+    def send_alerts(self, alerts):
+        """Send alerts to the Alert Processor and to the alerts Dynamo table.
 
         Args:
             alerts (list): A list of dictionaries representing json alerts.
         """
-        self._sink_lambda(alerts)
+        self._send_to_lambda(alerts)
 
         # While we are testing this, exceptions should be logged but not raise errors.
         try:
-            self._sink_dynamo(alerts)
+            self._send_to_dynamo(alerts)
         except Exception:  # pylint: disable=broad-except
             LOGGER.exception('Error saving alerts to Dynamo')
--- a/stream_alert/rule_processor/alert_forward.py
+++ b/stream_alert/rule_processor/alert_forward.py
@@ -48,7 +48,7 @@ class AlertForwarder(object):
         self.unprocessed_items = None
 
     def _send_to_lambda(self, alerts):
-        """DEPRECATED: Invoke Alert Processor directly
+        """Invoke Alert Processor directly
 
         Sends a message to the alert processor with the following JSON format:
             {
@@ -148,7 +148,7 @@ class AlertForwarder(object):
                               on_backoff=backoff_handlers.backoff_handler,
                               on_success=backoff_handlers.success_handler,
                               on_giveup=backoff_handlers.giveup_handler)
-        def decorated_batch_write(cls):
+        def decorated_batch_write(forwarder):
             """batch_write_item with the unprocessed_items from the AlertForwarder instance.
 
             There are 2 different errors to handle here:
@@ -160,14 +160,15 @@ class AlertForwarder(object):
                     batch_write will run again, but only with the remaining unprocessed items.
 
             Args:
-                cls (AlertForwarder): Instance of the AlertForwarder
+                forwarder (AlertForwarder): Instance of the AlertForwarder
 
             Returns:
                 (bool) True if the batch write succeeded, False if there were UnprocessedItems.
             """
-            response = cls.client_dynamo.batch_write_item(RequestItems=cls.unprocessed_items)
-            cls.unprocessed_items = response['UnprocessedItems']
-            return len(cls.unprocessed_items) == 0
+            response = forwarder.client_dynamo.batch_write_item(
+                RequestItems=forwarder.unprocessed_items)
+            forwarder.unprocessed_items = response['UnprocessedItems']
+            return len(forwarder.unprocessed_items) == 0
 
         return decorated_batch_write(self)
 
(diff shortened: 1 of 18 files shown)


stream_alert/rule_processor/sink.py
```diff
@@ -32,14 +36,13 @@ def __init__(self, env):
...
         self.client_lambda = boto3.client('lambda', region_name=self.env['lambda_region'])
         self.function = os.environ['ALERT_PROCESSOR']
+        self.table = os.environ['ALERT_TABLE']
 
-    def sink(self, alerts):
-        """Sink triggered alerts from the StreamRules engine.
-
-        Args:
-            alerts (list): a list of dictionaries representating json alerts
+    def _sink_lambda(self, alerts):
```
ryandeivert: We should have just done away with this `sink` moniker now 😂

terraform/modules/tf_lambda/iam.tf
```diff
@@ -17,27 +17,11 @@ resource "aws_iam_role" "role" {
...
-resource "aws_iam_role_policy" "logs_metrics_policy" {
-  count  = "${var.enabled}"
-  name   = "LogsAndMetrics"
-  role   = "${aws_iam_role.role.id}"
-  policy = "${data.aws_iam_policy_document.logs_metrics_policy.json}"
+// Attach write permissions for CloudWatch logs
+resource "aws_iam_role_policy_attachment" "logs_metrics_policy" {
+  count      = "${var.enabled}"
+  role       = "${aws_iam_role.role.id}"
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
```
jacknagz: For reference: `AWSLambdaBasicExecutionRole – Grants permissions only for the Amazon CloudWatch Logs actions to write logs. You can use this policy if your Lambda function does not access any other AWS resources except writing logs.`

stream_alert_cli/terraform/generate.py
```diff
@@ -163,12 +163,16 @@ def generate_main(**kwargs):
...
-    # Configure global resources like Firehose alert delivery
+    # Configure global resources like Firehose alert delivery and alerts table
     main_dict['module']['globals'] = {
         'source': 'modules/tf_stream_alert_globals',
         'account_id': config['global']['account']['aws_account_id'],
         'region': config['global']['account']['region'],
-        'prefix': config['global']['account']['prefix']
+        'prefix': config['global']['account']['prefix'],
+        'alerts_table_read_capacity': (
+            config['global']['infrastructure']['alerts_table']['read_capacity']),
```
jacknagz: Are these required on purpose? If a user forgot to add them to the `conf`, it would throw `KeyErrors`

terraform/modules/tf_stream_alert_globals/main.tf
```diff
@@ -4,3 +4,26 @@ module "alerts_firehose" {
...
+
+  // Only the hash key and range key attributes need to be defined here.
+
+  attribute {
+    name = "RuleName"
+    type = "S"
+  }
+  attribute {
+    name = "Timestamp"
+    type = "S"
```
jacknagz: Is this intended to be in `Date/Calendar` format as specified [here](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.DataTypes.html)?

stream_alert/rule_processor/sink.py
jacknagz: +1, let's move away from `sink` altogether in classes/methods/filenames

stream_alert/rule_processor/sink.py
```diff
@@ -126,7 +127,10 @@ def _alert_batches(self, alerts, batch_size=25):
                 ]
             }
 
-    @backoff.on_exception(backoff.expo, ClientError, max_tries=5, jitter=backoff.full_jitter)
+    @backoff.on_exception(backoff.expo, ClientError, max_tries=5, jitter=backoff.full_jitter,
```
jacknagz: awesome

terraform/modules/tf_lambda/iam.tf
jacknagz: It might be worth it to add this as a comment ^

stream_alert/rule_processor/sink.py
```diff
@@ -92,3 +95,88 @@ def sink(self, alerts):
...
+        """
+        for attempt in xrange(1, max_attempts + 1):
+            response = self.client_dynamo.batch_write_item(RequestItems=request_items)
+
+            # If Dynamo experiences an internal error, unprocessed items are listed in the response.
+            # AWS recommends retrying unprocessed items in a loop with exponential backoff.
+            request_items = response['UnprocessedItems']
+            if request_items:
+                LOGGER.warn(
+                    'Batch write failed: %d alerts were not written (attempt %d/%d)',
```
ryandeivert: Do you think we should add a custom metric for this right away so we have some insight during the testing phase?

stream_alert/rule_processor/sink.py
```diff
@@ -92,3 +95,88 @@ def sink(self, alerts):
...
+
+        Args:
+            alerts (list): A list of dictionaries representing json alerts.
+        """
+        self._sink_lambda(alerts)
+
+        # While we are testing this, exceptions should be logged but not raise errors.
+        try:
+            self._sink_dynamo(alerts)
+        except Exception:  # pylint: disable=broad-except
```
ryandeivert: Is this a loose `except` for now just so we don't bring the whole ship down? Can I assume this will be refined after testing, etc? If so I'd add a TODO so we don't forget

terraform/modules/tf_alert_processor_iam/main.tf
```diff
@@ -3,6 +3,13 @@
 locals {
   firehose_arn_prefix = "arn:aws:firehose:${var.region}:${var.account_id}"
   lambda_arn_prefix   = "arn:aws:lambda:${var.region}:${var.account_id}:function"
+
+  // Terraform is upset if you try to index into an empty list, even if the resource count = 0.
+  // https://github.com/hashicorp/terraform/issues/11210
+  // As a workaround, we append an unused dummy element to the output lists.
```
ryandeivert: Interesting - thanks for adding the comment :)

tests/unit/stream_alert_rule_processor/test_sink.py
```diff
@@ -13,28 +13,35 @@
...
 from nose.tools import assert_equal
 
 from stream_alert.rule_processor.config import load_env
 from stream_alert.rule_processor.sink import StreamSink
 from tests.unit.stream_alert_rule_processor.test_helpers import get_mock_context
 
 
-class TestStreamSink(object):
+@patch.dict(os.environ, {'CLUSTER': 'corp'})
+class TestStreamSink(unittest.TestCase):
```
ryandeivert: What's the benefit of making this a subclass of `unittest.TestCase` here?

terraform/modules/tf_lambda/iam.tf
austinbyers: We are attaching this policy in addition to other policies (that comment refers to choosing the managed policy when creating the Lambda role from the AWS console). So every Lambda function will have this basic policy attached.

stream_alert_cli/terraform/generate.py
austinbyers: Yes, they should be required IMO. Dynamo cost is directly related to capacity, and it should be an error if you don't have the table capacity as part of your config.

This requirement could be removed when we add support for autoscaling, for example.

terraform/modules/tf_stream_alert_globals/main.tf
austinbyers: Yes, but `Date` is not actually a Dynamo type. It's really [just a String](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html):

_You can use the string data type to represent a date or a time stamp. One way to do this is by using ISO 8601 strings_

You can confirm this by looking at the Dynamo UI as well:
screen shot 2018-02-27 at 4 23 38 pm

No `Date`, only `String`. I chose ISO 8601 so it was easy to sort and could be loaded by other applications as needed

tests/unit/stream_alert_rule_processor/test_sink.py
```diff
@@ -96,17 +103,131 @@ def test_streamsink_sink_success(self, log_mock):
...
+        """StreamSink - Dynamo Sink - Retry unprocessed alerts"""
+
+        def mock_batch_write_item(**kwargs):
+            """Mock client_dynamo.batch_write_item to always return all items unprocessed."""
+            return {
+                'ResponseMetadata': {'HTTPStatusCode': 200},
+                'UnprocessedItems': {self.ALERT_TABLE: kwargs['RequestItems']}
+            }
+
+        self.sinker.client_dynamo.batch_write_item = mock_batch_write_item
```
ryandeivert: I'd suggest using `patch` for this if possible (which I think it should be). If not, ignore me.. :)

You should be able to do this like so:

```
@patch('boto3.client')  ## add your other patches also...
def test_dynamo_sink_unprocessed_alerts(boto_mock):
    """Use return_value to specify what you want a patched object to return"""
    alerts = self._generate_alerts(1)
    value = {
        'ResponseMetadata': {'HTTPStatusCode': 200},
        'UnprocessedItems': {self.ALERT_TABLE: alerts}
    }
    
    boto_mock.return_value.batch_write_item.return_value = value

    self.sinker._sink_dynamo(self._generate_alerts(1))
...
```

austinbyers:
@jacknagz 🚰 🚫

stream_alert/rule_processor/sink.py
```diff
@@ -92,3 +95,88 @@ def sink(self, alerts):
...
                             response['ResponseMetadata']['RequestId'])
+
+    def _alert_batches(self, alerts, batch_size=25):
+        """Group alerts into batches of 25, the maximum allowed by Dynamo batch_write_item.
+
+        Yields:
+            (dict) The constructed request for batch_write_item, containing <= 25 alerts.
+                Maps table name to a list of requests.
+        """
+        for i in xrange(0, len(alerts), batch_size):
```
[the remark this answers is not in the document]
austinbyers: That's fine, I can stick with  `range`

stream_alert/rule_processor/sink.py
```diff
@@ -92,3 +95,88 @@ def sink(self, alerts):
...
+                                # Compact JSON encoding (no extra spaces)
+                                'Record': {'S': json.dumps(alert['record'], separators=(',', ':'))}
+                            }
+                        }
+                    }
+                    for alert in batch
+                ]
+            }
+
+    @backoff.on_exception(backoff.expo, ClientError, max_tries=5, jitter=backoff.full_jitter,
```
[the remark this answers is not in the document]
austinbyers: Great idea, thanks for the examples! How did you do that?

stream_alert/rule_processor/sink.py
austinbyers: Nah - there are no metrics in the alert processor right now, and adding support for them would be nontrivial. We can just search `Batch write failed` in the logs - it shouldn't ever happen during normal operation.

But I'm amenable if you feel strongly we should add metrics sooner than later. I think as the alert merger develops, then it may make more sense

stream_alert/rule_processor/sink.py
austinbyers: Yeah, this is to just prevent the rule processor from failing because of a problem sending to Dynamo. I added a TODO to remove/refine it after testing

tests/unit/stream_alert_rule_processor/test_sink.py
austinbyers: You can leverage the built-in test utilities. For example, setting `self.maxDiff = None` will show all diffs (not just snippets) during equality checks. In fact, I think we can use the built in `self.assertEqual()` instead of using the one from `nose`, but still be able to run the tests with nose, but that's a project for another day.

It looks like we only inherit from `unittest.TestCase` in the tests I created :P So I'll remove them to be consistent

stream_alert/rule_processor/sink.py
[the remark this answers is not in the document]
ryandeivert: just pasted a link to github /shrug :)

stream_alert/rule_processor/sink.py
```diff
@@ -92,3 +95,88 @@ def sink(self, alerts):
...
+
+            # If Dynamo experiences an internal error, unprocessed items are listed in the response.
+            # AWS recommends retrying unprocessed items in a loop with exponential backoff.
+            request_items = response['UnprocessedItems']
+            if request_items:
+                LOGGER.warn(
+                    'Batch write failed: %d alerts were not written (attempt %d/%d)',
+                    len(request_items[self.table]), attempt, max_attempts)
+                # Simple exponential backoff: Sleep 0.5, 1, 2, 4 and 8 seconds.
+                time.sleep(0.25 * 2 ** attempt)
```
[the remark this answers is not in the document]
ryandeivert: Austin explained offline - no need to comment here

stream_alert/rule_processor/sink.py
ryandeivert: @austinbyers This code is in the rule processor :)

stream_alert/rule_processor/sink.py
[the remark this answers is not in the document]
austinbyers: There are 2 different kinds of failures here:
1. If `Dynamo` is unavailable, a boto `ClientError` is retried in the backoff.
2. If some (but not all) of the records are not processed for some reason, these are returned in `UnprocessedItems`. I retried this latter one myself because the items being passed to `batch_write` change between successive retries.

After looking at this again, a much simpler solution would be to use `backoff` for both error cases and just store the `unprocessed_items` in the class.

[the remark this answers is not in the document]
austinbyers: I'll also add some comments, since this is definitely confusing

stream_alert/rule_processor/sink.py
austinbyers: This has now been replaced with a @backoff decorator, which has its own logging which we can search for

tests/unit/stream_alert_rule_processor/test_sink.py
austinbyers: The client was already patched in the test setup, I was setting the mock return value to the patched object. However, I switched it to reference `boto_mock` instead of `self.forwarder.client_dynamo` to make it clearer (they are the same object).

austinbyers:
@jacknagz @ryandeivert Thanks for the great feedback! It's all been addressed. @ryandeivert great idea about enabling the TTL so as to not blow up the table, I've gone ahead and implemented that.

stream_alert/rule_processor/alert_forward.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # Keep track of unprocessed items when retrying batch_write_item()
+        self.unprocessed_items = None
+
+    def _send_to_lambda(self, alerts):
+        """DEPRECATED: Invoke Alert Processor directly
+
+        Sends a message to the alert processor with the following JSON format:
+            {
+                "record": record,
```
jacknagz: this comment is outdated, the schema is:

```
                'record': record,
                'rule_name': rule.rule_name,
                'rule_description': rule.rule_function.__doc__ or DEFAULT_RULE_DESCRIPTION,
                'log_source': str(payload.log_source),
                'log_type': payload.type,
                'outputs': rule.outputs,
                'source_service': payload.service(),
                'source_entity': payload.entity,
                'context': rule.context}
```

ryandeivert:
thanks @austinbyers I'll take a look shortly

stream_alert/rule_processor/alert_forward.py
```diff
@@ -0,0 +1,199 @@
...
+            except ClientError as err:
+                LOGGER.exception('An error occurred while sending alert to '
+                                 '\'%s:production\'. Error is: %s. Alert: %s',
+                                 self.function,
+                                 err.response,
+                                 data)
+                continue
+
+            if response['ResponseMetadata']['HTTPStatusCode'] != 202:
+                LOGGER.error('Failed to send alert to \'%s\': %s',
```
jacknagz: side note: this should really add to a metric instead of only logging

stream_alert/rule_processor/alert_forward.py
```diff
@@ -0,0 +1,199 @@
...
+        return decorated_batch_write(self)
+
+    def _send_to_dynamo(self, alerts):
+        """Write alerts in batches to Dynamo."""
+        for batch_num, batch in enumerate(self._alert_batches(alerts), start=1):
+            LOGGER.info('Sending batch %d to Dynamo with %d alert(s)',
+                        batch_num, len(batch[self.table]))
+            self.unprocessed_items = batch
+            if not self._batch_write():
+                LOGGER.error('Unable to save alert batch; unprocessed items remain: %s',
```
jacknagz: this should also log a metric

stream_alert/rule_processor/alert_forward.py
austinbyers: I'll go ahead and change it, but this entire method will be deleted in a few days :) (rule processor will no longer invoke alert processor at all)

stream_alert/rule_processor/alert_forward.py
austinbyers: Again, this will be deleted in a few days once we've tested the table and added a Lambda function to dispatch alerts from the table

stream_alert/rule_processor/alert_forward.py
```diff
@@ -0,0 +1,199 @@
...
+                    The backoff.on_predicate will automatically retry with any Falsey value, and
+                    batch_write will run again, but only with the remaining unprocessed items.
+
+            Args:
+                cls (AlertForwarder): Instance of the AlertForwarder
+
+            Returns:
+                (bool) True if the batch write succeeded, False if there were UnprocessedItems.
+            """
+            response = cls.client_dynamo.batch_write_item(RequestItems=cls.unprocessed_items)
```
ryandeivert: IMHO we should shy away from using `cls` as the var name here and it should be reserved for use in classmethods. I was fairly confused when looking at this since I _thought_ this was within a classmethod... but I knew `unprocessed_items` was a instance property, so I couldn't wrap my head around it. Then I realized you were passing `self` as an arg and it made more sense :)

tests/unit/stream_alert_rule_processor/test_alert_forward.py
```diff
@@ -0,0 +1,209 @@
...
+        """AlertForwarder - Dynamo - Retry unprocessed alerts"""
+
+        def mock_batch_write_item(**kwargs):
+            """Mock client_dynamo.batch_write_item to always return all items unprocessed."""
+            return {
+                'ResponseMetadata': {'HTTPStatusCode': 200},
+                'UnprocessedItems': {self.ALERT_TABLE: kwargs['RequestItems']}
+            }
+
+        self.boto_mock.return_value.batch_write_item.side_effect = mock_batch_write_item
```
ryandeivert: RE your previous (now not shown) comment about this needing to be dynamic - the value of `kwargs['RequestItems']` is always going to be the same thing as what is returned from the `self._generate_alerts(1)` function that is passed in (in this case). IE - this is effectively the same:

```
        ...
        items = self._generate_alerts(1)
        def mock_batch_write_item(**kwargs):
            """Mock client_dynamo.batch_write_item to always return all items unprocessed."""
            return {
                'ResponseMetadata': {'HTTPStatusCode': 200},
                'UnprocessedItems': {self.ALERT_TABLE: items}
            }

        self.boto_mock.return_value.batch_write_item.side_effect = mock_batch_write_item

        self.forwarder._send_to_dynamo(items)
        ...
```

tests/unit/stream_alert_rule_processor/test_alert_forward.py
```diff
@@ -0,0 +1,209 @@
...
+@patch.dict(os.environ, {'CLUSTER': 'corp'})
+class TestAlertForwarder(object):
+    """Test class for AlertForwarder"""
+    ALERT_PROCESSOR = 'corp-prefix_streamalert_alert_processor'
+    ALERT_TABLE = 'corp-prefix_streamalert_alerts'
+
+    @classmethod
+    def setup_class(cls):
+        """Setup the class before any methods"""
+        patcher = patch('stream_alert.rule_processor.alert_forward.boto3.client')
```
ryandeivert: not really an issue, but you don't have to use the fully-qualified path to the imported module here. `boto3.client` is sufficient

stream_alert/rule_processor/alert_forward.py
austinbyers: Good idea; done

stream_alert/rule_processor/alert_forward.py
```diff
@@ -0,0 +1,199 @@
...
+        self.client_dynamo = boto3.client('dynamodb', region_name=self.env['lambda_region'])
+        self.client_lambda = boto3.client('lambda', region_name=self.env['lambda_region'])
+        self.function = os.environ['ALERT_PROCESSOR']
+        self.table = os.environ['ALERT_TABLE']
+
+        # Keep track of unprocessed items when retrying batch_write_item()
+        self.unprocessed_items = None
+
+    def _send_to_lambda(self, alerts):
+        """DEPRECATED: Invoke Alert Processor directly
```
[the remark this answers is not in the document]
austinbyers: Good point, thanks

stream_alert/rule_processor/alert_forward.py
```diff
@@ -0,0 +1,199 @@
...
+                              max_tries=self.BACKOFF_MAX_RETRIES, jitter=backoff.full_jitter,
+                              on_backoff=backoff_handlers.backoff_handler,
+                              on_success=backoff_handlers.success_handler,
+                              on_giveup=backoff_handlers.giveup_handler)
+        @backoff.on_exception(backoff.expo, ClientError,
+                              max_tries=self.BACKOFF_MAX_RETRIES, jitter=backoff.full_jitter,
+                              on_backoff=backoff_handlers.backoff_handler,
+                              on_success=backoff_handlers.success_handler,
+                              on_giveup=backoff_handlers.giveup_handler)
+        def decorated_batch_write(cls):
```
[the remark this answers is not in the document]
austinbyers: Thanks!

stream_alert/rule_processor/alert_forward.py
austinbyers: Ah, good point. This is why code reviews are great - we can clarify anything which is confusing

tests/unit/stream_alert_rule_processor/test_alert_forward.py
austinbyers: It is not anymore - after adding the metrics I updated the test so that it invoked multiple times with different batch sizes

tests/unit/stream_alert_rule_processor/test_alert_forward.py
austinbyers: This was just copied from the other file, but I can fix that

PR #619, opened 2018-03-05
S3Payload Ignore Empty Files, NormalizedRecords Metric

jacknagz:
to: @ryandeivert 
cc: @airbnb/streamalert-maintainers @chunyong-lin 
size: small

## Background

Currently when the rule processor receives S3 event notifications, it will still try and download the file, but instead pre processing should just exit at this point.  I have also added a quick metric for NormalizedRecords

## Changes

* Ignore empty S3 files
* Add NormalizedRecords metric
* Update tests

## Testing

Local CI

### DIFF
--- a/stream_alert/rule_processor/payload.py
+++ b/stream_alert/rule_processor/payload.py
@@ -152,9 +152,12 @@ class S3Payload(StreamPayload):
                 returning a generator, providing the ability to support
                 multi-record like this (s3).
         """
-        s3_file = self._get_object()
+        s3_file_path = self._get_object()
+        if not s3_file_path:
+            return
+
         line_num, processed_size = 0, 0
-        for line_num, data in self._read_downloaded_s3_object(s3_file):
+        for line_num, data in self._read_downloaded_s3_object(s3_file_path):
 
             self._refresh_record(data)
             yield self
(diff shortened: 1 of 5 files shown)


stream_alert/rule_processor/payload.py
```diff
@@ -200,7 +203,11 @@ def _download_object(self, region, bucket, key):
         """
         size_kb = self.s3_object_size / 1024.0
         size_mb = size_kb / 1024.0
-        if size_mb > 128:
+
+        # File size checks before downloading
+        if size_kb == 0:
+            return None
```
ryandeivert: explicit `None` is not required (could be `False` if you'd like to be verbose)

stream_alert/rule_processor/payload.py
```diff
@@ -233,9 +242,9 @@ def _get_object(self):
         """
         # Use the urllib unquote method to decode any url encoded characters
         # (ie - %26 --> &) from the bucket and key names
-        def unquoted(data):
-            return unquote(data).decode('utf-8')
+        unquoted = lambda(data): unquote(data).decode('utf-8')
```
ryandeivert: ++

PR #620, opened 2018-03-05
Add SNS and SQS outputs

austinbyers:
to: @ryandeivert 
size: medium

## Background

SNS and SQS are the most common services for "Application Integration" within AWS, and they should be supported as StreamAlert outputs.

SNS (the "Simple Notification Service") supports a variety of endpoints including HTTP(S), Email Lambda, SMS, etc. In particular, this will allow setting email and SMS as StreamAlert outputs (via an SNS topic)!

SQS (the "Simple Queue Service") is useful for buffering a large number of events to be processed by another application. For example, SQS could be used to buffer alerts sent to [BinaryAlert](https://github.com/airbnb/binaryalert) for YARA processing.

## Changes

* Add SNS and SQS output support in the alert processor and CLI
* Remove the redundant `outputs` section of the alert processor config. There is only 1 alert processor, and its outputs are specified in `outputs.json`
* Adds the output descriptor to alert processor logging, since there could be more than one output of a given type. So `Successfully sent alert to aws-sns` becomes `Successfully sent alert to aws-sns:topic-descriptor`

## Testing

From a fresh clone of this branch, with a test rule which sends outputs to every AWS service:

Setup:
```
./manage.py configure account_id ...
./manage.py configure prefix ...
./manage.py terraform init

# Test alert - lands in Lambda and S3

./manage.py output new --service aws-sns  # Modifies outputs.json
./manage.py terraform build  # Adds new SNS:publish permission
./manage.py lambda deploy -p alert

# Test alert - lands in Lambda, S3, and SNS

./manage.py output new --service aws-sqs  # Modifies outputs.json
./manage.py terraform build  # Adds new SQS:send permission
./manage.py lambda deploy -p alert

# Test alert - lands in Lambda, S3, SNS, and SQS
```


tests/unit/stream_alert_cli/test_outputs.py
```diff
@@ -38,6 +38,8 @@ def test_load_output_config():
         u'aws-firehose',
         u'aws-lambda',
         u'aws-s3',
+        u'aws-sns',
```
ryandeivert: While you're adding to this - can you remove the `u` unicode prefix from these since it's not required/is confusing.

mime-frame:
Great! Quick favor - can you update the list here: https://streamalert.io/outputs.html

austinbyers:
@mime-frame I've added SNS and SQS to the output list in the documentation, but keep in mind the following:

* This change lands in the release branch, but the documentation builds from `master`. These changes will not show up on the website until we release v2.0.0 (which is by design)
* When we do release v2.0.0, we're planning on a complete documentation overhaul. As such, I've been deprioritizing documentation updates until then

Issue #572, opened 2018-01-16
Bug: terraform doesn't change DynamoDB correctly when AutoScaling enabled


chunyong-lin:
## Background
Terraform failed to update DynamoDB AutoScaling target due to invalid value set  `provisionedThroughput.readCapacityUnits` and `provisionedThroughput.writeCapacityUnits` to `-1`. 

### Description
When terraform is making change to `provisionedThroughput.readCapacityUnits` and `provisionedThroughput.writeCapacityUnits`, it will set it to `-1` first, and then change it to new value we set. However, `-1` is invalid value to AWS API. The error message is sent to CloudWatch event, and there is no error output from `terraform plan` or `terraform apply`. 

The error message in CloudWatch event is `2 validation errors detected: Value '-1' at 'provisionedThroughput.writeCapacityUnits' failed to satisfy constraint: Member must have value greater than or equal to 1; Value '-1' at 'provisionedThroughput.readCapacityUnits' failed to satisfy constraint: Member must have value greater than or equal to 1`

## Desired Change
Prevent terraform to change `provisionedThroughput.readCapacityUnits` and `provisionedThroughput.writeCapacityUnits` set to `-1` before new value kicks in.

chunyong-lin:
Open the issue in terraform repo for support https://github.com/terraform-providers/terraform-provider-aws/issues/3155

chunyong-lin:
I found out there is [a post](https://forums.aws.amazon.com/thread.jspa?threadID=270113) in AWS development discussion forums (you need to login to your aws console to have read permission), it mentioned that
`- Application Auto Scaling automatically creates the SLR in your account when you call RegisterScalableTarget or PutScalingPolicy if you have the IAM CreateServiceLinkedRole permission.`

This requirement affects dynamodb table when the autoscale is enabled also. Because in our [terraform code](https://github.com/airbnb/streamalert/blob/master/terraform/modules/tf_threat_intel_downloader/dynamodb.tf#L30-L34), it creates a IAM role for dynamodb autoscale. But eventually, aws will use the service role it created when registering a scale target for autoscale. 

The idea to solve this problem is we probably will force to use DynamoDB Autoscaling which is configurable currently. So we can remove the IAM we create for DynamoDB table and use auto-created one instead.

austinbyers:
Closed via #623

vijaypatel1985:
[...]

Issue #267, opened 2017-08-16
RuntimeWarning: Parent module not found while handling absolute import


austinbyers:
All of the StreamAlert Lambda functions output the following RuntimeWarning when running in AWS Lambda:

```
/var/task/stream_alert/alert_processor/main.py: RuntimeWarning: Parent module 'stream_alert.alert_processor' not found while handling absolute import
```

The warning happens because the full `stream_alert/[function]` folder structure in the repo does not exist when running in the Lambda environment. Each Lambda function sees only its own code, not the entire package structure. These errors pollute the logs, making to harder to find real errors.

One solution is to conditionally import based on whether the code is part of a package ([example](https://github.com/airbnb/binaryalert/blob/2d93266604140971df87e0e8ded71fb4be1ba095/lambda_functions/analyzer/main.py#L14-L20)). So Lambda imports like

```python
from stream_alert.alert_processor.helpers import validate_alert
```

could be replaced with
```python
if __package__:
    # Imported by other packages (e.g. unit tests)
    from stream_alert.alert_processor.helpers import validate_alert
else:
    # Running in AWS Lambda
    from helpers import validate_alert
```

jacknagz:
> The warning happens because the full stream_alert/[function] folder structure in the repo does not exist when running in the Lambda environment.

However, the deployment package includes this structure:

![screen shot 2017-08-16 at 3 09 20 pm](https://user-images.githubusercontent.com/11466941/29387462-f5f9dbe8-8294-11e7-9ed7-21d44d804cc1.png)

austinbyers:
Interesting. I wonder if Lambda's context is restricted to the folder containing the main entry point?

jacknagz:
@austinbyers we might want to reach out to AWS on this

strcrzy:
I'd like to know the answer to this too

juhamust:
Any findings regarding to this?

austinbyers:
Closed via #623

PR #623, opened 2018-03-07
A collection of small fixes

austinbyers:
to: @chunyong-lin 
size: small|medium|large
resolves: #267
resolves: #572  

## Background

A collection of small fixes and features that I've accumulated

## Changes

* #620 added support for SNS and SQS outputs, but did not update the alert tests to mock those outputs. This has been fixed
* Per #572 , Dynamo autoscaling works best with the service role automatically generated by AWS. We use that instead of creating our own IAM role
* Run `terraform fmt` to fix a spacing issue which snuck in
* The `third_party_libraries` config option is generally only applicable to the rule processor, where users can specify additional libraries to install for their rules to work. To reduce confusion, this config option has been removed from the other default lambda configs (although it will still work for other functions if it is defined). 
* Adds a BinaryAlert YARA match rule (the log schema was added in #556, but no rules were added)
* One of the unit tests was consuming 5-10% of the test time because it was not properly mocked:

```
[success] 10.41% tests.unit.stream_alert_athena_partition_refresh.test_main.TestStreamAlertAthenaGlobals.test_handler_no_unique_buckets: 0.9611s
```

* The ever-annoying RuntimeWarning import error (#267) can be fixed with a magical

```python
from __future__ import absolute_import
```

I have no idea why it works, but I found this solution in the [PyCharm blog](https://youtrack.jetbrains.com/issue/PY-20171#comment=27-1802805)

## Testing

* Deployed to a test account - alerts worked end-to-end
* `terraform validate`
* CI

### DIFF
--- a/stream_alert/alert_processor/main.py
+++ b/stream_alert/alert_processor/main.py
@@ -13,6 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+from __future__ import absolute_import  # Suppresses RuntimeWarning import error in Lambda
 from collections import OrderedDict
 import json
 
--- a/terraform/modules/tf_threat_intel_downloader/dynamodb.tf
+++ b/terraform/modules/tf_threat_intel_downloader/dynamodb.tf
@@ -27,17 +27,15 @@ resource "aws_dynamodb_table" "threat_intel_ioc" {
 }
 
 resource "aws_appautoscaling_target" "dynamodb_table_read_target" {
-  count              = "${var.autoscale ? 1 : 0}"
   max_capacity       = "${var.max_read_capacity}"
   min_capacity       = "${var.min_read_capacity}"
-  resource_id        = "table/${var.prefix}_streamalert_threat_intel_downloader"
+  resource_id        = "table/${aws_dynamodb_table.threat_intel_ioc.name}"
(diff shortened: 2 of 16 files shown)


terraform/modules/tf_threat_intel_downloader/dynamodb.tf
```diff
@@ -26,89 +26,12 @@ resource "aws_dynamodb_table" "threat_intel_ioc" {
...
-      "cloudwatch:SetAlarmState",
-      "cloudwatch:DeleteAlarms",
-    ]
-
-    resources = ["*"]
-  }
-}
-
 resource "aws_appautoscaling_target" "dynamodb_table_read_target" {
   count              = "${var.autoscale ? 1 : 0}"
```
chunyong-lin: Can you also help remove this line and `autoscale` variable? Because deleting custom static IAM role, it implies the autoscaling is enable all the time and can not be disabled.

chunyong-lin: There is another line below should be deleted.

stream_alert/alert_processor/main.py
```diff
@@ -13,6 +13,7 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+from __future__ import absolute_import  # Suppresses RuntimeWarning import error in Lambda
```
chunyong-lin: Magic! 💥

terraform/modules/tf_threat_intel_downloader/dynamodb.tf
austinbyers: Good idea, thanks!

PR #624, opened 2018-03-07
Required Sub-key Bug Fix, Community Rules Update, and More

jacknagz:
to: @ryandeivert 
size: medium

## Background

While writing rules, discovered that `req_subkeys` is not checking that each sub-key is a truthy value, which ends up resulting in KeyErrors.

Wanted to also standardize on a default output for community rules, and also apply `yapf` formatting to them.

## Changes

* Required Sub-keys bug fix + docs update
* Community rules formatting with `yapf` and update default outputs
* New `safe_json_loads` helper with unit test
* Docs update

## Testing

Local CI

### DIFF
--- a/helpers/base.py
+++ b/helpers/base.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 from fnmatch import fnmatch
 import logging
+import json
 import time
 
 from netaddr import IPAddress, IPNetwork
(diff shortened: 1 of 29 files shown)


helpers/base.py
```diff
@@ -217,3 +218,19 @@ def data_has_value_from_substring_list(data, needle_list):
...
+    """Safely load a JSON string into a dictionary
+
+    Args:
+        data (str): A JSON string
+
+    Returns:
+        dict: The loaded JSON string or empty dict
+    """
+    try:
+        json_data = json.loads(data)
```
ryandeivert: you could just do the return here directly instead of in the else below:  `return json.loads(data)`

PR #594, opened 2018-02-13
Add rule and test for GuardDuty

0xdabbad00:
to: @airbnb/streamalert-maintainers
size: small
resolves #503 

## Background

Adds a rule for GuardDuty and a test for it

## Testing

Setup StreamAlert. Enabled GuardDuty. Added this.

PR #625, opened 2018-03-08
Trusted Advisor support

### LINKED ISSUE #618: Create community rule for Trusted Advisor events
By default, StreamAlert monitors all CloudWatch events, but does not alert on all of them.  The AWS Trusted Advisor service supposedly sends events to CloudWatch events by default.  The TrustedAdvisor service checks for things like opening Security Groups publicly to port 3306 (mysql), making S3 buckets public, and a number of other checks.  Some of these are free, and for those with Business support plans, there are alerts for things like Access Keys being leaked to Github.

We should have a Community rule to detect these, similar to what we have for GuardDuty: https://github.com/airbnb/streamalert/blob/master/rules/community/guardduty/guard_duty_all.py

Examples of these events are here: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/EventTypes.html#trusted-advisor-event-types

0xdabbad00:
size: small
resolves #618

Trusted Advisor events go to CloudWatch Events.  This rule is similar to the recent GuardDuty rule in https://github.com/airbnb/streamalert/pull/594

There are dozens of Trusted Advisor events, so this filters on just a few of the more important security ones, otherwise you will be flooded with hundreds regularly as Trusted Advisor tries to save you a few pennies.

### DIFF
--- a/rules/community/trustedadvisor/trusted_advisor.py
+++ b/rules/community/trustedadvisor/trusted_advisor.py
@@ -1,0 +1,48 @@
+"""Alert on Trusted Advisor"""
+from stream_alert.rule_processor.rules_engine import StreamRules
+
+rule = StreamRules.rule
+disable = StreamRules.disable()
+
+
+@rule(logs=['cloudwatch:events'],
+      matchers=['trusted_advisor'],
+      outputs=['slack:sample-channel'])
+def trusted_advisor(rec):
+    """
+    author:         spiper
+    description:    Alert on Trusted Advisor events.
+                    Note that these will only go to CloudWatch events if you
+                    have a Business support contract or better for the account.
+    playbook:       Depends on the alert. Some of these are misconfigurations,
+                    some are warnings that you are approaching limits, and some
+                    and critical alerts that need to be responded to (such as
+                    access keys being pushed to github).
+    testing:        Create an S3 bucket with a random name and no contents,
+                    and make it publicly read-able.
+    """
+    detail = rec['detail']
+    check_name = detail.get('check-name', '')
+
+    # Ignore checks that tell us things are OK
+    if detail.get('status', '') == 'OK':
+        return False
+
+    # The check are somewhat documented at: https://aws.amazon.com/premiumsupport/ta-iam/
+
+    # Known checks to alert on
+    if check_name == 'Security Groups - Specific Ports Unrestricted':
+        return True
+    if check_name == 'MFA on Root Account':
+        return True
+    if check_name == 'IAM Password Policy':
+        return True
+    if check_name == 'AWS CloudTrail Logging':
+        return True
+    if check_name == 'Exposed Access Keys':
+        return True
+    if check_name == 'Amazon Route 53 MX Resource Record Sets and Sender Policy Framework':
+        return True
+
+    # Ignore everything else
+    return False
--- a/matchers/matchers.py
+++ b/matchers/matchers.py
@@ -20,3 +20,7 @@ matcher = StreamRules.matcher()
 @matcher
 def guard_duty(record):
     return record['detail-type'] == 'GuardDuty Finding'
+
+@matcher
+def trusted_advisor(record):
+    return record['source'] == 'aws.trustedadvisor'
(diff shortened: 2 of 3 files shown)


rules/community/trustedadvisor/trusted_advisor.py
```diff
@@ -0,0 +1,48 @@
...
+    check_name = detail.get('check-name', '')
+
+    # Ignore checks that tell us things are OK
+    if detail.get('status', '') == 'OK':
+        return False
+
+    # The check are somewhat documented at: https://aws.amazon.com/premiumsupport/ta-iam/
+
+    # Known checks to alert on
+    if check_name == 'Security Groups - Specific Ports Unrestricted':
```
jacknagz: Could you just add all of these to a set outside the scope of the function and then reference it here?

rules/community/trustedadvisor/trusted_advisor.py
```diff
@@ -0,0 +1,48 @@
...
+                    and critical alerts that need to be responded to (such as
+                    access keys being pushed to github).
+    testing:        Create an S3 bucket with a random name and no contents,
+                    and make it publicly read-able.
+    """
+    detail = rec['detail']
+    check_name = detail.get('check-name', '')
+
+    # Ignore checks that tell us things are OK
+    if detail.get('status', '') == 'OK':
```
jacknagz: This can be added to `req_subkeys`

rules/community/trustedadvisor/trusted_advisor.py
```diff
@@ -0,0 +1,48 @@
...
+                    have a Business support contract or better for the account.
+    playbook:       Depends on the alert. Some of these are misconfigurations,
+                    some are warnings that you are approaching limits, and some
+                    and critical alerts that need to be responded to (such as
+                    access keys being pushed to github).
+    testing:        Create an S3 bucket with a random name and no contents,
+                    and make it publicly read-able.
+    """
+    detail = rec['detail']
+    check_name = detail.get('check-name', '')
```
jacknagz: This field can also be added to req subkeys

jacknagz:
Hey @0xdabbad00 sorry for the delay! just left some comments

ryandeivert:
closing due to staleness - happy to revisit if desired.

0xdabbad00:
This was such a good rule.  I'm setting up StreamAlert for a company now and having this rule is one of the reasons for doing so, so I'm putting it in their environment.

@ryandeivert Anything that should be done to get this merged into master for folks? It's been awhile since I touched StreamAlert. Are the comments from Jack still relevant to how the rules work?

ryandeivert:
hey @0xdabbad00 - you're right, this is a good rule. addressing the comments will just optimize this rule a bit (global set that isn't re-created each time the function is called, using required subkeys, etc..). Example of sorta what we'd expect for this:

```python
"""Alert on Trusted Advisor"""
from rules.matchers import trusted_advisor
from stream_alert.shared.rule import rule

_CHECKS = {
    'Security Groups - Specific Ports Unrestricted',
    'MFA on Root Account',
    'IAM Password Policy',
    'AWS CloudTrail Logging',
    'Exposed Access Keys',
    'Amazon Route 53 MX Resource Record Sets and Sender Policy Framework',
}


@rule(logs=['cloudwatch:events'],
      matchers=[trusted_advisor],
      req_subkeys={'detail': ['check_name', 'status']},
      outputs=['slack:sample-channel'])
def trusted_advisor(rec):
    """
    author:         spiper
    description:    Alert on Trusted Advisor events.
                    Note that these will only go to CloudWatch events if you
                    have a Business support contract or better for the account.
    playbook:       Depends on the alert. Some of these are misconfigurations,
                    some are warnings that you are approaching limits, and some
                    and critical alerts that need to be responded to (such as
                    access keys being pushed to github).
    testing:        Create an S3 bucket with a random name and no contents,
                    and make it publicly read-able.
    """
    return rec['detail']['check_name'] in _CHECKS and rec['detail']['status'] != 'OK'
```

ryandeivert:
feel free to open up another PR with this and the matcher/test event to go with it

PR #626, opened 2018-03-09
Add alert UUID and use it in SNS subject

austinbyers:
to: @ryandeivert 
cc: @airbnb/streamalert-maintainers  @mime-frame 
size: small

## Background

We want to generate unique alert IDs to track them throughout their lifetime. For example, we can use them in the subject of messages sent via SNS

## Changes

* Adds alert UUID to alert dictionary
* Adds a subject to SNS messages with rule name + alert ID
* [unrelated] Fix BinaryAlert rule docstring

## Testing

* BuildKite CI
* Deploy to a test account - the SNS alert showed up in my inbox:

`binaryalert_yara_match triggered alert 091b1fd6-51f7-4be4-982e-9378b0af8cf6`

### DIFF
--- a/stream_alert/alert_processor/main.py
+++ b/stream_alert/alert_processor/main.py
@@ -62,6 +62,7 @@ def run(alert, region, account_id, function_name, config):
             following structure:
 
             {
+                'id': uuid,
                 'record': record,
                 'rule_name': rule.rule_name,
                 'rule_description': rule.rule_function.__doc__,
@@ -84,7 +85,7 @@ def run(alert, region, account_id, function_name, config):
         LOGGER.error('Invalid alert format:\n%s', json.dumps(alert, indent=2))
         return
 
-    LOGGER.debug('Sending alert to outputs:\n%s', json.dumps(alert, indent=2))
+    LOGGER.info('Sending alert %s to outputs %s', alert['id'], alert['outputs'])
 
     # strip out unnecessary keys and sort
     alert = _sort_dict(alert)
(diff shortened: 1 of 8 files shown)


stream_alert/alert_processor/main.py
```diff
@@ -84,7 +85,7 @@ def run(alert, region, account_id, function_name, config):
         LOGGER.error('Invalid alert format:\n%s', json.dumps(alert, indent=2))
         return
 
-    LOGGER.debug('Sending alert to outputs:\n%s', json.dumps(alert, indent=2))
+    LOGGER.info('Sending alert %s to outputs %s', alert['id'], alert['outputs'])
```
jacknagz: 👏

stream_alert/rule_processor/rules_engine.py
```diff
@@ -443,10 +444,12 @@ def rule_analysis(record, rule, payload, alerts):
             if StreamRules.check_alerts_duplication(record, rule, alerts):
                 return
 
-            LOGGER.info('Rule [%s] triggered an alert on log type [%s] from entity \'%s\' '
-                        'in service \'%s\'', rule.rule_name, payload.log_source,
+            alert_id = str(uuid.uuid4())  # Random unique alert ID
+            LOGGER.info('Rule [%s] triggered alert [%s] on log type [%s] from entity \'%s\' '
```
jacknagz: Isn't this a redundant logger message since `main.py` has a similar one when dispatching?

austinbyers: This log happens when the alert is built in the rule processor, the `main.py` you commented on above is in the alert processor. I like having both because then the alert ID is in the logs of both processors (and the rule processor / alert processor will be completely decoupled soon)

PR #627, opened 2018-03-12
[output] catch requests timeout exceptions

chunyong-lin:
to: @ryandeivert or @javuto 
size: small
resolves N/A

## Background
During an internal triage, I found out we need to catch ReadTimeout and ConnectTimeout while sending [requests post request](https://github.com/airbnb/streamalert/blob/master/stream_alert/alert_processor/outputs/output_base.py#L398). [requests.execptions.Timeout](http://docs.python-requests.org/en/master/_modules/requests/exceptions/) will catch both `ReadTimeout` and `ConnectTimeout`.

## Changes

* Add `requests.exceptions.Timeout` to default exceptions in output base class
* Update unit test cases. 

## Testing
* Unit test
```
./tests/scripts/unit_tests.sh
...
Ran 548 tests in 9.324s

OK
```
* Rule test
```
...
StreamAlertCLI [INFO]: (61/61) Successful Tests
StreamAlertCLI [INFO]: (33/33) Alert Tests Passed
```

### DIFF
--- a/stream_alert/alert_processor/outputs/output_base.py
+++ b/stream_alert/alert_processor/outputs/output_base.py
@@ -19,6 +19,7 @@ import json
 import os
 import tempfile
 import requests
+from requests.exceptions import Timeout as ReqTimeout
 import urllib3
 
 import backoff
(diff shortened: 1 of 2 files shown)


jacknagz:
🚢

PR #636, opened 2018-03-15
Merge release branch into master

austinbyers:
to: @ryandeivert 
size: large

## Background

In preparation for an upcoming backwards-incompatible release (v2.0.0), we experimented with a separate release branch. However, it proved to be time-consuming to manage bug fixes on `master` and big features on `release` and our own internal version and syncing the 3 disparate versions.

(It didn't help that I was doing `rebase` instead of `merge`, which muddied the commit history.)

Instead, @ryandeivert has created a `stable` tag which allow users to easily clone the latest stable release and all development going forward will continue on `master`.

This is NOT the v2 release, we're just merging the branch early.

## Changes

* Merge release-2.0.0 into `master`
  * Re-usable Lambda Terraform module
  * Consolidated alert processor
  * Alerts Dynamo table
  * SNS/SQS output support
  * Required Athena function
  * Faster/simpler deploy process

NOTE: This will be squashed into a single commit in the master branch

## Testing

* `release` was already up-to-date with master, so there are no merge conflicts
* `release` has been tested with a live test
* CI


jacknagz:
🚢

PR #638, opened 2018-03-16
More generalization for the Lambda module

austinbyers:
to: @jacknagz 
size: medium

## Background

In preparation for adding a new Lambda function (the alert merger), we do some work to generalize the `tf_lambda` module and its generation. Looking forward, we also want this module to be as general as possible to work across more of our projects.

## Changes

* The `tf_lambda` module now supports:
  * Custom alias name (instead of hardcoding `production`)
  * Invocation via CloudWatch events ("cronjob")
  * Reserved concurrency
  * Toggling individual metric alarms
* Refactor Terraform generation to be re-usable for Lambda functions
* Rename `_common.py` to `common.py` in CLI/terraform (it generates warnings in some IDEs and we don't follow that format anywhere else)
* Upgrade AWS provider to 1.11.0 (to support Lambda reserved concurrency)
* Removes extraneous logs during `./manage.py lambda test -p all`

## Testing

* Deploy to a test account
* Enabled concurrency and cloudwatch cron of the alert processor to verify they worked as expected


terraform/modules/tf_lambda/variables.tf
```diff
@@ -68,29 +73,39 @@ variable "auto_publish_versions" {
...
+  default     = "production"
+  description = "An alias with this name is automatically created which points to aliased_version"
+}
+
 variable "aliased_version" {
   default     = ""
   description = "Alias points to this version (or the latest published version if not specified)"
 }
 
+variable "invocation_frequency_minutes" {
```
jacknagz: Could we change this to just accept either a `rate(...)` or a cron based expression?  That way, users have the freedom to use hours, days, etc.

terraform/modules/tf_lambda/variables.tf
```diff
@@ -41,6 +41,11 @@ variable "source_object_key" {
   description = "S3 object key pointing to the function source code"
 }
 
+variable "concurrency_limit" {
+  default     = ""
```
jacknagz: What does Lambda default to?

terraform/modules/tf_lambda/cloudwatch.tf
```diff
@@ -1,5 +1,42 @@
+/* CloudWatch event rules, log group, and metric alarms */
+
+// CloudWatch event to trigger Lambda in a cronjob (if applicable)
+
+resource "aws_cloudwatch_event_rule" "cronjob" {
```
jacknagz: Could we change the verbiage to be 'scheduled' instead of 'cronjob'? It may be slightly confusing since CWE supports `rate()` expressions and `cron` based expressions

stream_alert_cli/terraform/lambda_module.py
```diff
@@ -0,0 +1,128 @@
...
+        result['{}_alarm_enabled'.format(alarm_type)] = True
+        result['{}_alarm_evaluation_periods'.format(alarm_type)] = settings.get(
+            'evaluation_periods', 1)
+        result['{}_alarm_period_secs'.format(alarm_type)] = settings.get('period_secs', 120)
+        result['{}_alarm_threshold'.format(alarm_type)] = settings.get('threshold', 0)
+
+    return result
+
+
+def generate_lambda(function_name, config, environment=None):
```
jacknagz: 🔥

stream_alert_cli/terraform/lambda_module.py
```diff
@@ -0,0 +1,128 @@
...
+        'source': 'modules/tf_lambda',
+        'function_name': '{}_streamalert_{}'.format(config['global']['account']['prefix'],
+                                                    function_name),
+        'description': 'StreamAlert {}'.format(function_name.replace('_', ' ').title()),
+        'handler': lambda_config['handler'],
+        'memory_size_mb': lambda_config['memory'],
+        'timeout_sec': lambda_config['timeout'],
+        'source_bucket': lambda_config['source_bucket'],
+        'source_object_key': lambda_config['source_object_key'],
+        'concurrency_limit': lambda_config.get('concurrency_limit', ''),
```
jacknagz: Isn't the module default `''`? Should we declare it in both places?

terraform/modules/tf_lambda/variables.tf
austinbyers: Good idea - I was thinking about this, too. The more generalized form would also allow the `apps` to migrate to this module

terraform/modules/tf_lambda/variables.tf
austinbyers: It defaults to disabled, which is no value at all. However, Terraform [currently interprets 0 as disabled](https://github.com/terraform-providers/terraform-provider-aws/issues/3803), which is technically a bug.

I updated the default to `0` and added a comment and more complete description

terraform/modules/tf_lambda/cloudwatch.tf
austinbyers: Yeah, good idea

stream_alert_cli/terraform/lambda_module.py
austinbyers: 🍦

stream_alert_cli/terraform/lambda_module.py
austinbyers: This is one of the negative side effects of marshaling our custom config to the Terraform variables - there is a lot of duplication. (I'm exploring other ways to do this.)

In the meantime, you're right - our general strategy has been not to marshal data which is not present in the config (which will then use the module default). I've updated the Lambda module generation logic to follow the same pattern (at the cost of some complexity)

austinbyers:
@jacknagz PTAL - I've addressed your comments and tested again with a new deploy and all the config option combinations

PR #640, opened 2018-03-16
Use context manager for Dynamo batch_write

austinbyers:
to: @ryandeivert 
size: small

## Background

Batch writing in Dynamo is far easier than I previously thought, because boto3 provides a [context manager](http://boto3.readthedocs.io/en/latest/guide/dynamodb.html#batch-writing):

```python
with table.batch_writer() as batch:
    batch.put_item(Item={})
```

The batch writer automatically handles buffering and retrying items which failed to write. So we can remove a bunch of custom logic we added to do this.

## Changes

* Use `batch_writer()` for the alerts table instead of the underlying client
* Replace the `UnprocessedAlerts` custom metric with `FailedDynamoWrites`
* Rename `ALERT_TABLE` environment variable in the rule processor to `ALERTS_TABLE` (makes more sense to me - I kept naming variables "alerts_table")
* Change the alert as it appears in the Dynamo table
    * `AlertID` is the range key instead of `Timestamp` (which allows efficient lookup of a specific ruleName + alertID)
    * `Timestamp` renamed to `Created`
    * Add `LogSource`, `LogType`, `SourceEntity`, `SourceService` (these are keys present in the alert record but not yet saved to Dynamo)

## Testing

* Live test end-to-end! The alert in Dynamo now looks like this:

screen shot 2018-03-16 at 2 44 50 pm

### DIFF
--- a/stream_alert/rule_processor/alert_forward.py
+++ b/stream_alert/rule_processor/alert_forward.py
@@ -16,21 +16,16 @@ limitations under the License.
 from datetime import datetime
 import json
 import os
-import time
 
-import backoff
 import boto3
 from botocore.exceptions import ClientError
 
-from stream_alert.shared import backoff_handlers
-from stream_alert.shared.metrics import MetricLogger
-from stream_alert.rule_processor import FUNCTION_NAME, LOGGER
+from stream_alert.rule_processor import LOGGER
 
 
 class AlertForwarder(object):
     """Sends alerts to the Alert Processor and the alerts Dynamo table."""
     # TODO: Do not send to Alert Processor after Alert Merger is implemented
-    BACKOFF_MAX_RETRIES = 6
 
     def __init__(self, env):
         """Initialize the Forwarder with the boto3 clients and resource names.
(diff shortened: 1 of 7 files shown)


stream_alert/rule_processor/alert_forward.py
```diff
@@ -192,9 +134,10 @@ def send_alerts(self, alerts):
...
-        # For now, don't blow up the rule processor if there is a problem sending to Dynamo.
-        # TODO: Remove/refine broad exception handling once tested.
         try:
             self._send_to_dynamo(alerts)
-        except Exception:  # pylint: disable=broad-except
+        except ClientError:
+            # The batch_writer() automatically retries transient errors - any raised ClientError
+            # is likely unrecoverable. Log an exception and metric
             LOGGER.exception('Error saving alerts to Dynamo')
+            MetricLogger.log_metric(FUNCTION_NAME, MetricLogger.FAILED_DYNAMO_WRITES, 1)
```
ryandeivert: Thanks for adding this :)

PR #641, opened 2018-03-16
Rule Test Templates

jacknagz:
to: @ryandeivert @fusionrace @jbussing 
size: large

## Background

Today, writing StreamAlert rule tests requires a copy of a 'real' log that will classify properly in the rules engine.  This can be a little arduous if your rule only tests one or two conditions.

The purpose of this pull request it to introduce a new functionality where a user can supply a partial test event, and the testing framework will fill the rest of the fields in with default data.

Here's an example:

For the given rule:

```python
@rule(logs=['duo:authentication'], outputs=['aws-firehose:alerts'])
def duo_anonymous_ip_failure(rec):
    """
    author:       airbnb_csirt
    description:  Alert on Duo auth logs marked as a failure due to an Anonymous IP.
    reference:    https://duo.com/docs/policy#anonymous-networks
    playbook:     N/A
    """
    return rec['result'] == 'FAILURE' and rec['reason'] == 'Anonymous IP'
```

The only fields checked are `result` and `reason`.  If we look at the intended schema of these logs, we see:

```
  "duo:authentication": {
    "schema": {
      "access_device": {},
      "device": "string",
      "factor": "string",
      "integration": "string",
      "ip": "string",
      "location": {},
      "new_enrollment": "boolean",
      "reason": "string",       # used in the rule
      "result": "string",         # used in the rule
      "timestamp": "integer",
      "username": "string"
    },
    "parser": "json"
  }
```

There a lot of fields not used in the rule.  Thankfully, the test event can now be declared with an `override_record` key, which contains the fields you want to override from a 'default' log generated from the `duo:authenticaiton` schema.

```json
  {
    "override_record": {
      "reason": "Location restricted",
      "result": "FAILURE"
    },
    "description": "Duo authentication log marked as failure as a result of 'Location restricted' that will not create an alert",
    "log": "duo:authentication",
    "service": "stream_alert_app",
    "source": "prefix_cluster_duo_auth_sm-app-name_app",
    "trigger_rules": []
  }
```

## Changes

* Support a new rule test format, where it can simply be a JSON list, instead of needing to wrap the object with a 'records' key.
* Support override_records, where default events are generated and then merged with the user supplied keys.
* Move generation of default schema values into `ParserBase`.
* Update unit tests, format files, cleanup as needed.

## Testing

Local rule tests, pylinting, and yapf


jacknagz:
After code review, I'll update documentation

tests/integration/rules/duo/duo_anonymous_ip_failure.json
```diff
@@ -1,68 +1,50 @@
...
+        "trusted_endpoint_status": "not trusted",
+        "os": "Mac OS X",
+        "browser": "Chrome"
       },
-      "description": "Duo authentication log marked as failure as a result of 'Anonymous IP' that will create an alert",
-      "log": "duo:authentication",
-      "service": "stream_alert_app",
-      "source": "prefix_cluster_duo_auth_sm-app-name_app",
-      "trigger_rules": ["duo_anonymous_ip_failure"]
+      "timestamp": 1505316499,
```
mime-frame: to confirm, you really only needed to have 2 entries here, `result` and `reason`, since that's only what you care to override, correct?: https://github.com/airbnb/streamalert/blob/master/rules/community/duo_authentication/duo_anonymous_ip_failure.py

jacknagz: yep, I just left one of the remainder fields in

stream_alert_cli/helpers.py
```diff
@@ -635,28 +631,45 @@ def user_input(requested_info, mask, input_restrictions):
...
-
-        # Make sure the test event file is formatted in the way we expect
-        if not (isinstance(contents, dict) and 'records' in contents):
-            message = message_template.format(path, 'File must be a dict (JSON '
-                                              'object) with top level key \'records\'')
-
-            return False, message
+            return [], message
+        else:
+            # Check for legacy format, return a list
```
ryandeivert: Let's log an error (or warning at least) here mentioning that the events should be updated to the new format, and add a TODO that flags this for removal sometime in the future. we should probably discuss creating a specific `TECHDEBT` (or similar) todo tag that we can search for in the future.

stream_alert_cli/test.py
```diff
@@ -361,11 +376,42 @@ def check_keys(self, test_event):
...
+        # Setup the parser to access default optional values
+        self.parsers[parser] = self.parsers.get('parser', get_parser(parser))
+
+        # Setup the modified test event
+        default_test_event = {}
+        data = test_event['override_record'].copy()
+
+        # Add apply default values based on the declared schema
+        for key, value in schema.iteritems():
+            default_test_event[key] = self.parsers[parser].default_optional_values(value)
```
ryandeivert: since `default_optional_values` is now a static method on `ParserBase` you can make this an easy comprehension using:

`default_test_event = {key: ParserBase.default_optional_values(value) for key, value in schema.iteritems()`

ryandeivert:
Also thanks for the great PR body - really helped me understand much better!

stream_alert_cli/test.py
```diff
@@ -361,11 +376,42 @@ def check_keys(self, test_event):
...
+        """
+        event_log = self.cli_config['logs'].get(test_event['log'])
+
+        parser = event_log['parser']
+        schema = event_log['schema']
+        configuration = event_log.get('configuration', {})
+
+        # Add envelope keys
+        if configuration.get('envelope_keys'):
+            schema.update(configuration['envelope_keys'])
```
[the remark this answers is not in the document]
jacknagz: what's the benefit of the first option? doesn't it have the same effect?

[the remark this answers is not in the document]
ryandeivert: nothing really :) use the second one

stream_alert_cli/test.py
```diff
@@ -361,11 +376,42 @@ def check_keys(self, test_event):
...
+
+        # Setup the parser to access default optional values
+        self.parsers[parser] = self.parsers.get('parser', get_parser(parser))
```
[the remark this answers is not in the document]
jacknagz: it only fetches it if it doesn't exist in the dict, but there's actually a bug where the `.get()` should be for `parser` and not `'parser'`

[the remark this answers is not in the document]
ryandeivert: aha! I see what you did there :) 👍

jacknagz:
@ryandeivert PTAL, just pushed some final fixes

docs/source/rule-testing.rst
```diff
@@ -1,30 +1,100 @@
...
+.. note:: Multiple tests can be included in one file by adding them to the array above.
+
+When specifying the test data, it can be either of two fields:
+
+1. ``"data"``: An entire example record, with all necessary fields to properly classify
+2. ``"override_record"``: A subset of the example record, where only relevant fields are populated
+
+The advantage of option #2 is that the overall test event is much smaller.
+
+The testing framework will auto-populate the records behind the scenes with the remainder fields for that given log type.
```
ryandeivert: `remainder` --> `remaining`

docs/source/rule-testing.rst
```diff
@@ -1,30 +1,100 @@
...
+      "log": "cloudwatch:events",
+      "service": "s3",
+      "source": "test-s3-bucket-name",
+      "trigger_rules": [
+        "my_fake_rule"
+      ]
+    }
+  ]
+
+Let's say a rule is only checking the value of ``source`` in the test event.  In that case, it's not valuable to fill in all the other data.  Here is what the event would  look like with ``override_record``:
```
ryandeivert: there is an extra space between  `would` and `look`.

I'd suggest changing `it's not valuable to fill in all the other data` to something like `there is no added benefit to filling in all of the other data`

jacknagz:
# FYI

I have committed a script that can convert the old style test events to new style!

```
$ ./tests/scripts/convert_test_events.py
```

mime-frame:
image

PR #642, opened 2018-03-19
Create alert merger which dispatches alerts from Dynamo to alert processors

austinbyers:
to: @ryandeivert 
size: large

## Background

This new Lambda function (the Alert Merger) will, as its name implies, eventually be responsible for alert merging. For now, it just dispatches alerts from the alerts Dynamo table and sends them to the alert processor.

The lifecycle of an alert is now as follows:

1. The rule processor generates alerts and saves them in batches to Dynamo
2. The alert merger runs every minute, finding all alerts in Dynamo which have not yet been sent or which need to be retried
3. The alert merger invokes the alert processor once for each alert, providing the Dynamo record as the payload
4. The alert processor deletes the Dynamo record once it is sent successfully, or else updates the list of outputs which need to be retried

Note that this PR does *not* implement alert merging, it just sets up the last piece of infra required to make it happen

## Changes

* Rule processor no longer invokes alert processor
* Add a new Alert Merger Lambda function, which runs every minute
* Update the alert processor with the new invocation format
* Support metric filters in the `tf_lambda` module
* Add a new `AlertAttempts` custom metric emitted by the Alert MErger

## Testing

* 100% coverage for the alert merger main
* Deploy to a test account and verify alerts end-to-end
* Deploy just the alert merger: `./manage.py lambda deploy -p alert_merger`
* Rollback just the alert merger: `./manage.py lambda rollback -p alert_merger`
* Try an alert which can't be sent and verify it is retried
* Enable and disable custom metrics


stream_alert/alert_merger/main.py
```diff
@@ -0,0 +1,151 @@
...
+    def dispatch(self):
+        """Find and dispatch all pending alerts to the alert processor."""
+        for rule_name in self.alerts_db.rule_names():
+            for alert in self.alerts_db.pending_alerts(rule_name, self.alert_proc_timeout):
+                self._dispatch_alert(alert)
+
+
+def handler(event, context):  # pylint: disable=unused-argument
+    """Entry point for the alert merger."""
+    global MERGER  # pylint: disable=global-statement
```
ryandeivert: I thought the `global` keyword was supposed to avoided whenever possible. We could favor class properties (not instance properties) that would keep the client active for the life of the container. I thought we were doing this elsewhere much I'm having trouble finding it.. but basically this:

```python
class Test(object):
    CLS_PROP = None

    def __init__(self):
        if not Test.CLS_PROP:
            Test.CLS_PROP = "now created"

    def get_prop(self):
        return self.CLS_PROP

for i in range(2):
    # when i == 1 the object is already cached
    print 'Class prop before init #%d: %s' % (i+1, Test.CLS_PROP)
    tester = Test()
    print 'Class prop after init #%d: %s' % (i+1, Test.CLS_PROP)
    print 'Instance prop after init #%d: %s' % (i+1, tester.get_prop())
```

I like this because it still encapsulates the client within the class that actually _uses_ it.

**EDIT**: Pardon my naivety - I was thinking this was a client wrapper but quickly realized it is not.

stream_alert/alert_merger/main.py
```diff
@@ -0,0 +1,151 @@
...
+            self.table.update_item(
+                Key={'RuleName': rule_name, 'AlertID': alert_id},
+                UpdateExpression='SET Dispatched = :now ADD Attempts :one',
+                ExpressionAttributeValues={':now': int(time.time()), ':one': 1},
+                ConditionExpression='attribute_exists(AlertID)'
+            )
+        except ClientError as error:
+            if error.response['Error']['Code'] == 'ConditionalCheckFailedException':
+                LOGGER.warn('Conditional update failed: %s', error.response['Error']['Message'])
+            else:
```
ryandeivert: Can you add some context as to why this particular error is okay and only logs, while others will be raised?

austinbyers:
@ryandeivert Thanks for the feedback, I've addressed it! I also rebased with your recent change (#643 ) and added support for custom metrics in the alert merger. I felt these were substantial enough code changes to require another review when you get a chance!

stream_alert/alert_merger/main.py
austinbyers: No, that's a good point. After some discussion, I decided to go with your suggestion and cache the instantiation in the class itself. It makes the handler itself super clean:

```python
def handler(event, context):
    return AlertProcessor.get_instance(context.invoked_function_arn).run(event)
```

stream_alert/alert_merger/main.py
austinbyers: It's possible (though very unlikely) that the alert processor could delete the entry from the table before the alert merger updates it with its additional state. If that happens, that's totally fine - we want the rows to be deleted when we don't need them anymore.

I removed the logging entirely - it shouldn't even be a warning, it's just an unusual order of operations. I also clarified with some more comments

stream_alert/alert_merger/main.py
```diff
@@ -145,7 +152,4 @@ def dispatch(self):
 
 def handler(event, context):  # pylint: disable=unused-argument
     """Entry point for the alert merger."""
-    global MERGER  # pylint: disable=global-statement
-    if not MERGER:
-        MERGER = AlertMerger()
-    MERGER.dispatch()
+    AlertMerger.get_instance().dispatch()
```
ryandeivert: this is excellent! :) thanks for dealing with my requests!

stream_alert/shared/metrics.py
```diff
@@ -60,6 +64,9 @@ class MetricLogger(object):
     FIREHOSE_FAILED_RECORDS = 'FirehoseFailedRecords'
     NORMALIZED_RECORDS = 'NormalizedRecords'
 
+    # Alert Merger metric names
+    ALERT_ATTEMPTS = 'AlertAttempts'
```
ryandeivert: 🎉

PR #643, opened 2018-03-20
[outputs] adding support for required alerting outputs

ryandeivert:
to: @austinbyers 
size: medium

## Background

A previous change was made to enable [athena as part of the default deployment](https://github.com/airbnb/streamalert/pull/599). Due to this change, StreamAlert must enforce sending alerts to required outputs even if they are not defined in the actual rule.

## Changes

* Required alerting outputs will be hard-coded in the the `stream_alert/shared/resource.py` module.
  * Currently, this only includes the default `aws-firehose` output for `alerts`
* The rule processor will combine the default output names with the user-defined outputs for a rule.
* The alert processor config will now be passed to the resources function to merge in `required` outputs.
* Updating rules as necessary.
* Adding suppression for noisy log statement during tests.

## Testing

Updating unit tests that broke because of this change. Adding tests for the shared functions that return the required outputs.


ryandeivert:
@austinbyers just fixed the integration testing mocks, PTAL

stream_alert/rule_processor/rules_engine.py
```diff
@@ -19,7 +19,7 @@
 
 from stream_alert.rule_processor import LOGGER
 from stream_alert.rule_processor.threat_intel import StreamThreatIntel
-from stream_alert.shared import NORMALIZATION_KEY
+from stream_alert.shared import resources, NORMALIZATION_KEY
```
austinbyers: In the alert processor, these are imported in the other order. Not sure which way is the "proper" alphabetization, but let's make it the same either way

stream_alert/rule_processor/rules_engine.py
```diff
@@ -53,6 +53,7 @@ class StreamRules(object):
     def __init__(self, config):
         """Initialize a StreamRules instance to cache a StreamThreatIntel instance."""
         self._threat_intel = StreamThreatIntel.load_from_config(config)
+        self._required_outputs = set(resources.get_required_outputs())
```
austinbyers: `get_required_outputs()` already returns a set, it doesn't need to be recreated as such. However, for clarity, you could rename `self._required_outputs` to `self._required_output_set` if you wanted

stream_alert/shared/resources.py
```diff
@@ -0,0 +1,65 @@
...
+def get_required_outputs():
+    """Iterates through the required outputs and collapses to the right format
+
+    Returns:
+        set: Set of required output services and names in the form service:name
+    """
+    outputs = set()
+    for service, value in REQUIRED_OUTPUTS.iteritems():
+        if not isinstance(value, dict):
+            continue
```
austinbyers: Since `REQUIRED_OUTPUTS` is hardcoded, I don't think we need any error-checking logic. I would rather fail with an exception here than just silently skip what is likely a problem with the `REQUIRED_OUTPUTS` definition

stream_alert/shared/resources.py
```diff
@@ -0,0 +1,65 @@
...
+        prefix (str): Prefix for this StreamAlert deployment to be injected into
+            resource names
+
+    Returns:
+        dict: Entire formatted outputs dictionary, including required items and
+            user defined outputs
+    """
+    for service, value in REQUIRED_OUTPUTS.iteritems():
+        if not isinstance(value, dict):
+            continue
```
austinbyers: Ditto - I don't think we need this

stream_alert/shared/resources.py
```diff
@@ -0,0 +1,65 @@
...
+        outputs.update({'{}:{}'.format(service, output) for output in value.keys()})
+    return outputs
+
+
+def merge_required_outputs(user_config, prefix):
+    """Iterates through the required outputs and merges them with the user outputs
+
+    Args:
+        user_config (dict): Loaded user outputs dictionary from conf/outputs.json
```
austinbyers: Can you add a line explaining that the incoming config is modified? Just reading the docstring I was under the impression a new dictionary was returned

stream_alert/rule_processor/rules_engine.py
```diff
@@ -443,14 +439,20 @@ def rule_analysis(record, rule, payload, alerts):
             LOGGER.info('Rule [%s] triggered alert [%s] on log type [%s] from entity \'%s\' '
                         'in service \'%s\'', rule.rule_name, alert_id, payload.log_source,
                         payload.entity, payload.service())
+
+            # Combine the required alert outputs with the ones for this rule
+            all_outputs = self._required_outputs
+            if rule.outputs:
+                all_outputs = all_outputs.union(set(rule.outputs))
```
austinbyers: When I saw this, I thought it was a bug because it looks like modifying `all_outputs` should modify `self._required_outputs` (they are initially the same reference). But this line actually rebinds `all_outputs` to reference a new set created by the union.

For clarity, and to avoid possible future bugs with this, can we just do
```python
all_outputs = self._required_outputs | set(rule.outputs)
```

Ideally, `rule.outputs` is always a list or set, and that will work fine. But if not, you can do
```python
all_outputs = self._required_outputs | set(rule.outputs or [])
```

stream_alert/rule_processor/rules_engine.py
```diff
@@ -443,14 +439,20 @@ def rule_analysis(record, rule, payload, alerts):
...
+
             alert = {
                 'id': alert_id,
                 'record': record,
                 'rule_name': rule.rule_name,
                 'rule_description': rule.rule_function.__doc__ or DEFAULT_RULE_DESCRIPTION,
                 'log_source': str(payload.log_source),
                 'log_type': payload.type,
-                'outputs': rule.outputs,
+                'outputs': list(all_outputs),
```
austinbyers: We actually are [saving the outputs as a `StringSet` in Dynamo](https://github.com/airbnb/streamalert/blob/e0839db8d79944f2568c1137bdd206995b88394b/stream_alert/rule_processor/alert_forward.py#L53), so I'd say let's keep this as a set

stream_alert/shared/__init__.py
```diff
@@ -1,6 +1,7 @@
 """Define some shared resources."""
 import logging
 import os
+import string
```
austinbyers: I'm surprised pylint didn't complain about this! Just of curiosity, where did it come from?

ryandeivert: oh interesting - I think pylint won't complain about imports within `__init__` since it probable assumes you're doing package imports like `from . import *`

stream_alert/rule_processor/rules_engine.py
ryandeivert: adding a TODO to change this :)

stream_alert/rule_processor/rules_engine.py
ryandeivert: ah thank you! I was actually trying to figure out the best way to ensure rule outputs wasn't None

stream_alert/rule_processor/rules_engine.py
ryandeivert: good catch - the initial implementation of `get_required_outputs` actually returned a dict, which needed to be cast to `set`. updating now

stream_alert/rule_processor/rules_engine.py
ryandeivert: good catch - this was a lazy find/replace since I used to have `resources` named `common` but switch the naming convention

stream_alert/shared/resources.py
ryandeivert: I will actually change this to return a copy, not modify in place

ryandeivert:
hey @austinbyers just addressed your feedback. PTAL when you get a chance

stream_alert/rule_processor/rules_engine.py
austinbyers: Yeah, you still need JSON-compatibility for the alert processor (but not for much longer!)

stream_alert/shared/resources.py
```diff
@@ -26,12 +26,9 @@ def get_required_outputs():
...
     """
-    outputs = set()
-    for service, value in REQUIRED_OUTPUTS.iteritems():
-        if not isinstance(value, dict):
-            continue
-        outputs.update({'{}:{}'.format(service, output) for output in value.keys()})
-    return outputs
+    return {'{}:{}'.format(service, output)
+            for service, value in REQUIRED_OUTPUTS.iteritems()
+            for output in value.keys()}
```
austinbyers: So Pythonic and elegant! 🐍

PR #649, opened 2018-03-22
small optimization to avoid multiple recursive calls

ryandeivert:
to: @austinbyers or @jacknagz 
size: small

## Background

Identified a function that used backoff that does recursive record sanitization. If the function ever backs-off, this would result in unnecessary calls to sanitize keys.

## Changes

* Moving the record dumping and sanitization to outside of the backed-off function.
* Changing another small instance of something similar in the threat_intel module.

## Testing

Unit tests passing, etc

### DIFF
--- a/stream_alert/rule_processor/firehose.py
+++ b/stream_alert/rule_processor/firehose.py
@@ -164,25 +164,27 @@ class StreamAlertFirehose(object):
                               on_backoff=backoff_handler,
                               on_success=success_handler,
                               on_giveup=giveup_handler)
-        def firehose_request_wrapper():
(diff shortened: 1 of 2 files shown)


stream_alert/rule_processor/firehose.py
```diff
@@ -164,25 +164,27 @@ def _firehose_request_helper(self, stream_name, record_batch):
                               on_backoff=backoff_handler,
                               on_success=success_handler,
                               on_giveup=giveup_handler)
-        def firehose_request_wrapper():
+        def firehose_request_wrapper(data):
```
jacknagz: 👍

PR #651, opened 2018-03-22
Add error handling

### LINKED ISSUE #648: Improve error handling of non-json sent to kinesis stream.
## Background

Have you tried pinging us on Slack? No
Are you on the latest version of StreamAlert? Yes

### Description

I was testing something and sent the string "hello" as a record to my Kinesis Stream.  StreamAlert kept error'ing every time it tried to read from the stream, due to an `AttributeError` here: https://github.com/airbnb/streamalert/blob/a97fb604f0d6498219f3599f786745fd7544c992/stream_alert/rule_processor/parsers.py#L154

Full crash was:
```
'unicode' object has no attribute 'keys': AttributeError
File "/var/task/stream_alert/rule_processor/main.py", line 59, in handler
StreamAlert(context).run(event)
File "/var/task/stream_alert/rule_processor/handler.py", line 121, in run
payload_with_normalized_records.extend(self._process_alerts(payload))
File "/var/task/stream_alert/rule_processor/handler.py", line 178, in _process_alerts
self.classifier.classify_record(record)
File "/var/task/stream_alert/shared/stats.py", line 27, in timed
result = func(*args, **kw)
File "/var/task/stream_alert/rule_processor/classifier.py", line 140, in classify_record
parse_result = self._parse(payload)
File "/var/task/stream_alert/rule_processor/classifier.py", line 257, in _parse
schema_matches = self._process_log_schemas(payload)
File "/var/task/stream_alert/shared/stats.py", line 27, in timed
result = func(*args, **kw)
File "/var/task/stream_alert/rule_processor/classifier.py", line 226, in _process_log_schemas
parsed_data = parser.parse(schema, payload.pre_parsed_record)
File "/var/task/stream_alert/shared/stats.py", line 27, in timed
result = func(*args, **kw)
File "/var/task/stream_alert/rule_processor/parsers.py", line 324, in parse
if not self._key_check(schema, json_records):
File "/var/task/stream_alert/rule_processor/parsers.py", line 140, in _key_check
json_keys = set(json_records[index].keys())
AttributeError: 'unicode' object has no attribute 'keys'
```

Every time new data came into my Kinesis Stream, StreamAlert's rule lambda would trigger and attempt to read all the data in the stream, starting back at the point of this failure, causing yet another failure.  I suspect this would continue until the retention time of the data expired.

I eventually bypassed this by adding:
```
if type(json_records[index]) is unicode:
                LOGGER.info('Record was unicode, skipping')
                return False
```

This allowed StreamAlert to catch back up by skipping that record.  There could be better handled by checking if it is a `dict` probably and skipping otherwise.

### Steps to Reproduce

Do a `put-record` against your stream of just `hello`.  You're StreamAlert setup will no longer be able to evaluate anymore incoming records until the retention of the data expires.

## Desired Change

Add error handling to skip malformed records and record somehow that malformed data was received so it can be investigated without blocking all other records.

I don't know if this is by design, so let me describe the two options:
1. Keep things as they are.  This means you block on bad data so someone can fix the problem and that record and all records after that which are similarly malformed could be handled correctly.  The negative is no records will be evaluated until this is resolved.
2. Skip the malformed record.  The negative of this is that potentially valuable records will be skipped and cannot easily be re-evaluated.  The positive is that your StreamAlert setup continues functioning for other records.

I prefer Option 2.

0xdabbad00:
cc: @ryandeivert 
size: small
resolves #648

## Background

Adds some simple error handling.  I was not able to get my instance of StreamAlert to fail in the same way it had previously, to retest this change, as now when I do:

```
aws kinesis put-record --stream-name test_prod_stream_alert_kinesis --partition-key 0 --data string_test
```
I get the error in the logs:

```
[ERROR]	2018-03-22T20:39:44.315Z	480fda82-041e-45f4-934b-4246aec7f9cd	Record does not match any defined schemas: <KinesisPayload valid:False log_source:None entity:test_prod_stream_alert_kinesis type:None record:None>
```
So that situation is being handled somewhere else, and I couldn't figure out how to put things back into the bad situation I was experiencing previously.

However, I did make this change, deploy this code, and ensure StreamAlert still functions correctly, so it looks like this should be harmless for normal operation.

### DIFF
--- a/stream_alert/rule_processor/parsers.py
+++ b/stream_alert/rule_processor/parsers.py
@@ -155,7 +155,8 @@ class JSONParser(ParserBase):
                              'but received %s. Value: %s',
                              type(json_records[index]),
                              json_records[index])
-                return False
+                del json_records[index]
+                continue
             schema_match = False
             json_keys = set(json_records[index].keys())
             if json_keys == schema_keys:
--- a/stream_alert/rule_processor/parsers.py
+++ b/stream_alert/rule_processor/parsers.py
@@ -150,6 +150,12 @@ class JSONParser(ParserBase):
         # Because elements are deleted off of json_records during
         # iteration, this block uses a reverse range.
         for index in reversed(range(len(json_records))):
+            if not isinstance(json_records[index], dict):
+                LOGGER.error('Skipping record. Expected type dict, '
+                             'but received %s. Value: %s',
+                             type(json_records[index]),
+                             json_records[index])
+                return False
             schema_match = False
             json_keys = set(json_records[index].keys())
             if json_keys == schema_keys:


stream_alert/rule_processor/parsers.py
```diff
@@ -150,6 +150,12 @@ def _key_check(self, schema, json_records):
         # Because elements are deleted off of json_records during
         # iteration, this block uses a reverse range.
         for index in reversed(range(len(json_records))):
+            if not isinstance(json_records[index], dict):
+                LOGGER.error('Skipping record. Expected type dict, '
+                             'but received %s. Value: %s',
+                             type(json_records[index]),
+                             json_records[index])
+                return False
```
ryandeivert: should this return false or just continue to the next 'record'?

0xdabbad00:
Good catch @ryandeivert, I've hopefully fixed it.

jacknagz:
Hey @0xdabbad00, if we can't reproduce the error in a unit test, I'm not comfortable with this change.

The `parse` method actually catches non unicode/string input right at the beginning:

https://github.com/0xdabbad00/streamalert/blob/485da07f2dbf05f1f882463a1142951335a1016e/stream_alert/rule_processor/parsers.py#L310-L323

Also, the change in `key_check` is highly recursive and will most likely cause noticeable performance decreases

ryandeivert:
hey @0xdabbad00 sorry i just noticed you mentioned you weren't able to actually repro the error 🤔 . since this check happens within a recursive function we might not be super comfortable making the change

0xdabbad00:
That's reasonable.  I tried a couple of ways to reproduce the original exception, but have not been successful.  I no longer experience this issue, so I'm fine with closing this PR and related issue.  If I experience it again I'll hopefully be able to create a unit test, and I'll have this code to refer back to.

PR #653, opened 2018-03-27
More efficient helper functions and alert merger bug fix

austinbyers:
to: @javuto 
cc: @mime-frame @airbnb/streamalert-maintainers
size: medium

## Background

The `in_set` and `select_key` helper methods are rather inefficient, so we replace them with optimized alternatives.

In addition, #642 introduced a subtle bug where the alert processor could update a non-existent rule because Dynamo updates are eventually consistent.

## Changes

* Replace `in_set` helper function with `ends_with_any`, `contains_any`, `matches_any`,  and `starts_with_any`
* Replace `select_key` with `get_first_key` and `get_keys`
* Update community rules to leverage new helper functions and define global variables outside of the rule
* Force consistent reads when the alert merger is querying alert data

## Testing

* CI: unit tests, rule tests, lint, etc
* Deploy to a test account


javuto:
Code looks super clean! And remember to rebase. Do you think it makes sense to add some sort of benchmark for this helpers? It could prevent performance regressions in the future...
Great job! Thanks for adding this @austinbyers

PR #655, opened 2018-03-27
[threat intel] allow one record to be normalized multiple times

### LINKED ISSUE #654: Bug: Threat Intel prevent one record to be normalized more than once
## Background
A flag [normalized_record_appened](https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/rules_engine.py#L387) in rules engine prevent one record to be normalized more than once when `threat intel` feature is enabled.

### Description
* Requirement: threat intel is enabled.
* Consider following rules,
```
        @rule(datatypes=['fileHash'], outputs=['s3:sample_bucket'])
        def match_file_hash(rec): 
            """The rule to match file hash"""
            return 'streamalert:ioc' in rec and 'md5' in rec['streamalert:ioc']

        @rule(datatypes=['fileHash', 'sourceDomain'], outputs=['s3:sample_bucket'])
        def match_source_domain(rec):
            """The rule to match source domain and file hash"""
```
* With a record 
```
{
    "Field1": {
	"SubField1": {
 	  "key1": 17,
	  "key2_md5": "md5-of-file",
          "key3_source_domain": "evil.com"
        },
        "SubField2": 1
    },
    "Field2": {
        "Authentication": {}
    },
    "Field3": {},
    "Field4": {}
}
```
* This record will never match to 2nd rule `match_source_domain`, because the 1st rule triggers triggers to run data normalization against record and set the flag [normalized_record_appened](https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/rules_engine.py#L387) to True. So the record will be skipped to match other rules.

## Desired Change
* Allow one record to be normalized multiple times.

chunyong-lin:
to: @ryandeivert
size: small
resolves #654 

## Background

Please see issue #654 where has example and detail.

## Changes

* Remove the flag which prevent a record to be normalized more than once.
* Refactor `check_alerts_duplication` to remove duplicated records.
* Add unit test case to reproduce issue and verify fix.

## Testing
* Rule testing
```
...
StreamAlertCLI [INFO]: (63/63) Successful Tests
StreamAlertCLI [INFO]: (34/34) Alert Tests Passed
```

* Unit testing
```
./tests/scripts/unit_tests.sh
...
Ran 556 tests in 9.120s

OK
```

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -474,6 +474,13 @@ class StreamRules(object):
             if rule.rule_name == exist_alert['rule_name']:
                 record_copy = record.copy()
                 exist_alert_record_copy = exist_alert['record'].copy()
+                # Early return when two records are dumplicated.
+                if record_copy == exist_alert_record_copy:
+                    return True
+
+                # There is chance that there are different values in 'streamalert:normalization'
+                # or 'streamalert:ioc' fields. So do comparision again after removing
+                # two keys.
                 record_copy.pop(StreamThreatIntel.IOC_KEY, None)
                 record_copy.pop(NORMALIZATION_KEY, None)
                 exist_alert_record_copy.pop(StreamThreatIntel.IOC_KEY, None)
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -351,9 +351,6 @@ class StreamRules(object):
             return alerts, normalized_records
 
         for record in payload.records:
-            # One record may be added to normalized records list multiple time due
-            # to each record is processed by all rules.
-            normalized_record_appended = False
             for rule in rules:
                 # subkey check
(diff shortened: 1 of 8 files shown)


chunyong-lin:
@ryandeivert Thanks for the comment. To answer your questions.
> can we compare without removing the keys first?

Good suggestion. Yes, we can do this first and most cases it can exit earlier! I made the change.

> Is there a chance the existing record and the new one would have different values for these?

Yes, it may have different value in `streamalert:normalization` field after I removed the flag `normalized_record_appended` in `process`. After applying data normalization to a record, it will add a new key `streamalert:normalization` key with value in the record if any matching. One record maybe normalized by different rules with different datatypes and resulting different values in `streamalert:normalization` field.

Same situation will be occurred when different threat intel rules with different IOC types apply to one record.

>  Do we know if it has any notable performance impact?

I don't expect to have performance impact after this change. 
* We are doing this comparison in current code. The comparison will only be triggered when the rule returns `True`. 
* I made the change as your suggested to allow early exist before removing keys. I will also reduce performance impact.

PR #656, opened 2018-03-27
[tf] refactor apps terraform module to use the new `tf_lambda` module

ryandeivert:
to: @austinbyers 
size: medium

## Background

With @austinbyers' great addition of a `tf_lambda` terraform module, we have the ability to reduce a lot of redundancy with our lambda terraform modules.

## Changes

* Refactor apps terraform module to use tf_lambda module
  * This is now split into two parts - most of the module has been removed and will use the `tf_lambda` module for generation.
  * IAM and SSM configuration settings are what remain in the renamed `stream_alert_app_iam` module.
* Updates to the cli manage code for new app config format
  * This also includes updates for creating a new app with the new config format

## Other

* This has the added benefit of introducing support for `metric_alarms` into the apps functions.
  * These will be enabled by default when adding a new app.

## Testing
* Updating unit tests for new app config format
* Ran `terraform validate` on the newly pared down `stream_alert_app_iam` module.
* Performed `python manage.py terraform generate` and `terraform validate` on the files in `terraform`


stream_alert_cli/config.py
```diff
@@ -407,18 +412,41 @@ def add_app_integration(self, app_info):
...
+                'log_retention_days': 14,
+                'memory': app_info['memory'],
+                'metric_alarms': {
+                    'errors': {
+                        'enabled': True,
+                        'evaluation_periods': 1,
+                        'period_secs': 120
+                    }
+                },
+                'schedule_expression': app_info['interval'],
```
austinbyers: Where does the `app_info` object come from? Can we rename `interval` to `schedule_expression` to be consistent?

stream_alert_cli/manage_lambda/package.py
```diff
@@ -320,6 +333,19 @@ class AppIntegrationPackage(LambdaPackage):
     precompiled_libs = {'boxsdk[jwt]==2.0.0a11'}
     version = apps_version
 
+    def _update_config_options(self, full_package_name, package_sha256):
+        """Overriding the base class method since apps store these settings in the cluster config"""
```
austinbyers: Nice!

stream_alert_cli/manage_lambda/version.py
```diff
@@ -73,7 +73,7 @@ def _version_helper(**kwargs):
 
         return int(version)
 
-    def _publish_helper(self, **kwargs):
+    def _publish_helper(self, cluster=None):
```
austinbyers: 🙌

terraform/modules/tf_stream_alert_app_iam/iam.tf
```diff
@@ -0,0 +1,81 @@
...
+// IAM Role Policy: Allow the StreamAlert App function to invoke itself
+resource "aws_iam_role_policy" "stream_alert_app_invoke_self_lambda_role_policy" {
+  name   = "${var.cluster}_${var.type}_app_invoke_self_lambda_role_policy"
+  role   = "${var.role_id}"
+  policy = "${data.aws_iam_policy_document.app_invoke_self_policy.json}"
+}
+
+// IAM Policy Doc: Allow the StreamAlert App function to invoke itself
+data "aws_iam_policy_document" "app_invoke_self_policy" {
+  statement {
```
austinbyers: Would this be simpler as another `statement` with the other Lambda invocation policy? Or even a single `AllowLambdaInvocations` statement. I like grouping statements by service, personally, but happy with whatever!

stream_alert_cli/config.py
ryandeivert: this was actually a bug :) thanks for flagging!

terraform/modules/tf_stream_alert_app_iam/iam.tf
ryandeivert: good suggestion @austinbyers - I've added a commit to address this. can you PTAL?

PR #659, opened 2018-03-28
fixing help message for athena cli commands

ryandeivert:
to: @chunyong-lin 
size: small

## Changes

* Small fixes to `manage.py` help strings for athena subcommands.


ryandeivert:
yep @chunyong-lin thanks for flagging!! those updates are in the pipeline :)

PR #663, opened 2018-03-29
fix RP memory leak due to record copying

chunyong-lin:
to: @austinbyers or @ryandeivert 
size: small
resolves N/A

## Background
We found out there is chance of Rule Processor will run out of memory due to extra copy action when checking alerts duplication, even these are local variables. 

To solve this issue completely, we will iterate through two dictionaries instead of copy them. 

## Changes

* Don't copy records! 
* Iterator through two dictionaries to do comparison. 

## Testing
* Rule testing
```
...
StreamAlertCLI [INFO]: (63/63) Successful Tests
StreamAlertCLI [INFO]: (34/34) Alert Tests Passed
```

* Unit testing
```
./tests/scripts/unit_tests.sh
...
Ran 561 tests in 9.848s

OK
```

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -32,6 +32,8 @@ RuleAttributes = namedtuple('Rule', ['rule_name',
                                      'req_subkeys',
                                      'context'])
 
+_IGNORE_KEYS = {StreamThreatIntel.IOC_KEY, NORMALIZATION_KEY}
+
 
 class StreamRules(object):
     """Container class for StreamAlert Rules
@@ -428,7 +430,7 @@ class StreamRules(object):
         """
         rule_result = StreamRules.process_rule(record, rule)
         if rule_result:
-            if StreamRules.check_alerts_duplication(record, rule, alerts):
+            if self.check_alerts_duplication(record, rule, alerts):
                 return
 
             alert_id = str(uuid.uuid4())  # Random unique alert ID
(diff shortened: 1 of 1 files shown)


stream_alert/rule_processor/rules_engine.py
```diff
@@ -470,22 +470,31 @@ def check_alerts_duplication(record, rule, alerts):
...
 
-                # There is chance that there are different values in 'streamalert:normalization'
-                # or 'streamalert:ioc' fields. So do comparision again after removing
-                # two keys.
-                record_copy.pop(StreamThreatIntel.IOC_KEY, None)
-                record_copy.pop(NORMALIZATION_KEY, None)
-                exist_alert_record_copy.pop(StreamThreatIntel.IOC_KEY, None)
-                exist_alert_record_copy.pop(NORMALIZATION_KEY, None)
-                if record_copy == exist_alert_record_copy:
+                found_duplication = False
```
ryandeivert: suggestion: could we simplify this and have this bool default to `True` then `break` like so:

```
found_duplication = True
for key in record_keys:
    if record[key] != exist_alert['record'][key]:
        found_duplication = False
        break
```

ryandeivert: I actually think we could improve this by inverting our logic.. so instead of checking if it IS a duplicate, let's check if it's _not_ a duplicate

This gives us the ability to exit early the moment we find a key/value that is NOT a match.. instead of having to look at ALL key/values to see that they all match before returning true. let's talk more about this before this merges :)

ryandeivert: I just did a quick test and inverse logic gets us a ~2x improvement

austinbyers: I thought this at first too, but we already terminate early when a duplicate is found. We can't return early if a duplicate is *not* found, because there could still be a duplicate later on.

Since this is difficult to understand, @chunyong-lin I recommend moving the inner-loop logic into its own method which returns `True` if one alert matches the given record. Then the code will be more readable, and you can write lots of unit tests for that inner method to make sure its working as expected

ryandeivert: yep sorry @austinbyers and @chunyong-lin after discussion this morning I take all fo that back :)

chunyong-lin:
This fix doesn't solve the memory leak issue. Will look for other solution.

PR #667, opened 2018-03-31
[Data Normalization] Fix memory exhausting issue

### LINKED ISSUE #654: Bug: Threat Intel prevent one record to be normalized more than once

chunyong-lin:
to: @jacknagz 
size: medium
resolves #654

## Background
The fix to allow one record to be normalized multiple times with **different set of datatypes** in PR #655 causes memory exhausting issue. We can reproduce the issue by profiling memory usage using library `memory_profile`.

The root cause of memory exhaustion is caused by one record is normalized multiple times with **same set of datatypes** and normalized records list is grown in linear speed when threat intel feature enabled. User may not notice the memory issue unless there are thousands records been normalized in one lambda invocation.

The solution is to check if a record has been normalized by the **same set of datatypes** or not. If so, use normalized information from existing normalized records, otherwise apply normalization to the record.

## Changes
* Extract Data Normalization and Threat Intel logic from `process` method.
* Check alert duplication.
* Add helper functions to prevent a record being normalized multiple times by the **same set of datatypes**.

## Testing
* Testing environment is locally:
  * A json file contains 7622 logs. Majority of logs are related to system processes. So the datatypes `command`, `fileHash` are heavily applied to normalize records.

* One record can  only be normalized once
  * the memory usage is 52.6MB
  * Generated 7622 normalized records
```
Line #    Mem usage    Increment   Line Contents
================================================
    68     85.8 MiB     85.8 MiB       @profile
    69                                 def run(self, event):
    70                                     """StreamAlert Lambda function handler.
  ...
   111                                         # Cache the log sources for this service and entity on the classifier
   112     87.0 MiB      0.0 MiB               if not self.classifier.load_sources(service, entity):
   113                                             continue
   114
   115                                         # Create the StreamPayload to use for encapsulating parsed info
   116     87.0 MiB      0.0 MiB               payload = load_stream_payload(service, entity, raw_record)
   117     87.0 MiB      0.0 MiB               if not payload:
   118                                             continue
   119
   120    120.6 MiB     33.6 MiB               payload_with_normalized_records.extend(self._process_alerts(payload))
   121
...
   159    121.2 MiB      0.0 MiB           if self._firehose_client:
   160    139.6 MiB     18.4 MiB               self._firehose_client.send()
...
   164    139.6 MiB      0.0 MiB           return self._failed_record_count == 0
```
  * After merge PR #655, one record can be normalized multiple time
    * The memory usage will be increased to 1,227.5 MB!! (1.2GB)
    * Generated 110,483 normalized records
```
Line #    Mem usage    Increment   Line Contents
================================================
    68     86.0 MiB     86.0 MiB       @profile
    69                                 def run(self, event):
...
    116     86.7 MiB      0.0 MiB               payload = load_stream_payload(service, entity, raw_record)
    117     86.7 MiB      0.0 MiB               if not payload:
    118                                             continue
    119
    120   1290.9 MiB   1204.2 MiB               payload_with_normalized_records.extend(self._process_alerts(payload))
...
    159   1297.4 MiB      0.0 MiB           if self._firehose_client:
    160   1314.2 MiB     16.8 MiB               self._firehose_client.send()
...
    164   1314.2 MiB      0.0 MiB           return self._failed_record_count == 0
```
* Apply current fix which is only allow a record to be normalized multiple times by the **different set of datatypes**
  * The memory usage looks reasonable, **143.4MB**
  * Generated 27,534 normalized records
```
Line #    Mem usage    Increment   Line Contents
================================================
    68     86.0 MiB     86.0 MiB       @profile
    69                                 def run(self, event):
...
   116     87.2 MiB      0.0 MiB               payload = load_stream_payload(service, entity, raw_record)
   117     87.2 MiB      0.0 MiB               if not payload:
   118                                             continue
   119
   120    212.6 MiB    125.5 MiB               payload_with_normalized_records.extend(self._process_alerts(payload))
...
   159    218.8 MiB      0.0 MiB           if self._firehose_client:
   160    230.6 MiB     11.8 MiB               self._firehose_client.send()
...
   164    230.6 MiB      0.0 MiB           return self._failed_record_count == 0
...

```
* Runtime profiling: No noticeable running time changed.

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -352,8 +352,6 @@ class StreamRules(object):
             return alerts, normalized_records
 
         for record in payload.records:
-            # One record may be added to normalized records list multiple time due
-            # to each record is processed by all rules.
             for rule in rules:
                 # subkey check
                 has_sub_keys = self.process_subkeys(record, payload.type, rule)
@@ -368,13 +366,15 @@ class StreamRules(object):
                 if rule.datatypes:
                     # When rule 'datatypes' option is defined, rules engine will
                     # apply data normalization to all the record.
-                    self._apply_normalization(record, normalized_records, rule, payload, alerts)
+                    record_copy = self._apply_normalization(record, normalized_records,
+                                                            rule, payload)
+                    self.rule_analysis(record_copy, rule, payload, alerts)
                 else:
                     self.rule_analysis(record, rule, payload, alerts)
 
         return alerts, normalized_records
 
-    def _apply_normalization(self, record, normalized_records, rule, payload, alerts):
+    def _apply_normalization(self, record, normalized_records, rule, payload):
         """Apply data normalization to current record
 
         Args:
@@ -414,7 +414,7 @@ class StreamRules(object):
             else:
                 record_copy = record
 
-        self.rule_analysis(record_copy, rule, payload, alerts)
+        return record_copy
 
     @staticmethod
     def _update_normalized_record(payload, record_copy, normalized_records):
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -32,6 +32,7 @@ RuleAttributes = namedtuple('Rule', ['rule_name',
                                      'req_subkeys',
                                      'context'])
 
+_IGNORE_KEYS = {StreamThreatIntel.IOC_KEY, NORMALIZATION_KEY}
 
 class StreamRules(object):
     """Container class for StreamAlert Rules
@@ -353,7 +354,6 @@ class StreamRules(object):
         for record in payload.records:
             # One record may be added to normalized records list multiple time due
             # to each record is processed by all rules.
-            normalized_record_appended = False
             for rule in rules:
                 # subkey check
                 has_sub_keys = self.process_subkeys(record, payload.type, rule)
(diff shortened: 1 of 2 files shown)


stream_alert/rule_processor/rules_engine.py
```diff
@@ -458,30 +527,42 @@ def rule_analysis(self, record, rule, payload, alerts):
...
+        Returns:
+            bool: Return True only when two records have same keys and values.
+        """
+        record1_keys = set(record1.keys()) - _IGNORE_KEYS
+        record2_keys = set(record2.keys()) - _IGNORE_KEYS
+
+        if record1_keys != record2_keys:
+            return False
+
+        return all(record1[key] == record2[key] for key in record1_keys)
```
jacknagz: This only checks if all `record_1` key/values are present in `record2`, but it doesn't check the other way around.  There's also a built-in `cmp()` method that you could use, but you'd have to pop out the ignore keys prior to comparison

stream_alert/rule_processor/rules_engine.py
```diff
@@ -365,33 +365,100 @@ def process(self, input_payload):
...
+                                            payload.normalized_types,
+                                            rule.datatypes)
+
+            if types_result:
+                record_copy = record.copy()
+
+                # Insert normalization result to the record copy.
+                record_copy[NORMALIZATION_KEY] = types_result
+
+                if self._threat_intel:
```
jacknagz: Is there a reason we need to couple Threat Intel to Data Normalization? This essentially means a user could not use normalization on its own

stream_alert/rule_processor/rules_engine.py
```diff
@@ -353,7 +354,6 @@ def process(self, input_payload):
         for record in payload.records:
             # One record may be added to normalized records list multiple time due
             # to each record is processed by all rules.
-            normalized_record_appended = False
```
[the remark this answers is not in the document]
chunyong-lin: 😄 good catch. I will remove these comments.

stream_alert/rule_processor/rules_engine.py
```diff
@@ -365,33 +365,100 @@ def process(self, input_payload):
...
-                        payload_copy = copy(input_payload)
-                        payload_copy.pre_parsed_record = record_copy
-                        payload_copy.records = None
-                        payload_copy.raw_record = None
-                        normalized_records.append(payload_copy)
-                        normalized_record_appended = True
+                    # When rule 'datatypes' option is defined, rules engine will
+                    # apply data normalization to all the record.
+                    self._apply_normalization(record, normalized_records, rule, payload, alerts)
                 else:
```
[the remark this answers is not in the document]
chunyong-lin: The else part covers to apply rule analysis logic to the rules don't have data normalization applied.

stream_alert/rule_processor/rules_engine.py
```diff
@@ -365,33 +365,100 @@ def process(self, input_payload):
...
-                        # The payload metadata includes log_source, type, service,
-                        # and entity. The metadata will be returned to along with
-                        # normalized record for threat detection.
-                        payload_copy = copy(input_payload)
-                        payload_copy.pre_parsed_record = record_copy
-                        payload_copy.records = None
-                        payload_copy.raw_record = None
-                        normalized_records.append(payload_copy)
-                        normalized_record_appended = True
+                    # When rule 'datatypes' option is defined, rules engine will
```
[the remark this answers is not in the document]
chunyong-lin: The comment here is actually helpful that when `datatypes` is defined in the rule, it triggers to apply data normalization to the records. The logic in `else` is for rules without data normalization.

stream_alert/rule_processor/rules_engine.py
chunyong-lin: A user can use normalization on its own by just define `datatypes` option in the Rule. Threat Intel is based on Data Normalization so threat intel will only be triggered when the normalization key existed.

stream_alert/rule_processor/rules_engine.py
```diff
@@ -365,33 +365,100 @@ def process(self, input_payload):
...
+
+                if self._threat_intel:
+                    # If threat intel is enabled, add newly normalized record to
+                    # the list for future threat detection.
+                    self._update_normalized_record(payload, record_copy, normalized_records)
+
+            else:
+                record_copy = record
+
+        self.rule_analysis(record_copy, rule, payload, alerts)
```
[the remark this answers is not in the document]
chunyong-lin: 👍 Good catch! I will make the change.

stream_alert/rule_processor/rules_engine.py
chunyong-lin: You are right! I thought about this. Two lines above where comparing if two set of keys from `record_1` and `record_2` are same. If they are not same, return False immediately.  It prevents us to check the other way around.

chunyong-lin:
@jacknagz Thanks for the comments. PTAL.

stream_alert/rule_processor/rules_engine.py
```diff
@@ -365,33 +365,100 @@ def process(self, input_payload):
...
+                have been normalized.
+            rule (namedtuple): Contains alerting logic.
+            payload (namedtuple): Contains parsed logs.
+            alerts (list): A list of triggered alerts.
+        """
+        normalized, normalized_payload = self._is_normalized(record,
+                                                             normalized_records,
+                                                             rule.datatypes)
+
+        record_copy = None
```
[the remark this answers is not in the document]
jacknagz: Maybe I didn't explain that well enough. Does this needs to be defined in this line if it's re-defined below?

stream_alert/rule_processor/rules_engine.py
jacknagz: > A user can use normalization on its own by just define datatypes option in the Rule.

But the normalized record is only added to the `normalized_records` list if `self._threat_intel` is True, meaning it's tightly coupled to having Threat Intelligence enabled

jacknagz: Wouldn't we want the `self._threat_intel` check to happen outside the scope of normalization to get the benefits of both?

stream_alert/rule_processor/rules_engine.py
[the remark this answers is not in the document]
chunyong-lin: Aha, yes. It needs to be defined in the line, otherwise it will have error `Unbound variable`.

stream_alert/rule_processor/rules_engine.py
chunyong-lin: This block is only to add normalized records to a list. So it has to be happened right after data normalization. The list contains normalized records will be used when threat intel enabled in [handler.py](https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/handler.py#L125)

PR #670, opened 2018-04-05
[cli] adding support for alert table rebuilding

ryandeivert:
to: @austinbyers
size: small

## Background

The CLI currently lacks the ability to rebuild the athena 'alerts' table. This abstracts away the differentiators between 'data' tables and the 'alerts' table so the same code can be used for rebuilding both.

**NOTE**: This PR is currently being compared to @austinbyers `austin-alert-class` branch since it is dependent on some of his changes. Once his PR merges, I will switch this to compare against master.

## Changes

* Updating athena cli code to handle refreshing both 'data' and 'alerts' table types.
* This uses @austinbyers `Alert` class to determine the most up-to-date set of keys and value types in an alert and transforms that into a schema that reflects said keys/values.
* Removing '--table-type' requirement from athena cli commands since it can be deduced from the '--table-name' instead

## Testing

* Adding tests for simple `record_to_schema` helper function that was added.

### DIFF
--- a/stream_alert_cli/athena/handler.py
+++ b/stream_alert_cli/athena/handler.py
@@ -16,22 +16,12 @@ limitations under the License.
 from stream_alert.athena_partition_refresh.main import StreamAlertAthenaClient
 from stream_alert.athena_partition_refresh import helpers as athena_helpers
 from stream_alert.rule_processor.firehose import StreamAlertFirehose
+from stream_alert.shared.alert import Alert
 from stream_alert_cli.logger import LOGGER_CLI
-from stream_alert_cli.helpers import continue_prompt
+from stream_alert_cli.helpers import continue_prompt, record_to_schema
 from stream_alert_cli.athena import helpers as handler_helpers
 
-ALERTS_TABLE_STATEMENT = ('CREATE EXTERNAL TABLE alerts ('
-                          'log_source string,'
-                          'log_type string,'
-                          'outputs array<string>,'
-                          'record string,'
-                          'rule_description string,'
-                          'rule_name string,'
-                          'source_entity string,'
-                          'source_service string)'
-                          'PARTITIONED BY (dt string)'
-                          'ROW FORMAT SERDE \'org.openx.data.jsonserde.JsonSerDe\''
-                          'LOCATION \'s3://{bucket}/alerts/\'')
+
 CREATE_TABLE_STATEMENT = ('CREATE EXTERNAL TABLE {table_name} ({schema}) '
                           'PARTITIONED BY (dt string) '
                           'ROW FORMAT SERDE \'org.openx.data.jsonserde.JsonSerDe\' '
@@ -41,7 +31,7 @@ CREATE_TABLE_STATEMENT = ('CREATE EXTERNAL TABLE {table_name} ({schema}) '
 MAX_QUERY_LENGTH = 262144
 
 
-def rebuild_partitions(table, bucket, table_type, config):
+def rebuild_partitions(table, bucket, config):
     """Rebuild an Athena table's partitions
 
     Steps:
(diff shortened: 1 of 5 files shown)


tests/unit/stream_alert_cli/test_helpers.py
```diff
@@ -95,3 +101,59 @@ def test_get_rules_from_test_events():
...
+    result = helpers.record_to_schema(record, recursive=False)
+
+    assert_is_instance(result['dict'], dict)
+    assert_is_instance(result['list'], list)
+    assert_equal(result['string'], 'string')
+    assert_equal(result['float'], 'float')
+    assert_equal(result['integer'], 'integer')
+    assert_equal(result['boolean'], 'boolean')
+    # Check to make no recursion occurred
+    assert_equal(len(result['dict']), 0)
```
austinbyers: Instead of all these different statements, can you have just a single `assert_equal`? I find that easier to read and understand

```python
result = helpers.record_to_schema(record, recursive=False)

expected = {
    'dict': {},
    'list': [],
    'string': 'string',
    ...,
}
assert_equal(result, expected)
```

Also down below

stream_alert_cli/helpers.py
```diff
@@ -34,6 +34,41 @@
...
+    bool: 'boolean',
+    float: 'float',
+    int: 'integer',
+    str: 'string',
+    dict: dict(),
+    list: list()
+}
+
+
+def record_to_schema(record, recursive=False):
```
austinbyers: We're doing 2 layers of translation: The schema is converted here to 'boolean', 'float', etc, and then again in a downstream helper function to the underlying Hive types:

https://github.com/airbnb/streamalert/blob/master/stream_alert_cli/athena/helpers.py#L18-L25

Is there any way to skip this intermediate representation?

ryandeivert: @austinbyers can you clarify what you mean by `intermediate representation`? do you mean combine them into one? I added the record_to_schema because it could be handy elsewhere (for example: we had discussed potentially auto-generating schema suggestions for users if there is ever a 'failed parse')

stream_alert_cli/athena/handler.py
```diff
@@ -199,31 +185,37 @@ def create_table(table, bucket, table_type, config, schema_override=None):
...
-                                          config['logs'])
+    sa_firehose = StreamAlertFirehose(config['global']['account']['region'],
+                                      config['global']['infrastructure']['firehose'],
+                                      config['logs'])
+
+    # Convert special characters in schema name to underscores
+    sanitized_table_name = sa_firehose.firehose_log_name(table)
+
+    # Check that the log type is enabled via Firehose
+    if sanitized_table_name not in sa_firehose.enabled_logs:
```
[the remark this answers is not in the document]
ryandeivert: good catch - I'll add a check for that

stream_alert_cli/helpers.py
austinbyers: Ah, `record_to_schema` is expecting the same schema type as we use in `conf/logs.json`. That makes more sense, sorry I missed that. Can you add a comment somewhere clarifying that the `schema` being returned here is compatible with the log schema?

ryandeivert: yes sorry for the ambiguity/confusion :) I will add a comment to clarify - thanks for flagging!

ryandeivert: added the requested comment!

ryandeivert:
for peace of mind, here's the output for rebuilding a data table:
```
$ python manage.py athena rebuild-partitions --bucket rdrefr003.streamalert.data --table-name cloudwatch:events
StreamAlertCLI [INFO]: Dropping table cloudwatch_events
StreamAlertCLI [INFO]: Dropped table cloudwatch_events
StreamAlertCLI [INFO]: Creating table cloudwatch_events
StreamAlertCLI [INFO]: The cloudwatch_events table was successfully created!
StreamAlertCLI [INFO]: Creating 1 new partitions for cloudwatch_events
StreamAlertCLI [INFO]: Successfully rebuilt partitions for cloudwatch_events
```

chunyong-lin:
Well done 👏 👏 👏 👏 👏

PR #671, opened 2018-04-09
[app] salesforce app integration

chunyong-lin:
to: @ryandeivert @jacknagz 
size: large
resolves N/A

## Background
Many companies are using Salesforce service to support customer relationship management. It is important to audit user access and prevent data loss from security perspective. Salesforce provides [42 types of event](https://developer.salesforce.com/docs/atlas.en-us.208.0.object_reference.meta/object_reference/sforce_api_objects_eventlogfile.htm) in latest API version.

New StreamAlert App for Salesforce will collect *Console*, *Login*, *LoginAs*, *Report* and *ReportAs* event logs and send to StreamAlert Rule Processor for further analysis. Later on, we can write rules to alert on any unexpected behaviors.

To collect Salesforce logs, we will use [Event Monitoring](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/using_resources_event_log_files.htm) which is an API only feature. Each organization's event log files are stored in an API object called EventLogFile. 
* The `SalesforceApp` will first retrieve a list of EventLogFiles with filters by `LogDate` and `EventType`. 
* Then it will send another GET requests to fetch log events by log file locations. 
* `SalesforceApp` will segment and forward log events to the rule processor. 

## prerequisite
* A user account in our Salesforce account.
* The security token associated to your user account.
* Create a connected app, following [Salesforce documentation](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/intro_defining_remote_access_applications.htmhttps://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/intro_defining_remote_access_applications.htm). 
 
## Changes

* Create a new app for **Salesforce** following the [Developing a New App](https://streamalert.readthedocs.io/app-development.html) documentation. 
* Add unit test cases.

## Testing
* Unit test coverage is 100%!
```
./tests/scripts/unit_tests.sh
...
Name                                                    Stmts   Miss  Cover   Missing
...
app_integrations/apps/salesforce.py                       131      0   100%
...
Ran 618 tests in 9.930s

OK
```
* Tested in testing accounts.


app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+            bool: Returns True if update auth headers and instance url successfully.
+        """
+        headers = {'Content-Type': 'application/x-www-form-urlencoded'}
+        # required credentials when request for a token.
+        data = {
+            'grant_type': 'password',
+            'client_id': self._config.auth['client_id'],
+            'client_secret': self._config.auth['client_secret'],
+            'username': self._config.auth['username'],
+            'password': self._config.auth['password']+self._config.auth['security_token'],
```
ryandeivert: please use str.format() for concatenation instead of `+` like: `'{}{}'.format(..., ...)`

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+            'response_type': 'code',
+            'redirect_uri': self._SALESFORCE_TOKEN_URL
+        }
+        success, response = self._make_post_request(self._SALESFORCE_TOKEN_URL, headers, data)
+
+        if not success:
+            return False
+
+        if not response or not response.get('access_token') or not response.get('instance_url'):
```
ryandeivert: there's a lot of `not` negative logic here.. can we do this instead:

`if not (response and response.get('access_token') and response.get('instance_url'):...`

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+        elif resp.status_code > 200:
+            LOGGER.error('Unexpected status code %d detected, error message %s',
+                         resp.status_code,
+                         resp.json())
+            return False
+
+    def _make_get_request(self, full_url, headers, params=None):
+        """Make GET request with backoff logic
+
+        This method overrides the method in AppIntegration base class to handle
```
ryandeivert: I think we can probably modify the base class's method to handle returning data as json or raw response object. you could still have something in here for doing the backoff, but ideally all get requests would go through the base class if possible. we can chat more about how to achieve this (i'm thinking just have a `kwargs` argument that could include a `return_json` param or something)

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+            return False
+
+        bearer = 'Bearer {}'.format(response.get('access_token'))
+        self._auth_headers = {
+            'Content-Type': 'application/json',
+            'Authorization': bearer
+        }
+        self._instance_url = response.get('instance_url')
+        LOGGER.debug('Successfully obtain OAuth token and instance URL')
+        return True
```
ryandeivert: consider having this return value be `return self._instance_url and self._auth_headers` (would be `False` if either of those are `None`)

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+            return None
+        return [log for log in reader]
+
+    def _gather_logs(self):
+        """Gather all log events. There are 32 event types.
+
+        Returns:
+            list: A list of dictionaries contains log events.
+        """
+        self._request_token()
```
ryandeivert: it looks like `_request_token` returns a bool - can we do a check here and return if this returns False (instead of doing the check in `_get_latest_api_version` for the `_instance_url`, etc)?

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+            },
+            {
+                "version": "26.0",
+                "label": "Winter '13",
+                "url": "/services/data/v26.0"
+            }
+        ]
+        """
+        # instance_url identifies the Salesforce instance to which API calls should be sent.
+        if not self._instance_url:
```
ryandeivert: remove this check if you take my suggestion for doing this check in `_request_token`

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+            LOGGER.error('No instance url found for %s %s', self.service, self._type)
+            return
+
+        url = '{}/services/data/'.format(self._instance_url)
+        success, response = self._make_get_request(url, self._auth_headers)
+
+        if not success:
```
ryandeivert: make this `if not (success and response):...` and then you can remove the `if response` check below

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+            LOGGER.error('Failed to fetch lastest api version')
+            return
+
+        if response:
+            versions = [float(version.get('version', 0)) for version in response]
+            if versions:
+                self._latest_api_version = str(sorted(versions)[-1])
+                LOGGER.debug('Successfully obtain latest API version %s', self._latest_api_version)
```
ryandeivert: should we log an error if this fails? ie if `versions` is empty?

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+
+        Returns:
+            list: A list of dictionaries contains log events.
+        """
+        self._request_token()
+        self._get_latest_api_version()
+
+        log_files = self._list_log_files()
+        if not log_files:
+            return None
```
ryandeivert: `None` return is implicit and can be omitted

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+        url = '{}/{}'.format(self._instance_url, log_file_path)
+        data = None
+        try:
+            success, resp = self._make_get_request(url, self._auth_headers)
+            if not success:
+                LOGGER.error('Failed to get event logs')
+                return None
+
+            data = StringIO.StringIO(resp)
+            reader = csv.DictReader(data)
```
ryandeivert: isn't this going to be doing more work than we need to and increasing memory/size of logs we're sending to the rule processor? this could be problematic if we're running through thousands of lines and the rule processor already has logic for doing the csv -> json transformation. let's discuss this more if you'd like, but I'd be more comfortable if we just stripped off the header line (skip line 1) and send the lines as raw csv data. the rule processor will then use the schema to remap the line to it's header values and create the dict object

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+                    "EventType": "Console",
+                    "LogFile": "/services/data/v32.0/sobjects/EventLogFile/0ATD00001SdOAI/LogFile",
+                    "LogDate": "2014-03-13T00:00:00.000+0000",
+                    "LogFileLength": 1345.0
+                }
+            ]
+        }
+        """
+        log_files = []
+        for event_type in self._SALESFORCE_EVENT_TYPES:
```
ryandeivert: +++ beautiful!!

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+import backoff
+import requests
+
+from app_integrations import LOGGER
+from app_integrations.apps.app_base import StreamAlertApp, AppIntegration
+
+
+class SalesforceAppError(Exception):
+    """Salesforce App Error class"""
+    pass
```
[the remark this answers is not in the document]
jacknagz: Good to know

app_integrations/apps/salesforce.py
jacknagz: @ryandeivert could we just use `all()` in this case?

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,395 @@
...
+            'Content-Type': 'application/json',
+            'Authorization': bearer
+        }
+        self._instance_url = response.get('instance_url')
+        LOGGER.debug('Successfully obtain OAuth token and instance URL')
+        return True
+
+    def _sleep_seconds(self):
+        """Salesforce doesn't provide rate limit information. We can default it to 0"""
+        return 0
```
jacknagz: @chunyong-lin do you see this as something we'll have to tune in the future?

app_integrations/apps/salesforce.py
jacknagz: Agreed.  The rule processor also has additional safety checks, etc for parsing out CSV data.  We also haven't really used it much yet and this would be a great opportunity to do so!

jacknagz:
Looking good!!!!

app_integrations/apps/salesforce.py
chunyong-lin: I will take @ryandeivert's suggestion. If use `all()`, it will check all conditions.
if use `if not (response and response.get('access_token') and response.get('instance_url')`, it will not check rest conditions if `response` if `None` or `False`.

app_integrations/apps/salesforce.py
chunyong-lin: We may tune this in the future but low chance. I don't see the detail document regarding rate limit in Salesforce. I have error logging when it happens which is `status_code` 403.
Also, this method is base class, it is required to be defined in subclass.

app_integrations/apps/salesforce.py
chunyong-lin: Yes, I agree with you two. I will make the change. 😄

app_integrations/apps/salesforce.py
ryandeivert: yeah I think `all()` is great for generators or iterables but not great when you're just passing in a bunch of vars

app_integrations/apps/salesforce.py
ryandeivert: this is great added context - thank you for clarifying!! in this case, I think you can leave it as is. we can investigate ways to make the subclass more flexible/adaptable in the future :)

app_integrations/apps/salesforce.py
ryandeivert: yeah this is an abstractmethod that's currently enforced for subclasses - I may remove that requirement as it seems this is a less commonly used feature

app_integrations/apps/salesforce.py
ryandeivert: if you want to speak more about subclassing your new app to handle different log types let me know!!

chunyong-lin:
@ryandeivert @jacknagz PTAL 😄 😸 😃

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,456 @@
...
+
+        Args:
+            log_file_path (str): log file location.
+
+        Returns:
+            list: a list of event logs or None.
+        """
+        url = '{}/{}'.format(self._instance_url, log_file_path)
+        data = None
+        try:
```
ryandeivert: what do you think about changing this try.. I think the StringIO stuff might be unnecessary 🤔. isn't `resp` just a string?

you might be able to just do:
```
        try:
            success, resp = self._make_get_request(url, self._auth_headers)
        except SalesforceAppError:
            LOGGER.exception('An exception occurred when getting event logs')

        if not (success and resp):
            LOGGER.error('Failed to get event logs')
            return

        return resp.splitlines()[1:]
```

app_integrations/apps/salesforce.py
```diff
@@ -327,24 +329,23 @@ def _list_log_files(self):
...
-                api_version=self._latest_api_version,
-                query=self._SALESFORCE_QUERY_FILTERS,
-                start_time=self._SALESFORCE_CREATE_AFTER.format(self._last_timestamp),
-                event_type=event_type_format
-            )
-            success, response = self._make_get_request(url, self._auth_headers)
-            if not success:
-                LOGGER.error('Failed to get a list of log files.')
-                return log_files
+        event_type_format = 'AND+EventType+=+\'{}\''.format(self._type)
```
ryandeivert: you can probably just move this var declaration right to the `event_type` keyword argument right?

app_integrations/apps/salesforce.py
```diff
@@ -0,0 +1,456 @@
...
+                data.close()
+            return
+
+    def _gather_logs(self):
+        """Gather all log events. There are 32 event types.
+
+        Returns:
+            list: A list of dictionaries contains log events.
+        """
+        if not (self._request_token() and self._get_latest_api_version()):
```
ryandeivert: woohoo!!

app_integrations/apps/salesforce.py
```diff
@@ -396,3 +397,60 @@ def _gather_logs(self):
...
+@StreamAlertApp
+class SalesforceReportExport(SalesforceApp):
+    """Salesforce ReportExport Events app integration
+
+    ReportExport events contain details about reports that a user exported
+    """
+
+    @classmethod
+    def _type(cls):
+        return 'ReportExport'
```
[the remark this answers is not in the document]
chunyong-lin: Actually, I just did test, you are right, it is case-insensitive. I will make the change.

app_integrations/apps/salesforce.py
chunyong-lin: You are right! Using StringIO is over kill.

app_integrations/apps/salesforce.py
```diff
@@ -360,24 +358,19 @@ def _fetch_event_logs(self, log_file_path):
...
-                data.next()
-                return data.readlines()
-        except (SalesforceAppError, IOError):
+        except SalesforceAppError:
             LOGGER.exception('Failed to get event logs')
-            if data:
-                data.close()
             return
 
+        if not (success or resp):
```
ryandeivert: don't you want `and` here?

chunyong-lin: Aha, good catch!

PR #673, opened 2018-04-10
[athena] adding execution ID to logger output message in case of error

ryandeivert:
to: @jacknagz 
size: small

## Background

We have no context on executions IDs if an error occurs. This adds it.

## Changes

* Logging execution ID to log error output.


ryandeivert:
Actually I think I misunderstood what you were requesting, I can add this today

PR #678, opened 2018-04-11
[athena] adding StateChangeReason to logged error in athena

ryandeivert:
to: @jacknagz 
size: small

## Background

Iterating on #673

## Testing


stream_alert/athena_partition_refresh/main.py
```diff
@@ -182,18 +182,28 @@ def check_query_status(self, query_execution_id):
...
+        def _check_status(query_execution_id):
+            resp = self.athena_client.get_query_execution(
                 QueryExecutionId=query_execution_id
-            )['QueryExecution']['Status']['State']
+            )
+
+            state = resp['QueryExecution']['Status']['State']
+            in_progress = state in states_to_backoff
+
+            if not in_progress and state != 'SUCCEEDED':
```
jacknagz: Could we just check if the state is `'FAILED'|'CANCELLED'` vs inverse logic?

stream_alert/athena_partition_refresh/main.py
```diff
@@ -182,18 +182,28 @@ def check_query_status(self, query_execution_id):
...
+                reason = resp['QueryExecution']['Status']['StateChangeReason']
+                LOGGER.error('Bad state recveived: %s, reason: %s', state, reason)
```
jacknagz: Can we adjust the verbiage to say something like: "The query {id} {state} with response {reason}" or something along those lines

ryandeivert:
hey @jacknagz please take another look. I tweaked this to act a bit differently than before - please confirm the logic makes sense

jacknagz:
sorry for the delay!!

PR #679, opened 2018-04-11
[tf] add table for storing rule information

ryandeivert:
to: @austinbyers 
size: small

## Background

We need to begin storing some general information related to rules to enable staging/promotion/etc. This is the beginning of changes to enable that.

## Changes

* Creating `rules_tables` DynamoDB table in terraform with a hash key of `RuleName`
* Updating terraform generation for read/write capacity for this table.
* General cleanup of some generation code where kwargs is unnecessary and/or being abused.

## Incoming

* Another PR is incoming that will actually add information to this table upon deploy.

## Testing

Updating unit tests for new `rules_table` settings.

### DIFF
--- a/conf/global.json
+++ b/conf/global.json
@@ -22,7 +22,7 @@
     },
     "rules_table": {
       "read_capacity": 20,
-      "write_capacity": 2
+      "write_capacity": 5
     }
   },
   "terraform": {
--- a/conf/global.json
+++ b/conf/global.json
@@ -19,6 +19,10 @@
     },
     "monitoring": {
       "create_sns_topic": true
+    },
+    "rules_table": {
+      "read_capacity": 20,
+      "write_capacity": 2
     }
   },
   "terraform": {
--- a/stream_alert_cli/manage_lambda/version.py
+++ b/stream_alert_cli/manage_lambda/version.py
@@ -32,46 +32,15 @@ class LambdaVersion(object):
     All StreamAlert setups should start with "$LATEST".
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, config, package):
         """Initialize the version publishing
 
         Keyword Args:
             config (CLIConfig): Loaded StreamAlert CLI Config
             package (LambdaPackage): The created Lambda Package
         """
-        self.config = kwargs['config']
-        self.package = kwargs['package']
-
-    @staticmethod
-    def _version_helper(**kwargs):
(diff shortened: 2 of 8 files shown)


conf/global.json
```diff
@@ -19,6 +19,10 @@
     },
     "monitoring": {
       "create_sns_topic": true
+    },
+    "rules_table": {
+      "read_capacity": 20,
+      "write_capacity": 2
```
austinbyers: My personal preference - let's not go below 5 for capacity. It's the default, it's cheap, and it works well for most simple use cases. Of course, we should enable autoscaling eventually, but for now 5 seems fine with some wiggle room

stream_alert_cli/manage_lambda/version.py
```diff
@@ -32,15 +32,15 @@ class LambdaVersion(object):
...
         """Initialize the version publishing
 
         Keyword Args:
             config (CLIConfig): Loaded StreamAlert CLI Config
             package (LambdaPackage): The created Lambda Package
         """
-        self.config = kwargs['config']
-        self.package = kwargs['package']
+        self.config = config
+        self.package = package
```
austinbyers: 👍 😸

conf/global.json
ryandeivert: okay good suggestion - I had set it at 2 because this will be written to fairly rarely, but will set to 5 at minimum

stream_alert_cli/manage_lambda/version.py
ryandeivert: we've talked about this before but:

1) kwargs adds weird complexity like having to do a dict lookup inside of the called function. this is especially the case when there are only a few args or the function doesn't have a bunch of varied arguments it should expect
2) the caller can still refer to the argument by name, (without using kwargs) so the interface doesn't change.
3) not using kwargs has the added advantage of ensuring a function is actually used properly... usage of kwargs exclusively would allow you to call the function with no args at all which would fail in most areas use it throughout code. this is bad design and the use would never know what they were doing wrong
4) here is a good use of kwargs (ignore docstring that I have to update): https://github.com/airbnb/streamalert/pull/679/files#diff-e92d6322b319cf27f8967bef12e023d4R51
5) bad use of kwargs: https://github.com/airbnb/streamalert/pull/679/files#diff-e92d6322b319cf27f8967bef12e023d4L410

PR #681, opened 2018-04-12
Create and deliver merged alerts!

austinbyers:
to: @ryandeivert 
size: large

## Background

#666 implemented the algorithm to create merged alerts, but the logic is not currently used. In this PR, the Alert Merger actually creates and delivers merged alerts!

### Alert Lifecycle
The lifecycle of an alert with merge configuration is now as follows:

1. An alert is created by the Rule Processor and saved to the Dynamo alerts table as usual
2. The Alert Merger sees the new alert and immediately forwards it to the Alert Processor
3. Because the alert has merge configuration, the Alert Processor **only sends to required outputs**  (`aws-firehose:alerts`).
    - In the future, the user will be able to specify which outputs should receive merged vs unmerged alerts.
4. After the alert is sent to firehose, the Alert Processor updates `SentOutputs` in Dynamo to indicate that the alert has finished sending.
5. In the next Alert Merger table scan, it sees the alert with merge information that has already sent to firehose. This alert is now eligible for merging! If at least the merge window has passed since the alert was created, the Alert Merger will create a new Alert which combines our alert with any others.
6. The new, merged alert is saved to Dynamo and forwarded to the Alert Processor
7. The Alert Merger deletes all of the original alerts (which have now merged)
8. Because the merged alert does not itself have a merge configuration, the Alert Processor sends to all of the specified outputs.
9. Once the alert has been delivered to all outputs, the Alert Processor deletes the alert (since it isn't going to be merged later).

An alert which does not have merge configuration follows steps 1-2 and 8-9

## Changes

* Minimize the number of interactions with the Dynamo table.
    * This means the new alerts and deleted alerts are all batched until the end of the alert merger run.
    * #666 changed the alert merger to send only the Dynamo key so that the Alert Processor could pull the full record, but in many (maybe even most) cases, this results in a redundant query. Instead, _the full alert record is sent if possible_, falling back to just the key only if the payload is too large to send in an asynchronous invocation.
* Dates and durations in the `Alert` class have been changed to `datetime` and `timedelta` objects, respectively. These higher-level Python libraries make computing time differences super easy.
* The `RetryOutputs` column in the alerts table has been replaced with `OutputsSent`. The Alert Processor needs a way to indicate when an alert has finished sending, since now it can't just always delete it. Rather than creating a new `FinishedSending` column, I just inverted the logic of `RetryOutputs`.
* The `Alert` class raises a new `AlertCreationError` if an alert can't be created because of an invalid format. This error is caught and logged by the Alert Merger. This provides additional robustness against corrupted alert rows and makes future alert schema changes easier (since the old schema will just log an error instead of triggering an exception)
* `merge_by_keys` and `merge_window_mins` have been added to the `@rule` decorator

## Testing

- 100% test coverage for most modified files:
    - `alert_merger/main.py`
    - `alert_processor/main.py`
    - `shared/alert.py`
    - `shared/alert_table.py`

### Deploy to Test Account
1. Setup SNS inbound topic to invoke Rule Processor
2. Create 2 rules, one which merges and one which does not

```python
@rule(logs=['binaryalert'], outputs=['aws-sns:sample-topic', 'slack:mychannel'],
      merge_by_keys=['MatchedRules'], merge_window_mins=2)
def binaryalert_yara_match(rec):
    return rec['NumMatchedRules'] > 0

@rule(logs=['binaryalert'], outputs=['aws-sns:sample-topic', 'slack:mychannel'])
def binaryalert_yara_match_no_merge(rec):
    return rec['NumMatchedRules'] > 0
```

3. Run this script 10 times in rapid succession:
```python
import boto3, json, random

topic = boto3.resource('sns').Topic('...')

topic.publish(
    Message=json.dumps({
        'NumMatchedRules': random.randint(1, 100),
        'FileInfo': {
            'there': None,
            'nested': [1, 2, 'hi']
        },
        'MatchedRules': {
            'Rule1': 'Stuff'
        }
    })
)
```

4. This creates 20 alerts - 10 which should be merged and 10 which should not. 
5. The 10 `*_no_merge` alerts landed immediately in Slack and SNS
6. A few minutes later, this merged alert landed in Slack and SNS (technically, the `record` is the merged alert):
```json
{
  "cluster": "prod", 
  "context": {}, 
  "created": "2018-04-12T00:41:06.576999Z", 
  "id": "9abf60c9-9b9d-4c3e-bef8-4dd276042792", 
  "log_source": "binaryalert", 
  "log_type": "json", 
  "outputs": [
    "aws-firehose:alerts", 
    "aws-sns:sample-topic", 
    "slack:mychannel"
  ], 
  "record": {
    "AlertCount": 10, 
    "AlertTimeFirst": "2018-04-12T00:38:48.974498Z", 
    "AlertTimeLast": "2018-04-12T00:38:55.615220Z", 
    "MergedBy": {
      "MatchedRules": {
        "Rule1": "Stuff"
      }
    }, 
    "OtherCommonKeys": {
      "FileInfo": {
        "nested": [
          1, 
          2, 
          "hi"
        ], 
        "there": null
      }
    }, 
    "ValueDiffs": {
      "2018-04-12T00:38:48.974498Z": {
        "NumMatchedRules": 68
      }, 
      "2018-04-12T00:38:49.556785Z": {
        "NumMatchedRules": 6
      }, 
      "2018-04-12T00:38:50.014491Z": {
        "NumMatchedRules": 10
      }, 
      "2018-04-12T00:38:50.415742Z": {
        "NumMatchedRules": 67
      }, 
      "2018-04-12T00:38:51.334674Z": {
        "NumMatchedRules": 78
      }, 
      "2018-04-12T00:38:52.135900Z": {
        "NumMatchedRules": 47
      }, 
      "2018-04-12T00:38:53.036954Z": {
        "NumMatchedRules": 83
      }, 
      "2018-04-12T00:38:53.774258Z": {
        "NumMatchedRules": 83
      }, 
      "2018-04-12T00:38:54.736934Z": {
        "NumMatchedRules": 65
      }, 
      "2018-04-12T00:38:55.615220Z": {
        "NumMatchedRules": 42
      }
    }
  }, 
  "rule_description": "\n    author:       Austin Byers (Airbnb CSIRT)\n...", 
  "rule_name": "binaryalert_yara_match", 
  "source_entity": "austin-sa-source", 
  "source_service": "sns", 
  "staged": false
}
```
7. 21 alerts were visible in Athena - the 20 original alerts as well as the merged alert


stream_alert/alert_merger/main.py
```diff
@@ -42,30 +65,120 @@ def __init__(self):
...
         MetricLogger.log_metric(ALERT_MERGER_NAME, MetricLogger.ALERT_ATTEMPTS, alert.attempts)
 
+        record_payload = json.dumps(
+            alert.dynamo_record(), cls=Alert.AlertEncoder, separators=(',', ':'))
+        if len(record_payload) <= self.MAX_LAMBDA_PAYLOAD_SIZE:
+            # The entire alert fits in the Lambda payload - send it all
+            payload = record_payload
+        else:
+            # The alert is too big - the alert processor will have to pull it from Dynamo
+            payload = json.dumps(alert.dynamo_key)
```
ryandeivert: This is super cool !!

stream_alert/alert_processor/main.py
```diff
@@ -140,36 +138,47 @@ def _update_table(self, alert, output_results):
...
-                Invalid outputs are excluded
+            dict: Maps output (str) to whether it sent successfully (bool).
+                An empty dict is returned if the Alert was improperly formatted.
         """
-        alert = self.alerts_table.get_alert(rule_name, alert_id)
-        if not alert:
-            LOGGER.error('Alert %s does not exist', alert_id)
-            return
+        # Grab the alert record from Dynamo (if needed).
+        if set(event) == {'AlertID', 'RuleName'}:
```
ryandeivert: ++

stream_alert/shared/alert.py
```diff
@@ -81,12 +91,15 @@ def __init__(self, rule_name, record, outputs, **kwargs):
         self.attempts = int(kwargs.get('attempts', 0)) or 0  # Convert possible Decimal to int
         self.cluster = kwargs.get('cluster') or None
         self.context = kwargs.get('context') or {}
-        self.dispatched = int(kwargs.get('dispatched', 0)) or 0
+
+        # datetime.min isn't supported by strftime, so use Unix epoch instead for default value
+        self.dispatched = kwargs.get('dispatched') or datetime(year=1970, month=1, day=1)
```
ryandeivert: tip: you can also use `time.gmtime(0)` to accomplish this. but not a necessary change since `time` is not imported here

stream_alert/shared/alert.py
```diff
@@ -152,27 +175,34 @@ def create_from_dynamo_record(cls, record):
...
+                merge_by_keys=record.get('MergeByKeys'),
+                merge_window=timedelta(minutes=int(record.get('MergeWindowMins', 0))),
+                outputs_sent=set(record.get('OutputsSent') or []),
+                rule_description=record.get('RuleDescription'),
+                source_entity=record.get('SourceEntity'),
+                source_service=record.get('SourceService'),
+                staged=record.get('Staged')
+            )
+        except (KeyError, TypeError, ValueError) as error:
+            raise AlertCreationError(error)
```
ryandeivert: good thinking!

stream_alert/shared/alert_table.py
```diff
@@ -159,30 +161,32 @@ def mark_as_dispatched(self, alert):
...
+                ':attempts': alert.attempts,
+                ':dispatched': alert.dispatched.strftime(Alert.DATETIME_FORMAT)
+            },
             ConditionExpression='attribute_exists(AlertID)'
         )
 
     @ignore_conditional_failure
-    def update_retry_outputs(self, alert):
-        """Update the table with a new set of outputs to be retried.
+    def update_sent_outputs(self, alert):
```
ryandeivert: I like this new logic :) 👏

tests/scripts/convert_test_events.py
```diff
@@ -1,21 +1,18 @@
...
+"""One-off script to convert all rule test events from the legacy format to the new test format."""
 import json
 import os
 
 TEST_FILES_PATH = 'tests/integration/rules'
 
 
-def convert_test_event(file):
-    with open(file, 'r+') as test_event:
+def convert_test_event(path):
```
ryandeivert: thanks for changing this - my ide would flag this as well

stream_alert/shared/alert_table.py
```diff
@@ -100,38 +100,40 @@ def pending_alerts(self, rule_name, alert_proc_timeout_sec):
                 This is used to determine whether an alert could still be in progress
 
         Yields:
-            Alert: An alert instance for each row in the Dynamo table
+            dict: Each row in the Dynamo table which is not being worked on by the alert processor.
         """
+        time_to_retry = datetime.utcnow() - timedelta(seconds=alert_proc_timeout_sec)
```
ryandeivert: as we discussed in person, it might be helpful to rename this var since I was confused by it - your call!

stream_alert/alert_merger/main.py
```diff
@@ -42,30 +65,120 @@ def __init__(self):
         self.alert_proc_timeout = int(os.environ['ALERT_PROCESSOR_TIMEOUT_SEC'])
         self.lambda_client = boto3.client('lambda')
 
+    def _sorted_alerts(self, rule_name):
+        """Build a list of Alert instances sorted by creation time."""
```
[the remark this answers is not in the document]
austinbyers: Good catch - the sorting used to happen here, but now it happens downstream (in `_merge_groups`). I'll rename it

stream_alert/alert_merger/main.py
austinbyers: 🍦

stream_alert/shared/alert.py
austinbyers: Good idea! Unfortunately, `time.gmtime(0)` returns a `time.struct_time`, which isn't compatible with `timedelta`. So it would have to be explicitly converted to a `datetime` object, like so:

```python
dispatched = datetime.fromtimestamp(time.mktime(time.gmtime(0)))
```

My thought was it was just easier to keep everything as `datetime` and `timedelta` (both members of the `datetime` class)

stream_alert/shared/alert.py
austinbyers: Thanks. I wasn't sure if this was overkill or not, but I kept catching the same tuple of errors and figured it might be nice if it had a name

stream_alert/shared/alert_table.py
austinbyers: Yeah, me too. It's more explicit

stream_alert/shared/alert_table.py
austinbyers: Yeah, I'll rename to `in_progress_threshold` and add some more comments

austinbyers:
I tested end-to-end again after addressing feedback and updating with latest changes to master. All clear!

PR #690, opened 2018-04-16
adding support for specifying rule import paths

ryandeivert:
to: @austinbyers 
size: medium

## Background

Currently based largely on changes in #689, this enables specifying arbitrary paths from within which to import rule and matcher files.

## Changes

* Adding `rule_locations` and `matcher_locations` options to global config (see [Up for Disucssion](#up-for-discussion) below for more info).
* Moving rule import logic to the `stream_alert.shared.rule` module so it is decoupled from the lambda handler code.
* Calling the import code from within the new `RulesEngine` class, which will load the various paths in the config.
* Updating `stream_alert_cli/test.py` to remove the formerly required `import stream_alert.rule_processor.main` line to force rule loading out of band.
* Updating message in a `NameError` that was improperly formatting output.

## Up for Discussion

* Naming convention for the `rule_locations` - I've currently put this under a `general` category in the global.json config but I'm not 100% in love with that. Any ideas?
* We could now support 'global' rules as well as having rules specific to clusters (in the cluster's rule processor config). This could actually result in a slight optimization. There are surely clusters we have now that use 'matchers' to restrict on environment when we could just restrict what rules are loaded for a given environment.

## Testing

Migrating rule loading unit tests to the `stream_alert_shared/test_rule.py` tests file.
Removing parts of TestRulesEngine that were not actually doing anything.

### DIFF
--- a/conf/global.json
+++ b/conf/global.json
@@ -5,6 +5,14 @@
     "prefix": "PREFIX_GOES_HERE",
     "region": "us-east-1"
   },
+  "general": {
+    "rule_locations": [
+      "rules"
+    ],
+    "matcher_locations": [
+      "matchers"
+    ]
+  },
   "infrastructure": {
     "alerts_table": {
       "read_capacity": 5,
--- a/stream_alert/rule_processor/handler.py
+++ b/stream_alert/rule_processor/handler.py
@@ -57,9 +57,12 @@ class StreamAlert(object):
         self._processed_size = 0
         self._alerts = []
 
+        rule_import_paths = [item for location in {'rule_locations', 'matcher_locations'}
+                             for item in self.config['global']['general'][location]]
+
         # Create an instance of the StreamRules class that gets cached in the
         # StreamAlert class as an instance property
-        self._rules_engine = RulesEngine(self.config)
+        self._rules_engine = RulesEngine(self.config, *rule_import_paths)
 
         # Firehose client attribute
         self._firehose_client = None
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -21,7 +21,7 @@ from stream_alert.rule_processor import LOGGER
 from stream_alert.rule_processor.threat_intel import StreamThreatIntel
 from stream_alert.shared import NORMALIZATION_KEY, resources
 from stream_alert.shared.alert import Alert
-from stream_alert.shared.rule import Rule
+from stream_alert.shared.rule import import_folders, Rule
 
 
 _IGNORE_KEYS = {StreamThreatIntel.IOC_KEY, NORMALIZATION_KEY}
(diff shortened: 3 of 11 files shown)


stream_alert/rule_processor/handler.py
```diff
@@ -57,9 +57,12 @@ def __init__(self, context):
         self._processed_size = 0
         self._alerts = []
 
+        rule_import_paths = [item for location in {'rule_locations', 'matcher_locations'}
+                             for item in self.config['global']['general'][location]]
+
         # Create an instance of the StreamRules class that gets cached in the
         # StreamAlert class as an instance property
-        self._rules_engine = RulesEngine(self.config)
+        self._rules_engine = RulesEngine(self.config, *rule_import_paths)
```
austinbyers: Is there any advantage to expanding the paths as arguments instead of leaving them as a list? I have no strong preference either way, but I'm not sure why we do this here and not elsewhere

conf/global.json
```diff
@@ -5,6 +5,14 @@
     "prefix": "PREFIX_GOES_HERE",
     "region": "us-east-1"
   },
+  "general": {
```
austinbyers: I'm fine with `general`. The only other thing I can think of is `file_locations`

stream_alert/rule_processor/rules_engine.py
```diff
@@ -21,18 +21,19 @@
...
 
 
 class RulesEngine(object):
     """Class to act as a rules engine that processes rules"""
-    def __init__(self, config):
+    def __init__(self, config, *rule_paths):
         """Initialize a RulesEngine instance to cache a StreamThreatIntel instance."""
         self._threat_intel = StreamThreatIntel.load_from_config(config)
         self._required_outputs_set = resources.get_required_outputs()
+        import_folders(*rule_paths)
```
austinbyers: Since the `config` is passed in here and rule paths are computed directly from the config, do you think it should be the job of the `RulesEngine` to figure out the rule paths rather than the caller's responsibility?

stream_alert/rule_processor/handler.py
ryandeivert: we don't leverage arbitrary len arguments very often (if at all?) that I'm aware of which might be why it looks odd :)

stream_alert/rule_processor/rules_engine.py
ryandeivert: I had it this way before but changed it to this. I did so because I don't love that we pass a config object around in general and would rather stick to using primitive types like a basic string/list of strings.

I say this because there are places where we're passing around a whole config dict to end up only extracting/using one or two things from it.. and now for anything else to use that class properly, they'd have to construct the whole/valid config object to be passed to it. Thoughts on this? Happy to change back to the previous version.

austinbyers: Hmm I see your point. In general, I'm happy with either, but maybe if we ever do a config rewrite we can work on better consistency

ryandeivert: ++ for config rewrite +++++++

PR #691, opened 2018-04-16
[shared] consolidate backoff handler code into one place

ryandeivert:
to: @jacknagz 
size: medium

## Background

We currently have two implementation of various `backoff` handlers throughout the code. One of these exists in the `athena_partition_refresh` function, and one exists in the `stream_alert.shared` module. One is implemented (mostly) with `debug` level logging statements, while the other is implemented with `info` statements. This consolidates the need for two different handler types.

## Changes

* Updating the `backoff` pinned version to be `1.5.0` as this version supports the helpful `elapsed` attribute in the details that tracks how much time has elapsed since backing off. Prior to this, there was only `tries` and `wait` to have an idea of how much backoff has occurred.
* Updating the `backoff_handlers` in the `shared` module as wrappers that take an optional argument of `debug_only`. This bool arg dictates how a caller should do logging (info vs debug).
  * These updated handlers will now log `elapsed` time for both `on_giveup` and `on_success` handlers.
  * This allows us to have the `on_giveup` and `on_success` handlers default to `info` output so we always how long a backoff operation took to either succeed or quit. The `on_backoff` handler will now default to the `debug` logging level, as it will be less important to know _how much_ backoff occurred each try vs how much backoff occurred overall.
* Removing the `backoff_handlers` from the `athena_partition_refresh.main` module and importing these new functions from the shared module.
* Updating all usage of the shared backoff handlers throughout the code base (alert processor, rule processor, etc)

## Testing

Removed backoff handler tests from the athena main tests file and added new tests for both log levels in `tests/unit/stream_alert_shared/test_backoff_handlers.py`


stream_alert/shared/backoff_handlers.py
```diff
@@ -17,41 +17,64 @@
 from stream_alert.shared import LOGGER
 
 
-def backoff_handler(details):
+def backoff_handler(debug_only=True):
```
jacknagz: can we rename to just `debug`? I was a little confused what the purpose of `only` is in that var name since it just does one or the other

ryandeivert: I chose `debug_only` here because `debug` seems misleading to me... for instance, if I call a function:

`run_this(debug=True)`

That gives me the impression that *_extra logging_* will occur (like a verbose mode), when in fact thats not at all what happens.

So `debug_only` seemed more clear to me in that `run_this(debug_only=True)` implies it would only have an affect if debug mode is enabled.

Does that make sense?

jacknagz: makes sense

PR #696, opened 2018-04-20
support for adding rule information to dynamo table

ryandeivert:
to: @austinbyers 
size: large

## Background

In order to support rule staging, we need to implement a 'front end' for the rules DynamoDB table. This is the first iteration on a RuleTable performing that interaction.

### RuleTable
Adding initial `RuleTable` that currently handles operations such as:
* Loading the state of all rules currently in the table
* Diff'ing the existing remote rules against the local rule set.
* Updating an empty table with rules (presumably adding rules for the first time). This operation will have rules bypass the 'staging' state and go straight to 'production'
* Updating a table that has previously existing entries with new rules. This operation will have rules go into the staging state for a default window of 48 hours (for now).
* Printing a human-readable format of the remote rule state. Optional `verbose` flag can be passed to print both staging status and all staging information.

### CLI
The cli now supports loading the `RuleTable` and printing the status of all rules within it.
* Future support will come for dynamic toggling of staging status in the RuleTable from the CLI

Example (non-verbose):
```
Rule                                                        Staged?
  1: binaryalert_yara_match                                 False
  2: bogus_new_rule                                         True
  3: cloudtrail_critical_api_calls                          False
  4: cloudtrail_mfa_policy_abuse_attempt                    False
  5: cloudtrail_network_acl_ingress_anywhere                False
```

Example (verbose):
```
Rule                                                        Staged?
  1: binaryalert_yara_match                                 False
  2: bogus_new_rule                                         True
     - StagedAt:      2018-04-19T02:23:13.332223Z
     - NewlyStaged:   True
     - StagedUntil:   2018-04-21T02:23:13.332223Z
  3: cloudtrail_critical_api_calls                          False
  4: cloudtrail_mfa_policy_abuse_attempt                    False
  5: cloudtrail_network_acl_ingress_anywhere                False
```

### Other Notes
Upon deploy, these changes will not have an effect as far as the rule processor is concerned.
A config flag is currently set that disables updating the RuleTable with rule information. This will likely be removed once this is feature complete.

## Testing

Adding 100% test coverage for the `RuleTable`. Tested locally with a local DynamoDB instance.
Will also test with an actual test account deploy before merging.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1625,6 +1625,63 @@ Examples:
         '--debug', action='store_true', help=ARGPARSE_SUPPRESS
     )
 
+def _add_rule_table_subparser(subparsers):
+    """Add the rule database helper subparser: manage.py rule-table [subcommand] [options]"""
+    rule_table_usage = 'manage.py rule-table [subcommand] [options]'
+    rule_table_description = ("""
+StreamAlertCLI v{}
+Print the status of or update remote StreamAlert rule information within the rule database
+
+Available Subcommands:
+
+    manage.py rule-table status          List the current staging status from the rules databse
+    manage.py rule-table update          Update the staging status of rules
+
+""".format(version))
+    rule_table_parser = subparsers.add_parser(
+        'rule-table',
+        usage=rule_table_usage,
+        description=rule_table_description,
+        formatter_class=RawTextHelpFormatter,
+        help=ARGPARSE_SUPPRESS)
+
+    # Set the name of this parser to 'app'
+    rule_table_parser.set_defaults(command='rule-table')
+
+    rule_table_subparsers = rule_table_parser.add_subparsers()
+
+    _add_rule_table_status_subparser(rule_table_subparsers)
+
+def _add_rule_table_status_subparser(subparsers):
+    """Add the rule db status subparser: manage.py rule-table status"""
+    rule_table_list_usage = 'manage.py rule-table status'
+    rule_table_list_desc = ("""
+StreamAlertCLI v{}
+List all rules within the rule database and their staging status
+
+Command:
+
+    manage.py rule-table status       List rules in the rule database with their staging status
+
+Optional Arguments:
+
+    --verbose           Output additional information for rules in the database
+    --debug             Enable Debug logger output
+
+""".format(version))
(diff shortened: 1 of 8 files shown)


manage.py
```diff
@@ -1625,6 +1625,63 @@ def _validate_memory(val):
...
+
+""".format(version))
+    rule_table_parser = subparsers.add_parser(
+        'rule-table',
+        usage=rule_table_usage,
+        description=rule_table_description,
+        formatter_class=RawTextHelpFormatter,
+        help=ARGPARSE_SUPPRESS)
+
+    # Set the name of this parser to 'app'
```
austinbyers: Copy-pasta comment?

stream_alert/shared/rule_table.py
```diff
@@ -0,0 +1,200 @@
...
+        paginator = self._table.meta.client.get_paginator('scan')
+        page_iterator = paginator.paginate(TableName=self.name, ConsistentRead=True)
+        return {
+            item['RuleName']: {
+                key: value for key, value in item.iteritems()
+                if key != 'RuleName'
+            }
+            for page in page_iterator
+            for item in page['Items']
+        }
```
austinbyers: Nice! I'll adopt this for the AlertTable class as well

tests/unit/stream_alert_shared/test_rule_table.py
```diff
@@ -0,0 +1,281 @@
...
+        self.dynamo_mock.stop()
+
+    @classmethod
+    def _create_local_rules(cls, count=5):
+        """Helper to create N fake local rules"""
+        for i in range(count):
+            cls._create_local_rule_with_name('fake_rule_{:02d}'.format(i))
+
+    @classmethod
+    def _create_local_rule_with_name(cls, name):
```
austinbyers: Could this be `@staticmethod` instead of `@classmethod`?

manage.py
ryandeivert: yep good catch :)

tests/unit/stream_alert_shared/test_rule_table.py
ryandeivert: it could probably yeah - will make the change

PR #697, opened 2018-04-20
fix to only update rule table if rule proc is deployed

ryandeivert:
to: @chunyong-lin 
size: small

## Background

The rule table should only really be updated if the rule processor is being deployed (not, say, if only the alert processor is deployed).

## Changes

* Ensuring rule processor is being deployed before updating rules tables.


ryandeivert:
that gif is incredible

PR #700, opened 2018-04-23
exclude specified threat intel from dynamo lookups

strcrzy:
to: @chunyong-lin 
size: medium

## Background
Previously we submit all non-RFC1918 ip addresses for lookup in the dynamo table. 
This change adds an explicit config exclusion list that is checked against before batching for a dynamo query. All IOCs that we wish to bypass lookups for will go here.

## Changes

*  fetch `excluded_iocs` dict (`ip`,`domain`,`md5` subkeys)  from the global `threat_intel` config subsection
* new function [`is_excluded_ioc`](https://github.com/airbnb/streamalert/blob/jamie/threat_intel_exclude/stream_alert/rule_processor/threat_intel.py#L455-L476) to report if an IOC has been excluded
* replace the [existing is_public_ip](https://github.com/airbnb/streamalert/blob/0cb6f865affe1f1aec4f3bea3d2b89dc5ff5ebd1/stream_alert/rule_processor/threat_intel.py#L169) call to the new  [`is_excluded_ioc`](https://github.com/airbnb/streamalert/blob/jamie/threat_intel_exclude/stream_alert/rule_processor/threat_intel.py#L171)

## Testing
Testing is good. Right now we fail a test. I want to add more coverage and fix the failed test, will work with @chunyong-lin if she has a moment


jacknagz:
hey @strcrzy looks like there's some CI failures to look into

strcrzy:
yeah, looking into that with @chunyong-lin -- the [old](https://github.com/airbnb/streamalert/blob/0cb6f865affe1f1aec4f3bea3d2b89dc5ff5ebd1/stream_alert/rule_processor/threat_intel.py#L195) version actually gated off the attempts to access `config['global']` with an `if` check -- my revision [does not](https://github.com/airbnb/streamalert/pull/700/files#diff-17fc33610d6a4cad02cc7adfde6fb9c5R196) (it makes the assumption that a `config['global']` exists, which i thought was reasonable). i think somehow the plumbing of the test config to this test is not quite working, will investigate with chunyong

strcrzy:
beffae8 should cover the CI issues -- tests also caught a mistake on my part, fixed in a6e0119

stream_alert/rule_processor/threat_intel.py
```diff
@@ -411,11 +414,63 @@ def normalized_type_mapping(cls):
...
+            False if IOC should be looked up
+        """
+        if ioc_type == 'ip':
+            try:
+                ip_addr = IPAddress(str(ioc_value))
+                excluded_networks = StreamThreatIntel.excluded_iocs.get("ip", set())
+                return StreamThreatIntel.in_network(ip_addr, excluded_networks)
+            except: # pylint: disable=bare-except
+                LOGGER.error('IP IOC Exclusion Failed: %s', ioc_value)
+                return True
```
chunyong-lin: We also need to handle `AddrFormatError` exception. In `cloudtrail_api` events, many of `source ipaddress` are `s3.amazonaws.com`, `ecs.amazonaws.com`, `cloudtrail.amazonaws.com`, etc. If we add such unit test cases will be super helpful.

stream_alert/rule_processor/threat_intel.py
```diff
@@ -411,11 +414,63 @@ def normalized_type_mapping(cls):
...
+        else:
+            excluded = StreamThreatIntel.excluded_iocs.get(ioc_type, set())
```
chunyong-lin: Change this method to instance method, so can use `self.excluded_iocs`

stream_alert/rule_processor/threat_intel.py
strcrzy: i believe this is already tested in `test_extract_ioc_from_record_with_private_ip` from #557 
i can extract it into its own test though if you think it should be its own test

tests/unit/stream_alert_rule_processor/test_threat_intel.py
```diff
@@ -298,6 +298,37 @@ def test_extract_ioc_from_record_with_private_ip(self):
...
+                'region': '123456123456',
+                'detail': {
+                    'eventType': 'AwsConsoleSignIn',
+                    'eventName': 'ConsoleLogin',
+                    'userIdentity': {
+                        'userName': 'alice',
+                        'type': 'Root',
+                        'principalId': '12345',
+                    },
+                    'sourceIPAddress': '8.8.8.8',
```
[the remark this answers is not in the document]
strcrzy: done: https://github.com/airbnb/streamalert/blob/932089eb23cb61b7a50e3b6a137c3eb1646cf79f/tests/unit/stream_alert_rule_processor/test_threat_intel.py#L274-L294

strcrzy:
Thanks for the super helpful review, @chunyong-lin 
I've incorporated your feedback, could you PTAL when you have a moment?

stream_alert/rule_processor/threat_intel.py
```diff
@@ -19,17 +19,16 @@
...
 from stream_alert.shared import NORMALIZATION_KEY
 from stream_alert.shared.backoff_handlers import (
     backoff_handler,
     success_handler,
     giveup_handler
 )
-
 from stream_alert.rule_processor import LOGGER
-
+from helpers.base import in_network
```
ryandeivert: I know @chunyong-lin suggested this but I actually would not do this. These helpers were intended for use with `rules` and not core code.. If this is a requirement within threatintel, I'd suggest reimplementing a helper in this class. we can talk more since you've been pulled in multiple directions on this

ryandeivert: alternatively we could put some helper in `stream_alert/shared/..` that could be imported

stream_alert/rule_processor/threat_intel.py
```diff
@@ -164,14 +164,20 @@ def _extract_ioc_from_record(self, record):
...
+                        # Threat Intel will skip searching for known excluded IOCs
+                        if self.is_excluded_ioc(ioc_type, value):
                             continue
                         ioc_value_type_tuples.add((value, ioc_type))
         return [StreamIoc(value=str(value).lower(), ioc_type=ioc_type, associated_record=record)
                 for value, ioc_type in ioc_value_type_tuples]
 
+    def _setup_excluded_iocs(self, config_excluded_iocs=None):
+        if config_excluded_iocs:
+            self.excluded_iocs = config_excluded_iocs
```
ryandeivert: This function sets the `excluded_iocs` instance property from a call to `_setup_excluded_iocs` within the `__init__`. This can lead to confusion, since the property is not defined within the `__init__` itself (and could result in users/maintainers not realizing the property exists.

What I would suggest is have this function _return_ the `excluded_iocs` that would then get set on `self` within the init.. for example:


```
def __init__(self, excluded_iocs, table, region='us-east-1'):
    self.excluded_iocs = self._setup_excluded_iocs(excluded_iocs)

...

def _setup_excluded_iocs(self, config_excluded_iocs=None):
    if not config_excluded_iocs:
        return {'ip': set(), 'domain': set(), 'md5': set()}

    # Transform the IPs into IPNetwork objects
    config_excluded_iocs['ip'] = {IPNetwork(ip) for ip in config_excluded_iocs.get('ip', [])}
    return config_excluded_iocs
```

This way all instance properties are defined within `__init__` as opposed to within the _scope_ of `__init__`

stream_alert/rule_processor/threat_intel.py
```diff
@@ -164,14 +164,20 @@ def _extract_ioc_from_record(self, record):
...
+        else:
+            self.excluded_iocs = {"ip": set(), "domain": set(), "md5": set()}
+        self.excluded_iocs["ip"] = {IPNetwork(ip) for ip in self.excluded_iocs["ip"]}
```
ryandeivert: Please stick to single quotes (`'`) vs double quotes (`"`) for string literals to be consistent with the rest of the codebase

stream_alert/rule_processor/threat_intel.py
```diff
@@ -411,11 +416,42 @@ def normalized_type_mapping(cls):
...
+            True if IOC lookup should be bypassed for this value
+            False if IOC should be looked up
+        """
+        if ioc_type == 'ip':
+            try:
+                ip_addr = IPAddress(str(ioc_value))
+                excluded_networks = self.excluded_iocs.get('ip', set())
+                return in_network(ip_addr, excluded_networks)
+            except: # pylint: disable=bare-except
+                LOGGER.error('IP IOC Exclusion Failed: %s', ioc_value)
```
ryandeivert: Please use `LOGGER.exception(...)` here so we know what the error was if we need to troubleshoot

stream_alert/rule_processor/threat_intel.py
```diff
@@ -79,7 +78,8 @@ class StreamThreatIntel(object):
     # Class variable stores mapping between CEF normalized types and IOC types
     __normalized_ioc_types_mapping = {}
 
-    def __init__(self, table, region='us-east-1'):
+    def __init__(self, excluded_iocs, table, region='us-east-1'):
```
ryandeivert: since `excluded_iocs` could presumably be `None` or an empty dict `{}` (ie optional?) here, my preference would be to have this be the second arg (after `table`) or introduce `kwargs` to this initializer

stream_alert/rule_processor/threat_intel.py
```diff
@@ -411,11 +416,42 @@ def normalized_type_mapping(cls):
...
+        """
+        if ioc_type == 'ip':
+            try:
+                ip_addr = IPAddress(str(ioc_value))
+                excluded_networks = self.excluded_iocs.get('ip', set())
+                return in_network(ip_addr, excluded_networks)
+            except: # pylint: disable=bare-except
+                LOGGER.error('IP IOC Exclusion Failed: %s', ioc_value)
+                return True
+        else:
```
ryandeivert: Since the above `if` block always returns, this `else` block could be backed out (and the `else` left off completely). ie:
```
    def is_excluded_ioc(self, ioc_type, ioc_value):
        """
        check if we should bypass IOC lookup for specified IOC
        Args:
            ioc_type (string): the type of IOC to evaluate (md5, ip, domain)
            value (string): the value of IOC to evaluate
        Returns:
            True if IOC lookup should be bypassed for this value
            False if IOC should be looked up
        """
        if ioc_type == 'ip':
            try:
                ip_addr = IPAddress(str(ioc_value))
                excluded_networks = self.excluded_iocs.get('ip', set())
                return in_network(ip_addr, excluded_networks)
            except: # pylint: disable=bare-except
                LOGGER.error('IP IOC Exclusion Failed: %s', ioc_value)
                return True

        return ioc_value in self.excluded_iocs.get(ioc_type, set())
```

stream_alert/rule_processor/threat_intel.py
strcrzy: this one's trickier, most of the failures are from when the source ip is not in fact an ip (it's an amazon domain). i think this might cause a lot of exceptions, though i could write code to short circuit it out before it reaches this stage

stream_alert/rule_processor/threat_intel.py
strcrzy: nice, thanks!

stream_alert/rule_processor/threat_intel.py
strcrzy: i think i might go with kwargs here -- it felt weird placing the excluded IOCs sandwiched between the two dynamo options, and I couldn't place it at the end because of the default value for region present in the args

ryandeivert: yep - I'd go with kwargs in this case too :). theoretically you could add the `region` as a kwarg here to since that is only necessary in some circumstance (for example - it's only really need for local testing, not when deployed in aws lambda)

stream_alert/rule_processor/threat_intel.py
strcrzy: yeah, i think it would be noisy unless we dump the domain "ips" earlier

stream_alert/rule_processor/threat_intel.py
```diff
@@ -191,13 +198,12 @@ def load_from_config(cls, config):
...
-        if (config.get('global')
-                and config['global'].get('threat_intel')
-                and config['global']['threat_intel'].get('enabled')
-                and config['global']['threat_intel'].get('dynamodb_table')):
-            return cls(config['global']['threat_intel']['dynamodb_table'],
-                       config['global']['account'].get('region', 'us-east-1'))
+        threat_intel_config = config.get('global').get('threat_intel')
+        if threat_intel_config:
+            if threat_intel_config.get('enabled') and threat_intel_config.get('dynamodb_table'):
+                return cls(excluded_iocs=threat_intel_config['excluded_iocs'],
```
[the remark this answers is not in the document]
strcrzy: i tried a different approach here, PTAL

tests/unit/stream_alert_rule_processor/test_handler.py
```diff
@@ -279,7 +279,9 @@ def match_ipaddress(_): # pylint: disable=unused-variable
             """Testing dummy rule"""
             return True
 
-        mock_threat_intel.return_value = StreamThreatIntel('test_table_name', 'us-east-1')
+        mock_threat_intel.return_value = StreamThreatIntel(excluded_iocs={},
```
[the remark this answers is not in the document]
strcrzy: done: https://github.com/airbnb/streamalert/blob/jamie/threat_intel_exclude/tests/unit/stream_alert_rule_processor/test_handler.py#L317-L353

strcrzy:
@ryandeivert @chunyong-lin i've incorporated your feedback, when you have a moment can you PTAL one more time

thanks!

PR #703, opened 2018-04-24
[cli] adding stage subcommand to rule-table command

ryandeivert:
to: @chunyong-lin 
size: small

## Background

Users need the ability to run a command to stage or unstage a current rule in the rules db.

## Changes

* Adding a subcommand to support staging or unstaging of rules.
* Staging will work like so:
`$ python manage.py rule-table stage foo_rule bar_rule`
* Un-staging will work like so:
`$ python manage.py rule-table unstage foo_rule bar_rule`

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1660,6 +1660,7 @@ Available Subcommands:
     rule_table_subparsers = rule_table_parser.add_subparsers()
 
     _add_rule_table_status_subparser(rule_table_subparsers)
+    _add_rule_table_stage_subparser(rule_table_subparsers)
 
 def _add_rule_table_status_subparser(subparsers):
     """Add the rule db status subparser: manage.py rule-table status"""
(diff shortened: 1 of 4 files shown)


manage.py
```diff
@@ -1692,6 +1693,51 @@ def _add_rule_table_status_subparser(subparsers):
...
+    rule_table_stage_parser.add_argument(
+        'rules',
+        action=UniqueSetAction,
+        default=set(),
+        help=ARGPARSE_SUPPRESS,
+        nargs='+'
+    )
+
+    rule_table_stage_parser.add_argument(
+        '-u', '--unstage',
```
chunyong-lin: I am little hesitating to comment on this line. Because from users point, it is little confusing to unstage the rules after `stage` keyword
```
manage.py rule-table stage --unstage rule1 rule2
```

But if we change it to be a user friendly version, we will duplicate this subparser to support command like this
```
manage.py rule-table unstage rule1 rule2
```
What do you think?

ryandeivert:
@chunyong-lin done! PTAL

PR #707, opened 2018-04-26
config consolidation/removal of tech debt

ryandeivert:
to: @austinbyers 
size: large

## Background

We currently have JSON config loading logic in various places, each implemented for specific use-cases (for the most part). Some of these are untested but most are largely redundant. These changes are part of an effort to consolidate and simplify how we handle config loading.

## Changes

* Adding a new function (named simply `load_config`) to a new `stream_alert.shared.config` module.
* This function is flexible and can handle all of the following:
  * Loading the entire config (default)
  * Loading certain files only (through the use of an `include` keyword argument)
  * Loading the entire config _minus_ specific items (through the use of an `exclude` keyword argument)
  * Optionally 'validating' certain aspects of the config via a `validate` keyword argument.
* Updating all current config loading logic (that I was able to find) to use this new function.
  * Threat Intel Downloader function
  * Athena Partition Refresh function
  * Rule Processor
  * Alert Processor
  * CLIConfig
* The `CLIConfig` class has been refactored slightly, since it can use the new `load_config` function for loading instead of duplicating this effort internally.
* Updating some code related to configuring outputs that was doing its own config magic instead of leveraging code that currently exists.
* Standardizing lambda function ARN parsing to use the name `qualifier` instead of `lambda_alias`. 
  
### Future Work

* Ensure all places where lambda function arns are parsed use the new `parse_lambda_arn` function in the `stream_alert.shared.config` module.
* Move our config to be class-based, in which `CLIConfig` would inherit from in order to implement the 'write' functionality for writing the config back to disk.

## Testing

Updating all unit tests to utilize to the new `load_config` function and adding new unit tests for the `stream_alert.shared.config` module.


stream_alert_cli/config.py
```diff
@@ -31,12 +30,11 @@ class CLIConfigError(Exception):
...
-    DEFAULT_CONFIG_PATH = 'conf'
+    DEFAULT_CONFIG_PATH = 'conf/'
 
-    def __init__(self, **kwargs):
-        self.config_path = kwargs.get('config_path', self.DEFAULT_CONFIG_PATH)
-        self.config = {'clusters': {}}
-        self.load()
+    def __init__(self, config_path=DEFAULT_CONFIG_PATH):
+        self.config_path = config_path
+        self.config = config.load_config(config_path)
```
jacknagz: 👏

stream_alert/shared/config.py
```diff
@@ -0,0 +1,154 @@
...
+        include (set): Names of specific config files to only load
+        validate (bool): Validate aspects of the config to check for user error
+    """
+    default_files = {file for file in os.listdir(conf_dir) if file.endswith('.json')}
+    conf_files = kwargs.get('include', default_files).copy()
+    include_clusters = 'clusters' in conf_files
+
+    conf_files.intersection_update(default_files)
+    if not (conf_files or include_clusters):
+        raise ConfigError('No config files to load')
```
jacknagz: Could we add more context into this error?

stream_alert/shared/config.py
```diff
@@ -0,0 +1,156 @@
...
+    """Helper to return the loaded json from a given path"""
+    kwargs = {'object_pairs_hook': OrderedDict if ordered else None}
+    with open(path) as data:
+        try:
+            return json.load(data, **kwargs)
+        except ValueError:
+            raise ConfigError('Invalid JSON format for {}'.format(path))
+
+
+def _validate_config(config):
```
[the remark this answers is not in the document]
ryandeivert: The above `load_config` function allows _**anything**_ to be optional.. meaning `logs.json` or `sources.json` could be excluded by the caller using `exclude={'logs.json'}`, etc. These checks just ensure that the user did not do something like:

`load_config(exclude={'logs.json'}, validate=True)`

stream_alert/shared/config.py
```diff
@@ -0,0 +1,156 @@
...
+            cluster_path = os.path.join(conf_dir, 'clusters', cluster)
+            config['clusters'][os.path.splitext(cluster)[0]] = _load_json_file(cluster_path)
+
+    if kwargs.get('validate'):
+        _validate_config(config)
+
+    return config
+
+def _load_json_file(path, ordered=False):
+    """Helper to return the loaded json from a given path"""
```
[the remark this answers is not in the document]
ryandeivert: didn't we say no more 'nits'? 🤣

stream_alert/shared/config.py
```diff
@@ -0,0 +1,154 @@
...
+    rules to log fields.
+
+    Arguments:
+        conf_dir (str): [optional] Path from which to load the config
+
+    Keyword Arguemnts:
+        exclude (set): Names of config files or folders that should not be loaded
+        include (set): Names of specific config files to only load
+        validate (bool): Validate aspects of the config to check for user error
+    """
```
austinbyers: If there are only 3 keyword arguments, can we list them explicitly instead of using `**kwargs`? I'm nearly always a fan of explicit kwargs - it makes it easier for IDEs to autocomplete, etc

stream_alert/shared/config.py
```diff
@@ -0,0 +1,154 @@
...
+    `logs` declare the schema for the listed log types in `sources`.  Each
+    key denotes the name of the log type, and includes 'keys' used to match
+    rules to log fields.
+
+    Arguments:
+        conf_dir (str): [optional] Path from which to load the config
+
+    Keyword Arguemnts:
+        exclude (set): Names of config files or folders that should not be loaded
+        include (set): Names of specific config files to only load
```
austinbyers: Out of curiosity, what was the motivation for selectively loading parts of the config? Performance?

A potential downside to this approach is having to reason about which sections of the config are available to different Lambda functions instead of being able to reason about a single config object.

A potential future iteration of this (once we have a Config class) would be to lazy-load attributes. That way, callers don't need to explicitly say which config they need

ryandeivert: I suppose the motivation was largely to allow for supporting what we _currently_ do. For instance, the alert processor only loads the `outputs.json`, so I thought supporting a similar model was ideal. While I agree it could be confusing to have to reason about which files are loaded and which are not, the caller is the one who _restricts_ the loaded files (and thus should be responsible for knowing this..) otherwise all of them will be loaded.

After this initial iteration I'm open to building it out more (and anticipate doing so). I like the idea of lazy loading and think that we can benefit a lot from that approach!

stream_alert/shared/config.py
ryandeivert: no real reason not to - I can make the change :)

ryandeivert:
Just deployed this in test account and all is calm! merigng

PR #709, opened 2018-04-27
support comments in ioc exclude lists

strcrzy:
to: @chunyong-lin 
size: small

## Background
the IOC exclude list lives in the config json, and json does not natively support comments. 

## Changes

* support json dictionaries in `excluded_iocs` : `key : metadata`
* allow metadata, such as comments, sources, confidence, etc.. to be stored in metadata
* ~~maintain backwards compatibility with string entries for `excluded_iocs`~~
* update test fixtures to feature both types of exclusions

## Testing
unit tested


stream_alert/rule_processor/threat_intel.py
```diff
@@ -171,12 +171,15 @@ def _extract_ioc_from_record(self, record):
...
 
     @staticmethod
-    def _setup_excluded_iocs(config_excluded_iocs=None):
-        if not config_excluded_iocs:
+    def _setup_excluded_iocs(excluded=None):
+        if not excluded:
             return None
+
+        for (itype, iocs) in excluded.iteritems():
+            excluded[itype] = {ioc["indicator"] if isinstance(ioc, dict) else ioc for ioc in iocs}
```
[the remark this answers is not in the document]
strcrzy: i think i'll go with using the excluded ioc as the hash key as you recommended below. 
in that case, this line will be a little different

tests/unit/conf/global.json
```diff
@@ -30,11 +30,21 @@
     "enabled": false,
     "excluded_iocs": {
       "ip": [
-        "127.0.0.0/8",
-        "10.0.0.0/8",
-        "172.16.0.0/12",
-        "192.168.0.0/16",
-        "52.52.52.52/32"
+        {
```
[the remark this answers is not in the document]
strcrzy: even if we need to add keys in the future (one of the considerations i had when expanding this into a hash), we could have the value of the hash be another hash with metadata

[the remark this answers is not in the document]
ryandeivert: 👍 to this! an added benefit of being a dict is that we can do keyed lookups of the indicator now. previously, this would check for a value in a _list_ which is significantly slower if they list is large

stream_alert/rule_processor/threat_intel.py
```diff
@@ -171,12 +171,15 @@ def _extract_ioc_from_record(self, record):
                 for value, ioc_type in ioc_value_type_tuples]
 
     @staticmethod
-    def _setup_excluded_iocs(config_excluded_iocs=None):
-        if not config_excluded_iocs:
+    def _setup_excluded_iocs(excluded=None):
+        if not excluded:
             return None
+
+        for (itype, iocs) in excluded.iteritems():
```
ryandeivert: the parenths around this is not necessary

tests/unit/conf/global.json
```diff
@@ -29,13 +29,33 @@
...
     "excluded_iocs": {
-      "ip": [
-        "127.0.0.0/8",
-        "10.0.0.0/8",
-        "172.16.0.0/12",
-        "192.168.0.0/16",
-        "52.52.52.52/32"
-      ]
+      "domain": {
+        "not-evil.com": {
```
chunyong-lin: Can we just use
```
"domain": {
  "not-evil.com": "This domain is not evil"
}
```

We can discuss more if we have plan to add more keys other than `comment`.

strcrzy: this is possible, and more readable/writable, i guess the discussion is a good thing to have regarding whether or not we will associate more metadata than just the comment

mime-frame: I like the flexibility the current version affords us; ill bet we'll want to add more metadata in the future

chunyong-lin: Make sense then. Thanks for the clarification. 😸

PR #710, opened 2018-04-28
rule processor - adding check for staged rules

ryandeivert:
to: @austinbyers 
size: medium

## Background

Now that there is a rule table, the rule processor needs the ability to check the table for information on whether or not a rule is staged, etc.

## Changes

* Making a temporary update to the CLI so the default for deploy does NOT put new rules into staging.
  * Note: this will be reverted when the rule promotion lambda is in place & live.
* Fixing bug in firehose region lookup in the env.
* Adding a `_RULE_TABLE` class property to the `RulesEngine` that is cached throughout executions. It will be refreshed (aka recreated) if the configurable `refresh_minutes` have passed since the last refresh.
* Fixing stutter in `Rule.rule_name` attribute naming (renamed to `Rule.name`)
* Adding `is_staged` method to `Rule` class to allow rules to check if they are staged, given a `RuleTable` as a parameter.
* Updating the rules engine to only use required outputs if rule is staged.

## Testing

Adding unit tests for all new code and adjust old ones as necessary.

### DIFF
--- a/conf/global.json
+++ b/conf/global.json
@@ -30,7 +30,7 @@
     },
     "rules_table": {
       "enabled": false,
-      "refresh_minutes": 10,
+      "cache_refresh_minutes": 10,
       "read_capacity": 20,
       "write_capacity": 5
     }
--- a/conf/global.json
+++ b/conf/global.json
@@ -30,6 +30,7 @@
     },
     "rules_table": {
       "enabled": false,
+      "refresh_minutes": 10,
       "read_capacity": 20,
       "write_capacity": 5
     }
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -170,7 +170,7 @@ class RulesEngine(object):
             if not record.get(key):
                 LOGGER.debug(
                     'The required subkey %s is not found when trying to process %s: \n%s',
-                    key, rule.rule_name, record)
+                    key, rule.name, record)
                 return False
             if not all(record[key].get(x) for x in nested_keys):
                 return False
@@ -350,7 +350,7 @@ class RulesEngine(object):
         # Combine the required alert outputs with the ones for this rule
         all_outputs = self._required_outputs_set.union(rule.outputs_set)
         alert = Alert(
-            rule.rule_name, record, all_outputs,
+            rule.name, record, all_outputs,
             cluster=os.environ['CLUSTER'],
             context=rule.context,
             log_source=str(payload.log_source),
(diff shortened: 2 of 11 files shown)


conf/global.json
```diff
@@ -30,6 +30,7 @@
     },
     "rules_table": {
       "enabled": false,
+      "refresh_minutes": 10,
```
austinbyers: This makes me think this is a refresh of the table itself. But this refers to the rule processor refreshing its in-memory cache, right? If so, maybe `cache_refresh_minutes` would be clearer?

stream_alert/rule_processor/rules_engine.py
```diff
@@ -14,26 +14,66 @@
...
+        if not rt_config.get('enabled', False):
+            return
+
+        now = datetime.utcnow()
+        refresh_delta = timedelta(minutes=rt_config.get('refresh_minutes', 10))
+
+        # The rule table will need 'refreshed' if 1) the table or last refresh time
+        # do not exist, or 2) if the refresh interval has been surpassed
+        needs_refresh = (not (cls._RULE_TABLE and cls._RULE_TABLE_LAST_REFRESH) or
+                         cls._RULE_TABLE_LAST_REFRESH + refresh_delta < now)
```
austinbyers: If `cls._RULE_TABLE_LAST_REFRESH` is initialized to some date in the past (UNIX epoch, for example) instead of `None`, then I don't think you need this double conditional

stream_alert/rule_processor/rules_engine.py
```diff
@@ -170,7 +210,7 @@ def process_subkeys(record, payload_type, rule):
             if not record.get(key):
                 LOGGER.debug(
                     'The required subkey %s is not found when trying to process %s: \n%s',
-                    key, rule.rule_name, record)
+                    key, rule.name, record)
```
austinbyers: ++ 👍 Much cleaner 👍 ++

conf/global.json
ryandeivert: I like it! will make the change

stream_alert/rule_processor/rules_engine.py
ryandeivert: hey @austinbyers can you elaborate/show an example of what you mean? I'm not sure I'm understanding 🤔 are you suggesting that the class prop is instantiated with some default datetime value? ie here: 

https://github.com/airbnb/streamalert/pull/710/files/ebf8e8de667a5a7448cad6773fe3e4c898c42863#diff-30cc622fccfff0217d84af2ff5bcd7ffR33

austinbyers: Yes, exactly. For example, the `Alert` class sets the last dispatched time to the epoch:

https://github.com/airbnb/streamalert/blob/89c5b63115f3af23681e109c3d8260cb46239e1d/stream_alert/shared/alert.py#L94-L95

So that we can use a single condition to pull alerts from a given time window:

https://github.com/airbnb/streamalert/blob/89c5b63115f3af23681e109c3d8260cb46239e1d/stream_alert/shared/alert_table.py#L114-L117

ryandeivert: 👍 awesome thank you!

PR #711, opened 2018-04-30
updating athena helpers and adding some tests

ryandeivert:
to: @austinbyers 
size: medium

## Changes

* Migrating some athena helpers to the CLI where they are used, instead of within the athena lambda function package.
* Updating athena helper function name from `to_athena_schema` to `logs_schema_to_athena_schema` to be more clear.
* Improving `_construct_create_table_statement` logic.

## Testing

Updating tests, adding test for 'create table' statement construction code, adding test for 'add partition' statement construction code.


stream_alert_cli/athena/handler.py
```diff
@@ -157,26 +156,22 @@ def _construct_create_table_statement(schema, table_name, bucket):
...
+                '{0}:{1}'.format(sub_key, key_type[sub_key])
+                for sub_key in sorted(key_type.keys())
+            )
+            schema_statement.append('{0} struct<{1}>'.format(key_name, struct_schema))
 
     return CREATE_TABLE_STATEMENT.format(
         table_name=table_name,
-        # Remove the last comma from the generated statement
-        schema=schema_statement[:-1],
+        schema=', '.join(schema_statement),
```
austinbyers: This whole schema construction is much cleaner!

stream_alert_cli/athena/helpers.py
```diff
@@ -47,8 +86,25 @@ def to_athena_schema(log_schema):
...
+
+    Useful for listing tables, partitions, databases, enable_metrics
+
+    Args:
+        query_result (dict): The result of run_athena_query
+
+    Returns:
+        set: Unique values from the query result"""
+    return {
+        res
```
austinbyers: Can we just call this `val` or `value`? In my mind, it makes it clearer where it comes from and matches the description

ryandeivert: yep I can make that change @austinbyers (assuming you mean `val` instead of `res`?)

PR #712, opened 2018-04-30
assortment of small fixes to rule staging

ryandeivert:
to: @chunyong-lin or @austinbyers 
size: small

## Background

Discovered some small bugs related to rule staging that needed addressed.

## Changes

* Adding IAM perms so the rule processor can read rule table (`dynamodb:Scan`)
* Fixing CLI help message in `manage.py`
* Ensuring rule info is returned before looking up staged status

## Testing

Deployed to test

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -788,6 +788,8 @@ Optional Arguments:
 
     --skip-rule-staging                Skip staging of new rules so they go directly into
                                          production.
+    --stage-rules                      Stage the rules provided in a space-separated list
+    --unstage-rules                    Unstage the rules provided in a space-separated list
     --debug                            Enable debug logger output
 
 Examples:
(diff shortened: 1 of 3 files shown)


manage.py
```diff
@@ -816,18 +818,18 @@ def _add_lambda_deploy_subparser(subparsers):
             help=ARGPARSE_SUPPRESS
         )
 
-    # flag to manually bypass rule staging for specific rules during deploy
+    # flag to manually demote specific rules to staging during deploy
     lambda_deploy_parser.add_argument(
-        '--unstage-rules',
+        '--stage-rules',
```
chunyong-lin: I don't think precedence of argument `--stage-rules` and `--unstage-rules` matter. Isn't it?

PR #713, opened 2018-05-01
Add a Carbon Black output for hash banning using the CBAPI

fusionrace:
to: @ryandeivert @austinbyers 
cc: @chunyong-lin 
size: small

## Background
This feature aims to automate binary hash banning, using the CarbonBlack API through the use of an output. 

## Workflow
1. The output receives the `md5hash value` from the StreamAlert rule processor
2. The Carbon Black API (cbapi) takes the `md5hash value` and checks whether it is currently banned on the CB Enterprise Response Cloud.
3. If True, the cbapi then checks if the banned hash is enabled.  If True, the automation is exited.
4. If the `md5hash value` is not enabled (False), a command is sent to enable the hash.
5. If the `md5hash value` does not exist (False), an object is created that includes the binary hash value, banned hash text and checks the bit for the enabled feature.
6. If unsuccessful, a log is recorded.

## API Reference
http://cbapi.readthedocs.io/en/latest/response-examples.html#ban-a-binary

## Changes
* Import the CB Response API (cbapi=1.3.6)
* Created `outputs/carbonblack.py` that receives an `md5hash value` that is passed to the `cbapi`
* Created a `test_outputs/test_carbonblack.py`
* Updated supporting files: `/manage_lambda/package.py`, `test_outputs/test_output_base.py`

## Testing
**Unit Tests**
screen shot 2018-04-30 at 9 05 46 pm
screen shot 2018-04-30 at 9 05 27 pm

**Patched tested rules locally**

screen shot 2018-04-30 at 8 50 00 pm

## Thanks
@austinbyers @chunyong-lin @ryandeivert

### DIFF
--- a/stream_alert/alert_processor/outputs/carbonblack.py
+++ b/stream_alert/alert_processor/outputs/carbonblack.py
@@ -77,9 +77,9 @@ class CarbonBlackOutput(OutputDispatcher):
         client = CbResponseAPI(**creds)
 
         # Get md5 hash 'value' from streamalert's rule processor
-        action = alert.context.get('action')
+        action = alert.context.get('carbonblack', {}).get('action')
         if action == 'ban':
-            binary_hash = alert.context.get('value')
+            binary_hash = alert.context.get('carbonblack', {}).get('value')
             # The binary should already exist in CarbonBlack
             binary = client.select(Binary, binary_hash)
 
@@ -88,7 +88,7 @@ class CarbonBlackOutput(OutputDispatcher):
             if binary.banned:
                 # Determine if the banned action is enabled, if true exit
                 if binary.banned.enabled:
-                    return
+                    return True
                 # If the binary is banned and disabled, begin the banning hash operation
                 banned_hash = client.select(BannedHash, binary_hash)
                 banned_hash.enabled = True
@@ -105,4 +105,4 @@ class CarbonBlackOutput(OutputDispatcher):
             return self._log_status(success, descriptor)
         else:
             LOGGER.error('[%s] Action not supported: %s', self.__service__, action)
-            return self._log_status(binary.banned.enabled, self.__service__)
+            return self._log_status(False, descriptor)
--- a/stream_alert/alert_processor/outputs/carbonblack.py
+++ b/stream_alert/alert_processor/outputs/carbonblack.py
@@ -47,11 +47,6 @@ class CarbonBlackOutput(OutputDispatcher):
              OutputProperty(description='URL to the CB Response server [https://hostname]:',
                             mask_input=False,
                             cred_requirement=True)),
-            ('ssl_verify',
-             OutputProperty(description='Use SSL/TLS certificate validation'
-                                        ' (answer "N" if using self-signed certs) [Y/N]:',
-                            mask_input=False,
-                            cred_requirement=True)),
             ('token',
              OutputProperty(description='API token (if unknown, leave blank):',
                             mask_input=True,
@@ -84,8 +79,6 @@ class CarbonBlackOutput(OutputDispatcher):
             binary_hash = alert.context.get('carbonblack', {}).get('value')
             # The binary should already exist in CarbonBlack
             binary = client.select(Binary, binary_hash)
-
-            success = True
             # Determine if the binary is currenty listed as banned
             if binary.banned:
                 # Determine if the banned action is enabled, if true exit
@@ -104,7 +97,7 @@ class CarbonBlackOutput(OutputDispatcher):
                 banned_hash.enabled = True
                 banned_hash.save()
 
-            return self._log_status(success, descriptor)
+            return self._log_status(banned_hash.enabled is True, descriptor)
         else:
             LOGGER.error('[%s] Action not supported: %s', self.__service__, action)
             return self._log_status(False, descriptor)
--- a/stream_alert/alert_processor/outputs/carbonblack.py
+++ b/stream_alert/alert_processor/outputs/carbonblack.py
@@ -14,9 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 from collections import OrderedDict
-from cbapi.response import BannedHash, Binary, CbResponseAPI
-from stream_alert.alert_processor import LOGGER
 
+from cbapi.response import BannedHash, Binary, CbResponseAPI
+
+from stream_alert.alert_processor import LOGGER
 from stream_alert.alert_processor.outputs.output_base import (
     OutputDispatcher,
     OutputProperty,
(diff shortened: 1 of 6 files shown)


stream_alert/alert_processor/outputs/carbonblack.py
```diff
@@ -0,0 +1,108 @@
...
+
+        creds = self._load_creds(descriptor)
+        if not creds:
+            return self._log_status(False, descriptor)
+
+        client = CbResponseAPI(**creds)
+
+        # Get md5 hash 'value' from streamalert's rule processor
+        action = alert.context.get('carbonblack', {}).get('action')
+        if action == 'ban':
```
jacknagz: Should we extract the different actions into private instance methods? Do we see the list growing?

stream_alert/alert_processor/outputs/carbonblack.py
```diff
@@ -0,0 +1,108 @@
...
+            binary_hash = alert.context.get('carbonblack', {}).get('value')
+            # The binary should already exist in CarbonBlack
+            binary = client.select(Binary, binary_hash)
+
+            success = True
```
jacknagz: Is this mean't to change based on the if/else logic below? If not, you could just hard code `True` in the return statement

ryandeivert: yeah like jack mentioned - we should probably check the result of the save() to ensure that the binary was in fact marked properly

stream_alert/alert_processor/outputs/carbonblack.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        return OrderedDict([
+            ('descriptor',
+             OutputProperty(description='a short and unique descriptor for this'
+                                        ' carbonblack output')),
+            ('url',
+             OutputProperty(description='URL to the CB Response server [https://hostname]:',
+                            mask_input=True,
+                            cred_requirement=True)),
+            ('ssl_verify',
```
ryandeivert: this could actually be problematic, since these inputs are always `str` types.. basically, when this prompts the user they'll enter something like `true`. this will get serialized to a string and saved to s3 like:
```json
{
    "url": "whatever.url",
    "ssl_verify": "true",
    "token": "your.token"
}
```

When this is passed to the API below ([here](https://github.com/airbnb/streamalert/pull/713/files#diff-57f77691016d4f3f105484b13ad81280R77) it should be `bool` == `True` or `False` and instead will be a string.

I'm actually not sure of the best way to handle this, since these input don't have any validation right now. The StreamAlert apps have a 'format' param that can be a validation function that I would like to adopt here (eventually). An example can be seen here: https://github.com/airbnb/streamalert/blob/master/app_integrations/apps/box.py#L190-L210

stream_alert/alert_processor/outputs/carbonblack.py
ryandeivert: this suggestion could also potentially make testing a bit easier

stream_alert/alert_processor/outputs/carbonblack.py
fusionrace: We could hard code this.

stream_alert/alert_processor/outputs/carbonblack.py
```diff
@@ -0,0 +1,108 @@
...
+"""
+from collections import OrderedDict
+from cbapi.response import BannedHash, Binary, CbResponseAPI
+from stream_alert.alert_processor import LOGGER
+
+from stream_alert.alert_processor.outputs.output_base import (
+    OutputDispatcher,
+    OutputProperty,
+    StreamAlertOutput
+)
```
austinbyers: Can we add spacing so that the imports look like this:

```python

from cbapi.response import BannedHash, Binary, CbResponseAPI

from stream_alert.alert_processor import LOGGER
from stream_alert.alert_processor.outputs.output_base import (
    OutputDispatcher,
    OutputProperty,
    StreamAlertOutput
)
```

stream_alert/alert_processor/outputs/carbonblack.py
austinbyers: Yeah, that's a good idea. I'm ok if it doesn't make it in this PR, but I think it's a pretty easy fix

stream_alert/alert_processor/outputs/carbonblack.py
austinbyers: As far as I can tell, `save()` doesn't actually return anything useful:

https://github.com/carbonblack/cbapi-python/blob/c8cf5a5cf730068a481bbd4ffcd82124aa87db36/src/cbapi/models.py#L447-L453

So if it didn't raise an exception, I think it was successful

stream_alert/alert_processor/outputs/carbonblack.py
fusionrace: I do not see the list going in the near future.  Given other priorities we can slate this as an improvement in a future PR.

stream_alert/alert_processor/outputs/carbonblack.py
fusionrace: Are we looking to make this change now or later?

stream_alert/alert_processor/outputs/carbonblack.py
fusionrace: Update will be pushed shortly

stream_alert/alert_processor/outputs/carbonblack.py
ryandeivert: that's a good questions - what are the odds that anyone would ever want ssl_verify to be `False`? my short term fix would be to make the ssl_verify attribute hardcoded to the boolean of `True`. thoughts?

fusionrace: For now I think this would be the ideal situation.  We could fix later when we come up with a better solution.  The `ssl_verify` flag is also optional.

stream_alert/alert_processor/outputs/carbonblack.py
austinbyers: Oh interesting. So you think the API could modify the properties of the object if unsuccessful? That would be strange, but you might be right.

Easy enough fix: in that case, it's successful if `banned_hash.enabled == True`

fusionrace:
@ryandeivert @jacknagz @austinbyers 

## Changes
- Removed the `ssl_verify` entry from the credential setup.  It should now default to True.  Will will update this code in the future after a more in-depth discussion.  

- Added a check to for ensuring the the banned hash is enabled.  `banned-hash.enabled is True` 

### Question
- Are there any other items that you see as a blocker for this PR?  Thanks!

PR #716, opened 2018-05-02
updates to make athena client portable for use in other packages

ryandeivert:
to: @austinbyers 
size: large

## Background

The upcoming rule promotion lambda function is going to have to perform various queries against athena in order to generate alert statistics, etc. Instead of writing all of logic again, we're adapting the `StreamAlertAthenaClient` to be more flexible/portable.

## Note to Reviewers

* The `StreamAlertSQSClient` was migrated from the `stream_alert/athena_partition_refresh/main.py` file to the `clients.py` file without many code changes at all. **Please** do not review this too heavily as this code is mostly not mine. The exception is some small updates and bug fixes that were made during the migration/testing phase. See majority of them in [this commit](https://github.com/airbnb/streamalert/commit/354e827d4def35590d41ee15f5153f54af946dd3).

## Changes

* Migrating the `StreamAlertAthenaClient` from `stream_alert/athena_partition_refresh/main.py` to `stream_alert/shared/athena.py`.

### AthenaRefresher

Adding an `AthenaRefresher` class that is now responsible for:
* 1) polling sqs via the sqs client for new event notifications
* 2) using the new `StreamAlertAthenaClient` to update the partitions in athena given the buckets/keys from sqs.
* Removing a previous `add_partitions` method from the `StreamAlertAthenaClient` and adding this as a method to the `AthenaRefresher` class (since the logic embedded in this function was not related to athena directly ).
* Breaking the `add_partitions` method into two functions: `_get_partitions_from_keys` and `_add_partitions`. This helped with testing quite a bit.

### StreamAlertAthenaClient
* The client now supports a number of new operations:
  * The `query_result_paginator` returns a generator and allows for paginating over the results of a query that is being executed. Prior to this only one page could be retrieved for any query ran.
  * Simple standard sql operations like dropping tables, showing partitions, etc are now implemented as methods to improve ease of use (see: `drop_table`, `drop_all_tables`, and `get_table_partitions`)
  * New query operation that allows for executing and verifying the execution was successful (useful for things like dropping tables, etc where there is no actual results returned).
* Moving the helper function of `_unique_values_from_query` to be a static method of this class.
* Now catching `ClientErrors` that could be the result of a bad query string, etc.

### CLI Changes

* Updating the `stream_alert_cli/athena/handler.py` file to use the new athena client along with its new functionality.

## Testing

* Unit tests: migrating and updating some/adding new to cover all new code.
* Deployed in testing account to ensure all systems go. Alerts and data are available in athena as expected.


jacknagz:
🎉

PR #719, opened 2018-05-02
[Data Normalization] Improve performance

### LINKED ISSUE #717: Improvement: Data Normalization performance
## Background
In current Data Normalization implementation, Data Normalization is applied to a record right before rule matching. So one record may be normalized multiple times when there are different rules using different `datatypes`. For example,
* Rule 1 
```
@rule(datatypes=['command'], ...)
def rule_get_commands(rec):
...
```
* Rule 2
```
@rule(datatypes=['sourceAddress'], ...)
def rule_get_ip_address(rec):
...
```
* Rule 3
```
@rule(datatypes=['command', `sourceAddress`], ...)
def rule_get_command_and_ip_address(rec):
...
```
In this case, one record will be normalized 3 times!!! The performance will become bottleneck when you will thousands records read from files from S3 bucket, and it will result lambda function to be timed out.

## Desired Change
In rules engine, aggregate datatypes from all rules and normalize records only once before any rules matching.

chunyong-lin:
to: @ryandeivert @austinbyers 
size: median
resolves #717 

## Background
In current Data Normalization implementation, Data Normalization is applied to a record right before rule matching. So one record may be normalized multiple times when there are different rules using different `datatypes`. For example,
* Rule 1 
```
@rule(datatypes=['command'], ...)
def rule_get_commands(rec):
...
```
* Rule 2
```
@rule(datatypes=['sourceAddress'], ...)
def rule_get_ip_address(rec):
...
```
* Rule 3
```
@rule(datatypes=['command', `sourceAddress`], ...)
def rule_get_command_and_ip_address(rec):
...
```
In this case, one record will be normalized 3 times!!! The performance will become bottleneck when you will thousands records read from files from S3 bucket, and it will result lambda function to be timed out.

## Changes
* In rules engine, aggregate datatypes from all rules and normalize records only once before any rules matching.
* This change also update `get_keys` helper to skip normalization information.

## Testing
The testing environment has multiple rules using data normalization with different datatypes. And running time is based on `timeit`.
* 2761 carbonblack netconn records read from S3 file 
  * Before improvement, the running time is 24s
  * Apply improvement, the running time is 10s
  
* 7713 carbonblack event records read from S3 file
  * Before improvement, the running time is 37s
  * Apply improvement, the running time is 25s
  
* 1521 cloudtrail records read from S3 file
  * Before improvement, the running time is 584s
  * Apply improvement, the running time is 3s

Cloudtrail records seem to have significant benefit because they contains nested keys and ip addresses which will be normalized in my test environment.

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -73,8 +73,7 @@ class RulesEngine(object):
         cls._RULE_TABLE = RuleTable(table_name)
         cls._RULE_TABLE_LAST_REFRESH = now
 
-    @staticmethod
-    def match_types(record, normalized_types, datatypes):
+    def match_types(self, record, normalized_types, datatypes):
         """Match normalized types against record
 
         Args:
@@ -102,10 +101,9 @@ class RulesEngine(object):
         if not (datatypes and normalized_types):
             return
 
-        return RulesEngine.match_types_helper(record, normalized_types, datatypes)
+        return self.match_types_helper(record, normalized_types, datatypes)
 
-    @staticmethod
-    def match_types_helper(record, normalized_types, datatypes):
+    def match_types_helper(self, record, normalized_types, datatypes):
         """Helper method to recursively visit all subkeys
 
         Args:
@@ -121,8 +119,8 @@ class RulesEngine(object):
             if key == NORMALIZATION_KEY:
                 continue
             if isinstance(val, dict):
-                nested_results = RulesEngine.match_types_helper(val, normalized_types, datatypes)
-                RulesEngine.update(results, key, nested_results)
+                nested_results = self.match_types_helper(val, normalized_types, datatypes)
+                self.update(results, key, nested_results)
             else:
                 for datatype in datatypes:
                     if datatype in normalized_types and key in normalized_types[datatype]:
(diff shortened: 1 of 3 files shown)


tests/unit/stream_alert_rule_processor/test_rules_engine.py
```diff
@@ -810,10 +825,8 @@ def match_source_domain(rec): # pylint: disable=unused-variable
         # One record will be normalized twice by two different rules with different
         # normalization keys. It should generate two alerts by two different rules
         # from same record.
-        assert_equal(len(normalized_records), 2)
+        assert_equal(len(normalized_records), 1)
```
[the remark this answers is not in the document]
chunyong-lin: Aha, good catch! I will update the comment!

stream_alert/rule_processor/rules_engine.py
```diff
@@ -170,15 +168,12 @@ def update(results, parent_key, nested_results):
                 for item in val:
                     item.insert(0, parent_key)
             else:
-                val.insert(0, parent_key)
```
austinbyers: So was this a bug before?

chunyong-lin: Yes, it was a bug.

PR #725, opened 2018-05-03
better alert processor output success/failure logging

ryandeivert:
to: @austinbyers 
size: medium

## Background

I noticed that the way logging 'success' or 'failures' during alert dispatching was sort of janky. This updates the way logging occurs to make it much easier for someone to implement an alert output and not reason about logging statuses.

## Changes

* The alert output's `_dispatch` method now just returns True or False, depending on success or failure.
* Updating the output base class's `dispatch` method to call the subclass `_dispatch` method and handle the status logging of the result.
* Updating the alert processor to remove a now unnecessary method (`_send_alert`).
* Updates to documentation to explain the new interface for this.

## Testing

* Updating all unit tests for the changes made.

### DIFF
--- a/stream_alert/alert_processor/outputs/aws.py
+++ b/stream_alert/alert_processor/outputs/aws.py
@@ -63,7 +63,7 @@ class AWSOutput(OutputDispatcher):
                     **{values['descriptor'].value: values['aws_value'].value})
 
     @abstractmethod
-    def dispatch(self, alert, descriptor):
+    def _dispatch(self, alert, descriptor):
         """Placeholder for implementation in the subclasses"""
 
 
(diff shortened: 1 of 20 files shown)


stream_alert/alert_processor/outputs/aws.py
```diff
@@ -214,16 +214,14 @@ def dispatch(self, alert, descriptor):
...
                                  Qualifier=qualifier)
-        else:
-            resp = client.invoke(FunctionName=function,
-                                 InvocationType='Event',
-                                 Payload=alert_string)
 
-        return self._log_status(resp, descriptor)
+        return client.invoke(FunctionName=function,
+                             InvocationType='Event',
+                             Payload=alert_string)
```
austinbyers: I thought `_dispatch` was supposed to return a bool? I know this (dict return) is the same functionality we had before, but I'd like to keep consistent return types (it happens in other outputs, too). Maybe we just return True at the end of this? Returning the dict doesn't make much sense anyway

ryandeivert:
thanks @austinbyers ! good suggestion, I've update the pr to address your feedback

PR #726, opened 2018-05-04
Update Requirements and Fix Testing Deprecations

jacknagz:
to: @austinbyers 
cc: @ryandeivert @airbnb/streamalert-maintainers 
size: medium

## Background

Update requirements, fix testing deprecations.

## Changes

* Added `ushlex` into the top level requirements
* FakeFS `CreateFile` is [being deprecated](https://github.com/jmcgeheeiv/pyfakefs/blob/a22863571b1e0a6eeb0629cf68e78ea2e98ecc42/pyfakefs/fake_filesystem.py#L2864) in favor of `create_file`
* FakeFS Teardown [is no longer needed](https://github.com/jmcgeheeiv/pyfakefs/blob/99029d88065b3fbbe95226008b54e6ef16b92bf6/pyfakefs/tests/example_test.py#L73-L75)
* Asserting that Exceptions will raise via a context manager to fix `TypeError: 'exceptions.DeprecationWarning' object is not callable` errors

## Testing

Locally with unit, rule, and linting scripts


requirements-top-level.txt
```diff
@@ -21,4 +21,5 @@ pylint==1.7.4
 requests
 Sphinx
 sphinx-rtd-theme
+ushlex==0.99
```
ryandeivert: pinning might be unnecessary here

tests/unit/stream_alert_shared/test_config.py
```diff
@@ -33,22 +33,22 @@ def setUp(self):
...
+        self.fs.create_file('conf/global.json', contents='{}')
+        self.fs.create_file('conf/lambda.json', contents='{}')
+        self.fs.create_file('conf/logs.json', contents='{}')
+        self.fs.create_file('conf/outputs.json', contents='{}')
+        self.fs.create_file('conf/sources.json', contents='{}')
+        self.fs.create_file('conf/types.json', contents='{}')
 
     def tearDown(self):
-        self.tearDownPyfakefs()
+        pass
```
ryandeivert: these teardowns can be removed altogether if you want (less code!! :) )

PR #731, opened 2018-05-07
fixing bug where slack messages would split on each character if no newline existed

ryandeivert:
to: @austinbyers 
size: small

## Background

We observed issues with the `SlackOutput` where many thousands of attachments could be included with a message.

## Changes

* Fixing an issue with message splitting that was a result of the message not containing any new line characters (or no new line characters within the max message size limit).
* If new lines do not exist in the range, we will fall back on splitting on the last _space_ in the chunk of text.
* If neither space nor new lines exists in the chunk of text, we will split on the max message size, ignoring delimiters altogether.
* Simplifying the attachment generation code significantly.

## Testing

* Adding tests for the various characters (or lack there of) we can split one (new line, space, none).
* Adding test for the max attachment logging as well.

### DIFF
--- a/stream_alert/alert_processor/outputs/slack.py
+++ b/stream_alert/alert_processor/outputs/slack.py
@@ -60,6 +60,83 @@ class SlackOutput(OutputDispatcher):
         ])
 
     @classmethod
+    def _split_attachment_text(cls, alert_record):
+        """Yield messages that should be sent to slack.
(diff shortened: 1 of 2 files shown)


ryandeivert:
lol if there was anything I expected you to comment on, it was that @austinbyers fixed!

PR #732, opened 2018-05-07
[payload] shred the tmp directoy before downloading s3 objects

jacknagz:
to: @austinbyers or @ryandeivert 
size: small

## Background

To ensure Lambda's disk space is cleared on each invocation, let's shred the temp directory.

## Changes

* Shred temp
* Update tests
* Round for file sizes

## Testing

Local, CI, and deployed in a side account

### DIFF
--- a/stream_alert_cli/test.py
+++ b/stream_alert_cli/test.py
@@ -89,6 +89,9 @@ class RuleProcessorTester(object):
         # Create a cache map of parsers to parser classes
         self.parsers = {}
 
+        # Patch the tmp shredding as to not slow down testing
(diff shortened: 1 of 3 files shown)


stream_alert_cli/test.py
```diff
@@ -89,6 +89,9 @@ def __init__(self, context, config, print_output):
         # Create a cache map of parsers to parser classes
         self.parsers = {}
 
+        # Patch the tmp shredding as to not slow down testing
```
austinbyers: Well, you also definitely don't want to shred peoples /tmp directories when running unit tests :)

PR #734, opened 2018-05-10
[tf] send message to SNS topic when alarms recover

jacknagz:
size: small

## Background

The Terraform module for `monitoring` is used to create CloudWatch alarms when certain thresholds are passed.  When alarms are triggered, a message is sent to an SNS topic, which can be subscribed by PagerDuty.

While digging through [PagerDuty documentation](https://www.pagerduty.com/docs/guides/aws-cloudwatch-integration-guide/), I discovered that the alarm wasn't configured to report back to SNS when alarms recover, which can auto resolve PagerDuty incidents.

## Changes

* Send messages to the cloudwatch monitoring SNS topic when the triggered alarm recovers

## Testing

Deployed in a test account


austinbyers:
We would also want to do this in the `tf_lambda` module, which is used by most of the StreamAlert Lambda functions

PR #737, opened 2018-05-15
[firehose] only retry failed records when re-calling put_record_batch

ryandeivert:
to: @jacknagz 
size: small

## Background

If firehose fails to send some records within a batch of records, the response will identify which ones failed to send. We can leverage this information to only retry items that failed, instead of retrying the entire batch.

## Changes

* Fixing issue with firehose client reset when a backoff exception occurs.  This resets the client with the proper config, which includes connect and read timeout information.
* Adding logic to only retry firehose records that have failed to send in the batch. This modifies the list of records by reference, and the backed-off function will see the new, modified list.

## Testing

Tested in a local script with backoff and adding unit tests for the function that performs the stripping of successful records from the batch.

### DIFF
--- a/stream_alert/rule_processor/firehose.py
+++ b/stream_alert/rule_processor/firehose.py
@@ -51,12 +51,12 @@ class StreamAlertFirehose(object):
     BOTO_TIMEOUT = 5
 
     def __init__(self, region, firehose_config, log_sources):
-        self._boto_config = client.Config(
+        boto_config = client.Config(
             connect_timeout=self.BOTO_TIMEOUT,
             read_timeout=self.BOTO_TIMEOUT,
             region_name=region
         )
-        self._firehose_client = self._create_firehose_client()
+        self._firehose_client = boto3.client('firehose', config=boto_config)
         # Expand enabled logs into specific subtypes
         self._enabled_logs = self._load_enabled_log_sources(firehose_config, log_sources)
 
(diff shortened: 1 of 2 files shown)


stream_alert/rule_processor/firehose.py
```diff
@@ -155,11 +154,8 @@ def sanitize_keys(cls, record):
 
     def _backoff_handler_firehose_reset(self, details):
         """Custom backoff handler to re-instantiate the Firehose Client"""
-        LOGGER.info('[Backoff]: Calling \'%s\' again in %f seconds with %d tries so far',
-                    details['target'].__name__,
-                    details['wait'],
-                    details['tries'])
-        self._reset_firehose_client()
+        backoff_handler(debug_only=False)(details)
+        self._firehose_client = self._create_firehose_client()
```
jacknagz: After some more thought, we may not need to do this on backoff. That was an attempt to fix the issue at hand, which didn't end up working.

ryandeivert: I was thinking the same but chose to leave it for.. happy to remove it if you think we should?

stream_alert/rule_processor/firehose.py
```diff
@@ -157,6 +157,23 @@ def _backoff_handler_firehose_reset(self, details):
         backoff_handler(debug_only=False)(details)
         self._firehose_client = self._create_firehose_client()
 
+    @staticmethod
+    def _strip_successful_records(batch, response):
+        """Inspect the response and remove any records records that have successfully to sent
```
jacknagz: Can we add this to the docstring:

``` For each record, the index of the response element is the same as the index used in the request array.```

jacknagz: From: http://boto3.readthedocs.io/en/latest/reference/services/firehose.html#Firehose.Client.put_record_batch

stream_alert/rule_processor/firehose.py
jacknagz: We can remove it to boost the perf

ryandeivert:
@jacknagz addressed comments - thanks!

PR #743, opened 2018-05-21
[parsers] better key-value parser support

jacknagz:
to: @ryandeivert 
size: small

## Background

The KV parser is fairly basic, and can't support parsing logs that have quoted fields, or spaces within values. Log types like `auditd`, `nginx`, and others use the style described in the unit test:

`key="value with spaces" another_key=2`

To support these logs, additional parser logic is now added. 

## Changes

* Support spaces in KV values, and strip quotes in data so rule writing is more intuitive
* Add unit test coverage

## Testing

### DIFF
--- a/stream_alert/rule_processor/parsers.py
+++ b/stream_alert/rule_processor/parsers.py
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 from abc import ABCMeta, abstractmethod
-from collections import OrderedDict
+from collections import defaultdict, OrderedDict
 import csv
 from fnmatch import fnmatch
 import json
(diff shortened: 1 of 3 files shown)


stream_alert/rule_processor/parsers.py
```diff
@@ -398,6 +398,40 @@ class KVParser(ParserBase):
     __default_separator = '='
     __default_delimiter = ' '
 
+    @staticmethod
+    def _string_to_list(data, delimiter, separator):
```
ryandeivert: Hey @jacknagz I'm curious if you looked at using `shlex` for this type of splitting? I think we can avoid rolling all of this logic on our own. For instance, does this provide what you're looking for? With the test data you provide of `'name=test result=OK date="Friday, January 01, 2018" count=5'`:

```
>>> import shlex
>>> shlex.split('name=test result=OK date="Friday, January 01, 2018" count=5')
['name=test', 'result=OK', 'date=Friday, January 01, 2018', 'count=5']
```

jacknagz: @ryandeivert  Good call, I didn't know about that library. I'll add it and re-commit.

ryandeivert: 👍 as an aside, you may want to cast to `str` first or use `ushlex` instead to enable better unicode support

austinbyers: Or you can convert everything to Python3 for better unicode support :trollface:

ryandeivert: trollolololol

PR #744, opened 2018-05-21
[tf][cli] add cross-account cloudwatch logs support

ryandeivert:
to: @jacknagz for CLI, @austinbyers for tf
size: medium

## Background

To enable cross-account ingestion from a CloudWatch Log group subscription, StreamAlert needs to create CloudWatch Log destination to collection the logs.

## Changes

* Adding a new terraform module (`tf_stream_alert_cloudwatch`) that will create a CloudWatch Log destination in each available region.
  * Creates the following in _each region_:
    * IAM role that will grant CloudWatch Logs the permission to put data into your Kinesis stream.
    * Permissions policy to define which actions CloudWatch Logs can perform.
      * `kinesis:PutRecord` for the Kinesis stream
      * `iam:PassRole` for the previously created IAM role
    * A CloudWatch Log destination that points to the cluster's default Kinesis stream.
    * Policy that defines who has write access to the destination
      * `logs:PutSubscriptionFilter` for the cross-account principals (account IDs)
* The destination that is created will point to the cluster's default kinesis stream, and therefore kinesis must be enabled in the cluster in order to enable this module.
* If needed, specific regions can be excluded from having a destination created via the config option of `excluded_regions`

## Testing

* CLI: Adding unit test for CLI generation code
* Terraform: ran `terraform validate` & `terraform plan`

### DIFF
--- a/terraform/modules/tf_stream_alert_cloudwatch/iam.tf
+++ b/terraform/modules/tf_stream_alert_cloudwatch/iam.tf
@@ -1,7 +1,7 @@
 # IAM Role: Allows CloudWatch Logs to put data into
 # this cluster's default Kinesis stream
 resource "aws_iam_role" "cloudwatch_subscription_role" {
-  name = "stream_alert_${var.cluster}_cloudwatch_subscription_role"
+  name = "stream_alert_${var.cluster}_cloudwatch_subscription_role_${var.region}"
 
   assume_role_policy = "${data.aws_iam_policy_document.cloudwatch_logs_assume_role_policy.json}"
 }
@@ -21,7 +21,7 @@ data "aws_iam_policy_document" "cloudwatch_logs_assume_role_policy" {
 
 // IAM Policy: Write to Kinesis
 resource "aws_iam_role_policy" "cloudwatch_kinesis_wo" {
-  name = "stream_alert_${var.cluster}_write_cloudwatch_to_kinesis"
+  name = "WriteCWLogsToKinesis"
   role = "${aws_iam_role.cloudwatch_subscription_role.id}"
 
   policy = "${data.aws_iam_policy_document.cloudwatch_put_kinesis_events.json}"
--- a/terraform/modules/tf_stream_alert_cloudwatch/iam.tf
+++ b/terraform/modules/tf_stream_alert_cloudwatch/iam.tf
@@ -1,0 +1,87 @@
+# IAM Role: Allows CloudWatch Logs to put data into
+# this cluster's default Kinesis stream
+resource "aws_iam_role" "cloudwatch_subscription_role" {
+  name = "stream_alert_${var.cluster}_cloudwatch_subscription_role"
+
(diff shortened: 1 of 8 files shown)


terraform/modules/tf_stream_alert_cloudwatch/iam.tf
```diff
@@ -0,0 +1,87 @@
...
+    principals {
+      type        = "Service"
+      identifiers = ["logs.${var.region}.amazonaws.com"]
+    }
+  }
+}
+
+// IAM Policy: Write to Kinesis
+resource "aws_iam_role_policy" "cloudwatch_kinesis_wo" {
+  name = "stream_alert_${var.cluster}_write_cloudwatch_to_kinesis"
```
jacknagz: Can we make this camel case? Maybe `WriteCLWToKinesis`

terraform/modules/tf_stream_alert_cloudwatch/iam.tf
```diff
@@ -0,0 +1,87 @@
...
+  policy = "${data.aws_iam_policy_document.cloudwatch_put_kinesis_events.json}"
+}
+
+// IAM Policy Document: Write to Kinesis
+data "aws_iam_policy_document" "cloudwatch_put_kinesis_events" {
+  statement {
+    effect = "Allow"
+
+    actions = [
+      "kinesis:PutRecord",
```
jacknagz: Does it need `PutRecords` as well?

jacknagz:
Looking good, I would apply this in a side account and verify end to end before merging

terraform/modules/tf_stream_alert_cloudwatch/iam.tf
ryandeivert: The docs only show needing `PutRecord`. See step # 5 here: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CreateDestination.html

I also tested this with the perms they describe and it works.

terraform/modules/tf_stream_alert_cloudwatch/iam.tf
ryandeivert: does it need to be unique? Meaning can I drop the cluster interpolation without hurting anything?

ryandeivert: also are you only suggesting using camel case for the role policy? or the other 'name' attributes as well?

jacknagz: It just has to be unique to the role itself. And just the role_policy since `name` shows up as the descriptor in the IAM console

terraform/modules/tf_stream_alert_cloudwatch/iam.tf
jacknagz: Interesting that it does no batching 🤔

ryandeivert:
applied in test account and addressed PR feedback. merging

PR #745, opened 2018-05-22
[parsers] add cloudtrail via cloudwatch logs support

ryandeivert:
to: @austinbyers 
size: small

## Background

Logs that are ingested via CloudWatch logs subscription filter have a unique format. The format is described [here in step 8 under Example # 1](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html). The `message` objects within the list of `logEvents` are stored as embedded json objects. However, currently, there is no easy way to use a json path in combination with loading an embedded json object. This adds support for loading of embedded json objects for json records extracted via a json path.

## Changes

* Adding new `logs.json` config option of `embedded_json` that is a boolean indicating if the extracted json object is an embedded object (encoded json within json).
* Adding json loading logic for embedded json objects via new method of `_extract_records`. This new method also handles the json regex extraction & loading.
* Adding CloudTrail log via CloudWatch logs schema support. See [here](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-record-contents.html) for more info on the contents of the log.

## Testing

* Adding unit tests for the extraction of the embedded json object.

### DIFF
--- a/stream_alert/rule_processor/parsers.py
+++ b/stream_alert/rule_processor/parsers.py
@@ -280,7 +280,7 @@ class JSONParser(ParserBase):
                     try:
                         record = json.loads(match.value)
                     except ValueError:
-                        LOGGER.debug('Embedded json is invalid')
+                        LOGGER.warning('Embedded json is invalid')
                         continue
                 if envelope:
                     record.update({ENVELOPE_KEY: envelope})
--- a/stream_alert/rule_processor/parsers.py
+++ b/stream_alert/rule_processor/parsers.py
@@ -211,7 +211,7 @@ class JSONParser(ParserBase):
             json_payload (dict): The parsed json data
 
         Returns:
-            list: A list of JSON recrods extracted via JSONPath.
+            list: A list of parsed JSON records
         """
         # Check options and return the payload if there is nothing special to do
         if not self.options:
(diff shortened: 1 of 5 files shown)


stream_alert/rule_processor/parsers.py
```diff
@@ -255,13 +275,19 @@ def _parse_records(self, schema, json_payload):
                 return False
             for match in matches:
                 record = match.value
+                embedded_json = self.options.get('embedded_json')
+                if embedded_json:
+                    try:
+                        record = json.loads(match.value)
+                    except ValueError:
+                        LOGGER.debug('Embedded json is invalid')
```
austinbyers: Should this be a warning? If I understand correctly, this logic parses embedded json (if specified) for each matched log schema. If that fails, then the config is probably wrong? Or will this happen more frequently?

ryandeivert: It's probably safe to have it as warning or error. just updated!

jacknagz:
👏  @ryandeivert could you also add the new `embedded_json` config option to our docs so other users have an understanding of how it works?

ryandeivert:
@jacknagz I will make a todo for it and address soon

PR #750, opened 2018-05-24
[tf] cloudtrail to cloudwatch to kinesis support

ryandeivert:
to: @austinbyers 
size: medium

## Background

The cloudtrail terraform module currently supports sending cloudtrail logs to an s3 bucket, but does not support sending to a CloudWatch logs group. Sending to a CloudWatch Logs group would allow for sending to kinesis (skipping s3 altogether). 

## Changes

* Updating the cloudtrail terraform module to optionally support sending logs to a CloudWatch logs group (along with the default s3 bucket).

## Testing

* Fully deployed and enabled in test account.
* Updated unit tests for new cloudtrail module generation.

### DIFF
--- a/stream_alert_cli/terraform/cloudtrail.py
+++ b/stream_alert_cli/terraform/cloudtrail.py
@@ -36,7 +36,7 @@ def generate_cloudtrail(cluster_name, cluster_dict, config):
 
     cloudtrail_enabled = modules['cloudtrail'].get('enable_logging', True)
     kinesis_enabled = modules['cloudtrail'].get('enable_kinesis', True)
-    send_to_cloudwatch = modules['cloudtrail'].get('send_to_cloudwatch', True)
+    send_to_cloudwatch = modules['cloudtrail'].get('send_to_cloudwatch', False)
     exclude_home_region = modules['cloudtrail'].get('exclude_home_region_events', False)
 
     account_ids = list(
--- a/stream_alert_cli/terraform/cloudtrail.py
+++ b/stream_alert_cli/terraform/cloudtrail.py
@@ -36,6 +36,8 @@ def generate_cloudtrail(cluster_name, cluster_dict, config):
 
     cloudtrail_enabled = modules['cloudtrail'].get('enable_logging', True)
     kinesis_enabled = modules['cloudtrail'].get('enable_kinesis', True)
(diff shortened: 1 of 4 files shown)


stream_alert_cli/terraform/cloudtrail.py
```diff
@@ -36,6 +36,8 @@ def generate_cloudtrail(cluster_name, cluster_dict, config):
 
     cloudtrail_enabled = modules['cloudtrail'].get('enable_logging', True)
     kinesis_enabled = modules['cloudtrail'].get('enable_kinesis', True)
+    send_to_cloudwatch = modules['cloudtrail'].get('send_to_cloudwatch', True)
```
austinbyers: Was True your intended default here? If I understand this correctly, this means all existing cloudtrail modules in users' configs will now send to cloudwatch by default unless the user explicitly overrides it. Maybe that's ok, I don't know. Just wanted to check

terraform/modules/tf_stream_alert_cloudtrail/main.tf
```diff
@@ -1,8 +1,10 @@
-// StreamAlert CloudTrail
+// StreamAlert CloudTrail, also sending to CloudWatch Logs group
 resource "aws_cloudtrail" "streamalert" {
-  count                         = "${var.existing_trail ? 0 : 1}"
+  count                         = "${!var.existing_trail && var.send_to_cloudwatch ? 1 : 0}"
```
austinbyers: Silly thing, but can we switch this order? It makes more sense in English to say `send_to_cloudwatch and not existing_trail` (also below)

stream_alert_cli/terraform/cloudtrail.py
austinbyers: How do they specify just S3 as a destination? Both?

ryandeivert: Sending to s3 is required and on by default. I will make the cloudwatch stuff disabled by default as well - thanks!

terraform/modules/tf_stream_alert_cloudtrail/main.tf
ryandeivert: sure thing!

ryandeivert:
bypassing coveralls requirement since it seems to be hung

PR #752, opened 2018-05-25
[tf] fixing bug with policy resource value

ryandeivert:
to: @jacknagz or @austinbyers 
size: small

## Background

A bug was found with the policy doc that should allow `logs:PutSubscriptionFilter` for the cloudwatch destination.

## Changes

* Fixing bug by using the destination arn and not the kinesis stream arn the destination points to.
* Using the output of the destination arn as the destination in the subscription filter of the CloudTrail module.

## Testing

* Applied changes in test deploy with much success.


ryandeivert:
bypassing coveralls since it appears to be hung

PR #753, opened 2018-05-25
[cli][tf] adding ability to send cloudtrail to another cluster's cloudwatch destination

ryandeivert:
to: @austinbyers 
size: small

## Background

It's possible to enable the `cloudtrail` module in a different cluster than the `cloudwatch` module, but the cloudtrail module needs access to a destination arn from the cloudwatch module in order to send events to it.

## Changes

* Adding support for a config option to supply a destination arn that is different than the current cluster's destination arn.

## Testing

Applied and deployed 👏


ryandeivert:
bypassing coveralls.. again

PR #754, opened 2018-05-25
[parsers] support csv embedded within json logs

ryandeivert:
to: @austinbyers 
size: medium

## Background

Some logs are csv format, but are embedded within a json object. Ideally, these could be extracted and still parsed as csv.

## Changes

* Adding support to the csv parser to allow for extracting csv data via a json_path.
* Migrating some of this parsing from the `JSONParser` to the `ParserBase` class to be used across the subclasses.
* Adding support for extracting 'envelope' keys for csv-in-json logs.
  * This logic was largely migrated from the `JSONParser` class to `ParserBase` with some exceptions. This can be improved at a later date.
* Adding log patterns to `cloudwatch:control_message` so it does not get misclassified as another log type.
* Adding `cloudwatch:rds_aurora` schema that leverages the new csv-in-json logic to extract the csv records and parser them.

## Testing

* Updating unit tests to include tests for the embedded csv within json.
* Adding integration tests for aurora rds csv logs within json.
* All current rule tests passing.

### DIFF
--- a/conf/logs.json
+++ b/conf/logs.json
@@ -1192,6 +1192,13 @@
         "owner": "string",
         "subscriptionFilters": []
       },
+      "log_patterns": {
+        "streamalert:envelope_keys": {
+          "messageType": [
+            "CONTROL_MESSAGE"
+          ]
+        }
+      },
       "json_path": "logEvents[*]"
     }
   },
--- a/conf/logs.json
+++ b/conf/logs.json
@@ -1243,6 +1243,31 @@
       "json_path": "logEvents[*].extractedFields"
     }
   },
+  "cloudwatch:rds_aurora": {
+    "schema": {
+      "timestamp": "integer",
+      "serverhost": "string",
+      "username": "string",
+      "host": "string",
+      "connectionid": "integer",
+      "queryid": "integer",
+      "operation": "string",
+      "database": "string",
+      "object": "string",
+      "retcode": "integer"
+    },
+    "parser": "csv",
+    "configuration": {
+      "envelope_keys": {
+        "logGroup": "string",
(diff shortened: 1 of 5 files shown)


conf/logs.json
```diff
@@ -1236,6 +1243,31 @@
...
+    "parser": "csv",
+    "configuration": {
+      "envelope_keys": {
+        "logGroup": "string",
+        "logStream": "string",
+        "messageType": "string",
+        "owner": "string",
+        "subscriptionFilters": []
+      },
+      "json_path": "logEvents[*].message"
```
austinbyers: Does `logEvents` need to be specified in the `envelope_keys` or anywhere else in the schema config?

stream_alert/rule_processor/parsers.py
```diff
@@ -76,6 +76,56 @@ def type(self):
...
+            dict: Key/values extracted from the log to be used as the envelope
+        """
+        if not isinstance(json_payload, dict):
+            json_payload = json.loads(json_payload)
+        LOGGER.debug('Parsing envelope keys')
+        schema.update({ENVELOPE_KEY: envelope_schema})
+        envelope_keys = envelope_schema.keys()
+        envelope_jsonpath = jsonpath_rw.parse("$." + ",".join(envelope_keys))
+        envelope_matches = [match.value for match in envelope_jsonpath.find(json_payload)]
+        return dict(zip(envelope_keys, envelope_matches))
```
austinbyers: I know this was just migrated, but it's cool! I like it

conf/logs.json
ryandeivert: No - good catch though!

PR #757, opened 2018-06-03
[apps] fixing terraform bug where app iam would not be targeted during deploy

ryandeivert:
to: @austinbyers 
size: small

## Background

There is a bug where the IAM resources for apps are not added to the terraform targets during a deploy of the apps functions.

## Changes

* Fixing bug by adding the iam tf targets alongside the lambda tf targets during deploy of the apps functions.

## Testing

No unit tests but tested by doing a generate locally to make sure all the proper targets are included.

### DIFF
--- a/stream_alert_cli/manage_lambda/deploy.py
+++ b/stream_alert_cli/manage_lambda/deploy.py
@@ -107,7 +107,8 @@ def _create_and_upload(function_name, config, cluster=None):
         ),
         'apps': PackageMap(
             stream_alert_packages.AppIntegrationPackage,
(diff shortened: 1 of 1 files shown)


austinbyers:
Should we do this for the other functions too? Anything that uses the new `tf_lambda` module will have a separate module for its IAM permissions

ryandeivert:
@austinbyers good question - I'm not very familiar with how this is done for the other functions right now (is it not being done at all?)

austinbyers:
No worries, I'll get the other functions (it's the same idea)

Issue #758, opened 2018-06-04
Encrypt all data at rest (Dynamo, S3, SQS)


austinbyers:
StreamAlert stores and processes potentially sensitive security logs. As such, all data should be encrypted at rest. In particular, Dynamo, S3, and SQS support server-side encryption, and it's relatively easy to enable in Terraform

PR #761, opened 2018-06-05
Remove Lambda source bucket

austinbyers:
to: @ryandeivert 
size: medium

## Background

StreamAlert currently creates an S3 bucket to store Lambda source code, and part of the deploy process is uploading the `.zip` files for each Lambda function to S3.

This adds unnecessary complexity to the CLI and the config files; you can just specify the deployment package directly in Terraform. Besides being simpler (and possibly faster), this means there's one less bucket we need to encrypt for #758 

## Changes

* Remove Lambda source bucket
* Lambda `.zip` files are now saved to the `terraform` directory. This mimics BinaryAlert's setup, making it easy for both Terraform and human users to inspect the deployment package locally.
* Remove all references to source buckets and object keys from default config, docs, terraform modules, and tests
* Update IAM permissions for alert processor and alert merger on a deploy (follow-up to #757) 
* Add `oauth2client` to app deployment package to fix an import error when running a base app
     * This was caused by an update to `google-api-python-client`, which dropped the dependency on `oauth2client`. As a result, we're also (temporarily) pinning the version of the python client

A neat side effect of this cleanup is that you can now run `terraform` operations directly from the `terraform/` directory without any extra flags. For example, now you can `cd terraform && terraform plan` instead of `cd terraform && terraform plan -var-file=../conf/lambda.json`

_A future PR will also remove the Lambda version information from the config_

## Testing

* Applying the change to an existing deployment:

screen shot 2018-06-04 at 3 53 24 pm

* Tested deploy of core functions (rule processor, alert merger, alert processor, athena refresh), and alerts work end to end
* Deployed apps and threat intel downloader
    * Verified the apps no longer have an import error
* `./manage.py terraform build` shows no pending changes
* `./manage.py lambda deploy -p all` - able to deploy all Lambda functions
* `./manage.py lambda rollback -p all` - able to rollback all Lambda functions
* There are no remaining references to `source_bucket`, `source_current_hash`, nor `source_object_key` anywhere in the codebase


stream_alert_cli/manage_lambda/version.py
```diff
@@ -68,18 +68,17 @@ def _publish_helper(self, cluster=None):
 
         # Configure the Lambda client
         client = boto3.client('lambda', region_name=region)
-        code_sha_256 = self.config['lambda'][self.package.config_key]['source_current_hash']
 
         # Publish the function(s)
         # TODO: move the extra logic into the LambdaPackage subclasses instead of this
         if self.package.package_name == 'stream_alert_app':
-            if not 'stream_alert_apps' in self.config['clusters'][cluster]['modules']:
+            if 'stream_alert_apps' not in self.config['clusters'][cluster]['modules']:
```
ryandeivert: ++ I assume your editor flagged this? 😂

austinbyers: Yep. Both evaluate the same way, but this is the more proper form because it reads less ambiguous

ryandeivert:
bypassing hung coveralls

PR #762, opened 2018-06-05
Remove Lambda versions and handlers from config

austinbyers:
to: @ryandeivert or @jacknagz 
size: medium

## Background

As a follow-up to #761 , we can further simplify the config and the deploy process by removing Lambda versions and handlers. Terraform can auto-publish versions (so we need not do it explicitly) and handlers can be hard-coded in the `LambdaPackage` subclasses instead of visible in userland config files.

After this change, deploys will no longer change the config at all! There won't be any more "dirty state" after a deploy, or having to submit pull requests just for a deploy

## Changes

* Remove `current_version` and `handler` from config files
* Terraform auto-publishes versions for all Lambda functions instead of a `boto3` publish in the CLI
* The `LambdaPackage` class is slightly simplified and now includes the handler
* `./manage.py lambda rollback` is now done via `boto3` calls instead of modifying the config and running Terraform

One nice side effect of this change is that rollbacks are now blazing fast compared to a deploy, since Terraform isn't involved at all. I was able to rollback all 9 Lambda functions in my test deployment in just 2 seconds!

## Testing

* Deployed to a test account, with both apps and threat intel downloader enabled. Alerts flow end-to-end, and all functions are executing normally
* Various combinations of `./manage.py lambda deploy` and `./manage.py lambda rollback`


stream_alert_cli/manage_lambda/package.py
```diff
@@ -30,23 +27,13 @@
...
-    config_key = None
-    package_folders = set()
-    package_files = set()
-    package_name = None
-    package_root_dir = '.'
-    precompiled_libs = set()
-    third_party_libs = set()
-    version = None
+    """Build a deployment package for a StreamAlert Lambda function."""
+    config_key = None         # Configuration key to access conf/lambda.json
```
jacknagz: Nice simplification here

stream_alert_cli/manage_lambda/package.py
```diff
@@ -66,7 +53,7 @@ def create(self):
             exit(1)
 
         # Extract any precompiled third-party libs for this package
-        if not self._extract_precompiled_libs(temp_package_path):
+        if self.precompiled_libs and not self._extract_precompiled_libs(temp_package_path):
```
jacknagz: 👍

stream_alert_cli/manage_lambda/package.py
```diff
@@ -160,67 +146,89 @@ def _resolve_third_party(self, temp_package_path):
 class RuleProcessorPackage(LambdaPackage):
     """Deployment package class for the StreamAlert Rule Processor function"""
     config_key = 'rule_processor_config'
-    package_folders = {
-        'stream_alert/rule_processor', 'stream_alert/shared', 'rules', 'matchers', 'helpers', 'conf'
+    lambda_handler = 'stream_alert.rule_processor.main.handler'
+    package_files = {
```
jacknagz: Does it make sense to create a set with common package files between all functions?

stream_alert_cli/manage_lambda/rollback.py
```diff
@@ -13,128 +13,59 @@
...
-def _rollback_downloader(config):
-    """Decrement the current_version for the Threat Intel Downloader function."""
-    lambda_config = config['lambda'].get('threat_intel_downloader_config')
-    if lambda_config and _try_decrement_version(lambda_config, 'threat_intel_downloader_config'):
-        return['module.threat_intel_downloader']
 
+def _rollback_production(lambda_client, function_name):
+    """Rollback the production alias for the given function name."""
+    response = lambda_client.get_alias(FunctionName=function_name, Name='production')
+    current_version = int(response['FunctionVersion'])
```
jacknagz: This will fail if the current version is `$LATEST`

stream_alert_cli/manage_lambda/rollback.py
```diff
@@ -13,128 +13,59 @@
...
+    if current_version <= 1:
```
jacknagz: The version will never be less than 1

jacknagz: And it's ok to roll back to `$LATEST`

stream_alert_cli/manage_lambda/rollback.py
jacknagz: And keep in mind, since clusters exist, there's the possibility that a rule processor on `cluster 1` could be a different version than `cluster 2`

terraform/modules/tf_stream_alert/main.tf
```diff
@@ -31,7 +32,7 @@ resource "aws_lambda_alias" "rule_processor_production" {
   name             = "production"
   description      = "Production StreamAlert Rule Processor Alias"
   function_name    = "${aws_lambda_function.streamalert_rule_processor.arn}"
-  function_version = "${var.rule_processor_version}"
+  function_version = "${aws_lambda_function.streamalert_rule_processor.version}"
```
jacknagz: This interpolation enables the rollbacks to work properly right? I'm assuming Terraform refreshes state after the boto3 call is made

stream_alert_cli/manage_lambda/package.py
austinbyers: This logic just migrated out of `_extract_precompiled_libs` for simplicity

stream_alert_cli/manage_lambda/package.py
austinbyers: I just checked - there actually isn't a single file used by all functions. The apps only use a single folder (`app_integrations`)! We could create a common set for the rest of the functions, but I personally am in favor of the explicit list for each package - it makes it very easy to read

stream_alert_cli/manage_lambda/rollback.py
austinbyers: The current version should never be `$LATEST`, because Terraform will publish a new (numbered) version for every change. However, it's still worth something worth checking for (in case it was manually set) and handling the error gracefully. Will do!

As for the clusters, we're covered! This function is called for each individual Lambda function: get the current version, decrement by 1. If clusters are at different versions, they will all be decremented individually.

stream_alert_cli/manage_lambda/rollback.py
austinbyers: True, less than 1 is a silly check, I'll get rid of it (to avoid any confusion)

Because Terraform publishes new versions for every change, `$LATEST` will always be identical to the most recent published version. And if it were not (in case of a manual change), `$LATEST` will be, by definition, newer than any other version. So I would argue it's impossible to "roll back" to `$LATEST`

terraform/modules/tf_stream_alert/main.tf
austinbyers: Yep! After a rollback, a `terraform apply` would restore the aliases to the latest published version (as long as `-refresh` was enabled, which it is by default and in all of our CLI). A deploy after a rollback would first publish a new version and then skip ahead from the rollback to the most recent (jumping forward 2 versions).

I just tested both of these exact scenarios and confirm they work as expected

austinbyers:
@jacknagz PTAL when you get a chance - I added an error log in the very rare event someone is trying to roll back `$LATEST`

conf/lambda.json
```diff
@@ -58,18 +54,13 @@
...
   "rule_processor_config": {
-    "handler": "stream_alert.rule_processor.main.handler",
     "third_party_libraries": [
       "pathlib2"
     ]
   },
-  "stream_alert_apps_config": {
-    "handler": "app_integrations.main.handler"
-  }
+  "stream_alert_apps_config": {}
```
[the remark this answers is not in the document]
austinbyers: Nope! Good catch, I removed it

austinbyers:
I tested with another fresh deploy to make sure removing the unused config and terraform variables didn't break anything. We are 🚀

PR #768, opened 2018-06-19
adding optional gzip decompression for lookup file loading from s3

ryandeivert:
to: @chunyong-lin 
cc: @airbnb/streamalert-maintainers, @strcrzy, @jbussing 
size: small

## Background

As an optimization, gzip compressing the lookup files stored in S3 will have shorter latency and result in less S3 cost over time.

## Changes

* Adding decompression capability when loading lookup files from S3.
* If the file is not gzip'd, we will load the original raw data into json.

## Testing

Adding unit test to confirm decompression works as expected.

### DIFF
--- a/tests/unit/stream_alert_shared/test_lookup_tables.py
+++ b/tests/unit/stream_alert_shared/test_lookup_tables.py
@@ -16,6 +16,7 @@ limitations under the License.
 from datetime import datetime
 import json
 import os
+import zlib
 
 from botocore.vendored.requests.packages.urllib3.exceptions import ReadTimeoutError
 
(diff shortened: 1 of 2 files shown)


tests/unit/stream_alert_shared/test_lookup_tables.py
```diff
@@ -127,3 +128,21 @@ def test_load_lookup_tables(self, mock_logger):
...
+            'bucket_name',
+            'bar.json',
+            zlib.compress(json.dumps({'compressed_key': 'compressed_val'})),
+            self.region
+        )
+        result = lookup_tables.download_s3_objects()
+
+        assert_equal(result.get('bar'), {'compressed_key': 'compressed_val'})
+        assert_equal(result.get('foo'), {'bucket_name_key': 'foo_value'})
+        mock_logger.assert_any_call('Data in \'%s\' is not compressed', 'foo.json')
```
chunyong-lin: The s3 object is compressed, in this try/except block, the debug message should not be triggered. 
```
+                 # The lookup data can optionally be compressed, so try to decompress
+                 # This will fall back and use the original data if decompression fails
+                try:
+                    data = zlib.decompress(data, 47)
+                except zlib.error:
+                    LOGGER.debug('Data in \'%s\' is not compressed', json_file)
```
Do I miss anything?

ryandeivert: @chunyong-lin there are two files (see your __init__) above - the `bar.json` is compress, the `foo.json` is not. The logger call we check for is `'Data in \'%s\' is not compressed', 'foo.json'`

chunyong-lin: Thanks for the explanation! STAMP!

PR #776, opened 2018-06-29
add tf generate, packaging, and deploy code for rule promotion function

ryandeivert:
to: @austinbyers 
size: medium

## Background

A [previous PR](https://github.com/airbnb/streamalert/pull/774) added the core of the new rule promotion function, but lacked the code necessary to set it up.

## Changes

* If a user tries to re-stage a rule that is already staged, the CLI will log a message indicating the rule is already staged and that the rule staging window will just be updated.
* Adding helper for updating the local rule table cache that can be used when remote state is changed and should be synced locally.
* Updating StatsPublisher to support user-configurable sns topic to send digest to
* Adding rule promotion lambda function terraform generation code.
* Adding all terraform iam/misc code for rule promotion function. This code includes permission for:
  * Executing athena queries and getting results
  * Scanning the rules dynamodb table and updating items
  * Publishing sns digest message to configured sns topic
* Updating deploy logic, including the CLI portion and the packaging logic, to support the rule promotion function.
* Reworking the config slightly to be more explicit that the settings relate to rule staging, not just generically to the 'rules table'.

## Testing

* Added unit tests for terraform generation.
* Added additional test for sns topic arn formatting.
* Will do an actual test deploy soon.


stream_alert_cli/terraform/generate.py
```diff
@@ -42,6 +42,7 @@
 )
 from stream_alert_cli.terraform.monitoring import generate_monitoring
 from stream_alert_cli.terraform.streamalert import generate_stream_alert
+from stream_alert_cli.terraform.rule_promotion import generate_rule_promotion
```
austinbyers: Misalphetization: `r < s`. I don't know why that sticks out to me so much, but it does

ryandeivert: fixed! thanks ;)

PR #777, opened 2018-06-29
allow for null/NoneType values during type conversion

ryandeivert:
to: @austinbyers 
size: small

## Background

Empty values should be allowed during type conversion.

## Changes

* Allowing for null/NoneType values during type conversion in classification.

## Testing

N/A


austinbyers:
PR 777! ✈️

PR #778, opened 2018-06-30
updating rule promo perms to allow reading athena data from s3

ryandeivert:
to: @chunyong-lin 
cc: @airbnb/streamalert-maintainers
size: small

## Background

The Rule Promotion lambda logged an error of:
`Query 61aa0909-9608-427a-9df3-c7def32f516f FAILED with reason Access Denied (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: 12345), exiting!`

## Changes

* Fixing s3 permission for reading athena data

## Testing

Updating unit test and deployed.

PR #779, opened 2018-07-02
updates to raise appropriate errors from athena client, etc

ryandeivert:
to: @austinbyers 
size: small

## Background

When I observed the error noted in #778, I realized athena errors could be handled/propagated better. For instance, errors such as that one should have resulted in a AWS Lambda error, but instead was just logged as an error.

## Changes

* Introducing an `AthenaQueryExecutionError` exception class that should be used to raise any errors that occur during the execution process.
* Updating all applicable areas where errors/exceptions that were previously just logged should probably bubble up and either cause and exception or be handled by the caller.
* Updating calling code to be more concise and handle exceptions where necessary.
* Removing some now unnecessary return result checks, since exceptions will be raised when the return result would be bad.
* Updates to docstrings to include new exception that could be raised.
* Adding 'falsey' method to `StagingStatistic` class to simplify some checks.

## Testing

Updating all unit tests to ensure the proper exception(s) get raised.

### DIFF
--- a/stream_alert/rule_promotion/statistic.py
+++ b/stream_alert/rule_promotion/statistic.py
@@ -18,6 +18,8 @@ limitations under the License.
 class StagingStatistic(object):
     """Store information on generated alerts."""
 
+    _ALERT_COUNT_UNKOWN = 'unknown'
+
     _COUNT_QUERY_TEMPLATE = ("SELECT '{rule_name}' AS rule_name, count(*) AS count "
                              "FROM alerts WHERE dt >= '{date}-{hour:02}' AND "
                              "rule_name = '{rule_name}'")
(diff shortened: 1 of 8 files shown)


stream_alert/rule_promotion/statistic.py
```diff
@@ -34,9 +36,15 @@ def __init__(self, staged_at, staged_until, current_time, rule):
         self._current_time = current_time
         self._staged_at = staged_at
         self.staged_until = staged_until
-        self.alert_count = 'unknown'
+        self.alert_count = self._ALERT_COUNT_UNKOWN
         self.execution_id = None
 
+    def __nonzero__(self):
+        return self.alert_count not in {0, self._ALERT_COUNT_UNKOWN}
```
austinbyers: Cool! I didn't know you could do that!

PR #781, opened 2018-07-02
updates to rule promotion function to use second schedule for sending digest

ryandeivert:
to: @austinbyers 
size: medium

## Background

The Rule Promotion lambda currently uses a somewhat hacky approach in order to determine when the staging statistics digest should be sent to SNS. This removes that logic.

## Changes

* Introducing a second CloudWatch events rule/target to invoke the Rule Promotion lambda on a rate/cron schedule.
* The new rule will invoke the lambda with a constant input of `{"send_digest": true}`, which tells the function to send the rule staging digest after doing the normal rule promotion flow.
* Removing the now unnecessary SSM parameter that was read/written to track the digest publishing.
* Updating all terraform for the above changes.

## Testing

Removing old tests for the code removal and updating existing ones.
Deployed in test account and functioning as expected.


ryandeivert:
thanks @austinbyers ! fixed pylint error, should be good now

PR #785, opened 2018-07-07
[app] increase app default request timeout

chunyong-lin:
to: @austinbyers 
size: small
resolves N/A

## Background
Increase app default request timeout to 5 seconds.

## Testing
Deployed in internally, and all Apps are working correctly.

### DIFF
--- a/app_integrations/apps/app_base.py
+++ b/app_integrations/apps/app_base.py
@@ -101,7 +101,7 @@ class AppIntegration(object):
     _POLL_BUFFER_MULTIPLIER = 1.5
     # _DEFAULT_REQUEST_TIMEOUT indicates how long the requests library will wait before timing
     # out for both get and post requests. This applies to both connection and read timeouts
-    _DEFAULT_REQUEST_TIMEOUT = 3.05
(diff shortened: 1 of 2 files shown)


austinbyers:
Can you provide any context for why 5 is better?

chunyong-lin:
During internal deploy of Slack Apps, the default timeout 3.05s causes connection to `https://slack.com/api/` to timeout. Simply increase the default timeout to 5s fixing the issue.

ryandeivert:
hey @chunyong-lin thanks for adding this! since this mostly seems to impacting Slack, can we just override the `_DEFAULT_REQUEST_TIMEOUT` class constant in the `SlackApp` subclass?

chunyong-lin:
Hey @ryandeivert Great suggestion! I made the change!
Per conversation with Ryan, it is better to overwrite the request timeout in app individually, rather than making change in the base app.

We will keep `3.05` seconds as default request timeout in base class `AppIntegration`. It is recommended to set the request timeout slightly larger than a multiple of 3 according to the doc [here](http://docs.python-requests.org/en/master/user/advanced/#timeouts).

PR #788, opened 2018-07-11
[core] allow req_subkeys with none and false values

### LINKED ISSUE #787: Required subkey check fails when the subkey exists, but the value is false
## Background

When using required subkeys in a rule, the subkey check fails if the subkey exists, but the value is `false`.

For example, in the test for critical_api_calls:
```
{
  "data":{
    "Records":[
      {
        ...
        "responseElements":{
          "storageEncrypted":false,
        }
        ...
      }
    ]
  },
  "description":"Deleting a database cluster (DeleteDBCluster) will create an alert.",
  "log":"cloudtrail:events",
  "service":"s3",
  "source":"prefix.cluster.sample.bucket",
  "trigger_rules":[
  ]
}
```
The subkey `storageEncrypted` will cause the rule not to run if included as a required subkey.

Using `stable`, but it looks like the issue would also be present on `master`.

### Description

The suspected piece of code is:
https://github.com/airbnb/streamalert/blob/6da3b52994ec4d4751245cdcf61ab23172de0670/stream_alert/rule_processor/rules_engine.py#L238

Because `record[key].get(x)` will return `false` if the key exists, but value is actually `false`. Then `all()` returns `false` for any values of `None` and `false`.

### Steps to Reproduce

1. Clone master
2. Verify all of the following tests pass
```
python manage.py lambda test -p rule -r cloudtrail_critical_api_calls
```
3. Modify the rule `cloudtrail_critical_api_calls` to include the following subkey, which does exist in the test, but is set to `false`.
```
@rule(logs=['cloudtrail:events'], req_subkeys={'responseElements':['storageEncrypted']})
def cloudtrail_critical_api_calls(rec):
```
4. Verify that all the tests fail, even though the test `Deleting a database cluster (DeleteDBCluster) will create an alert.` should have passed.
5. For extra credit, change the `storageEncrypted` value from `false` to `"abc"` and watch it pass

## Desired Change

A subkey with value `false` should still pass the req_subkey check.

stoggi:
size: small
resolves #787 Required subkey check fails when the subkey exists, but the value is false

## Background

Subkeys with False or None values should still cause the rule to be run

## Changes

* Required subkeys with none or false values should still cause the rule to be run.
* Using `any()` instead of `all()` should be faster, as `any()` can exit early if a subkey is missing.
* Check the type of the subkey item to make sure it is a dictionary
* Added `test_process_subkeys_false_none_value` to test `None` and `False` values for subkeys.
* Added `test_process_subkeys_non_dict` to make sure non dictionary subkey objects are handled.

## Testing

1. Created a new unit tests `test_process_subkeys_false_none_value` and `test_process_subkeys_non_dict`
2. Tests fail
3. Implemented fix in `rules_engine`
4. Tests pass

### DIFF
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -235,6 +235,11 @@ class RulesEngine(object):
                     'The required subkey %s is not found when trying to process %s: \n%s',
                     key, rule.name, record)
                 return False
+            if not isinstance(record.get(key), dict):
+                LOGGER.debug(
+                    'The required subkey %s is not a dictionary when trying to process %s: \n%s',
+                    key, rule.name, record)
+                return False
             if any(x not in record[key] for x in nested_keys):
                 return False
 
--- a/stream_alert/rule_processor/rules_engine.py
+++ b/stream_alert/rule_processor/rules_engine.py
@@ -235,7 +235,7 @@ class RulesEngine(object):
                     'The required subkey %s is not found when trying to process %s: \n%s',
                     key, rule.name, record)
                 return False
-            if not all(record[key].get(x) for x in nested_keys):
+            if any(x not in record[key] for x in nested_keys):
                 return False
 
         return True
--- a/tests/unit/stream_alert_rule_processor/test_rules_engine.py
+++ b/tests/unit/stream_alert_rule_processor/test_rules_engine.py
@@ -281,6 +281,57 @@ class TestRulesEngine(object):
         assert_equal(alerts[0].rule_name, 'web_server')
         assert_equal(alerts[1].rule_name, 'data_location')
 
+    def test_process_subkeys_false_none_value(self):
+        """Rules Engine - Req Subkeys with False and None values"""
+        @rule(logs=['test_log_type_json_nested'],
+              outputs=['s3:sample_bucket'],
+              req_subkeys={'data': ['value']})
+        def value_false(rec):  # pylint: disable=unused-variable
+            return rec['data']['value'] == False
+
+        @rule(logs=['test_log_type_json_nested'],
+              outputs=['s3:sample_bucket'],
+              req_subkeys={'data': ['value']})
+        def value_none(rec):  # pylint: disable=unused-variable
+            return rec['data']['value'] is None
+
+        kinesis_data_items = [
+            {
+                'date': 'Dec 01 2016',
+                'unixtime': '1483139547',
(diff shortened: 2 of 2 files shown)


ryandeivert:
hey @stoggi thanks for filing the issue and for the quick fix!!!

tests/unit/stream_alert_rule_processor/test_rules_engine.py
```diff
@@ -281,6 +281,57 @@ def web_server(rec):  # pylint: disable=unused-variable
         assert_equal(alerts[0].rule_name, 'web_server')
         assert_equal(alerts[1].rule_name, 'data_location')
 
+    def test_process_subkeys_false_none_value(self):
+        """Rules Engine - Req Subkeys with False and None values"""
+        @rule(logs=['test_log_type_json_nested'],
+              outputs=['s3:sample_bucket'],
+              req_subkeys={'data': ['value']})
+        def value_false(rec):  # pylint: disable=unused-variable
+            return rec['data']['value'] == False
```
ryandeivert: it looks like this is the line that is causing CI (pylint) to fail - changing to the following should fix:

`return rec['data']['value'] is False`

stream_alert/rule_processor/rules_engine.py
```diff
@@ -235,7 +235,7 @@ def process_subkeys(record, payload_type, rule):
                     'The required subkey %s is not found when trying to process %s: \n%s',
                     key, rule.name, record)
                 return False
-            if not all(record[key].get(x) for x in nested_keys):
+            if any(x not in record[key] for x in nested_keys):
```
ryandeivert: I meant to respond on the issue you filed, but this is also be my preferred fix to this 👍

tests/unit/stream_alert_rule_processor/test_rules_engine.py
stoggi: Ah, should have run pylint locally. Cheers, will fix this.

stream_alert/rule_processor/rules_engine.py
ryandeivert: I'm also just realizing that we don't do anything to check that `record[key]` is a dict 🤔 I suppose it should be up to the rule-writer to know that keys in their log will reference dicts (objects) and not use, say, a top level key that references some other type. that said, it might not hurt to add that safety check here..

ryandeivert:
thanks @stoggi for the quick fix!

PR #792, opened 2018-07-25
[apps] Aliyun ActionTrail app

GarretReece:
cc: <optional-cc-to-specific-users>
size: medium

## Background

SteamAlert doesn't have an app that collects ActionTrail events from Aliyun

## Changes

* Adds an app that collects ActionTrail events from Aliyun
* Schema definition for the new log type is included
* Unit tests for the app as well as schema validation tests are included
* Documentation for authorizing the app is included in the README in the apps directory as well as in a section of the app-configuration.rst in the docs directory

## Testing

Barebones StreamAlert session created, the Aliyun app configured and deployed, and verifying that events are listed as captured in the log

### DIFF
--- a/app_integrations/apps/aliyun.py
+++ b/app_integrations/apps/aliyun.py
@@ -45,7 +45,7 @@ class AliyunApp(AppIntegration):
         https://www.alibabacloud.com/help/doc-detail/28849.htm
     """
 
-    _PAGE_SIZE = 50
+    _MAX_RESULTS = 50
 
     def __init__(self, config):
         super(AliyunApp, self).__init__(config)
@@ -60,21 +60,21 @@ class AliyunApp(AppIntegration):
         return 'aliyun'
 
 
-    #@classmethod
-    #def date_formatter(cls):
-    #    """Return a format string for a date, ie: 2014-05-26T12:00:00Z
+    @classmethod
+    def date_formatter(cls):
+        """Return a format string for a date, ie: 2014-05-26T12:00:00Z
 
-    #    This format is consisten with the format used by the Aliyun API:
-    #        https://www.alibabacloud.com/help/doc-detail/28849.htm
-    #    """
-    #    return '%y-%m-%dT%h:%m:%sZ'
+        This format is consisten with the format used by the Aliyun API:
+            https://www.alibabacloud.com/help/doc-detail/28849.htm
+        """
+        return '%Y-%m-%dT%H:%M:%SZ'
 
     def _gather_logs(self):
         auth = self._config.auth
         client = AcsClient(auth['access_key_id'], auth['access_key_secret'], auth['region_id'])
 
         request = LookupEventsRequest.LookupEventsRequest()
-        request.set_MaxResults(self._PAGE_SIZE)
+        request.set_MaxResults(self._MAX_RESULTS)
         request.set_StartTime = self._last_timestamp
         if self._next_token:
             request.set_NextToken(self._next_token)
@@ -101,19 +101,33 @@ class AliyunApp(AppIntegration):
     @classmethod
     def _required_auth_info(cls):
         """Required credentials for access to the resources"""
-        #TODO: support more regions, and lay out documentation for which regions are supported.
+
+        def region_validator(region):
+            """Region names pulled from https://www.alibabacloud.com/help/doc-detail/40654.htm"""
+
+            if region in ["cn-qingdao", "cn-beijing", "cn-zhangjiakou", "cn-huhehaote",
+                          "cn-hangzhou", "cn-shanghai", "cn-shenzhen", "cn-hongkong",
+                          "ap-southeast-1", "ap-southeast-2", "ap-southeast-3", "ap-southeast-5",
+                          "ap-northeast-1", "ap-south-1", "us-west-1", "us-east-1",
+                          "eu-central-1", "me-east-1"]:
+                return region
+            return False
+
         return {
             'access_key_id': {
-                'description': ('The access key id generated for a RAM user.'),
+                'description': ('The access key id generated for a RAM user. This '
+                                'should be a string of alphanumeric characters.'),
                 'format': re.compile(r'.*')
                 },
             'access_key_secret': {
-                'description': ('The access key secret generated for a RAM user.'),
+                'description': ('The access key secret generated for a RAM user. This '
+                                'should be a string of alphanumeric characters.'),
                 'format': re.compile(r'.*')
                 },
             'region_id': {
-                'description': ('The region identifier to collect data from.'),
-                'format': lambda x: x == 'ap-northeast-1'
+                'description': ('The region for the Aliyun API. This should be '
+                                'a string like \'ap-northeast-1\'.'),
+                'format': region_validator
                 },
             }
 
--- a/app_integrations/apps/aliyun.py
+++ b/app_integrations/apps/aliyun.py
@@ -1,0 +1,134 @@
+"""
+Copyright 2018-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import json
+import re
+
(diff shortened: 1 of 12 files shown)


app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import json
+import re
+
+from aliyunsdkcore.client import AcsClient
+from aliyunsdkcore.acs_exception.exceptions import ClientException
+from aliyunsdkcore.acs_exception.exceptions import ServerException
```
ryandeivert: can you group this exception with the import above pls

app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+from aliyunsdkactiontrail.request.v20171204 import LookupEventsRequest
+
+from app_integrations import LOGGER
+from app_integrations.apps.app_base import StreamAlertApp, AppIntegration
+
```
ryandeivert: add an extra new line here please (2 lines between imports and class declarations)

app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+        """Return a format string for a date, ie: 2014-05-26T12:00:00Z
+
+        This format is consisten with the format used by the Aliyun API:
+            https://www.alibabacloud.com/help/doc-detail/28849.htm
+        """
+        return '%Y-%m-%dT%H:%M:%SZ'
+
+    def _gather_logs(self):
+        auth = self._config.auth
+        client = AcsClient(auth['access_key_id'], auth['access_key_secret'], auth['region_id'])
```
ryandeivert: the `_gather_logs` method could be called a bunch of times in an execution - therefore, would it make sense to have the `AcsClient` created upon `__init__` and cached as an instance property?

app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+            if 'NextToken' in json_response:
+                self._more_to_poll = True
+                self._next_token = json_response['NextToken']
+            else:
+                self._more_to_poll = False
+                self._last_timestamp = json_response['EndTime']
+
+            return json_response['Events']
+
+        except ServerException as e:
```
ryandeivert: These two `except` blocks could be consolidated into one with:

```
except (ServerException, ClientException) as err:
    LOGGER.exception("%s exception occurred", err.get_error_type())
    return False
```

also, there's no need to include the actual content of the exception (with `str(e)`) when using `LOGGER.exception` - this call handles all of that for you.

app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+            return False
+
+    @classmethod
+    def _required_auth_info(cls):
+        """Required credentials for access to the resources"""
+
+        def region_validator(region):
+            """Region names pulled from https://www.alibabacloud.com/help/doc-detail/40654.htm"""
+
+            if region in ["cn-qingdao", "cn-beijing", "cn-zhangjiakou", "cn-huhehaote",
```
ryandeivert: please use single (`'`) quotes on string literals and not double (`"`) quotes

app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+                          "ap-northeast-1", "ap-south-1", "us-west-1", "us-east-1",
+                          "eu-central-1", "me-east-1"]:
+                return region
+            return False
+
+        return {
+            'access_key_id': {
+                'description': ('The access key id generated for a RAM user. This '
+                                'should be a string of alphanumeric characters.'),
+                'format': re.compile(r'.*')
```
ryandeivert: are there no restrictions similar to AWS for their access and secret keys? ie - AWS access keys are 20 alphanumeric all-caps characters and must start with `A`, etc

app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+                },
+            'access_key_secret': {
+                'description': ('The access key secret generated for a RAM user. This '
+                                'should be a string of alphanumeric characters.'),
+                'format': re.compile(r'.*')
+                },
+            'region_id': {
+                'description': ('The region for the Aliyun API. This should be '
+                                'a string like \'ap-northeast-1\'.'),
+                'format': region_validator
```
ryandeivert: ++ this is great

app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+        limit is the general limit on Aliyun API requests, which is no more than
+        100 per second. We can set this value to 0 safely.
+
+        Resource:
+            https://www.alibabacloud.com/help/doc-detail/29474.htm
+
+        Returns:
+            int: Number of seconds the polling function should sleep for
+        """
+        return 0
```
ryandeivert: I think if this is 0 it can be left off completely. I do like the context you've added in the docstring, though, so I'm not sure if I feel strongly about removing the method

stream_alert_cli/manage_lambda/package.py
```diff
@@ -203,6 +203,8 @@ class AppIntegrationPackage(LambdaPackage):
     package_name = 'stream_alert_app'
     precompiled_libs = {'boxsdk[jwt]==2.0.0a11'}
     third_party_libs = {
+        'aliyun-python-sdk-core==2.8.5',
```
ryandeivert: this is great - thank you for pinning the version!

app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+
+    @classmethod
+    def _type(cls):
+        return 'events'
+
+    @classmethod
+    def service(cls):
+        return 'aliyun'
+
+
```
ryandeivert: Looks like there's an extra new line here

requirements.txt
```diff
@@ -1,3 +1,5 @@
+aliyun-python-sdk-core==2.8.5
+aliyun-python-sdk-actiontrail==2.0.0
```
chunyong-lin: Can you add these two packages to `requirements-top-level.txt` as well?

app_integrations/apps/aliyun.py
chunyong-lin: Small request, can we use `set` instead of `list`, it is not big a deal though.

app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+        This format is consisten with the format used by the Aliyun API:
+            https://www.alibabacloud.com/help/doc-detail/28849.htm
+        """
+        return '%Y-%m-%dT%H:%M:%SZ'
+
+    def _gather_logs(self):
+        auth = self._config.auth
+        client = AcsClient(auth['access_key_id'], auth['access_key_secret'], auth['region_id'])
+
+        request = LookupEventsRequest.LookupEventsRequest()
```
chunyong-lin: I think we also can cache `request` to `__init__`.

conf/logs.json
```diff
@@ -1666,5 +1666,40 @@
         "reason"
       ]
     }
+  },
+  "aliyun:events": {
```
chunyong-lin: Remind to update schema name if you agree with me renaming `_type` to `actiontrail`.

app_integrations/apps/aliyun.py
GarretReece: I can't find any documentation anywhere regarding the format. The values I have are a mix of digits and upper and lower case letters, with the id 16 characters in length and the secret 30 characters in length. I'm perfectly willing to put those in as restrictions, but I don't have any reference for that being the supported format.

app_integrations/apps/aliyun.py
```diff
@@ -0,0 +1,148 @@
...
+      "Events": [entries],
+    }
+
+    If there are no more events in the queried range, the 'NextToken' element is not present.
+
+    Resource:
+        https://www.alibabacloud.com/help/doc-detail/28849.htm
+    """
+
+    _MAX_RESULTS = 50
```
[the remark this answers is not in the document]
GarretReece: The documentation lists the range for the MaxResult parameter as 0-50; I'll admit I haven't tested with higher values to see if it worked anyhow.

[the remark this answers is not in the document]
chunyong-lin: Can I have one testing request? Can you add a test case when `NextToken` is available in your testing environment? (I am not talking about unit test)

A while ago, I have a simple script to retrieve ActionTrail events, and set max results per request to `2`. I encountered the issue that `NextToken` doesn't take effect, and the 2nd request returned as same events as 1st request.
Can you verify this in the testing environment? Thanks!

PR #794, opened 2018-07-27
V2 Doc Overhaul: Part 1 - Intro

austinbyers:
to: @ryandeivert and @chunyong-lin 
size: medium

## Background

As we get ready to release StreamAlert V2 🙌, the documentation needs some serious updating. Each PR will focus on a specific section; this one focuses on the landing page and the introduction.

## Changes

* Landing Page (`index.html`)
    - Replaced the high-level schematic with one that is more accurate and easier to read
    - Rewrote the list of features (which replaces the `overview` and `benefits` pages)
* Architecture (new page)
    - Added a more complete architecture diagram and a step-by-step overview of the StreamAlert data lifecycle
* Getting Started
    - Migrated details from the `requirements` page
    - Added steps for testing StreamAlert with an example after it is deployed. The goal is that the Getting Started page is a 1-page tutorial for a brand-new user to go from nothing to seeing an alert generated by their StreamAlert deployment.
* Removed the `roadmap` page, which just points to GitHub milestones. We already link to GitHub elsewhere for issues and source code, I don't think this needs its own page of documentation.
* Disable Kinesis and CloudTrail by default. These are expensive to run, and generate lots of error messages in CloudWatch logs in a fresh deployment that may make it harder to understand what is happening for first-time users.

## Testing

* Built the documentation locally: `sphinx-build -W docs/source docs/build` (this is also how I recommend the reviewers to look at it)
* I tested the steps in "Getting Started" from start to finish with a new deployment


ryandeivert:
this is great - thank you for leading this charge @austinbyers !!! I'll get your branch checked and and build locally to review

docs/source/index.rst
```diff
@@ -5,40 +5,40 @@ StreamAlert
   :align: center
   :alt: StreamAlert
 
-StreamAlert is a serverless, realtime data analysis framework which empowers you to ingest, analyze, and alert on data from any environment, using datasources and alerting logic you define.
+StreamAlert is a serverless, real-time data analysis framework which empowers you to ingest, analyze, and alert on data from any environment, using data sources and alerting logic you define.
+Security teams use StreamAlert to scan terabytes of log data every day for proactive threat detection and prevention.
```
ryandeivert: I'd caution against saying `and prevention` here. SA doesn't do any prevention, and acts solely as an intrusion detection system (IDS) vs intrusion prevention system (IPS). you could, however, say `detection and response`

docs/source/index.rst
```diff
@@ -5,40 +5,40 @@ StreamAlert
...
-Other Links
------------
+Features
+--------
 
-* `Github <https://github.com/airbnb/streamalert>`_
-* `Slack (unofficial) <https://streamalert.herokuapp.com>`_
+* Rules are written in Python; they can utilize any Python libraries or functions
+* Ingested logs and generated alerts can be retroactively searched for compliance and research
+* Serverless design is cheaper and easier to maintain and scales to terabytes per day
```
ryandeivert: please add a comma here to break up this sentence a bit:

`Serverless design is cheaper and easier to maintain, and scales to terabytes per day`

docs/source/index.rst
```diff
@@ -5,40 +5,40 @@ StreamAlert
...
+* Deployment is automated: simple, safe and repeatable for any AWS account
+* Secure by design: least-privilege execution, containerized analysis, and encrypted data storage
+* Merge similar alerts and automatically promote new rules if they are not too noisy
+* Includes built-in support for dozens of log types and schemas
```
ryandeivert: Drop the `includes ` here and just go with `Built-in support...` could probably do the same on the following line

docs/source/index.rst
```diff
@@ -5,40 +5,40 @@ StreamAlert
...
+* Includes a collection of broadly applicable community rules
+* Fully open source and customizable: add your own log schemas, rules, and alert outputs
 
-.. note:: Docs are under construction, don't mind the dust!
+Resources
```
ryandeivert: Can we maybe throw a link to the `Getting started/quick start` here so it's on the landing page?

docs/source/getting-started.rst
```diff
@@ -1,48 +1,246 @@
...
+4. Run unit tests to make sure everything is installed correctly:
+
+.. code-block:: bash
+
+  tests/scripts/unit_tests.sh
+
+Set AWS Credentials
+-------------------
+
+1. Create an AWS account and an IAM user with permissions for at least the following services:
```
ryandeivert: can we link to AWS docs on this? ie: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_create.html

docs/source/getting-started.rst
```diff
@@ -1,48 +1,246 @@
...
+.. code-block:: bash
+
+  aws s3 ls | grep streamalert
+
+You can also login to the AWS web console and see StreamAlert's CloudWatch logs, Lambda functions, etc.
+
+Live Test
+---------
+Now let's upload some data and trigger an alert to see StreamAlert in action! Our example uses
+SNS for both sending the log data and receiving the alert but StreamAlert also supports many other
```
ryandeivert: please add a comma before the `but` here

docs/source/getting-started.rst
```diff
@@ -1,48 +1,246 @@
...
+    "time": "now",
+    "version": "2018"
+  }
+
+This is only a rough approximation of what the real log might look like, but good enough for our purposes.
+Then send it off to the data SNS topic:
+
+.. code-block:: bash
+
+  aws sns publish --topic-arn arn:aws:sns:REGION:ACCOUNTID:streamalert-test-data --message "`cat cloudtrail-root.json`"
```
ryandeivert: if you didn't want to use back ticks here, you could use command expansion with `$` like:

`... --message "$(cat cloudtrail-root.json)"`

not necessary, just a suggestion :)

docs/source/getting-started.rst
```diff
@@ -1,48 +1,246 @@
...
+`data sources <datasources.html>`_ and `alert outputs <outputs.html>`_.
+
+1. Create 2 SNS topics:
+
+.. code-block:: bash
+
+  aws sns create-topic --name streamalert-test-data
+  aws sns create-topic --name streamalert-test-alerts
+
+2. Subscribe your email to the alerts SNS topic (replacing REGION, ACCOUNTID, and EMAIL with your own values):
```
ryandeivert: Suggestion: You could have them export these values into their shell env and then use then env var here (and in other commands), so they could just copy/paste and not reason about replacing values

docs/source/getting-started.rst
```diff
@@ -1,48 +1,246 @@
...
+
+This is only a rough approximation of what the real log might look like, but good enough for our purposes.
+Then send it off to the data SNS topic:
+
+.. code-block:: bash
+
+  aws sns publish --topic-arn arn:aws:sns:REGION:ACCOUNTID:streamalert-test-data --message "`cat cloudtrail-root.json`"
+
+If all goes well, an alert should arrive in your inbox within a few minutes!
+If not, look for any errors in the CloudWatch logs for the StreamAlert Lambda functions.
```
ryandeivert: capitalize the L on logs here since it's a service name

docs/source/getting-started.rst
```diff
@@ -1,48 +1,246 @@
...
 
-1. Define your `AWS account <account.html>`_
-2. Define your `clusters <clusters.html>`_
-3. Define your `datasources <conf-datasources.html>`_
-4. Define your `schemas <conf-schemas.html>`_
-5. Configure your `kinesis stream/firehose <kinesis.html>`_
-6. Configure your `lambda settings <lambda.html>`_
-7. Write your `rules <rules.html>`_
-8. Configure your `outputs <outputs.html#configuration>`_
+(Be sure to replace PREFIX with your own name prefix.) If no records are returned, look for errors
```
ryandeivert: I know this is annoying for you to update your screenshot - but the `prefix_streamalert` is actually not required as long as the user has the right database selected in the `Database` dropdown:

image

docs/source/getting-started.rst
```diff
@@ -1,48 +1,246 @@
...
+in the ``athena_partition_refresh`` function or try invoking it directly.
 
-Now it's time to `deploy <deployment.html>`_!
+And there you have it! Ingested log data is parsed, classified, and scanned by the StreamAlert rules
+engine and the resulting alert is delivered to your configured output within a matter of minutes.
```
ryandeivert: `output` --> `output(s)`

docs/source/index.rst
```diff
@@ -49,30 +49,16 @@ Table of Contents
...
-.. _getting-started:
-
-.. toctree::
-   :maxdepth: 2
-   :caption: Getting Started
-
    getting-started
-   deployment
-   deploy-rollback
+   architecture
```
ryandeivert: This structure is way better - thank you!! 🙏 🙌

docs/source/index.rst
```diff
@@ -5,40 +5,40 @@ StreamAlert
...
-* Supports different use-cases like security, infrastructure, compliance and more
-
-Components
-----------
+Overview
+--------
+Incoming log data is classified and scanned by the StreamAlert rules engine running in your AWS account.
+Rule matches are reported to one or more alert outputs:
 
 .. figure:: ../images/sa-high-level-arch.png
```
chunyong-lin: Can not add comment to the binary file, so I add it here.
* I would prefer simpler diagram, replacing `Alert Delivery` with `Output` and only includes PagerDuty, Slack, Amazon S3, AWS SNS.
* It is high-level architecture, including CloudWatch logs, kinesis firehose SNS, SQS is too detailed.

docs/source/getting-started.rst
```diff
@@ -1,48 +1,246 @@
 Getting Started
 ===============
 
-Perform the following steps on your laptop or development machine
+It only takes a few minutes to get StreamAlert up and running!
 
-Dependencies
-------------
+Install Dependencies
```
chunyong-lin: It is implied that these commands are tested in mac os. It would be nice if we add a comment here.

docs/source/getting-started.rst
```diff
@@ -1,48 +1,246 @@
...
+
+  pip install -r requirements.txt
+
+4. Run unit tests to make sure everything is installed correctly:
+
+.. code-block:: bash
+
+  tests/scripts/unit_tests.sh
+
+Set AWS Credentials
```
chunyong-lin: `Setup AWS Credentials`?

docs/source/architecture.rst
```diff
@@ -0,0 +1,37 @@
...
+.. figure:: ../images/sa-complete-arch.png
+  :alt: StreamAlert Architecture
+  :align: center
+  :target: _images/sa-complete-arch.png
+
+Data Lifecycle
+--------------
+
+1. Log data can come through any of the supported `data sources <datasources.html>`_.
+This includes Kinesis, S3, SNS, or using a `StreamAlert App <app-configuration.html>`_ to periodically
```
chunyong-lin: I would use `StreamAlert App Framework`.

docs/source/getting-started.rst
```diff
@@ -1,48 +1,246 @@
...
+3. At this point, StreamAlert is up and running! You can, for example, see the S3 buckets
+that were automatically created:
+
+.. code-block:: bash
+
+  aws s3 ls | grep streamalert
+
+You can also login to the AWS web console and see StreamAlert's CloudWatch logs, Lambda functions, etc.
+
+Live Test
```
chunyong-lin: 👍  This is awesome! Using SNS to test StreamAlert is very convenient and easy. Thank you for adding this!

docs/source/getting-started.rst
austinbyers: Great idea, will do. This doesn't work when modifying the config files, but would make the bash commands work without modification.

docs/source/getting-started.rst
austinbyers: Right, selecting the database from the dropdown auto-populates the query field with the database name. I'll add that to the screenshot

docs/source/getting-started.rst
austinbyers: Wasn't sure whether this sentence should be generic or specific to the test example. I'll go with your suggestion (generic) and reword slightly

docs/source/index.rst
austinbyers: I like listing at least all of the AWS alert targets because it demonstrates just how powerful the alerting mechanism can be. Someone looking at StreamAlert for the first time can immediately see that they could use alerts to trigger their own Lambda functions, feed their own analysis SQS queues or Firehoses, etc.

I like `Alert Delivery` instead of `Output` because the connotation of an *alert* is something anomalous or noteworthy, as opposed to an "output", which is just the result of an execution. Most StreamAlert invocations will not generate any alerts, so calling it an "output" feels like a misnomer to me.

But I'm happy to discuss further. @ryandeivert what are your thoughts on the diagram?

docs/source/getting-started.rst
austinbyers: Good idea

docs/source/getting-started.rst
austinbyers: Sure, changed it to `Configure AWS Credentials`

docs/source/architecture.rst
austinbyers: My understanding is that an `App` is the actual Lambda function which polls data (which I reference here), and the `App Framework` is the set of base classes that you can use to write your own `App`

docs/source/getting-started.rst
austinbyers: No problem! Thanks for the review

docs/source/index.rst
austinbyers: sa-high-level-arch

docs/source/architecture.rst
ryandeivert: agreed with Austin here, `Framework` is what they're built around, not the thing itself

docs/source/conf.py
```diff
@@ -62,9 +62,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = u'1.0'
+version = u'2.0'
```
ryandeivert: @austinbyers can this pull from https://github.com/airbnb/streamalert/blob/f803613f839ac2b97284481ca451fcb6b27c73ea/stream_alert/__init__.py (also can you bump this version?)

docs/source/getting-started.rst
ryandeivert: I don't think it auto-populates anything?? you can just query like `select * from alerts` when you have that selected

docs/source/conf.py
austinbyers: I'll go ahead and update that version, but I wasn't able to automatically pull it into the docs:

```python
# docs/source/conf.py
from stream_alert import __version__ as streamalert_version
# ...
version = streamalert_version
```

broke doc building:
```
$ sphinx-build -W docs/source docs/build
Running Sphinx v1.7.5

Configuration error:
There is a programable error in your configuration file:

  File ".../venv/lib/python2.7/site-packages/sphinx/config.py", line 161, in __init__
    execfile_(filename, config)
  File ".../venv/lib/python2.7/site-packages/sphinx/util/pycompat.py", line 150, in execfile_
    exec_(code, _globals)
  File ".../venv/lib/python2.7/site-packages/six.py", line 709, in exec_
    exec("""exec _code_ in _globs_, _locs_""")
  File "<string>", line 1, in <module>
  File "conf.py", line 23, in <module>
    from stream_alert import __version__ as streamalert_version
ImportError: No module named stream_alert
```

I also tried adding `__init__.py` everywhere to make the `docs` tree a Python package, same error

austinbyers:
@chunyong-lin @ryandeivert awesome feedback, thanks for the thorough review! I've addressed it, but since there were lots of changes and comments, I'll wait until at least one of you approves again.

Note that I removed even more unnecessary stuff from the default cluster, stripping it down to almost the bare minimum (cloudwatch metrics, rule processor)

docs/source/conf.py
ryandeivert: sorry I should have added more context on how to achieve this - you shouldn't actually import the module, use you should use a method similar to what is recommended by `setuptools`: 

https://github.com/ThreatResponse/margaritashotgun/blob/master/setup.py#L12-L16

for reference, see # 1 here: https://packaging.python.org/guides/single-sourcing-package-version/

docs/source/getting-started.rst
```diff
@@ -121,17 +126,25 @@ SNS for both sending the log data and receiving the alert but StreamAlert also s
   aws sns create-topic --name streamalert-test-data
   aws sns create-topic --name streamalert-test-alerts
 
-2. Subscribe your email to the alerts SNS topic (replacing REGION, ACCOUNTID, and EMAIL with your own values):
+2. Export some environment variables for easy re-use later:
+
+.. code-block:: bash
+
+  export SA_REGION=us-east-1        # StreamAlert deployment region
+  export SA_ACCOUNT=123412341234    # AWS account ID
```
chunyong-lin: Do you change the account ID to `111111111111` which is used as example at line 93.

docs/source/index.rst
ryandeivert: I'm fairly indifferent on this, but do see how showing the AWS ones is helpful, and slightly prefer alert delivery to output. that said, the amount of alert output services is going to continue grow, so trying to include all is going to be hard over time. ideally, we'd highlight 4 that serve different purposes:

1) Slack - chatops style message
2) PD - case/incident management
3) S3 - historical retention
4) SQS - downstream tooling

just my 2 cents

ryandeivert:
@austinbyers back to the versioning thing - this should work and will ensure we're in constant version parity with docs/code:

in `docs/source/conf.py`:
```python
# The short X.Y version.
version = re.search(
    r"^__version__ = ['\"]([^'\"]*)['\"]",
    open('../../stream_alert/__init__.py', 'r').read(),
    re.MULTILINE
).group(1)

# The full version, including alpha/beta/rc tags.
release = version
```

ryandeivert:
great changes, still LGTM @austinbyers :)

PR #795, opened 2018-07-30
[apps] slight refactor of apps config and migration to moto for SSM mocking

ryandeivert:
size: large

## Background

The apps functions utilize SSM for 3 different parameters right now: one param each for `state`, `auth` and `config` information. Since the Lambda function is invoked on a schedule, we can leverage [CloudWatch Event targets](https://www.terraform.io/docs/providers/aws/r/cloudwatch_event_target.html), with an 'input' value. This will reduce the number of ssm param resources we need to maintain by simply feeding the 'config' param value into the Lambda as the input event each time it is invoked.

## Changes

* Adding support to the `tf_lambda` terraform module for passing event to Lambda using CloudWatch event target.
* Updating all packaging and terraform generate code for the above changes.
* Refactor app config to use Lambda input event instead of SSM parameter.
  * This also reduces the complexity of this class a bit.
* The `AppIntegration` base app class has been refactored to fully encapsulate the loading of the config/app, as well as the clean exit logic (`try: ... finally: ...`).
  * This also includes better timing/looping logic.
  * Updating all app subclasses to conform to this change.
* Updating apps Lambda handler entry point for new config/app loading logic.
* Removing the Terraform resource for the `config` parameter.
  * Migrating `iam.tf` to just be `main.tf`.
  * Simplifying tf logic, removing some unneeded variables.
  * Updates to README.md

## Other Changes

* Updating the ThreatIntel test to leverage `moto`'s new-ish support for SSM and removing our `MockSSMClient` class entirely.

## Next Steps

* Migrating to the `stream_alert` package

## Testing

* New and updated unit tests
* Deployed in testing account to ensure all is working as expected.


app_integrations/apps/__init__.py
```diff
@@ -2,6 +2,45 @@
...
+    def get_app(cls, app_type):
+        """Return the proper app integration class for this service
+
+        Args:
+            config (AppConfig): Loaded configuration with service, etc
+            init (bool): Whether or not this class should be instantiated with
+                the config that has been passed in
+
+        Returns:
+            AppIntegration: Subclass of AppIntegration corresponding to the config
```
austinbyers: Can you also add a `Raises` section to the docstring, which we use when explicitly raising an exception

ryandeivert: @austinbyers done! thank you :)

ryandeivert:
@austinbyers also added a commit to address [CVE-2018-10903](https://nvd.nist.gov/vuln/detail/CVE-2018-10903)

app_integrations/apps/app_base.py
```diff
@@ -259,11 +233,11 @@ def _finalize(self):
                          'due to the subclass not setting this value.')
 
         if self._last_timestamp == self._config.start_last_timestamp:
-            LOGGER.error('Ending last timestamp is the same as the beginning last timestamp. '
-                         'This could occur if there were no logs collected for this execution.')
+            LOGGER.warning('Ending last timestamp is the same as the beginning last timestamp. '
```
chunyong-lin: I don't think this is considered to be an abnormal behavior. I would use `info` logging level.

austinbyers:
Good call, we'll need to make that change for BinaryAlert as well

app_integrations/config.py
```diff
@@ -95,101 +110,54 @@ def remaining_ms():
...
-        return {'type', 'app_name', 'prefix', 'cluster', 'schedule_expression'}
-
-    @classmethod
-    def _state_keys(cls):
-        return {cls._STATE_KEY, cls._TIME_KEY}
+        return {'app_type', 'destination_function_name', 'schedule_expression'}
 
     @classmethod
-    def load_config(cls, context, event):
+    def load_config(cls, event, context):
```
chunyong-lin: Can you add doc string for `event` arg?

requirements.txt
```diff
@@ -2,14 +2,14 @@ aliyun-python-sdk-core==2.8.5
...
+backoff==1.6.0
 bandit==1.4.0
-boto3==1.7.48
+boto3==1.7.66
 boxsdk==2.0.0a11
 cbapi==1.3.6
 coverage==4.5.1
 coveralls==1.3.0
-cryptography==2.2.2
+cryptography==2.3
```
chunyong-lin: Thanks for updating this! 👍

app_integrations/apps/app_base.py
ryandeivert: thanks - I will change it!

PR #797, opened 2018-08-02
[apps] migrate apps package to be within top level stream_alert package

ryandeivert:
to: @austinbyers 
size: medium

## Background

The `app_integration` lambda function package is currently outside of the root `stream_alert` package. This was initially done with the thought that it wouldn't be tightly coupled to the streamalert code, but this has since become moot.

## Changes

* Migrating the root `app_integrations` directory to `stream_alert/apps`.
* Updating packaging logic for new folder structure and lambda handler path.
* Changing import logic for apps to make implementing a new app a bit simpler.
* Removing most references to `app_integration` in favor of just `app` (including in the CLI).

## Testing

* Updating all unit tests imports for new path.
* Noticed a slow down in tests after migration, and dug deeper to figure out it was related to the combination of testing the `tests/unit/stream_alert_alert_processor` and `tests/unit/stream_alert_apps` packages at the same time. I discovered the `mock_kms` and `mock_s3` mocking was causing issues for some reason, so I moved these mockers into the setup/teardown.


ryandeivert:
@chunyong-lin PTAL

manage.py
```diff
@@ -31,11 +31,11 @@
 
 from stream_alert import __version__ as version
 from stream_alert.alert_processor.outputs.output_base import StreamAlertOutput
+from stream_alert.apps import StreamAlertApp
```
austinbyers: This does look a lot better! Great idea

stream_alert/apps/_apps/aliyun.py
```diff
@@ -20,9 +20,7 @@
 from aliyunsdkcore.acs_exception.exceptions import ServerException, ClientException
 from aliyunsdkactiontrail.request.v20171204 import LookupEventsRequest
 
-from app_integrations import LOGGER
-from app_integrations.apps import StreamAlertApp
-from app_integrations.apps.app_base import AppIntegration
+from . import AppIntegration, LOGGER, StreamAlertApp
```
austinbyers: 🙌 we haven't used relative imports much before, way to be a trailblazer

stream_alert/apps/_apps/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Import some package level items to make implementing subclasses a bit nicer"""
```
austinbyers: The path `stream_alerts/apps/_apps` seems potentially confusing. The `_apps` folder is for actual app instantiations, right? Maybe `stream_alert/apps/_integrations`?

It does distinguish itself with the underscore, so I'm happy with whatever you think is best

tests/unit/stream_alert_alert_processor/test_outputs/test_carbonblack.py
```diff
@@ -50,11 +48,20 @@ class TestCarbonBlackOutput(object):
 
     def setup(self):
         """Setup before each method"""
+        self._mock_s3 = mock_s3()
+        self._mock_s3.start()
+        self._mock_kms = mock_kms()
+        self._mock_kms.start()
```
austinbyers: I've had to do this before as well, not sure why. It does seem to be the more stable solution

PR #798, opened 2018-08-02
[rule-staging] only creating rules table if rule staging is enabled

ryandeivert:
to: @austinbyers 
size: small

## Background

If the rule staging feature is not enabled, the rules dynamo table does not need to be created.

## Changes

* Adding a conditional to only create rules table if staging is enabled.
* Updating other dependent resources.
* Adding check to only allow the rule promotion lambda to be created if the rule_staging feature is enabled

## Testing

* Performed a `python manage.py terraform build` with setting enabled and disabled, and ensured the right resources would be created.
* Added unit test to ensure rule promo lambda module is not created.


ryandeivert:
@austinbyers thanks for the feedback/advice - PTAL, I added a check and updated tests

PR #801, opened 2018-08-06
[docs] rule staging and promo documentation

ryandeivert:
to: @austinbyers 
size: medium

## Background

There is not documentation for the v2.0.0 feature of rule staging/promotion.

## Changes

* Renaming `rule-table` subcommand to `rule-staging`
  * Adding new command for enabling/disabling rule staging feature
* Adding documentation for the rule staging and rule promotion features.

## Testing

Built and tested docs locally. I advised doing the same when reviewing :)


docs/source/rule-staging.rst
```diff
@@ -0,0 +1,109 @@
...
+===========================  ========  =======  ===========
+Key                          Required  Default  Description
+---------------------------  --------  -------  -----------
+``cache_refresh_minutes``    ``No``    ``10``   Maximum amount of time (in minutes) the Rule Processor
+                                                should wait to force refresh the rule staging information.
+``table.read_capacity``      ``Yes``   ``20``   DynamoDB read capacity to allocate to the table that stores staging
+                                                information. The default setting should be sufficient in most use cases.
+``table.write_capacity``     ``Yes``   ``5``    DynamoDB write capacity to allocate to the table that stores staging
+                                                information. The default setting should be sufficient in most use cases.
+===========================  ========  =======  ===========
```
austinbyers: I've actually been getting rid of the `Required` column. If a value has a default, then it isn't required

docs/source/rule-staging.rst
```diff
@@ -0,0 +1,109 @@
+Rule Staging
+============
+
+Rule staging allows dynamic toggling of rules from a staging to production state, and vice versa.
+Staged rules will only send alerts for historical retention, and the alerts will not be sent to any
+user-defined outputs, such as Slack, PagerDuty, etc.
```
austinbyers: Can you add some context about why this is a useful feature? E.g. "To help ensure new rules do not flood your outputs with false positives, rules can be _staged_..."

docs/source/rule-staging.rst
ryandeivert: @austinbyers I think historically 'default' has actually meant "there is a default in the terraform generation code, and you can omit the config option entirely"

ryandeivert: but I don't actually like the 'required' column anyway tbh so will nix as well

ryandeivert:
@austinbyers I added one commit with more info - PTAL

manage.py
```diff
@@ -1482,39 +1482,71 @@ def _validate_memory(val):
...
+    rule_table_parser = _generate_subparser(subparsers, 'rule-staging', usage, description)
 
     rule_table_subparsers = rule_table_parser.add_subparsers()
 
+    _add_rule_table_enable_subparser(rule_table_subparsers)
     _add_rule_table_status_subparser(rule_table_subparsers)
     _add_rule_table_stage_subparser(rule_table_subparsers)
     _add_rule_table_unstage_subparser(rule_table_subparsers)
 
+def _add_rule_table_enable_subparser(subparsers):
```
chunyong-lin: Do you think it is more reasonable to rename this method and following three methods wiht prefix `_add_rule_table` to `_add_rule_staging` now as well. Because we rename `rule-table` to `rule-staging` in cli.

ryandeivert: good catch!! sorry I thought I had addressed all of these!

PR #804, opened 2018-08-13
[apps] De-duplicate GSuite Events

### LINKED ISSUE #800: GSuiteReports App processes same event in multiple scheduled invocations
## Background

The `startTime` that is passed to the activities API is set to the timestamp of the last event.

https://github.com/airbnb/streamalert/blob/a7eaa30516856e17303244fc99cb386be697e69f/stream_alert/apps/_apps/gsuite.py#L127-L132

But the response from the next call to the activity API includes that last event. The `startTime` must be inclusive.

https://developers.google.com/admin-sdk/reports/v1/reference/activities/list

> startTime | string
> Sets the beginning of the range of time shown in the report. The date is in the RFC 3339 format, for example 2010-10-28T10:26:35.000Z. The report returns all activities from startTime until endTime. The startTime must be before the endTime (if specified) and the current time when the request is made, or the API returns an error.

### Description

The `startTime` passed to the Google Reports API includes the last event already processed in the previous lambda execution.

### Steps to Reproduce

Observed in the CloudWatch event logs for the GSuiteReports app:

```
/var/task/app_integrations/main.py:16: RuntimeWarning: Parent module 'app_integrations' not found while handling absolute import
from app_integrations.apps.app_base import StreamAlertApp
START RequestId: ... Version: 1
[INFO]	2018-08-05T01:39:20.841Z	...	Starting last timestamp set to: 2018-08-03T16:53:00.927Z
[INFO]	2018-08-05T01:39:21.119Z	...	App starting for service 'gsuite_admin'.
[INFO]	2018-08-05T01:39:21.119Z	...	App executing as a successive invocation: False
[INFO]	2018-08-05T01:39:22.873Z	...	Starting batch send of 1 logs to the rule processor
[INFO]	2018-08-05T01:39:23.191Z	...	Sent 1 logs to 'PREFIX_CLUSTER_streamalert_rule_processor' with Lambda request ID '...'
[INFO]	2018-08-05T01:39:23.193Z	...	Gather process for 'gsuite_admin' executed in 1.884724 seconds.
[INFO]	2018-08-05T01:39:23.197Z	...	Lambda remaining seconds: 114.62
[INFO]	2018-08-05T01:39:23.197Z	...	Lambda remaining seconds: 114.62
[ERROR]	2018-08-05T01:39:23.197Z	...	Ending last timestamp is the same as the beginning last timestamp. This could occur if there were no logs collected for this execution.
[INFO]	2018-08-05T01:39:23.197Z	...	App complete for service 'gsuite_admin'. Gathered 1 logs in 1 polls.
END RequestId: ...
REPORT RequestId: ...	Duration: 5505.05 ms	Billed Duration: 5600 ms Memory Size: 128 MB	Max Memory Used: 66 MB
```

Notice the last time stamp on the 3rd of August, even though the log is from the 5th of August. And there was 1 log sent to the rule_processing lambda. I observed this same log at every scheduled invocation where there were no new events.

## Desired Change

We could increment `_last_event_timestamp` by one millisecond, or exclude the last seen report id. What would happen if multiple events occurred on the same millisecond?

stoggi:
cc: @ryandeivert 
size: medium
resolves #800

## Background

The startTime passed to the Google Reports API includes the last event already processed in the previous lambda execution. So duplicate events are observed in subsequent API polls.

## Changes

    * Created new key "context" stored in the state parameter in ssm
    * Save at most 50 event ids from the most recent timestamp in the context. Limit of 4096 characters in ssm.
    * Correctly format dates with a fractional second component.
    * De-duplicate events based on last seen event ids.
    * Update unit tests to output unique timestamps per generated log.
    * Added unit tests

## Testing

Added new unit tests:
* test_gather_logs_duplicate_events_more_to_poll
* test_gather_logs_duplicate_events_next_poll
* test_set_context_new
* test_set_context_same
* test_set_context_not_a_dictionary
* test_set_context_not_serializable

### DIFF
--- a/stream_alert/apps/_apps/gsuite.py
+++ b/stream_alert/apps/_apps/gsuite.py
@@ -151,7 +151,13 @@ class GSuiteReportsApp(AppIntegration):
         # once during the first poll
         if not self._next_page_token:
             self._last_timestamp = activities[0]['id']['time']
+            self._context = [
+                activity['id']['uniqueQualifier']
+                for activity in activities
+                if activity['id']['time'] == self._last_timestamp
+            ]
             LOGGER.debug('Caching last timestamp: %s', self._last_timestamp)
+            LOGGER.debug('Caching last event ids: %s', self._context)
 
         self._next_page_token = results.get('nextPageToken')
         self._more_to_poll = bool(self._next_page_token)
--- a/stream_alert/apps/_apps/gsuite.py
+++ b/stream_alert/apps/_apps/gsuite.py
@@ -35,6 +35,11 @@ class GSuiteReportsApp(AppIntegration):
     _GOOGLE_API_EXCEPTIONS = (apiclient.errors.Error, client.Error, socket.timeout,
                               ssl.SSLError)
 
+    # The maximum number of unique event ids to store that occur on the most
+    # recent timestamp. These are used to de-duplicate events in the next poll.
+    # This is limited by AWS SSM parameter store maximum of 4096 characters.
+    _MAX_EVENT_IDS = 50
+
     def __init__(self, event, context):
         super(GSuiteReportsApp, self).__init__(event, context)
         self._activities_service = None
@@ -52,7 +57,7 @@ class GSuiteReportsApp(AppIntegration):
     @classmethod
     def date_formatter(cls):
         """Return a format string for a date, ie: 2010-10-28T10:26:35.000Z"""
-        return '%Y-%m-%dT%H:%M:%SZ'
+        return '%Y-%m-%dT%H:%M:%S.%fZ'
 
     @classmethod
     def _load_credentials(cls, keydata):
@@ -141,7 +146,12 @@ class GSuiteReportsApp(AppIntegration):
             LOGGER.error('[%s] No results received from the G Suite API request', self)
             return False
 
-        activities = results.get('items', [])
+        # Remove duplicate events present in the last time period.
+        last_event_ids = set(self._context.get("last_event_ids", []))
+        activities = [
+            activity for activity in results.get('items', [])
+            if activity['id']['uniqueQualifier'] not in last_event_ids
+        ]
         if not activities:
             LOGGER.info('[%s] No logs in response from G Suite API request', self)
             return False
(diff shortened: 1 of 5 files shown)


stream_alert/apps/_apps/gsuite.py
```diff
@@ -151,11 +161,24 @@ def _gather_logs(self):
...
             self._last_timestamp = activities[0]['id']['time']
-            LOGGER.debug('Caching last timestamp: %s', self._last_timestamp)
+            LOGGER.debug('[%s] Caching last timestamp: %s', self, self._last_timestamp)
+            # Store the event ids with the most recent timestamp to de-duplicate them next time
+            next_event_ids = [
+                activity['id']['uniqueQualifier']
+                for activity in activities
+                if activity['id']['time'] == self._last_timestamp
+            ]
+            self._context['next_event_ids'] = next_event_ids[:self._MAX_EVENT_IDS]
```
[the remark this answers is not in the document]
ryandeivert: or just something similar using a local property that then caches to the instance prop if you want to keep the warning message below

stream_alert/apps/_apps/gsuite.py
```diff
@@ -151,11 +161,24 @@ def _gather_logs(self):
...
+            if len(next_event_ids) > self._MAX_EVENT_IDS:
+                LOGGER.warning('[%s] More than %d next_event_ids. Unable to de-duplicate: %s',
+                               self, self._MAX_EVENT_IDS, next_event_ids[self._MAX_EVENT_IDS:])
 
         self._next_page_token = results.get('nextPageToken')
         self._more_to_poll = bool(self._next_page_token)
 
+        if not self._more_to_poll:
+            self._context['last_event_ids'] = self._context.get('next_event_ids', [])
```
ryandeivert: this could then just save the cached instance property instead of copying a value that already exist in the context

stream_alert/apps/config.py
```diff
@@ -273,9 +275,20 @@ def _determine_last_time(self, date_format):
...
+            AppStateError: If the parameter is unabled to be saved
+        """
+        try:
+            param_value = json.dumps({
+                self._TIME_KEY: self.last_timestamp,
+                self._STATE_KEY: self.current_state,
+                self._CONTEXT_KEY: self.context,
+            })
+        except TypeError as err:
+            raise AppStateError('Could not serialise state for name \'{}\'. Error: '
```
ryandeivert: I like the cultural diversity here :P but could we use the american english version of `serialise` here?

stream_alert/apps/_apps/gsuite.py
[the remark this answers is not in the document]
stoggi: The reason I needed to cache `_next_event_ids` and `_last_event_ids` is because the events are returned in reverse chronological order. The first few events contain the duplicates that we need to save, and the last events are the ones we want to de-duplicate. But in the case of a multiple data poll, there will be more than one invocation of `_gather_logs`, and the events at the end of the second poll will be the duplicates from the last polled time period, not necessarily the last lambda invocation. So we can't overwrite the cached event ids, and temporarily need to store both.

So, `_last_event_ids` are the event ids from the last poll, and `_next_event_ids` are the event ids for the next poll.

stream_alert/apps/_apps/gsuite.py
stoggi: It's true, at this point we no longer need to store `next_event_ids`. But only because there is nothing left to poll, and we know the next lambda invocation will be for a new time period. We could clear `next_event_ids`, but the important thing is to copy them across to `last_event_ids`.

stream_alert/apps/config.py
stoggi: Ah yep, good point 👍

stream_alert/apps/_apps/gsuite.py
stoggi: Refactored to only store one copy in the context since `_gather_logs` is called multiple times per execution. The case where it gets deferred to the next lambda execution is a very small edge case, and worst case you get some duplicate events.

stream_alert/apps/_apps/gsuite.py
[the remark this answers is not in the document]
stoggi: Refactored to only store one copy in the context since _gather_logs is called multiple times per execution. The case where it gets deferred to the next lambda execution is a very small edge case, and worst case you get some duplicate events.

stoggi:
@ryandeivert addressed your questions, and made the following changes:

* Refactored to only store one copy of last_event_ids because _gather_logs is called many times per lambda execution.
* Increased the _MAX_EVENT_IDS since only one copy of last_event_ids is stored.
* Fixed typos.

ryandeivert:
thanks @stoggi !! I appreciate you following up on this :) I'm going to try to get to this soon (by EOW)

PR #805, opened 2018-08-13
[rules] cloudtrail_mfa_policy_abuse_attempt rule logic improved

### LINKED ISSUE #803: errorCode logic not valid in rule cloudtrail_mfa_policy_abuse_attempt
## Background

In the rule `cloudtrail_mfa_policy_abuse_attempt` the errorCode is not properly checked as part of the rule logic:

https://github.com/airbnb/streamalert/blob/d5c658fc73f5dad6c6577bd462a85f38102b1c10/rules/community/cloudtrail/cloudtrail_mfa_policy_abuse_attempt.py#L62-L70

It seems the intent is only to alert if there is any error for a `CreateVirtualMFADevice` or `EnableMFADevice` event. But because errorCode is specified in optional_top_level_keys, it actually gets assigned a default value of `str()`, even if there is no error.

https://github.com/airbnb/streamalert/blob/d7a879b9b463d9a304328e1d170a27f118ff5cb0/stream_alert/rule_processor/parsers.py#L245-L248

### Description

The `errorCode` key is always set in records, regardless of an error or not, so the rule logic is not valid.

### Steps to Reproduce

1. Sign into the AWS Console without MFA
2. Activate a new virtual MFA device
3. Observe the alert
4. For bonus points, check the event in CloudWatch to double check the actual event had no key `errorCode` in the event JSON.

## Desired Change

Modify the rule to check if errorCode is not an empty string

stoggi:
cc: @strcrzy 
size: small
resolves #803 

## Background

Because `errorCode` is an optional parameter set in `optional_top_level_keys`. If the record does not contain it, it gets set to the default value for it's type. In this case `str()`.

So checking that the key exists is not enough to indicate an error, instead check the value evaluates to True.

## Changes

* Only alert if the contents of `errorCode` is not False.
* Added integration test.

## Testing

* Added new integration test: `Successfully creating an MFA device, while the user is not MFA'd should not create an alert.`

### DIFF
--- a/rules/community/cloudtrail/cloudtrail_mfa_policy_abuse_attempt.py
+++ b/rules/community/cloudtrail/cloudtrail_mfa_policy_abuse_attempt.py
@@ -66,7 +66,7 @@ def cloudtrail_mfa_policy_abuse_attempt(rec):
     # - 'AccessDenied'
     # - 'EntityAlreadyExists': Can't create another MFA device with the same name.
     # - 'LimitExceeded': Can't enable a second MFA device for the same user.
-    if 'errorCode' in rec and rec['eventName'] in _EVENT_NAMES:
+    if rec['errorCode'] and rec['eventName'] in _EVENT_NAMES:
         return True
 
     return False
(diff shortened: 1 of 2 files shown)


strcrzy:
awesome, thanks!

ryandeivert:
thanks for the change @stoggi - merging!

PR #806, opened 2018-08-15
updating the athena partition refresh function to use sqs as event source

ryandeivert:
to: @chunyong-lin or @austinbyers 
size: large

## Background

With the introduction of [SQS as an event source for Lambda](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html), we can get rid of a ton of tech debt in the Athena partition refresh function. This will also enable better scalability of this function.

## Changes

* Removing our now unnecessary SQS client class.
* Updating the `AthenaRefresher` class to read s3 bucket notifications from the input event.
* Updating terraform code for input event source mapping to the SQS queue.
* Bumping terraform aws provder version from `1.17.0` --> `1.26.0` (for lambda sqs support)

## Testing

Updating all unit tests/removing old ones.


stream_alert/athena_partition_refresh/main.py
```diff
@@ -152,44 +158,50 @@ def _add_partition(self, s3_buckets_and_keys):
                 )
 
             LOGGER.info('Successfully added the following partitions:\n%s',
-                        json.dumps({athena_table: partitions[athena_table]}, indent=4))
+                        json.dumps({athena_table: partitions[athena_table]}))
         return True
 
-    def run(self):
+    def run(self, event):
         """Poll the SQS queue for messages and create partitions for new data"""
```
austinbyers: This function no longer polls the SQS queue

stream_alert/athena_partition_refresh/main.py
```diff
@@ -152,44 +158,50 @@ def _add_partition(self, s3_buckets_and_keys):
...
+            for s3_rec in body['Records']:
+                if 's3' not in s3_rec:
+                    LOGGER.info('Skipping non-s3 bucket notification message: %s', s3_rec)
+                    continue
+
+                bucket_name = s3_rec['s3']['bucket']['name']
+
+                # Account for special characters in the S3 object key
+                # Example: Usage of '=' in the key name
+                object_key = urllib.unquote(s3_rec['s3']['object']['key']).decode('utf8')
```
austinbyers: Use `urllib.unquote_plus`, as spaces are encoded with `+`

ryandeivert:
Thanks @austinbyers !! I've addressed your feedback and set the concurrency limit to something fairly modest (10 for now). I checked some metrics and I think that should be more than enough headroom to operate here, but we can always update if the function gets throttled.

PR #807, opened 2018-08-16
[tf][firehose] support per firehose metric alarms for IncomingRecords

ryandeivert:
size: large

## Background

Currently, StreamAlert uses individual AWS Firehose delivery streams for each log type ingested and parsed. This enables a somewhat simple and reliable way to determine if processing of a specific log type has dropped off.

## Changes

* Slight refactor to StreamAlertFirehose client to make performing some actions easier.
  * The prior class required instantiation for performing operations such as determining enabled log types, etc, which was overkill.
  * Migrating a few methods to be class methods, since instantiation will create a boto client which is not needed for CLI operations.
  * Also renaming from `StreamAlertFirehose` to just `FirehoseClient` since it's less wordy.
  * Updating all areas that used this class.
* Migrating payload categorization logic to `FirehoseClient` class instead of handling in the main `StreamAlert`.
* Adding terraform code for minimum log count alarm support in Firehose DeliveryStreams
* Updating terraform generation code for firehose log count alarms configuration

## Testing

* Updating/renaming old `StreamAlertFirehose` unit tests.
* Adding unit tests for firehose terraform generation code (no tests previously existed).

### DIFF
--- a/conf/global.json
+++ b/conf/global.json
@@ -23,7 +23,7 @@
       "buffer_size": 128,
       "compression_format": "GZIP",
       "enabled": false,
-      "enabled_logs": []
+      "enabled_logs": {}
     },
     "monitoring": {
       "create_sns_topic": true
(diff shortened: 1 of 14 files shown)


ryandeivert:
thanks @austinbyers made the suggested changes :)

PR #811, opened 2018-08-24
fixing bug where apps could time out between gather calls

ryandeivert:
to: @chunyong-lin 
size: small

## Background

Identified bug from recent refactor that allowed for looping of gather calls to not take into account the time apps may have to 'sleep' between new api requests.

## Changes

* Adding the sleep seconds returned by the app to the calculated time that would be needed for another request to be successfully made.


austinbyers:
Awesome! Good find

PR #813, opened 2018-09-03
SNS s3 notifications

ts33:
cc: @austinbyers 
size: small
resolves ?

## Background

Streamalert currently allows adding an S3 bucket as a data source. This creates an S3 event notification for the S3 bucket. As S3 currently only supports one event notification per bucket, this also means that the source bucket would only be configured to notify Streamalert.

A common use case would be to have Streamalert subscribe to an existing S3 event notification channel such as an SNS topic. 

In this PR, Streamalert checks if an SNS payload is an S3 event notification and if so, extracts the S3 object for parsing.

## Changes

* Additional test case `test_pre_parse_sns_s3_notification`.
* Additional test helper method `make_sns_s3_notification_raw_record `.
* Modification of the `pre_parse` method of the `SnsPayload` payload class to detect and parse S3 event notifications.

## Testing

- Additional test case `test_pre_parse_sns_s3_notification` was added.
- A full test was also conducted with the following steps:
  - create a new cloudtrail with a target s3 bucket.
  - configure the target s3 bucket with an SNS s3 event notification.
  - spin up streamalert with the configuration templates below.
  - perform an action that triggers one of the community rules (create an IAM user using an IAM user profile that does not have MFA configured).

prod.json:
```
{
  "id": "prod",
  "modules": {
    "stream_alert": {
      "rule_processor": {
        "current_version": "$LATEST",
        "memory": 256,
        "timeout": 10,
        "inputs": {
          "aws-sns": [
            "arn:aws:sns:${AWS::Region}:${AWS::AccountId}:${sns_topic_name}"
          ]
        }
      },
      "alert_processor": {
        "current_version": "$LATEST",
        "memory": 128,
        "timeout": 25
      }
    }
  },
  "region": "ap-southeast-1"
}
```
source.json:
```
{
  "sns": {
    "${sns_topic_name}": {
      "logs": [
        "cloudtrail"
      ]
    }
  }
}
```

## Notes
You will have to configure the source bucket policy to allow access to the Streamalert AWS lambda function role for this to work:
```
{
    "Version": "2012-10-17",
    "Statement": [
        ...
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::${AWS::AccountId}:role/${stream_alert_prefix}_prod_streamalert_rule_processor_role"
            },
            "Action": [
                "s3:Get*",
                "s3:List*"
            ],
            "Resource": [
                "arn:aws:s3:::${s3_bucket_name}",
                "arn:aws:s3:::${s3_bucket_name}/*"
            ]
        }
    ]
}
```


willbengtson:
+1

ryandeivert:
hey @ts33 thank you for this contribution!! I apologize for not acknowledging this a bit sooner, but will be getting to look at it soon. please standby :)

stream_alert/rule_processor/payload.py
```diff
@@ -331,7 +332,19 @@ def pre_parse(self):
             self.raw_record['Sns']['MessageId'],
             self.raw_record['EventSubscriptionArn'])
 
-        self.pre_parsed_record = self.raw_record['Sns']['Message']
+        # check if SNS message is an Amazon S3 Notification type
+        if 'Subject' in self.raw_record['Sns'] and \
```
ryandeivert: Please instead use a dictionary lookup here to simplify this:

```
if self.raw_record['Sns'].get('Subject') == 'Amazon S3 Notification':
    ... 
```

Also, to be consistent with our current style, can you please switch to using single quotes for string literals instead of double? ie: `'Amazon S3 Notification'` vs the current `"Amazon S3 Notification"`

stream_alert/rule_processor/payload.py
```diff
@@ -331,7 +332,19 @@ def pre_parse(self):
...
+            self.raw_record['Sns']['Subject'] == "Amazon S3 Notification":
+            LOGGER.debug('Amazon S3 notification detected.. parsing message')
+
+            message = json.loads(self.raw_record['Sns']['Message'])['Records'][0]
+            # encapsulate in s3 stream payload
+            s3_payload = load_stream_payload('s3', 'sns-passthrough', message)
```
ryandeivert: This is a neat trick!! I would still suggest actually passing the instance's `entity` to the `load_stream_payload` for consistency about where the data came from:

`s3_payload = load_stream_payload('s3', self.entity, message)`

stream_alert/rule_processor/payload.py
```diff
@@ -331,7 +332,19 @@ def pre_parse(self):
...
+            s3_payload = s3_payload.pre_parse().next()
```
ryandeivert: It appears this code is assuming the file downloaded from s3 is only going to contain 1 line? 🤔 is that the expected logic here?

If not, I would suggest actually yielding all records from this, like so:


```
if self.raw_record['Sns'].get('Subject') == 'Amazon S3 Notification':
    ...
    # other logic above
    ...
    for record in s3_payload.pre_parse():
        yield record
else:
    self.pre_parsed_record = self.raw_record['Sns']['Message']
    yield self
```

Notice the `yield self` has been moved to within the `else` block as well.

ts33: My understanding of generators is pretty limited 😅 but when I tried the method you proposed, I get the following error in a manual integration test:

```
object of type 'NoneType' has no len(): TypeError
File "/var/task/stream_alert/rule_processor/main.py", line 26, in handler
StreamAlert(context).run(event)
File "/var/task/stream_alert/rule_processor/handler.py", line 122, in run
payload_with_normalized_records.extend(self._process_alerts(payload))
File "/var/task/stream_alert/rule_processor/handler.py", line 209, in _process_alerts
len(payload.records),
TypeError: object of type 'NoneType' has no len()
```
It seems like the return payload object will have an NoneType `records` parameter

ryandeivert:
this support has been added in a different way, but this is now supported in 2.1.0

PR #819, opened 2018-09-27
[lambda] support vpc attachment

strcrzy:
size: small

## Background
Typically network calls from SA lambdas egress over the internet. Sometimes we want them to egress from our VPC subnets.

## Changes

This change should allow users to attach their RuleProcessor lambda clusters to VPC subnets and security groups.

## Testing

I ran the unit tests

### DIFF
--- a/stream_alert_cli/terraform/streamalert.py
+++ b/stream_alert_cli/terraform/streamalert.py
@@ -44,6 +44,8 @@ def generate_stream_alert(cluster_name, cluster_dict, config):
     """
     account = config['global']['account']
     modules = config['clusters'][cluster_name]['modules']
+    vpc_subnets = config['clusters'][cluster_name].get("vpc_subnets", "")
+    vpc_security_groups = config['clusters'][cluster_name].get("vpc_security_groups", "")
 
     cluster_dict['module']['stream_alert_{}'.format(cluster_name)] = {
         'source': 'modules/tf_stream_alert',
@@ -59,6 +61,8 @@ def generate_stream_alert(cluster_name, cluster_dict, config):
         'rule_processor_memory': modules['stream_alert']['rule_processor']['memory'],
         'rule_processor_timeout': modules['stream_alert']['rule_processor']['timeout'],
         'rules_table_arn': '${module.globals.rules_table_arn}',
+        'vpc_subnets': vpc_subnets,
+        'vpc_security_groups': vpc_security_groups,
     }
 
     if (config['global'].get('threat_intel')
--- a/stream_alert_cli/terraform/streamalert.py
+++ b/stream_alert_cli/terraform/streamalert.py
@@ -44,8 +44,8 @@ def generate_stream_alert(cluster_name, cluster_dict, config):
     """
     account = config['global']['account']
     modules = config['clusters'][cluster_name]['modules']
-    vpc_subnets = config['clusters'][cluster_name].get("vpc_subnets", "")
-    vpc_security_groups = config['clusters'][cluster_name].get("vpc_security_groups", "")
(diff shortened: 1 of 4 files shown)


stream_alert_cli/terraform/streamalert.py
```diff
@@ -44,6 +44,8 @@ def generate_stream_alert(cluster_name, cluster_dict, config):
     """
     account = config['global']['account']
     modules = config['clusters'][cluster_name]['modules']
+    vpc_subnets = config['clusters'][cluster_name].get("vpc_subnets", "")
```
ryandeivert: maybe I'm misunderstanding what the value of these will/should be, but I think the default value here should be an empty list `[]` not empty string, otherwise the tf will fail since it's expecting a list: https://github.com/airbnb/streamalert/pull/819/files#diff-ef86a913d13a64356b590676f0fbce8eR73

terraform/modules/tf_stream_alert/main.tf
```diff
@@ -13,6 +13,11 @@ resource "aws_lambda_function" "streamalert_rule_processor" {
   source_code_hash = "${base64sha256(file(var.filename))}"
   publish          = true
 
+  vpc_config {
+    subnet_ids      = ["${var.vpc_subnets}"]
```
ryandeivert: since these vars are of list type, you should be able to omit the square brackets wrapping these (I could also be misinterpreting this)

tests/unit/stream_alert_cli/terraform/test_generate.py
```diff
@@ -283,6 +283,8 @@ def test_generate_stream_alert_test(self):
                     'rule_processor_memory': 128,
                     'rule_processor_timeout': 25,
                     'rules_table_arn': '${module.globals.rules_table_arn}',
+                    'vpc_subnets': "",
```
ryandeivert: Same comment here - should this be a single string of a list?

ryandeivert:
thanks for the changes @strcrzy I left a few comments

stream_alert_cli/terraform/streamalert.py
chunyong-lin: Can we use single quotes to be consistent? `"vpc_subnets"` --> `'vpc_subnets'`

strcrzy: i'm not sure -- [docs](https://www.terraform.io/docs/providers/aws/r/lambda_function.html#vpc_config) say:
 `NOTE: if both subnet_ids and security_group_ids are empty then vpc_config is considered to be empty or unset.`

stream_alert_cli/terraform/streamalert.py
```diff
@@ -44,6 +44,8 @@ def generate_stream_alert(cluster_name, cluster_dict, config):
     """
     account = config['global']['account']
     modules = config['clusters'][cluster_name]['modules']
+    vpc_subnets = config['clusters'][cluster_name].get('vpc_subnets', '[]')
```
ryandeivert: sorry @strcrzy - need to remove the quotes from around the lists here

tests/unit/stream_alert_cli/terraform/test_generate.py
```diff
@@ -283,6 +283,8 @@ def test_generate_stream_alert_test(self):
                     'rule_processor_memory': 128,
                     'rule_processor_timeout': 25,
                     'rules_table_arn': '${module.globals.rules_table_arn}',
+                    'vpc_subnets': '[]',
```
ryandeivert: unquote the list(s) here too pls

Ryxias:
Closing zombie'd PR

PR #821, opened 2018-09-29
function split part 1, migration and refactor of payload and parsers

ryandeivert:
to: @chunyong-lin , @strcrzy 

## Background

In the near future, StreamAlert's rule process will be split into 2 functions - one for classification and one for rule processing. This is the start of that migration.

## Changes

* Slight refactor to the `StreamPayload` class to better support loading of data from incoming records. This class now encapsulates all logic, instead of the logic being in multiple disparate places/tied to the 'classifier'.
* Updates to the `S3Payload` class
  * Adding support for various types of input. Now, any combination of the following will work:
    * Newline-delimited json data (one json object per line)
    * Formatted json data (not newline-delimited)
    * Lines of raw data (non-json, aka csv)
    * Gzipped data (of all above types)
    * Non-gzipped data (of all above types)
  * The above change now uses `jsonlines` to iterate over lines in newline-delimited json data. This is faster/more memory efficient then the alternative `ndjson` package, which loads the entire list of lines into memory and _then_ iterates over them. `jsonlines` does this upon read, so will use as little memory possible.
  * Utilizing `tempfile.TemporaryFile` to handle reading/garbage collection better.
* Complete overhaul of the `ParserBase` class to be far more clear and concise.
  * This change includes updates to support never missing unparsed data. Prior to this, [data could be deleted](https://github.com/airbnb/streamalert/blob/master/stream_alert/rule_processor/parsers.py#L219-L226) during parsing and we would never actually know if this occurred.
  * The parser will now track what records have successfully parsed with a given schema, along with ones that have failed.
  * No successful parses with a schema now means a complete failure, which will skip to trying the next schema.
* Removing usage of the `jsonpath_rw` library in parsing, since it is very slow. Replacing it with the simpler/faster `jmespath` library.
* Updates to requirements.txt and various other fixes.
* Fixes to various other bugs discovered during the rewrite.

## Testing

* Completely re-wrote unit tests for comprehensive coverage.

### Notes for reviewers

I apologize in advance that this is so large. This will be merged into an ongoing work-in-progress feature branch before merging into master.


stream_alert/classifier/parsers.py
```diff
@@ -0,0 +1,808 @@
...
+from abc import ABCMeta, abstractmethod
+from copy import deepcopy
+import csv
+from fnmatch import fnmatch
+import json
+import logging
+import re
+import StringIO
+
+import jmespath
```
strcrzy: thanks for this, sorry about jsonpath_rw

stream_alert/classifier/parsers.py
```diff
@@ -0,0 +1,808 @@
...
+            if isinstance(patterns, dict):
+                return cls._matches_log_patterns(record[field], patterns)
+
+            # Transform a flat pattern into a list
+            if isinstance(patterns, basestring):
+                LOGGER.debug('Transforming flat pattern \'%s\' into list', patterns)
+                patterns = [patterns]
+
+            # Ensure the pattern key is in the record
+            # TODO (ryandeivert): consider failing hard and returning False here
```
strcrzy: what about an exception?

stream_alert/classifier/parsers.py
```diff
@@ -0,0 +1,808 @@
...
+
+        schema_keys = set(schema)
+
+        keys = set(record) if not optionals else set(record).union(optionals)
+
+        if is_envelope and not schema_keys.issubset(keys):
+            LOGGER.debug('Missing keys in record envelope: %s', schema_keys - keys)
+            return False
+
+        if not is_envelope and keys != schema_keys:
```
strcrzy: again i wonder if this is somewhere an exception would be useful

stream_alert/classifier/parsers.py
```diff
@@ -0,0 +1,808 @@
...
+        # Handle jsonpath extraction of records
+        LOGGER.debug('Parsing records with JSONPath: %s', self.json_path)
+
+        return jmespath.search(self.json_path, payload)
+
+    def parse(self, data):
+        """Main parser method to be handle parsing of the passed data
+
+        Args:
+            data (str|dict): Data to be parsed.
```
strcrzy: when would this be a dict and not a string?

strcrzy: nm i see it calls itself now

stream_alert/classifier/parsers.py
```diff
@@ -0,0 +1,808 @@
...
+        # Add the optional envelope keys to record once at the beginning
+        self._add_optional_keys(envelope, self.envelope_schema, self.optional_envelope_keys)
+
+        for record, valid in self._parse(data_copy):
+            valid = valid and self._key_check(record, self.schema, self.optional_top_level_keys)
+            valid = valid and self._convert_type(record, self.schema, self.optional_top_level_keys)
+            valid = valid and self._matches_log_patterns(record, self.log_patterns)
+            self._add_parse_result(record, valid, envelope)
+
+        return self.valid
```
strcrzy: where is this set?

stream_alert/classifier/parsers.py
strcrzy: i take it back, i'm curious again

stream_alert/classifier/parsers.py
```diff
@@ -0,0 +1,808 @@
...
+                'Schema definition is not valid (%s):\n%s', self._schema_type, self.schema)
+            return False
+
+        data_copy = None
+        # If the data is a mutable object, copy it since
+        # the parse(s) can highly mutate the input
+        if isinstance(data, dict):
+            data_copy = deepcopy(data)
+
+        # If the data is a string, attempt to load it to json,
```
strcrzy: why do we do this in the base parser?

stream_alert/classifier/parsers.py
```diff
@@ -0,0 +1,808 @@
...
+                    kv_payload[key] = value
+
+        except UnicodeDecodeError:
+            return False
+
+        return kv_payload
+
+
+@parser
+class SyslogParser(ParserBase):
```
strcrzy: while we're factoring things, could this be a subclass of a more generic RegexParser?

stream_alert/classifier/payload/apps.py
```diff
@@ -0,0 +1,45 @@
...
+from stream_alert.classifier.payload.payload_base import (
+    PayloadRecord,
+    RegisterInput,
+    StreamPayload
+)
+from stream_alert.shared import CLASSIFIER_FUNCTION_NAME as FUNCTION_NAME
+from stream_alert.shared.metrics import MetricLogger
+
+
+@RegisterInput
```
strcrzy: no idea how this works, would love to learn

stream_alert/classifier/payload/kinesis.py
```diff
@@ -0,0 +1,59 @@
...
+        Decodes the base64 payload from within the record, and attempts to decompress
+        the data. If the data is not compressed, this falls back on the original content.
+
+        Yields:
+            Instances of PayloadRecord back to the caller containing the current log data
+        """
+        LOGGER.debug('Pre-parsing record from Kinesis. eventID: %s, eventSourceARN: %s',
+                     self.raw_record['eventID'], self.raw_record['eventSourceARN'])
+
+        # Kinesis records have to potential to be gzipped, so try to decompress
```
strcrzy: i remember this :)

stream_alert/classifier/payload/payload_base.py
```diff
@@ -0,0 +1,222 @@
...
+import logging
+
+from stream_alert.shared.logger import get_logger
+
+
+LOGGER = get_logger(__name__)
+LOGGER_DEBUG_ENABLED = LOGGER.isEnabledFor(logging.DEBUG)
+
+
+class PayloadRecord(object):
```
strcrzy: i'm trying to figure out where the line between the payload and the record is wrt deserialization. i think that's what the above line i was confused about has to do with, deserializing the wrapper payload. if that's the case, is that a parser concern, or a payload concern? if it's not the case and i'm confused then i'm sorry

stream_alert/classifier/payload/payload_base.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        Keyword Args:
+            raw_record (dict): The record to be parsed - in AWS event format
+        """
+        self.raw_record = raw_record
+        self.resource = resource
+        self.log_source = None
+        self.records = None
+        self.data_type = None
+        self.fully_classified = True
```
strcrzy: how does this get set to false?

stream_alert/classifier/payload/payload_base.py
```diff
@@ -0,0 +1,222 @@
...
+        if 'Sns' in raw_record:
+            # If this is an s3 event notification via SNS, extract the bucket from the record
+            if (raw_record['Sns'].get('Type') == 'Notification' and
+                    raw_record['Sns'].get('Subject') == 'Amazon S3 Notification'):
+                service = 's3'
+                raw_record = json.loads(raw_record['Sns']['Message'])['Records'][0]
+                resource = resource_mapper[service](raw_record)
+            else:
+                service = 'sns'
+                resource = raw_record['Sns']['TopicArn'].split(':')[-1]
```
strcrzy: why is this not in the resource_mapper?

stream_alert/classifier/payload/s3.py
```diff
@@ -0,0 +1,214 @@
...
+            tuple: line number, contents of the line being read
+        """
+        # Shred the temp dir before downloading
+        self._shred_temp_directory()
+        bucket = self._unquote(self.bucket)
+        key = self._unquote(self.key)
+
+        # Use tempfile.TemporaryFile to do the download
+        # This will automatically close/get garbage collected upon completion
+        with tempfile.TemporaryFile() as download:
```
strcrzy: right now it looks like you're only shredding before actually writing the sensitive data.
i might use a NamedTemporaryFile and shred it immediately after consumption (over shredding the directory at the end which may not run if the lambda terminates early)

stream_alert/classifier/parsers.py
ryandeivert: spoke offline a bit

stream_alert/classifier/parsers.py
ryandeivert: This is done here because there are options for parsing csv data embedded in json. front loading the json loading here ensures that the CSV parser get's the expected data

stream_alert/classifier/parsers.py
ryandeivert: this is a computed property (not set explicitly) and is considered true if this schema resulted in any parsed records. see:

https://github.com/airbnb/streamalert/pull/821/files#diff-39b8f0570d770502a95e06c76883fceeR121

stream_alert/classifier/parsers.py
ryandeivert: this is a good idea. I'll try to address now but may come in a later PR

stream_alert/classifier/payload/apps.py
ryandeivert: let's sync in person about this - essentially this class decorator acts as a factory to register the payload classes for lookups

stream_alert/classifier/payload/payload_base.py
ryandeivert: this is in the pt 2 PR in #822 - https://github.com/airbnb/streamalert/pull/822/files#diff-6845cff78cdbd876ae032a606d1237afR144

stream_alert/classifier/payload/s3.py
ryandeivert: good point regarding shredding after, not before. however, I'm confused what the benefit of using `NamedTemporaryFile` would be?

stream_alert/classifier/payload/payload_base.py
ryandeivert: make a change so the default for sns is still in the mapper

stream_alert/classifier/parsers.py
```diff
@@ -0,0 +1,807 @@
...
+        # Fall back on the data as default if extraction failed
+        data = extracted_data or [data]
+
+        return self._extract_records(data, self.schema)
+
+    def _extract_records(self, data, schema):
+        """Extract record(s) from the csv data using the specified schema
+
+        Args:
+            data (str): String representing a csv row.
```
chunyong-lin: Based on code below, argument `data` should be a list.

stream_alert/classifier/payload/payload_base.py
```diff
@@ -0,0 +1,222 @@
...
+        self._record_data = record_data
+        self.log_name = None
+        self.schema = None
+        self.data_type = None
+        self.classified = False
+        self._parsed_data = None
+
+    @property
+    def data(self):
+        return self._record_data
```
chunyong-lin: This is original data, right?

stream_alert/classifier/payload/payload_base.py
```diff
@@ -0,0 +1,222 @@
...
+    __bool__ = __nonzero__
+
+    def __repr__(self):
+        return '<{} valid:{} log_source:{} resource:{} type:{} record:{}>'.format(
+            self.__class__.__name__,
+            bool(self),
+            self.log_source,
+            self.resource,
+            self.data_type,
+            self.records
```
chunyong-lin: Should we also include `self.fully_classified` information here?

stream_alert/classifier/payload/payload_base.py
```diff
@@ -0,0 +1,222 @@
...
+        records (list): A list of parsed and typed PayloadRecord(s).
+        data_type (str): The data type of the record - json, csv, syslog, etc.
+        fully_classified (bool): Whether the payload has been successfully
+            and completely classified.
+    """
+    __metaclass__ = ABCMeta
+
+    def __init__(self, resource, raw_record):
+        """
+        Keyword Args:
```
chunyong-lin: `Args`

stream_alert/classifier/payload/s3.py
```diff
@@ -0,0 +1,214 @@
...
+            # Try to just load the reader as regular json first
+            yield 1, json.load(reader)
+            return
+        except ValueError:
+            pass
+
+        # Iterate over the lines, returning each
+        # This could be dicts from a jsonlines.Reader, or raw strings
+        for line_num, line in enumerate(cls._jsonlines_reader(reader), start=1):
+            yield line_num, line.strip() if isinstance(line, basestring) else line
```
chunyong-lin: It will yield raw string instead of a tuple (line number, content) if the `line` is not a `basestring`. Is it expected?

stream_alert/classifier/parsers.py
```diff
@@ -0,0 +1,807 @@
...
+    if not parser_type:
+        raise NotImplementedError(
+            '{}: Parser does not define a class property for \'_type\''.format(cls.__name__)
+        )
+
+    PARSERS[parser_type] = cls
+    return cls
+
+
+def get_parser(parserid):
```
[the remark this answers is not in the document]
ryandeivert: 👍

stream_alert/classifier/parsers.py
```diff
@@ -0,0 +1,807 @@
...
+            return True
+
+        result = True
+        for field, patterns in log_patterns.iteritems():
+            # handle nested log_patterns
+            if isinstance(patterns, dict):
+                return cls._matches_log_patterns(record[field], patterns)
+
+            # Transform a flat pattern into a list
+            if isinstance(patterns, basestring):
```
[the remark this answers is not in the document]
ryandeivert: hmm good catch. I'm hesitant to make the change, since I like the simplicity, and since the `2to3` conversion utility that we'll use when we 'upgrade' will actually handle the conversion for us

stream_alert/classifier/payload/payload_base.py
ryandeivert: yes - also FYI some of these properties have changed in later PRs

stream_alert/classifier/payload/payload_base.py
ryandeivert: this has actually significantly changed in a later PR: https://github.com/airbnb/streamalert/pull/822/files#diff-37c5fd6fd0ce64ec553cc7ac572a3173R187

stream_alert/classifier/payload/s3.py
ryandeivert: discussed offline

PR #823, opened 2018-10-03
function split part 3, adding firehose client for data retention

ryandeivert:
to: @strcrzy , @chunyong-lin 

## Background

Follow up to #822

This PR adds the `FirehoseClient` for historical data retention, and also implements the AWS Lambda handler for the `classifier` function. Note that this PR is being compared against the `ryandeivert-func-split-pt-2` branch. All of these will merge into the central `ryandeivert-function-split` branch upon review/approval.

## Changes

* Updating `FirehoseClient` batching logic to use jsonlines and simplifying it quite a bit through the use of a generator.
* Adding shared helper for boto configs to set default timeouts.

## Testing

* Adding comprehensive `FirehoseClient` tests and removing moto usage

### DIFF
--- a/stream_alert/classifier/clients/firehose.py
+++ b/stream_alert/classifier/clients/firehose.py
@@ -24,14 +24,14 @@ from botocore.exceptions import ClientError
 from botocore.vendored.requests.exceptions import ConnectionError, Timeout
 import jsonlines
 
-from stream_alert.shared import RULE_PROCESSOR_NAME as FUNCTION_NAME
+from stream_alert.shared import CLASSIFIER_FUNCTION_NAME as FUNCTION_NAME
 from stream_alert.shared.helpers import boto
 from stream_alert.shared.logger import get_logger
 from stream_alert.shared.metrics import MetricLogger
 from stream_alert.shared.backoff_handlers import (
     backoff_handler,
-    success_handler,
-    giveup_handler
+    giveup_handler,
+    success_handler
 )
 
 LOGGER = get_logger(__name__)
@@ -122,7 +122,7 @@ class FirehoseClient(object):
                 del current_batch[:]
 
             if line_len > cls.MAX_RECORD_SIZE:
-                LOGGER.error('Record too large to send to Firehose:\n%s', record)
+                LOGGER.error('Record too large (%d) to send to Firehose:\n%s', line_len, record)
                 cls._log_failed(1)
                 continue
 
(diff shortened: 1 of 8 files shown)


stream_alert/classifier/clients/firehose.py
```diff
@@ -0,0 +1,410 @@
...
+import re
+
+import backoff
+import boto3
+from botocore.exceptions import ClientError
+from botocore.vendored.requests.exceptions import ConnectionError, Timeout
+import jsonlines
+
+from stream_alert.shared import CLASSIFIER_FUNCTION_NAME as FUNCTION_NAME
+from stream_alert.shared.helpers import boto
```
chunyong-lin: Can we rename `boto` to `boto_config`? It is just client config for boto client.

stream_alert/classifier/clients/firehose.py
```diff
@@ -0,0 +1,410 @@
...
+        self._client = boto3.client('firehose', config=boto.default_config())
+        # Create a dictionary to hold parsed payloads by log type.
+        # Firehose needs this information to send to its corresponding
+        # delivery stream.
+        self._categorized_records = defaultdict(list)
+
+        self.load_enabled_log_sources(firehose_config, log_sources, force_load=True)
+
+    @classmethod
+    def _records_json_lines(cls, records):
```
chunyong-lin: Can we rename this method to something similar to `_records_json_encoding`? It represents what does this method do.

stream_alert/classifier/clients/firehose.py
```diff
@@ -0,0 +1,410 @@
...
+
+        Args:
+            records (list<dict>): Records to be written to line-delimited json
+
+        Returns:
+            list: JSON serialized records
+        """
+        # Write the json lines to the object in minimal form
+        json_lines = io.StringIO()
+        with jsonlines.Writer(json_lines, compact=True) as writer:
```
chunyong-lin: Talked offline. `jsonlines` won't bring benefit. We will switch back to `json.dumps`

stream_alert/classifier/clients/firehose.py
ryandeivert: I named it that so it could theoretically be expanded to other things.. for instance I was also thinking about adding factory function that returns a client/resource for a given service (and uses this default config). is it okay to leave as is?

stream_alert/classifier/clients/firehose.py
ryandeivert: how about `_records_to_json_list`?

stream_alert/classifier/clients/firehose.py
chunyong-lin: sgtm!

ryandeivert: addressed the confusion via updating the import/usage of this helper :)

PR #824, opened 2018-10-05
function split part 4, adding sqs client for function bridge and metrics updates

ryandeivert:
to: @strcrzy , @chunyong-lin 
size: large

Note: ignore the number of commits - I merged instead of rebasing.

## Background

Follow up to #823

This PR adds the `SQSClient` for sending data between the classifier and rules engine functions. It also introduces support for custom metrics via our metric filters logging.

## Changes

* Adding SQSClient class
* Updating `Classifier` class to send classified logs to SQS via client (before sending to firehose) with the relevant data needed by the rules engine.

## Testing

Updating all tests and adding coverage for `SQSClient`


stream_alert/classifier/clients/sqs.py
```diff
@@ -0,0 +1,319 @@
...
+            batched_messages (list): A list of messages that are already serialized to json
+                to be sent to the Rules Engine function
+        Returns:
+            bool: True if the request was successful, False otherwise
+        """
+        @backoff.on_predicate(backoff.fibo,
+                              lambda resp: len(resp.get('Failed', [])) > 0,
+                              max_tries=self.MAX_BACKOFF_ATTEMPTS,
+                              max_value=self.MAX_BACKOFF_FIBO_VALUE,
+                              jitter=backoff.full_jitter,
```
chunyong-lin: This is [default jitter algorithm](https://github.com/litl/backoff/blob/master/backoff/_decorator.py#L15). We can remove this line.

stream_alert/classifier/normalize.py
```diff
@@ -111,14 +113,9 @@ def load_from_config(cls, config):
         if cls._types_config:
             return cls  # config is already populated
 
-        if 'types_new' not in config:  # TODO: change this and below to types
+        if 'normalized_types' not in config:
```
chunyong-lin: Can we `normalized_types` as a constant?

stream_alert/shared/config.py
```diff
@@ -188,3 +188,20 @@ def _validate_config(config):
 
                 if not entity_attrs['logs']:
                     raise ConfigError('List of \'logs\' is empty for entity: {}'.format(entity))
+
+    if 'ioc_types' in config:
```
chunyong-lin: Can we have `ioc_types` as a constant, too?

stream_alert/shared/config.py
```diff
@@ -188,3 +188,20 @@ def _validate_config(config):
...
+        for ioc_type, normalized_keys in config['ioc_types'].iteritems():
+            for normalized_key in normalized_keys:
+                if not any(normalized_key in set(log_keys)
+                           for log_keys in config['normalized_types'].values()):
+                    raise ConfigError(
+                        'IOC key \'{}\' within IOC type \'{}\' must be defined for at least '
+                        'one log type in normalized types'.format(normalized_key, ioc_type)
+                    )
+
+    if 'normalized_types' in config:
```
chunyong-lin: Set `normalized_types` as a constant?

stream_alert/shared/metrics.py
```diff
@@ -84,6 +88,28 @@ class MetricLogger(object):
         },
         ALERT_PROCESSOR_NAME: {},   # Placeholder for future alert processor metrics
         ATHENA_PARTITION_REFRESH_NAME: {},  # Placeholder for future athena processor metrics
+        CLASSIFIER_FUNCTION_NAME: {
+            FAILED_PARSES: (_default_filter.format(FAILED_PARSES),
```
chunyong-lin: I believe we need to remove metrics in `CLASSIFIER_FUNCTION_NAME` from `RULE_PROCESSOR_NAME` below.

stream_alert/classifier/classifier.py
```diff
@@ -233,12 +238,15 @@ def run(self, records):
...
         if self.data_retention_enabled:
-            self._firehose_client.send(self._payloads)
+            self.firehose.send(self._payloads)
 
-        # Only log rule info here if this is not running tests
+        # Only log rule info here if this is deployed in Lambda
         # During testing, this gets logged at the end and printing here could be confusing
         # since stress testing calls this method multiple times
-        if self._verbose:
+        if self._in_lambda:
```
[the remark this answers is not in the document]
ryandeivert: good catch! must have messed up my rebase 😢

stream_alert/classifier/clients/sqs.py
ryandeivert: good catch!!

stream_alert/shared/metrics.py
ryandeivert: I will do this when all of the rule processor code is removed - this will happen in one later PR :)

PR #832, opened 2018-10-20
terraform for rules engine function and generation code

ryandeivert:
to: @chunyong-lin 

## Background

Building on previous PRs #821, #822, #823, #824, #826, #829

## Changes

* Adding `tf_rules_engine` terraform module that contains IAM and other resources needed for the rules engine function.
* Adding `RulesEnginePackage` class for handling rules engine packaging.

## Testing

Adding test for rules engine terraform generation.


terraform/modules/tf_rules_engine/main.tf
```diff
@@ -0,0 +1,19 @@
+// SNS Topic Subscription: Subscribe the Classifier to configured SNS topics
```
chunyong-lin: Looks these code is for Classifier. Can you remove it?

ryandeivert: yes!! good catch :)

PR #844, opened 2018-11-20
[misc] Standardizes the logs.json format

Ryxias:
cc: @chunyong-lin 
size: small
resolves #???

## Background

https://jira.airbnb.biz/browse/CSIRT-368

## Changes

* The `conf/logs.json` now is to be sorted alphabetically by schema key. This commit sorts it.
* A unit test has been added that ensures the `conf/logs.json` is sorted properly.
* A new script has been added that sorts the schema file, runnable with `$ python test/scripts/sort_configs.py`

### DIFF
--- a/conf/logs.json
+++ b/conf/logs.json
@@ -1,4 +1,40 @@
 {
+  "aliyun:actiontrail": {
+    "schema": {
+      "eventId": "string",
+      "eventVersion": "string",
+      "acsRegion": "string",
+      "requestParameters": {},
+      "referencedResources": {},
+      "recipientAccountId": "string",
+      "eventType": "string",
+      "eventTime": "string",
+      "apiVersion": "string",
+      "eventName": "string",
+      "userIdentity": {},
+      "userAgent": "string",
+      "eventSource": "string",
+      "requestId": "string",
+      "serviceName": "string",
+      "sourceIpAddress": "string",
+      "errorCode": "string",
+      "errorMessage": "string",
+      "additionalEventData": {}
+    },
+    "parser": "json",
+    "configuration": {
+      "optional_top_level_keys": [
+        "apiVersion",
+        "errorCode",
+        "errorMessage",
+        "acsRegion",
+        "referencedResources",
+        "recipientAccountId",
+        "userAgent",
+        "additionalEventData"
+      ]
+    }
+  },
   "binaryalert": {
     "schema": {
       "FileInfo": {},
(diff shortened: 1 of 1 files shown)


tests/scripts/sort_configs.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class JsonFileSorter(object):
+    """
+    Tests that the conf.json file is formatted properly and is sorted alphabetically
+    on the top-level key
+    """
+
+    def __init__(self):
+        self._logger = logging.getLogger(SCRIPT_NAME)
```
ryandeivert: since this is a standalone script (not a package), I'd suggest just using the `root` logger here:

`self._logger = logging.getLogger()`

tests/scripts/sort_configs.py
```diff
@@ -0,0 +1,42 @@
...
+    def sort_json_file(self, file_path):
+        self._logger.info('Sorting file: {}...'.format(file_path))
+
+        with open(file_path, 'r') as infile:
+            original_text = infile.read().strip()
+
+        # Load the JSON document using OrderedDict so that it preserves original ordering
+        schema = json.loads(original_text, object_pairs_hook=OrderedDict)
+
+        ordered_schema = OrderedDict(sorted(schema.items(), key=lambda k: k[0]))
```
ryandeivert: we need to be careful w/ csv-type log schemas. the reason OrderedDict is used in the first place here is that csv keys must retain their order. will your script account for keeping key order (regardless of canonical sort order) when `parser` == `csv`?

ryandeivert: I do notice the csv ones were not reordered in your logs.json, but don't see anything that accounts for it.. 🤔

Ryxias: I actually ran into this exact problem. Using `json.loads(....)` without the `OrderDict` hook results a bunch of nested `dicts` to be returned. Even if I wrap it with an `OrderedDict` later, the ordering is already destroyed on-load, so after "sorting" I noticed in the output that all of the internal keys were jumbled up.

I found using the `object_pairs_hook` will return a bunch of nested `OrderedDicts` with the original ordering of the internal keys preserved. The sorting is only concerned with sorting the top-level keys. To my understanding this *should* preserve original ordering of internal keys. 

I'll update the comment to be clearer—I didn't know about the CSV key ordering and I think that's an important detail..

Ryxias:
Ah crap. I accidentally merged it without actually committing my fixup...

PR #861, opened 2018-12-18
bumping version to 2.1.3

ryandeivert:
to: @chunyong-lin 

Version bump from `2.1.2` to `2.1.3`


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,11 +1,11 @@
-to: @airbnb/streamalert-maintainers
```
chunyong-lin: I would prefer to keep `@airbnb/streamalert-maintainers` either in `to` or `cc` field. Otherwise, it is difficult to find out who to tag to.

ryandeivert: sorry I didn't mean to remove that - adding back

ryandeivert: but adding to the cc line instead of to

Issue #240, opened 2017-08-07
Use SSM/Parameter Store for storing output config secrets


ryandeivert:
## Why
When adding a new output configuration via `stream_alert_cli.py`, secrets are currently encrypted with a KMS key and then stored in a `<prefix>.streamalert.secrets` bucket in S3. This involves a handful of operations and adds some unnecessary complexity. This would also make it easier to access/decrypt the secrets when sending to a given output. The encrypted blob from S3 would not have to be cached on disk and then decrypted with KMS - this could all happen in memory with the SSM get_parameter function.

## Requested Changes
Utilize [AWS's SSM via boto3](http://boto3.readthedocs.io/en/latest/reference/services/ssm.html) to store sensitive information.
The [`put_parameter`](http://boto3.readthedocs.io/en/latest/reference/services/ssm.html#SSM.Client.put_parameter) function can be used in conjunction with a KMS keyID to encrypt the sensitive integration key info being stored.
* This change should account for current secrets being stored in S3 and offer a way to migrate them to SSM.
* Support for deleting old output configurations via the cli should also be added.
  * Renaming of configurations should be considered as well.

jack1902:
Will start work on this tomorrow evening :). from my understand of SA, the following locations utilise this secret_store:
- alert_processor (via outputs.get_creds)
- streamalert_cli (configuring outputs)

Are there any locations this s3 secret_store is accessed ? (i could hunt through the code base but tribal knowledge is a quick win)

jack1902:
@ryandeivert I believe this can be closed?

ryandeivert:
fixed with #1142

PR #875, opened 2019-01-14
Separates CredentialProvider from OutputDispatcher

Ryxias:
to: @ryandeivert 

## Background

Implementing a new `OutputDispatcher` requires a lot of in-depth understanding of how the base class is implemented, and testing it has dependencies on AWS S3, AWS KMS, `os`, and the filesystem. To write a good test, you will need to know how to mock all three of these out which is lot of boilerplate.

Example: [`TestPagerDutyOutput.setup/teardown`](https://github.com/airbnb/streamalert/blob/master/tests/unit/stream_alert_alert_processor/test_outputs/test_pagerduty.py#L49-L64)

## Changes

I refactored `OutputDispatcher` and decoupled the credentials logic out of the base class and into a new class, `OutputCredentialsProvider`. This class now becomes solely responsible for providing `_load_creds()` functionality to the parent class. 

## Benefit?

Now, when writing unit tests, instead of having to mock out many different things as boilerplate, we can simply use mock to `patch` the `OutputCredentialsProvider` class to always return stub credentials.

Example: [Commit showing before/after](https://github.com/airbnb/streamalert/pull/875/commits/b3245b70096ddca6514d0b66813ec821be41fcb5)

This reduces the amount of code **slightly**, and makes it **slightly** easier to understand.

Additionally, this PR contains all changes in #878, with the new notion of `Drivers`. This opens up an easier path in the future to build an integration with SSM (#240).

## Other (Unrelated?) Changes

I [DRY'd out some code](https://github.com/airbnb/streamalert/pull/875/commits/f8d368ccd07c83644dd63c5ae6164a6096797d7b) here and there, and removed some private (?) methods from `OutputDispatcher` that I didn't see being used in production code, in order to simplify the class's interface and its inheritance model.

## Testing

I'm not quite done yet but I'm adding new unit tests which are passing. Will paste results here shortly.

`tests/scripts/unit_tests.sh`
```
stream_alert/threat_intel_downloader/__init__.py          0      0   100%
stream_alert/threat_intel_downloader/exceptions.py        4      0   100%
stream_alert/threat_intel_downloader/main.py            134      1    99%   348
-----------------------------------------------------------------------------------
TOTAL                                                  4414     94    98%
[success] 1.76% tests.unit.stream_alert_alert_merger.test_main.TestAlertMerger.test_dispatch: 0.3028s
[success] 1.45% tests.unit.streamalert.classifier.clients.test_firehose.TestFirehoseClient.test_send_batch: 0.2494s
[success] 1.42% tests.unit.stream_alert_shared.test_rule.RuleImportTest.test_path_to_module: 0.2436s
[success] 1.05% tests.unit.stream_alert_apps.test_apps.test_aliyun.TestAliyunApp.test_date_formatter: 0.1794s
[success] 0.99% tests.unit.stream_alert_alert_processor.test_outputs.test_pagerduty.TestPagerDutyIncidentOutput.test_dispatch_success_no_context: 0.1701s
[success] 0.99% tests.unit.stream_alert_shared.test_config.TestConfigLoading.test_load_invalid_file: 0.1696s
[success] 0.85% tests.unit.stream_alert_alert_processor.test_outputs.test_github.TestGithubOutput.test_dispatch_failure: 0.1454s
[success] 0.84% tests.unit.stream_alert_shared.test_rule.RuleImportTest.test_import_rules: 0.1447s
[success] 0.81% tests.unit.stream_alert_shared.test_config.TestConfigLoading.test_load_include: 0.1398s
[success] 0.81% tests.unit.stream_alert_shared.test_alert_table.TestAlertTable.test_get_alert_records: 0.1383s
Ran 882 tests in 17.284s

OK
```

`tests/scripts/rule_test.sh`
```
Summary:

Total Tests: 77
Pass: 77
Fail: 0

[INFO 2019-01-14 15:48:57,377 (stream_alert_cli.runner:85)]: Completed
```


Ryxias:
Per @ryandeivert 's comment worth considering #240

stream_alert/alert_processor/outputs/output_base.py
```diff
@@ -194,46 +172,75 @@ def _load_creds(self, descriptor):
...
-            client = boto3.client('s3', region_name=self.region)
+            client = boto3.client('s3', region_name=self._region)
             with open(cred_location, 'wb') as cred_output:
-                client.download_fileobj(self.secrets_bucket,
-                                        self.output_cred_name(descriptor),
-                                        cred_output)
+                client.download_fileobj(
+                    self._secrets_bucket,
+                    self.get_formatted_output_credentials_name(self._service_name, descriptor),
+                    cred_output
```
ryandeivert: I was thinking about this recently and think that we could actually benefit slightly from not ever writing these to disk (`cred_output`). Instead, we could use [tempfile.SpooledTemporaryFile](https://docs.python.org/2.7/library/tempfile.html#tempfile.SpooledTemporaryFile). This would store the blob in memory, which could then be passed to the `kms_decrypt` method for decryption.

Then, the actual output subclass could store a dict of outputs + decrypted creds in memory.

Thoughts on this?

Ryxias: This sounds like a good idea, and it doesn't sound too hard. I will take a look and add in the implementation if it doesn't take too long

stream_alert/alert_processor/outputs/output_base.py
```diff
@@ -116,62 +116,38 @@ def get_all_outputs(cls):
...
-        self.region = REGION
-        self.secrets_bucket = '{}.streamalert.secrets'.format(os.environ['STREAMALERT_PREFIX'])
-        self.config = config
-
-    @staticmethod
-    def _local_temp_dir():
-        """Get the local tmp directory for caching the encrypted service credentials
+       OutputDispatcher implementations may require credentials to authenticate with an external
+       gateway. All credentials for OutputDispatchers are to be stored in a single bucket on AWS S3
+       and are encrypted with AWS KMS. When the OutputDispatchers are booted, this these encrypted
```
ryandeivert: typo: `this these`

also 'booted' might not be the right verbiage :P

stream_alert/alert_processor/outputs/output_base.py
```diff
@@ -116,62 +116,38 @@ def get_all_outputs(cls):
...
-        if os.path.isfile(temp_dir):
-            os.remove(temp_dir)
+    """
 
-        # Create the folder on disk to store the credentials temporarily
-        if not os.path.exists(temp_dir):
-            os.makedirs(temp_dir)
+    def __init__(self, config, defaults, service_name):
+        self._config = config
+        self._region = REGION
```
ryandeivert: this might be silly - but since `REGION` is imported as a global, it doesn't really need to be set as a local/instance property. not a huge deal to leave it though

stream_alert/alert_processor/outputs/output_base.py
```diff
@@ -116,62 +116,38 @@ def get_all_outputs(cls):
...
+        self._account_id = os.environ['AWS_ACCOUNT_ID']
+        self._secrets_bucket = OutputCredentialsProvider.get_s3_secrets_bucket(
+            os.environ['STREAMALERT_PREFIX']
+        )
 
-    def _load_creds(self, descriptor):
+    def load_credentials(self, descriptor):
         """First try to load the credentials from /tmp and then resort to pulling
-        the credentials from S3 if they are not cached locally
+           the credentials from S3 if they are not cached locally
```
ryandeivert: I know this is old (my) code, but since you're in here an updating docstrings, now might be the best time to address this.

We've been trying to adhere to the [Google Python Style Guide](https://github.com/google/styleguide/blob/gh-pages/pyguide.md) as much as possible. That said, the 'format' for docstrings is usually a quick, one-line summary, followed by a blank line, followed by additional detail (if needed). See [here](https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings).

stream_alert/alert_processor/outputs/output_base.py
```diff
@@ -194,46 +172,75 @@ def _load_creds(self, descriptor):
...
         except ClientError as err:
             LOGGER.exception('credentials for \'%s\' could not be downloaded '
-                             'from S3: %s', self.output_cred_name(descriptor),
+                             'from S3: %s',
+                             self.get_formatted_output_credentials_name(self._service_name,
+                                                                        descriptor),
                              err.response)
 
-    def _kms_decrypt(self, data):
+    def kms_decrypt(self, data):
```
ryandeivert: I'm sure there is a reason for this but I can't find it.. why the change to a public method from protected?

stream_alert/alert_processor/outputs/output_base.py
```diff
@@ -243,12 +250,90 @@ def _kms_decrypt(self, data):
...
+        Args:
+            service_name (str): Service name on output class (i.e. "pagerduty", "demisto")
+            descriptor (str): Service destination (ie: slack channel, pd integration)
+
+        Returns:
+            str: Formatted credential name (ie: slack_ryandchannel)
+        """
+        cred_name = str(service_name)
+
+        # should descriptor be enforced in all rules?
```
ryandeivert: we should probably have a "validator" for the keyword arguments passed to the `@rule` decorator that would 'lint' or enforce this, but we don't yet, so please leave this :)

ryandeivert: I just filed an issue for this... #879

stream_alert/alert_processor/outputs/output_base.py
```diff
@@ -243,12 +250,90 @@ def _kms_decrypt(self, data):
...
+        if descriptor:
+            cred_name = '{}/{}'.format(cred_name, descriptor)
+
+        return cred_name
+
+    def get_aws_account_id(self):
```
ryandeivert: since this doesn't really need to be "computed", you could set `self._account_id` to a public property (`self.account_id`) if you'd like and remove this method. Same with the secrets bucket method. Will reduce amount of tests as well. Alternatively, you could use the `@property` for these to make read-only instance properties out of them

stream_alert/alert_processor/outputs/output_base.py
```diff
@@ -243,12 +250,90 @@ def _kms_decrypt(self, data):
...
+
+    # _DEFAULT_REQUEST_TIMEOUT indicates how long the requests library will wait before timing
+    # out for both get and post requests. This applies to both connection and read timeouts
+    _DEFAULT_REQUEST_TIMEOUT = 3.05
+
+    def __init__(self, config):
+        self.region = REGION
+        self.config = config
+
+        self._credentials_provider = OutputCredentialsProvider(config,
```
ryandeivert: I know this format is _technically_ valid (and likely not even your doing) but can we do this instead:

```
...
    self._credentials_provider = OutputCredentialsProvider(
        config,
        self._get_default_properties(),
        self.__service__
    )
...
```

stream_alert/alert_processor/outputs/output_base.py
```diff
@@ -243,12 +250,90 @@ def _kms_decrypt(self, data):
...
+                                                               self._get_default_properties(),
+                                                               self.__service__)
+
+        self.account_id = self._credentials_provider.get_aws_account_id()
```
ryandeivert: since a instance property already has these values on it (self._credentials_provider), I'd suggest using read-only computed properties that 'fall through' to these. like so:


```
@property
def account_id(self):
    return self._credentials_provider.get_aws_account_id()

@property
def secrets_bucket(self):
    return self._credentials_provider.get_secrets_bucket_name()
```

stream_alert/alert_processor/outputs/output_base.py
Ryxias: This method gets nuked in #878

stream_alert/alert_processor/outputs/output_base.py
Ryxias: This is much better imo. It makes it so much more clear that subclasses intended to use this. I stripped out a lot of other functionality but I noticed that **one** subclass used `account_id`.

stream_alert/alert_processor/outputs/output_base.py
Ryxias: I'll leave as-is; I'm a big proponent of classes having local references of everything they need to operate; a very *Dependency-Injection-centric* style.

stream_alert/alert_processor/outputs/output_base.py
Ryxias: At least for `OutputDispatcher`, what I'm trying to do is to reduce the amount of automatic inheritance.

I noticed that `OutputDispatcher` has a mix of methods; some are endemic to being an alert dispatch abstraction (e.g. `dispatch()`, `_dispatch()` or `get_user_defined_properties()` etc), but others are more like utilities (e.g. `_get_request_retry()`, or `_load_creds()` etc). For the latter, these methods are super convenient to have around for certain subclasses of `OutputDispatcher`, but are useless on the rest.

My plan is to break it up into a series of services that help `OutputDispatcher`. Then, instead of using automatic inheritance, we can use composition or mixins to assemble only the features necessary.

ryandeivert: 👍 I like the idea of using mixins for a lot of this functionality. thanks for the info!

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+        will auto-close.
+
+        Return:
+            file object
+        """
+        file_path = self.get_file_path(descriptor)
+        if not os.path.exists(file_path):
+            os.makedirs(os.path.dirname(file_path))
+
+        return open(file_path, 'a+b')  # read+write and in binary mode
```
ryandeivert: can you point to when we want/need to have `write` perms for the files?

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+        return True
+
+    def save_credentials_into_s3(self, descriptor, credentials, kms_key_alias):
+        """Takes the given credentials, encrypts them, and saves them to AWS S3.
+
+        Notably, this implementation is NOT for the CredentialsCachingDriver interface, as the
+        S3Driver is not a caching driver.
+
+        Args:
+            descriptor (str):
```
ryandeivert: missing some descriptions for these args

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+        return send_creds_to_s3(self._region, self._bucket, s3_key, encrypted_credentials)
+
+    def get_s3_key(self, descriptor):
+        """Returns an appropriate S3 bucket key for credentials relevant to this Output.
+
+        Args:
+            descriptor (str):
+
+        Returns:
+            string
```
ryandeivert: description of args and returns missing

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+        Return:
+            bool
+        """
+        if credentials.is_encrypted():
+            unencrypted_raw_creds = credentials.get_data_kms_decrypted()
+        else:
+            unencrypted_raw_creds = credentials.data()
+
+        key = self.get_storage_key(descriptor)
+        EphemeralUnencryptedDriver.CREDENTIALS_STORE[key] = unencrypted_raw_creds
```
ryandeivert: not a necessary change, but this is a good opportunity to highlight a neat 'trick' in python (that you very well may already be aware of)

this:
`EphemeralUnencryptedDriver.CREDENTIALS_STORE[key] = unencrypted_raw_creds`

can also be done with:
`type(self).CREDENTIALS_STORE[key] = unencrypted_raw_creds`

since the type function returns the _class_ of an instance

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+        return True
+
+    @staticmethod
+    def clear():
+        EphemeralUnencryptedDriver.CREDENTIALS_STORE = {}
```
ryandeivert: You could also just call `EphemeralUnencryptedDriver.CREDENTIALS_STORE.clear()` to flush the dict instead of re-instantiating to a new one

ryandeivert: this may also make more sense to be a `classmethod`. referencing via the class name (`EphemeralUnencryptedDriver`) within the class isn't necessarily _wrong_, it just feels clunky if it can be avoided

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+
+class SpooledTempfileDriver(CredentialsProvidingDriver, FileDescriptorProvider):
+    """Driver for fetching credentials that are stored in memory in file-like objects."""
+
+    SERVICE_SPOOLS = {}
+
+    def __init__(self, service_name, region):
+        self._service_name = service_name
+        self._region = region
+        self._spools = {}
```
ryandeivert: I think this may be a leftover/unused instance attribute.. potentially left after moving to a class property?

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+        """Saves the credentials into a temporary spool.
+
+        Args:
+            descriptor (str):
+            credentials (Credentials):
+
+        Return:
+            bool
+        """
+
```
ryandeivert: extra new line here. also update the docstring above with descriptions please

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+        The Drivers are sequentially checked in the order they are appended to the driver list.
+        """
+
+        # Ephemeral driver
+        ep_driver = EphemeralUnencryptedDriver(self._service_name)
+        self._drivers.append(ep_driver)
+
+        # Fall back onto downloading encrypted credentials from S3
+        s3_driver = S3Driver(self._prefix, self._service_name, self._region, cache_driver=ep_driver)
+        self._core_driver = s3_driver
```
ryandeivert: it seems like the `core_driver` is designed for 'remote' interaction.. ie, reading/writing from S3. might it be more appropriate to name something like `remote_driver` or similar? thoughts?

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+            LOGGER.exception('an error occurred during credentials decryption: %s', err.response)
+            return None
+
+    def encrypt(self, region, kms_key_alias):
+        """Encrypts the current Credentials.
+
+        Calling this method will entirely change the internals of this Credentials object.
+        Subsequent calls to .data() will return encrypted data.
+        """
+        self._region = region
```
ryandeivert: I think you could move this to after the `self.is_encrypted() call, since it's a no-op if that returns. also what does setting the region again here gain us (instead of just passing through to the `kms_encrypt` function

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+        if self.is_encrypted():
+            return
+
+        self._is_encrypted = True
+        if not self._data:
+            return
+
+        creds_json = json.dumps(self._data)
```
ryandeivert: we should use 'minimal' json here when dumping (`json.dumps(self._data, separators=(',', ':'))`) to reduce file size as much as possible. this isn't really a problem with s3, but could be with SSM due to sorta low limits. the gain won't be very huge anyway so maybe not a huge deal

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+import shutil
+import tempfile
+from abc import abstractmethod
+
+import boto3
+from botocore.exceptions import ClientError
+
+from stream_alert.shared.helpers.boto import default_config
+from stream_alert.shared.logger import get_logger
+from stream_alert_cli.outputs.helpers import kms_encrypt, send_creds_to_s3
```
ryandeivert: this import cannot be done here, as this will break when packaged and sent to lamdba. essentially, nothing from the `stream_alert_cli` package should ever be imported into the `stream_alert` module..

if this is something we need to share, you can move to somewhere within the `stream_alert/shared` package, as this _is_ bundled for deployment to lambda. then you can use import `stream_alert.shared` from the `stream_alert_cli` package if need-be

stream_alert_cli/outputs/handler.py
```diff
@@ -62,12 +65,13 @@ def output_handler(options, config):
...
         return output_handler(options, config)
 
-    secrets_bucket = '{}.streamalert.secrets'.format(prefix)
-    secrets_key = output.output_cred_name(props['descriptor'].value)
-
-    # Encrypt the creds and push them to S3
-    # then update the local output configuration with properties
-    if not encrypt_and_push_creds_to_s3(region, secrets_bucket, secrets_key, props, kms_key_alias):
+    provider = OutputCredentialsProvider(config=output.config,
+                                         defaults={},  # defaults are unnecessary for saving
```
ryandeivert: if defaults are 'optional', you can make this an optional keyword argument in the constructor of `OutputCredentialsProvider`

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,654 @@
...
+        load_credentials: Returns a dict of the credentials requested
+        get_local_credentials_temp_dir(): Returns full path to a temporary directory where all
+            encrypted credentials are cached.
+
+    """
+
+    def __init__(self, config, defaults, region, service_name, prefix=None, aws_account_id=None):
+        self._config = config
+        self._region = region
+        self._defaults = defaults
```
ryandeivert: if changing `defaults` to optional (see other comment), you could do `self._defaults = defaults if defaults else {}` if you want it to always be a dictionary

stream_alert/alert_processor/outputs/credentials/provider.py
Ryxias: In the S3Driver, the `offer_fileobj` is offered to the AWS Boto client in the `download_fileobj` API call, which writes the result into a file-like object.

ryandeivert: got it!

stream_alert/alert_processor/outputs/credentials/provider.py
Ryxias: It was a minor design quirk where if you have a `Credentials` object that you constructed in plaintext, if you called `.encrypt()` it would destroy the plaintext "internals", and then you'd pass it around and it'd just be ciphertext + the AWS region needed to decrypt it.

stream_alert/alert_processor/outputs/credentials/provider.py
Ryxias: ah.... I was wondering why there was duplicated code in the `helpers` module and the old Dispatcher code. TIL

stream_alert/alert_processor/outputs/credentials/provider.py
Ryxias: I thought of it like a "Driver who stores the source of Truth" kind of thing. I was thinking this would be where the SSMDriver would replace the S3 driver as the "primary" one. I'd think that different Outputs would have different Core drivers depending on where the credentials are configured (if we do a slow migration).

Ryxias:
@ryandeivert I've added a Fixup that addresses your comments, PTAL

(I also added `stream_alert.shared.helps.aws_api_client.py`)

Ryxias:
OH no i Messed up i put it on the wrong branch... hold on..

stream_alert_cli/outputs/helpers.py
```diff
@@ -15,17 +15,19 @@
 """
 import json
 
-import boto3
 from botocore.exceptions import ClientError
 
 from stream_alert.shared.logger import get_logger
+from stream_alert.shared.helpers.aws_api_client import AwsS3, AwsKms
```
ryandeivert: this import should go before the previous line (`h` < `l`)

stream_alert_cli/outputs/helpers.py
```diff
@@ -58,28 +62,25 @@ def kms_encrypt(region, data, kms_key_alias):
...
         bucket (str): The name of the s3 bucket to write the encrypted credentials to
         key (str): ID for the s3 object to write the encrypted credentials to
         blob_data (bytes): Cipher text blob from the kms encryption
     """
     try:
-        client = boto3.client('s3', region_name=region)
-        client.put_object(Body=blob_data, Bucket=bucket, Key=key)
-
-        return True
+        return AwsS3.put_object(blob_data, bucket=bucket, key=key, region=region)
```
ryandeivert: nice changes - if this function (`send_creds_to_s3`) is in fact not being anywhere else, you can freely remove it

Ryxias:
Done; I removed the deprecated code and added tests to replace the old ones that are gone

PR #878, opened 2019-01-16
Refactors Drivers for OutputCredentialsProvider

Ryxias:
to: @ryandeivert 
related to:
resolves:

## Background

This PR sits on top of #875, in response to [this comment](https://github.com/airbnb/streamalert/pull/875#discussion_r248055606). I split this into a separate PR to make it easier to understand.

## Changes

I split out the new `OutputCredentialsProvider` guts into this new concept—`CredentialsProvidingDriver`. The idea is these drivers each are strategies for fetching Credentials from **somewhere**.

### `CredentialsProvidingDriver`s
I implemented 4 drivers: S3, local files, spooled file, and ephemeral. I tried to design the interface to be plug-and-play-ready for a SSM migration in the future (#240).

### New Way
The new algorithm can be found in `setup_drivers()`. The pseudocode works like this:

* Ask Ephemeral Driver if credentials are already cached in memory. If so, load and return them.
* Ask S3 Driver to load credentials from AWS S3
* S3 credentials are downloaded, encrypted, and saved into a spool using the Spooled Driver
* Credentials are read from the spool, decrypted, and cached into Ephemeral Driver
* Credentials are returned

### Other stuff
I removed a lot of bloat from OutputDispatcher. These old methods were S3-specific when the OutputDispatcher **assumed** all of its credentials came from S3. Now that the OutputDispatchers are de-coupled from S3, these methods are no longer relevant.

## Testing

unit_test
```
--------------------------------------------------------------------------------------------
TOTAL                                                           4570    163    96%
[success] 6.55% tests.unit.streamalert.classifier.clients.test_firehose.TestFirehoseClient.test_send_batch: 0.8364s
[success] 2.40% tests.unit.stream_alert_alert_merger.test_main.TestAlertMerger.test_dispatch: 0.3069s
[success] 2.03% tests.unit.stream_alert_shared.test_config.TestConfigLoading.test_load_all: 0.2598s
[success] 1.57% tests.unit.stream_alert_alert_processor.test_outputs.test_output_base.TestOutputDispatcher.test_load_creds: 0.2000s
[success] 1.54% tests.unit.stream_alert_shared.test_rule_table.TestRuleTable.test_toggle_staged_state_false: 0.1970s
[success] 1.12% tests.unit.stream_alert_cli.test_cli_config.TestCLIConfig.test_aggregate_alarm_creation: 0.1433s
[success] 1.06% tests.unit.stream_alert_cli.test_cli_config.TestCLIConfig.test_add_threat_intel_downloader: 0.1360s
[success] 1.06% tests.unit.stream_alert_cli.test_cli_config.TestCLIConfig.test_toggle_metric: 0.1351s
[success] 1.05% tests.unit.stream_alert_cli.test_cli_config.TestCLIConfig.test_add_threat_intel_without_table_name: 0.1339s
[success] 1.04% tests.unit.stream_alert_cli.test_cli_config.TestCLIConfig.test_cluster_alarm_creation: 0.1332s
Ran 874 tests in 12.904s

OK

```

rule_test
```
Summary:

Total Tests: 77
Pass: 77
Fail: 0
```

### DIFF
--- a/stream_alert/alert_processor/outputs/credentials/provider.py
+++ b/stream_alert/alert_processor/outputs/credentials/provider.py
@@ -106,7 +106,7 @@ class OutputCredentialsProvider(object):
                  for (name, prop) in props.iteritems() if prop.cred_requirement}
 
         credentials = Credentials(creds, False, self._region)
-        return self._core_driver.save_credentials(descriptor, credentials, kms_key_alias)
+        return self._core_driver.save_credentials_into_s3(descriptor, credentials, kms_key_alias)
 
     def load_credentials(self, descriptor):
         """First try to load the credentials from /tmp and then resort to pulling
@@ -243,39 +243,62 @@ class CredentialsProvidingDriver(object):
     @abstractmethod
     def load_credentials(self, descriptor):
         """
+        Implementation that loads the given credentials into a new Credentials object. The
+        behavior can be nondeterministic if has_credentials() is false.
+
         Args:
             descriptor (string): Descriptor for the current output service
 
         Return:
-            Credentials
+            Credentials|None: Returns None when loading fails.
         """
-        pass
 
     @abstractmethod
     def has_credentials(self, descriptor):
         """
+        Implementation that determines whether the current driver is capable of loading the
+        requested credentials
+
         Args:
             descriptor (string): Descriptor for the current output service
 
         Return:
             bool: True if this driver has the requested Credentials, false otherwise.
         """
-        pass
 
 
 class FileDescriptorProvider(object):
+    """Interface for Drivers capable of offering file-handles to aid in download of credentials."""
+
     @abstractmethod
     def offer_fileobj(self, descriptor):
         """
+        Offers a file-like object. The caller is expected to call this method in a with block,
+        and this file-like object is expected to automatically close.
+
         Returns:
              file object
         """
-        pass
 
 
 class CredentialsCachingDriver(object):
+    """Interface for Drivers capable of being used as a caching layer to accelerate the speed
+    of credential loading."""
+
     @abstractmethod
     def save_credentials(self, descriptor, credentials):
+        """
+        Saves the given credentials, such that on a subsequent call of load_credentials(), the
+        same credentials will be loaded.
+
+        Args:
+            descriptor (str): OutputDispatcher descriptor
+            credentials (Credentials): The credentials object to save. Notably, certain drivers are
+                incapable of (or disallowed from) saving credentials that are unencrypted.
+
+        Return:
+            bool: True if saving succeeds. False otherwise.
+        """
         pass
 
 
@@ -375,7 +398,22 @@ class S3Driver(CredentialsProvidingDriver):
            guaranteed to be cold-stored."""
         return True
 
-    def save_credentials(self, descriptor, credentials, kms_key_alias):
+    def save_credentials_into_s3(self, descriptor, credentials, kms_key_alias):
+        """
+        Takes the given credentials, encrypts them, and saves them into a determined S3 bucket
+        and location.
+
+        Notably, this implementation is NOT for the CredentialsCachingDriver interface, as the
+        S3Driver is not a caching driver.
+
+        Args:
+            descriptor (str):
+            credentials (Credentials):
+            kms_key_alias (str):
+
+        Returns:
+            bool: True on success, False otherwise.
+        """
         if credentials.is_encrypted():
             # Don't try saving encrypted credentials, or you will doubly-encrypt them, since this
             # method will encrypt the credentials immediately prior to saving into S3.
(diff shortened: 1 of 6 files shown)


stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,589 @@
...
+            string
+        """
+        return '{}.streamalert.secrets'.format(self._prefix)
+
+
+class LocalFileDriver(CredentialsProvidingDriver, FileDescriptorProvider, CredentialsCachingDriver):
+    """Driver for fetching credentials that are saved locally on the filesystem."""
+
+    def __init__(self, region, service_name):
+        self._region = region
```
ryandeivert: I don't think region will ever be used for the 'local' one right?

stream_alert/alert_processor/outputs/credentials/provider.py
```diff
@@ -0,0 +1,589 @@
...
+    Public methods:
+        load_credentials: Returns a dict of the credentials requested
+        get_local_credentials_temp_dir(): Returns full path to a temporary directory where all
+            encrypted credentials are cached.
+
+    """
+
+    def __init__(self, config, defaults, region, service_name, prefix=None, aws_account_id=None):
+        self._config = config
+        self._region = region
```
ryandeivert: subclasses of `OutputCredentialsProvider` that use 'boto3' clients and need a region could use the `default_config` function to return a botocore config (see: https://github.com/airbnb/streamalert/blob/master/stream_alert/shared/helpers/boto.py#L28)

TBH, I'd advocate for using the `default_config` method anywhere we are creating boto3 clients, since it enforces a shorter timeout (which is better for efficiency/cost in lambda).

Ryxias: 👍 
So we would do something like

```
client = boto3.client('zzz', config=default_config(region=self._region))
```

Instead of
```
client = boto3.client('zzz', region_name=self._region)
```
?

stream_alert/alert_processor/outputs/credentials/provider.py
Ryxias: Unfortunately I need it to pass into the `Credentials` constructor, because that class requires the `region` to use KMS to decrypt the credentials. I honestly could've passed in a `CredentialsFactory` kind of thing that would avoid having to hot-potato around all of these `region`, `service_name` and so on things.

stream_alert/alert_processor/outputs/credentials/provider.py
ryandeivert: no you could omit region entirely from the call to `default_config` (unless it could/would ever be something other than the current region.. which I don't think it would be in this case)

ryandeivert: `client = boto3.client('zzz', config=default_config())`

stream_alert/alert_processor/outputs/credentials/provider.py
ryandeivert: but only the `S3` driver would need to do the decryption right? or am I misunderstanding this?

if that's the case - the S3 driver constructor can take different params than the base (super) class and just call the super's init with the right params also. I could be misunderstanding the inheritance model, though

stream_alert/alert_processor/outputs/credentials/provider.py
Ryxias: I remembered something when I was confirming whether or not region can be globally configured; what about on the [CLI handler](https://github.com/airbnb/streamalert/pull/878/files#diff-21436558def6ebba39502261900e910f)? It looks like the `region` (among other configurations) is passed in from ... somewhere. How does the CLI handler know which `REGION` the code runs in? Is it inferred from the `aws2fa.sh` session?

stream_alert/alert_processor/outputs/credentials/provider.py
Ryxias: Not exactly; the S3 driver (none of the "drivers") actually do any decryption on their own. I delegated **decryption** to this `Credentials` class.

I didn't actually use any traditional inheritance. The three "base" classes are:

* [FileDescriptorProvider](https://github.com/airbnb/streamalert/pull/878/files#diff-4463a710593722ea77bc4209c8fc8fcdR202)
* [CredentialsCachingDriver](https://github.com/airbnb/streamalert/pull/878/files#diff-4463a710593722ea77bc4209c8fc8fcdR216)
* [CredentialsProvidingDriver](https://github.com/airbnb/streamalert/pull/878/files#diff-4463a710593722ea77bc4209c8fc8fcdR172) (The important one)

The implementations of `CredentialsProvidingDriver` are supposed to return `Credentials`, which can be `is_encrypted()` or not. The idea is (was?) to allow credentials to be passed between drivers while still encrypted. The only thing the driver needs to know is whether or not `Credentials` are encrypted "at rest" or not.

This is why all of the Drivers need a `region`; because they need to pass it to the `Credentials` class to be able to decrypt the ciphertext after being read from disk/memory/whatever. The exception is the `EphemeralDriver`, which is the only unencrypted-at-rest Driver.

Issue #876, opened 2019-01-15
[bug] update-auth function fail due to empty return from handler.py


GavinElder:
### Description

There is an issue with `app_info['function_name']` being returned as a non-type for use in side of the `python manage.py app update-auth` function. 

https://github.com/airbnb/streamalert/blob/cb7e09100081799a944cb0ea29676526123bfda2/stream_alert_cli/apps/handler.py#L70

```shell 
python manage.py app update-auth -c test -n $appName e_app
[INFO 2019-01-14 17:02:11,959 (stream_alert_cli.runner:58)]: Issues? Report here: https://github.com/airbnb/streamalert/issues
Traceback (most recent call last):
  File "manage.py", line 1510, in <module>
    main()
  File "manage.py", line 1506, in main
    sys.exit(not cli_runner(options))
  File "/Users/gavin/internal-streamalert/stream_alert_cli/runner.py", line 84, in cli_runner
    result = cmds[args.command](args)
  File "/Users/gavin/internal-streamalert/stream_alert_cli/runner.py", line 61, in <lambda>
    'app': lambda opts: app_handler(opts, config),
  File "/Users/gavin/internal-streamalert/stream_alert_cli/apps/handler.py", line 70, in app_handler
    app_info['function_name'] = '_'.join([app_info.get(value) for value in func_parts] + ['app'])
TypeError: sequence item 2: expected string, NoneType found
```

### Steps to Reproduce

- Create App
- Attempt to change auth.
**Failure**
- Change line 70 in handler.py to `app_info['function_name'] = ['app']` 
**Works**

## Desired Change

Logic to handle empty returns.

PR #884, opened 2019-01-19
fixing bug related to updating app auth information

ryandeivert:
to: @chunyong-lin 
resolves: #876

## Background

See #876

## Changes

* Properly looking up the 'existing' app by name in the config, and using its function name to update the auth info.
* Fixing a type in description used when requesting auth info for Duo (and updating docs to match).

## Testing

Ran through this a currently deployed app and works as expected.


ryandeivert:
hi @GavinElder you should be able to check the _value_ of the parameters in SSM, here:

https://console.aws.amazon.com/systems-manager/parameters

PR #898, opened 2019-02-09
[OutputDispatchers] Standardizes all outputs on alert.publish_for()

Ryxias:
to: @ryandeivert 
cc: @airbnb/streamalert-maintainers

## Background

This is done as step 1 in preparation for *Alert Publishers*. In order to implement a proper interface between *Alerts* and *Outputs* we need to have this layer communicate across a common vocabulary.

### OutputDispatchers standardize around `publish_for()`
Instead of referencing fields on the alert directly, all `OutputDispatchers` now reference `alert.publish_for(---, ----)`. Currently it has an empty implementation, but in the near future *Alert Publishers* will be able to dynamically change the output.

### Publishers Control Certain Output Fields
Also I'm *experimenting* with having Publishers control more aspects of each Output layer... you can see a couple examples [here](https://github.com/airbnb/streamalert/pull/898/files#diff-6e737c09dbb041674665ee140c439226R52) and [here](https://github.com/airbnb/streamalert/pull/898/files#diff-0f851c392e56eb2deb2a2cec2fe04a61R231). I'm not sure if I like this *yet* but I think it's on track to fix the TIR issue of "Redundant 'StreamAlert Rule Triggered -'  headers".

### (WIP?) Remove `alert` from `_dispatch()` interface
**One thing** that I did that I'm not 100% certain of. For a lot direct references to `alert.rule_name` or alert.rule_description`, I changed them to use the `publish_for()`, knowing that an incorrectly coded Publisher might forget to include these fields. I did this because I wanted to eventually remove `alert` from the `_dispatch(alert, descriptor)` function and replace it with `_dispatch(alert_publication, descriptor)`... but it seems a little contrived. Thoughts? Example can be found [HERE](https://github.com/airbnb/streamalert/pull/898/files#diff-be02b00f457df67b4678d87aac903176R97). 

I could go ahead and keep `alert` in the `_dispatch()` interface. This seems more sensible when outputs are *directly* referencing Alert fields. The risk here is future Outputs might accidentally go around the `publish_for()` abstraction layer. The alternative is I could remove it completely, but now the risk is a badly written publisher can completely omit these fields which may make alerts render weirdly in outputs (which... might be as intended?).

## Testing

```
--------------------------------------------------------------------------------------------
TOTAL                                                           4754    104    98%
[success] 3.18% tests.unit.stream_alert_alert_processor.test_outputs.credentials.test_provider.TestS3Driver.test_save_credentials_into_s3_blank_credentials: 0.4081s
[...]

PR #899, opened 2019-02-11
[Output] Standardize outputs around `publish_for()` method

Ryxias:
to: @ryandeivert 

## Background

Small improvement over #898

This is done as step 1 in preparation for _Alert Publishers_. In order to implement a proper interface between _Alerts_ and _Outputs_ we need to have this layer communicate across a common vocabulary.

### OutputDispatchers standardize around `publish_for(output, descriptor)`
Instead of referencing fields on the alert directly, all `OutputDispatchers` now reference `alert.publish_for(---, ----)`. Currently it has an empty implementation, but in the near future _Alert Publishers_ will be used to modify the return value of this.

This (more or less) replaces `output_dict()`.

### OutputDispatchers offer a default presentation, but can be overridden using magic fields
All OutputDispatchers offer a simple default integration with some default summary/title/priority/severity. 

However, they also can be overridden by certain fields. You can see an example of this here: https://github.com/airbnb/streamalert/pull/899/files#diff-1102db3770def41b73a2a49ea0304899R347.

This allows Outputs to be *both* easy to use as well as highly customizable.

### Demisto Outputs change _slightly_
Some Demisto label keys have moved around. I'm fairly confident nobody is using Demisto yet, so this should be okay for now.

### DIFF
--- a/stream_alert/alert_processor/outputs/demisto.py
+++ b/stream_alert/alert_processor/outputs/demisto.py
@@ -71,7 +71,7 @@ class DemistoOutput(OutputDispatcher):
         if not creds:
             return False
 
-        request = DemistoRequestAssembler.assemble(alert.publish_for(self, descriptor), descriptor)
+        request = DemistoRequestAssembler.assemble(alert.publish_for(self, descriptor))
         integration = DemistoApiIntegration(creds, self)
 
         LOGGER.debug('Sending alert to Demisto: %s', creds['url'])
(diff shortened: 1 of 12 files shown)


stream_alert/alert_processor/outputs/demisto.py
```diff
@@ -213,53 +212,82 @@ def custom_fields(self):
...
         Args:
-            alert (Alert): Alert instance which triggered a rule
+            alert (Alert): Instance of the alert
+            alert_publication (Dict): Published alert data of the alert that triggered a rule
             descriptor (str): Output descriptor
 
         Returns:
             DemistoCreateIncidentRequest
         """
+        # Default presentation values
```
ryandeivert: This stuff for demisto looks new (not refactor) right? can you call it out in the PR body

Ryxias:
Thanks @ryandeivert

PR #900, opened 2019-02-12
Adds Alert Publisher framework

Ryxias:
to: @ryandeivert 

## Background

Introducing *Alert Publishers*! 

The concept behind Alert Publishers is to offer a flexible framework for transforming alerts prior to dispatching to output providers. This allows you to customize the look & feel of alerts that arrive in each 

## Changes

* I removed `publish_for()` because the `publish()` method cannot be on the `Alert` class. This is because the `Alert` class is *shared*, whereas all of the publisher stuff is specifically for `alert_processor` (and thus cannot be loaded by the lambda function).
* I refactored all outputs to use a new method, `publish_alert(alert: Alert, output: OutputDispatcher, descriptor: str) -> dict`. This is a convenience method that wraps all of the Alert publisher stuff
* I tweaked the Slack output integration slightly to accommodate for custom payloads; this is in alignment with other outputs
* There is a a new `publishers/` directory. Most Publisher code is located under this directory. It's now included in the `package.py` and is deployed to the rules engine and alert processor

## Design

* Define new publishers as *functions* or *classes* that inherit from `AlertPublisher`
* Register these publishers by putting them under the `publishers/` directory and decorating them with `@Register`
* Assign publishers to Rules on the `@Rule` annotation, using a new `publisher=` argument; use references to the *function* or *class*
* The RulesEngine automatically converts references into fully qualified string names, and saves them into the *DynamoDB* `alerts` table, on a new `publishers` field
* The AlertProcessor fetches the `alert` with the new field
* `publish_alert()` uses the requested Publishers to create a JSON dict for outputs to read

## Testing

I deployed and tested this on stage.


Ryxias:
## Outstanding problems

Currently, if you provide a blanket publisher:

```
@Rule( ...
  outputs=['demisto', 'pagerduty'],
  publishers='blank',
  ...)
```

This syntax applies the publisher to ALL outputs, including required outputs (AWS Firehose). We do not want this because we do not want to transform the record prior to saving into Firehose. We need to (1) force the required_output(s) to always use DefaultPublisher, or we need to (2) at time of creating the Alert record, immediately standardize the publisher syntax to `output —> [publishers]`.

stream_alert/alert_processor/helpers.py
```diff
@@ -30,3 +31,81 @@ def elide_string_middle(text, max_length):
...
+        #   apply all publishers to all outputs + descriptors
+        publisher_names += alert_publishers
+    elif isinstance(alert_publishers, dict):
+        # Case 3: The publisher is a dict mapping output strings -> strings or list of strings
+        #   apply only publishers under the correct output key. We look under 2 keys:
+        #   one key that applies publishers to all outputs for a specific output type, and
+        #   another key that applies publishers only to outputs of the type AND matching
+        #   descriptor.
+
+        # FIXME (derek.wang)
```
Ryxias: ❗️

publishers/core.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+        publisher_names = publisher_names or []
+        publishers = []
+
+        for publisher_name in publisher_names:
+            publisher = cls.get_publisher(publisher_name)
+            if publisher:
+                publishers.append(publisher)
+
+        if len(publishers) <= 0:
```
ryandeivert: python is silly and you could just do `if not publishers:`

stream_alert/alert_processor/helpers.py
```diff
@@ -30,3 +31,81 @@ def elide_string_middle(text, max_length):
...
+    Returns:
+        AlertPublisher
+    """
+
+    alert_publishers = alert.publishers
+    publisher_names = []
+    if isinstance(alert_publishers, basestring):
+        # Case 1: The publisher is a single string.
+        #   apply this single publisher to all outputs + descriptors
+        publisher_names.append(alert_publishers)
```
ryandeivert: is this still accurate? string literal publisher names and not classes/functions?

stream_alert/alert_processor/helpers.py
```diff
@@ -30,3 +31,81 @@ def elide_string_middle(text, max_length):
...
+        #   one key that applies publishers to all outputs for a specific output type, and
+        #   another key that applies publishers only to outputs of the type AND matching
+        #   descriptor.
+
+        # FIXME (derek.wang)
+        # this is here because currently the OutputDispatcher sits in a __init__.py module that
+        # performs on-demand loading of the other output classes. If you load this helper before
+        # the output classes are loaded, it creates a cyclical dependency
+        # helper.py -> output_base.py -> __init__ -> komand.py -> helper.py
+        # This is a temporary workaround
```
ryandeivert: we should address this :)

stream_alert/alert_processor/helpers.py
```diff
@@ -30,3 +31,81 @@ def elide_string_middle(text, max_length):
...
+            described_output_name = '{}:{}'.format(output_service_name, descriptor)
+            if described_output_name in alert_publishers:
+                publisher_name_or_names = alert_publishers[described_output_name]
+                if isinstance(publisher_name_or_names, list):
+                    publisher_names += publisher_name_or_names
+                else:
+                    publisher_names.append(publisher_name_or_names)
+
+            # Then load output-specific publishers second
+            if output_service_name and output_service_name in alert_publishers:
```
ryandeivert: seems like a lack of `output_service_name` should result in an immediate failure here no?

stream_alert/alert_processor/helpers.py
```diff
@@ -30,3 +31,81 @@ def elide_string_middle(text, max_length):
 
     half_len = (max_length - 5) / 2  # Length of text on either side.
     return '{} ... {}'.format(text[:half_len], text[-half_len:])
+
+
+def publish_alert(alert, output, descriptor):
```
ryandeivert: I know these are 'publishers' and all... but the calls below of `publication = publish_alert(...)` threw me for a second. `publish_alert` seems almost too close to `dispatch`.. would you be open to something a little less 'transactional', like `compose_alert` for this function name? (not super set on this, just thinking out loud).

stream_alert/rules_engine/rules_engine.py
```diff
@@ -211,6 +207,119 @@ def _rule_analysis(self, payload, rule):
...
+            """Standardizes a requested publisher into a string name
+
+            Requested publishers can be either the fully qualified string name, OR it can be a
+            direct reference to the function or class.
+            """
+            if not is_publisher_declaration(string_or_reference):
+                LOGGER.error('Invalid publisher requested: %s', string_or_reference)
+                return None
+
+            if isinstance(string_or_reference, basestring):
```
ryandeivert: Maybe I misunderstood during this whole review and the intention is to support both string literal references and objects.

stream_alert/rules_engine/rules_engine.py
```diff
@@ -211,6 +207,119 @@ def _rule_analysis(self, payload, rule):
...
+                )
+
+            if AlertPublisherRepository.has_publisher(publisher_name):
+                return publisher_name
+
+            LOGGER.warning('Requested publisher named (%s) is not registered.', publisher_name)
+
+        def is_publisher_declaration(string_or_reference):
+            """Returns TRUE if the requested publisher is valid (a string name or reference)"""
+            return isinstance(string_or_reference, basestring) \
```
ryandeivert: Trailing `\` are frowned upon (in the google python style guide). Please wrap this statement with parenthesis like:
```
return (
    isinstance(string_or_reference, basestring) or
    AlertPublisherRepository.is_valid_publisher(string_or_reference)
)
```

stream_alert/rules_engine/rules_engine.py
```diff
@@ -211,6 +207,119 @@ def _rule_analysis(self, payload, rule):
...
+            return None
+
+        configured_publishers = {}
+
+        for output in requested_outputs:
+            output_service = output.split(':')[0]
+
+            assigned_publishers = []
+
+            if is_publisher_declaration(requested_publishers):
```
ryandeivert: This code looks similar to code in the `_assemble_alert_publisher_for_output`. Any chance it could be made generic and shared? (maybe I'm totally wrong.. this is a long review).

stream_alert/shared/alert.py
```diff
@@ -215,29 +228,15 @@ def output_dict(self):
             'log_source': self.log_source or '',
             'log_type': self.log_type or '',
             'outputs': list(sorted(self.outputs)),  # List instead of set for JSON-compatibility
```
ryandeivert: This is your code, but the `sorted` function already returns a list.. casting to list again just copies the list again..

docs/source/publishers.rst
```diff
@@ -0,0 +1,254 @@
...
+    outputs=['pagerduty', 'slack'],
+    publishers=[publisher_1, publisher_2]
+  )
+  def my_rule(rec):
+    # ...
+
+The ``publishers`` argument is a structure containing references to **Publishers** and can follow any of the
+following structures:
+
+**String, Publisher Name**
```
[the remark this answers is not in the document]
Ryxias: I messed up and forgot to update the docs; the syntax now supports both string literals and function/class references. I will update the docs

docs/source/publishers.rst
```diff
@@ -0,0 +1,254 @@
...
+
+  publishers={
+    'pagerduty:analyst': [publisher_1, publisher_2],
+    'pagerduty': [publisher_3, publisher_4],
+    'demisto': other_publisher,
+  }
+
+When using this syntax, publishers under each key will be applied to their matching outputs. Publisher keys
+with specific outputs (e.g. ``pagerduty:analyst``) are loaded first, before publisher keys that pertain
+to groups of outputs (e.g. ``pagerduty``).
```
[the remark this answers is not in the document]
Ryxias: image

docs/source/publishers.rst
```diff
@@ -0,0 +1,254 @@
...
+
+When using this syntax, all given publishers will be applied to all outputs.
+
+
+**Dict mapping Output to Publisher Names**
+
+.. code-block:: python
+
+  publishers={
+    'pagerduty:analyst': [publisher_1, publisher_2],
```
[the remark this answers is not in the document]
Ryxias: Interesting... perhaps if `outputs` is omitted but `publishers` are provided in the *output*—>*publisher* dict format, then `outputs` defaults to inheriting all keys from the `publishers`?

I agree `publishers` and `outputs` can be redundant. I'll implement the following logic:

* Only applies if `publishers=` is in the DICT (let's call it "explicit") format
* If both `outputs=` and `publishers=` are provided, and a specific *output:descriptor* setting in `publishers=` is not provided in `outputs=`, then a `LOGGER.warning` is sent (or should this be a `raise`?)
* If `publishers=` is provided and `output=` is completely omitted, the alert's outputs will be implicitly loaded from any specific *output:descriptor* settings on `publishers=`

I'll open a jira ticket for this enhancement for next iteration

publishers/core.py
```diff
@@ -0,0 +1,227 @@
...
+    """A service that loads all publishers in the publishers/ directory."""
+    _is_imported = False
+    _is_importing = False
+
+    @classmethod
+    def import_publishers(cls):
+        if cls._is_imported or cls._is_importing:
+            return
+
+        cls._is_importing = True
```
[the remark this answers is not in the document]
Ryxias: I... err... *think* I was getting stuck on a cyclical module loading dependency at one point, where I implemented this as a safeguard. I actually do not think there's a danger of cycles anymore so I can remove this

publishers/core.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class AlertPublisherRepository(object):
+    """A repository mapping names -> publishers"""
+    _publishers = {}
+
+    @staticmethod
+    def is_valid_publisher(class_or_function):
+        """Returns TRUE if the given reference can be registered as a publisher"""
+        if isclass(class_or_function) and issubclass(class_or_function, AlertPublisher):
```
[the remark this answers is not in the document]
Ryxias: I'll test this... 

_The arduousness of python polymorphism makes me regret baking everything into classes._

publishers/core.py
```diff
@@ -0,0 +1,227 @@
...
+    def __init__(self, function):
+        self._function = function
+
+    def publish(self, alert, publication):
+        return self._function(alert, publication)
+
+
+class AlertPublisherRepository(object):
+    """A repository mapping names -> publishers"""
+    _publishers = {}
```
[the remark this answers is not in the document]
Ryxias: The reason they need to be referenced by name is `rule/` code does not exist in the Alert Processor Lambda. The publishers are saved into DynamoDb as the strings which are used to look up the publisher classes.

An alternative way of implementing this is we include all of the `rule/` code in the Alert Processor Lambda. If this is the case, we can directly look up publishers by using `Rule.get_rule(alert.rule_name).publishers`

stream_alert/alert_processor/helpers.py
Ryxias: The `AlertPublisherLoader` is a pattern that I implemented to prevent myself from making the same mistake. 

The publishers are loaded on-demand from the repository, rather than immediately when you import the base module. This gives up compile-time import because everything has to happen at runtime. Maybe we can apply this to outputs too. Thoughts?

stream_alert/alert_processor/helpers.py
Ryxias: I see your point. The choice of verb `publish` implies we're actively *publishing it to somewhere*. Maybe `get_alert_publication` is too wordy. `compose_alert` sounds about right so I can change that

stream_alert/rules_engine/rules_engine.py
Ryxias: Maybe I should only support references? The string literals are a relic of a bygone age at this point

docs/source/publishers.rst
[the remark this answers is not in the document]
Ryxias: https://jira.airbnb.biz/browse/CSIRT-650

docs/source/publishers.rst
```diff
@@ -0,0 +1,254 @@
...
+    def publish(alert: Alert, publication: dict) -> dict:
+      # ...
+      return {}
+
+
+**Recommended Implementation**
+
+Publishers should always return dicts containing only simple types (str, int, list, dict).
+
+Publishers are executed in series, each passing its published ``Alert`` to the next publisher. The ``publication``
```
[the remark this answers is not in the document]
Ryxias: I intended for this to fully error out and blow up the application because—supposing a bug in a publisher—I did not want a bunch of alerts to silently get dispatched with incomplete information. Far safer would be to have the publishers wig out and send the Alert into a retry-loop (until a human stepped in an intervened) which would reduce the change of lost data.

I'm also told that broadly `except Exception:` has to be exercised with caution in python code cuz it can interrupt certain super big bads, like interrupt signals or out-of-memory errors.

As a compromise, I implemented a catch on `KeyError` (since I anticipate this will be like 99% of cases) in `CompositePublisher`. It will error log then reraise the error.

publishers/core.py
```diff
@@ -0,0 +1,227 @@
+from abc import abstractmethod
+from copy import deepcopy
+from inspect import isclass
+
+from stream_alert.shared.logger import get_logger
+from stream_alert.shared.rule import import_folders
```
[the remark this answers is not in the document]
Ryxias: I created a new module:

```
from stream_alert.shared.importer import import_folders
```

publishers/core.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class AlertPublisher(object):
+    """Interface for a Publisher. All class-based publishers must inherit from this class."""
+
+    @abstractmethod
+    def publish(self, alert, publication):
+        """Publishes the given alert.
+
+        Publishers are not intended to MODIFY the given publication; It is preferable to use
```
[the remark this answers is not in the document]
Ryxias: I moved `deepcopy` to the `CompositePublisher` so now individual publishers no longer have to worry about having side effects in one another

publishers/core.py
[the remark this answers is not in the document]
Ryxias: I confirmed that chained inheritance does work with `issubclass` and I wrote a test for it: https://github.com/airbnb/streamalert/pull/900/files#diff-2e4a6d0ba4f31dc59fdb21bde1759ebdR60

publishers/core.py
```diff
@@ -0,0 +1,227 @@
...
+        Args:
+            name (str): The name of the publisher.
+
+        Returns:
+            AlertPublisher|None
+        """
+        if cls.has_publisher(name):
+            return cls._publishers[name]
+
+        LOGGER.error('Publisher [%s] does not exist', name)
```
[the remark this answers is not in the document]
Ryxias: This was (kind of..) in the name of compatibility in certain corner cases: 

## Case #1: Accidentally deploy rule without deploying alert
* Develop a new Publisher named `abc`
* `deploy --function rule`
* Since you forgot to deploy the publisher code to the `alert` lambda, alerts get stuck in *DynamoDB* and error like crazy until a deploy occurs.


## Case #2: Removing a publisher
* Suppose you wanted to remove a publisher, `xyz`
* Delete it and `deploy --function alert`
* If there were alerts in flight, queued up on *DynamoDB* now they get permanently stuck unless you modify the row.


In the current implementation, the above alerts will each still send *one* error but still get dispatched regardless, leveraging a default publisher implementation.

stream_alert/alert_processor/helpers.py
Ryxias: In this specific file, this asks about `alert.publishers` which is a field on DynamoDb, so they are always string literals and never classes/functions. They have to be looked-up and transformed into the actual `Publisher` classes using the `create_composite_publisher`

stream_alert/alert_processor/helpers.py
Ryxias: This permissive logic honestly was a bad idea and was mainly cuz I was lazy while writing tests.

I tweaked it to raise an Exception on invalid `output`.

stream_alert/rules_engine/rules_engine.py
Ryxias: To be clear, there's actually 2 places where code lives to support "publishers"

* On the @rule(publishers=...), which permissively supports both string + references, but I'm trending toward just references
* On any code that pulls from `alert.publishers`, which sits in *DynamoDb*; this code cannot support references at all and can only support strings

stream_alert/rules_engine/rules_engine.py
Ryxias: The structure of the code is similar but sadly the two of them do opposite things. This one converts a `dict(str|Publisher) —> list(strings)`, and the other one converts `dict(str) —> list(Publishers)`. DRY'ing it out is a bit tricky in this case..

docs/source/publishers.rst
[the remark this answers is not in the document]
ryandeivert: maybe not? I guess I didn't see exactly how you implemented it, so it probably works either way?

publishers/core.py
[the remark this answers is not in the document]
ryandeivert: great! 👏

publishers/core.py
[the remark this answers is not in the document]
ryandeivert: nice! thanks for checking :) I thought it would but wasn't sure either

docs/source/publishers.rst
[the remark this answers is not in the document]
Ryxias: (Github chewed up my response message or something...)

I decided to flip it around because I agree with your argument as it does make more intuitive sense.

stream_alert/rules_engine/rules_engine.py
ryandeivert: ++ for just references to reduce complexity and chance for fat-fingering typed literals

stream_alert/shared/publisher.py
```diff
@@ -0,0 +1,272 @@
...
+class AlertPublisherRepository(object):
+    """A repository mapping names -> publishers
+
+    As a usability optimization, using this Repository will eagerly load and register all
+    publishers in the application.
+    """
+    _publishers = {}
+
+    @staticmethod
+    def is_valid_publisher(thing):
```
ryandeivert: thing 😂

stream_alert/shared/publisher.py
```diff
@@ -0,0 +1,272 @@
...
+        Args:
+            thing (mixed): Any primitive or reference to be checked
+
+        Returns:
+            bool
+        """
+
+        # We have to put the isclass() check BEFORE the callable() check because classes are also
+        # callable!
+        if isclass(thing) and issubclass(thing, AlertPublisher):
```
ryandeivert: this could be a oneliner (if you'd want):

`return (isclass(thing) and issubclass(thing, AlertPublisher)) or callable(thing)`

stream_alert/shared/publisher.py
```diff
@@ -0,0 +1,272 @@
...
+
+        Args:
+             publisher (callable|AlertPublisher): An instance of a publisher class or a function
+        """
+        if not AlertPublisherRepository.is_valid_publisher(publisher):
+            LOGGER.error(
+                'Could not register publisher %s; Not callable nor subclass of AlertPublisher',
+                publisher
+            )
+            raise PublisherRegistrationError
```
ryandeivert: custom exceptions that are raised should be instantiated (preferably with a helpful message). In this case, since you are logging an error and then raising an exception, I'd just suggest not logging the error and using that error message when raising the exception. but.. do the message formatting outside of the exception otherwise it looks ugly. so:

```python
error = 'Could not register publisher {}; Not callable nor subclass of AlertPublisher'.format(publisher)
raise PublisherRegistrationError(error)
```

stream_alert/shared/publisher.py
```diff
@@ -0,0 +1,272 @@
...
+        if cls.has_publisher(name):
+            return cls._publishers[name]
+
+        LOGGER.error('Publisher [%s] does not exist', name)
+
+    @classmethod
+    def has_publisher(cls, name):
+        """Returns true if the given publisher name has been registered in this Repository
+        """
+        AlertPublisherImporter.import_publishers()
```
ryandeivert: it still looks like this could be called a bunch of times.. could this call just be put at the class level (since it doesn't appear that this class is ever instantiated) so it's only called once at import time? like:

```python
class AlertPublisherRepository(object):
    """A repository mapping names -> publishers
     As a usability optimization, using this Repository will eagerly load and register all
    publishers in the application.
    """
    _publishers = {}

    AlertPublisherImporter.import_publishers()  # do import now, once, at import time
```

TBH I don't know if you need a dedicated `AlertPublisherImporter` class since it mostly just seems to be a passthrough to the `import_folders` function anyway. Thoughts?

stream_alert/shared/publisher.py
ryandeivert: example of way to _not_ do the formatting:
```
Query '<id>' FAILED with reason 'FAILED: SemanticException Unable to fetch table cloudwatch_cloudtrail. null (Service: AmazonDataCatalog; Status Code: 500; Error Code: ServiceUnavailableException; Request ID: <id>)', exiting: AthenaQueryExecutionError
File "/var/task/stream_alert/athena_partition_refresh/main.py", line 219, in handler
AthenaRefresher().run(event)
File "/var/task/stream_alert/athena_partition_refresh/main.py", line 211, in run
if not self._add_partitions():
File "/var/task/stream_alert/athena_partition_refresh/main.py", line 166, in _add_partitions
success = self._ATHENA_CLIENT.run_query(query=query)
File "/var/task/stream_alert/shared/athena.py", line 270, in run_query
return bool(self._execute_and_wait(query))
File "/var/task/stream_alert/shared/athena.py", line 106, in _execute_and_wait
self.check_query_status(exeuction_id)
File "/var/task/stream_alert/shared/athena.py", line 218, in check_query_status
'Query \'
{}
\'
{}
with reason \'
{}
\', exiting'.format(execution_id, state, reason)
AthenaQueryExecutionError: Query '<id>' FAILED with reason 'FAILED: SemanticException Unable to fetch table cloudwatch_cloudtrail. null (Service: AmazonDataCatalog; Status Code: 500; Error Code: ServiceUnavailableException; Request ID: <id>)', exiting
```

docs/source/publishers.rst
```diff
@@ -0,0 +1,275 @@
...
+should freely add, modify, or delete fields from previous publications. However, publishers should avoid
+doing in-place modifications of the publications, and should prefer to copy-and-modify:
+
+.. code-block:: python
+
+  from stream_alert.shared.publisher import Register
+
+  @Register
+  def sample_publisher(alert, publication):
+    new_publication = deepcopy(publication)
```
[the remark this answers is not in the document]
Ryxias: No it's not, I'll get rid of it

stream_alert/alert_processor/outputs/slack.py
```diff
@@ -100,23 +101,25 @@ def _split_attachment_text(cls, alert_record):
             alert_text = alert_text[index+1:]
 
     @classmethod
-    def _format_attachments(cls, alert_publication, header_text):
+    def _format_default_attachments(cls, alert, alert_publication, fallback_text):
```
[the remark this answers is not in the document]
Ryxias: Will consider for next iteration

stream_alert/alert_processor/outputs/slack.py
```diff
@@ -160,8 +304,17 @@ def _format_message(cls, rule_name, alert_publication):
                     Record (Part 1 of 2):
                     ...
         """
-        header_text = '*StreamAlert Rule Triggered: {}*'.format(rule_name)
-        attachments = list(cls._format_attachments(alert_publication, header_text))
+        default_header_text = '*StreamAlert Rule Triggered: {}*'.format(alert.rule_name)
+        header_text = alert_publication.get('slack.text', default_header_text)
+
+        if 'slack.attachments' in alert_publication:
```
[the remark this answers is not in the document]
Ryxias: In my current design, publishers live on a gradient of ["more generic"](https://github.com/airbnb/streamalert/pull/911/files#diff-02659c3a8e560badc42bf9f5f12e80d6) to ["more output-specific"](https://github.com/airbnb/streamalert/pull/904/files#diff-450ac77ecedc2c339725de0fea83ae74). 

All outputs at the bare minimum has to hook up the `publication` to the API with some default implementation. Each publisher can choose to adopt this default (primarily for reverse-compatibility) or can choose to override it.

### Example #1: 
You're okay with what's going on in PagerDuty you just want to skim out certain fields

In this case you can use a more generic publishers. You opt to NOT provide those "path specs" or "magic fields" or whatever, and this signals to the Output to use a default implementation.

### Example #2:
You want to completely reformat what gets sent to Slack.

You can use slack-specific publishers (refer to above link) or implement your own. These publishers live closer to the output code and have fine-grained control of the outputs (when the APIs allow us), but also require significantly more knowledge of the outputs to implement.

stream_alert/shared/publisher.py
Ryxias: wtf you can put code that executes in a class outside of a function :mindblown:

Ryxias: Nope.. just tried... can't do this because it causes a cyclical require dependency in the following:

`stream_alert/shared/publisher` -> `publishers/community/generic` -> `stream_alert.shared.publisher`

I could probably do this if I split it into its own file, because the `publishers/community/generic` requires the `@Register` annotation, not the repository itself. Will think about it.

Ryxias:
For the sake of not continuing to bloat this PR, I'm going to merge into `release-2-2-0` and open a FIXUP PR for all of the remaining stuff to address, so it's easier to review

stream_alert/alert_processor/outputs/slack.py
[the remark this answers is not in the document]
Ryxias: Update: I opened #917 to try to better segregate these magic fields from the other publication top-level keys.

strcrzy:
👏

PR #904, opened 2019-02-22
[Alert Publishers] Add some community Slack Publishers

Ryxias:
to: @ryandeivert 

## Background
A peek into what publishers can do!

after clean

## Some neat details

I implemented a cute service that parses the `rule_description` and does fuzzy extraction of fields like `author:` and whatnot. I did this because I wanted to adhere to our current "conventions" of rule writing. It may not be perfect but it's workable

There are a bunch of mixable publishers for Slack now. Each of them corresponds with a particular colored-section in the above image. To get the exact image above:

```
from publishers.community.slack.slack_layout import PrettyLayout, AttachRuleInfo, AttachPublication, AttachFullRecord
from publishers.other import other_publisher

@rule(
  publishers=[
    other_publisher,   # presumably your own publisher
    Summary,      # Prints the first rausch colored section
    AttachRuleInfo,    # Prints the 2nd lima colored section
    AttachPublication, # Prints the babu/cyan colored section and fills it with the result of other_publisher
    AttachFullRecord,  # Prints the hackberry colored section
  ]
)
``` 

You can omit any section by simply not including the publisher.

## Testing

I tested this on stage. I got a new slack channel on our internal slack that sucks up data from stage. You need to configure the output with the webhook URL though.

### DIFF
--- a/publishers/community/slack/slack_layout.py
+++ b/publishers/community/slack/slack_layout.py
@@ -1,0 +1,268 @@
+"""
+Copyright 2017-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import cgi
+import json
+import time
+import urllib
+
+from stream_alert.shared.publisher import AlertPublisher, Register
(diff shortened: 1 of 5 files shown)


publishers/community/slack/slack_layout.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def publish(self, alert, publication):
+        publication['slack.attachments'] = publication.get('slack.attachments', [])
+
+        rule_description = alert.rule_description
+        rule_presentation = RuleDescriptionParser.present(rule_description)
+
+        publication['slack.attachments'].append({
+            'color': self._color(),
+            'fields': map(
```
ryandeivert: this might just be me, but I think the map usage can lead to confusion

Ryxias: Changed to 

```
'fields': [
  {'title': key.capitalize(), 'value': rule_presentation['fields'][key]}
  for key in rule_presentation['fields'].keys()
],
```

PR #917, opened 2019-03-04
[Alert Publishers] Standardizes magic fields with @-sign prefix

Ryxias:
to: @ryandeivert 

## Background

I decided that having magic fields like `pagerduty-incident:asdf` were too easy to get mixed up with normal top-level keys.

*NOTE* I did NOT update the PagerDuty output because I'm currently messing with that on #911  and I don't want to deal with the merge conflicts. They're all going into `2.2.0` anyway.

## Changes

I introduced a standard to publisher keys: All of them must start with an at-sign `@`.


Ryxias:
Much obliged @ryandeivert

PR #920, opened 2019-03-19
Fix some bugs introduced due to "default behavior" in pagerduty integration

Ryxias:
related to:
resolves:

## Background

We received some requests regarding changes in default behavior of certain pages. This tweaks them back to normal.

## Testing

CI?


stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -654,6 +651,18 @@ def _create_base_incident(self, alert, publication, rule_context):
                 assigned_key: assigned_value
             }
         }
+
+        # Urgency, if provided, must always be 'high' or 'low' or the API will error
+        if incident_urgency:
+            if incident_urgency in ['low', 'high']:
+                incident_urgency['incident']['urgency'] = incident_urgency
+            else:
+                LOGGER.error(
```
chunyong-lin: Use `warning` instead?

Ryxias: Changed to `LOGGER.warn`

PR #922, opened 2019-03-22
Adds a new publisher, improves to description parser

Ryxias:
to: @chunyong-lin  or @ryandeivert 
related to:
resolves:

## Background

I'm adding some minor enhancements to aid with deploying Custom Alert Payloads.

## Changes

* Add a new publisher for specifically white-listing and bubbling up fields
* Improve the rule description parser to handle some more corner cases
* Add a new publisher that converts scalar arrays inside of publications into strings
* Consolidates some directories to better separate community code from non-community code

## Testing

CI


publishers/community/generic.py
```diff
@@ -163,3 +164,46 @@ def _recursive_enumerate_fields(structure, output_reference, path=''):
     _recursive_enumerate_fields(publication, output)
 
     return OrderedDict(sorted(output.items()))
+
+
+@Register
+def bubble_fields(alert, publication):
```
chunyong-lin: Maybe use `populate_fields`?

publishers/community/generic.py
```diff
@@ -163,3 +164,46 @@ def _recursive_enumerate_fields(structure, output_reference, path=''):
...
+    """This publisher bubbles all requested fields to the top level and deletes everything else.
```
chunyong-lin: Does the publisher really delete non-requested fields? I think it just bubbles all requested fields to the top level.

publishers/community/generic.py
```diff
@@ -163,3 +164,46 @@ def _recursive_enumerate_fields(structure, output_reference, path=''):
...
+
+    NOTE: It is possible for bubbled fields to continue to contain nested dicts, so do not assume
+          this publisher will result in a flat dictionary publication.
+    """
+
+    new_publication = {}
+    for bubble_field in alert.context.get('bubble_fields', []):
+        extractions = get_keys(publication, bubble_field)
+
+        if len(extractions) > 1:
```
chunyong-lin: I don't think it is necessary to check the length of `extractions`. In current logic, if the length of extractions larger than 1, the value will be a list. If the length is 1, the value will be a string. Is it what you want?

publishers/community/slack/slack_layout.py
```diff
@@ -25,6 +25,7 @@
 BABU = '#00d1c1'
 LIMA = '#8ce071'
 HACKBERRY = '#7b0051'
+BEACH = '#ffb400'
```
chunyong-lin: What does this constant mean? Is it some special characters needs to be skipped or special processed? Can you add a comment?

stream_alert/shared/description.py
```diff
@@ -67,16 +82,25 @@ def parse(cls, rule_description):
             dict: A dict mapping fields to lists of strings, each corresponding to a line belonging
                   to that field. All field names are lowercase.
         """
-        tokens = rule_description.split('\n')
+        rule_description = '' if not rule_description else rule_description
+        tokens = rule_description.strip().split('\n')
```
chunyong-lin: We may `strip` after `split`, because
`The strip() method returns a copy of the string with both leading and trailing characters removed (based on the string argument passed).` So after split, the substring may contains leading or trailing characters.

publishers/community/generic.py
Ryxias: It deletes it from the `publication`, not from the `alert.record`. The way **publishers** work, they are supposed to chain the `publication` from one publisher to the next, adding/transforming fields as they go. This one is "special" because it ignores the previous publishers' work and just throws all of the existing fields away.

publishers/community/generic.py
Ryxias: Yeah that's what I wanted, since this is more of a function to generate user-facing data, instead of data that's being used by more code. I don't have a _strong_ opinion on this, since I think a 1 element array is not all that different.

I just don't like how PagerDuty renders 1 element arrays, since it shows up as:

```
[
  "element_here\n with newlines\noh no"
]
```

Versus the way I do it it'd show up as 

```
element here
with newlines 
oh no
```

Ryxias: That being said, it's easy for me to have another PagerDuty-specific publisher that transforms arrays into cleaner PD-readable strings.

publishers/community/slack/slack_layout.py
Ryxias: It's an airbnb brand color that I added... since it's not being used I can get rid of it

stream_alert/shared/description.py
Ryxias: Doesn't `split` return an array?

ryandeivert: I think she means on the values within the list returned by `split`.. like:

`tokens = [line.strip() for line in rule_description.strip().split('\n')]`

Ryxias: Python is :magic:

publishers/community/generic.py
Ryxias: I removed this logic and moved it into a new publisher.

Ryxias:
@ryandeivert @chunyong-lin pushed some changes. PTAL

publishers/community/generic.py
```diff
@@ -163,3 +164,89 @@ def _recursive_enumerate_fields(structure, output_reference, path=''):
...
+class StringifyArrays(AlertPublisher):
+    """Deeply navigates a dict publication and coverts all scalar arrays to strings
+
+    Any array discovered with only scalar values will be joined into a single string with the
+    given DELIMITER. Subclass implementations of this can override the delimiter to join the
+    string differently.
+    """
+    DELIMITER = '\n'
+
+    def publish(self, alert, publication):
```
chunyong-lin: Can you add docstring for this method?

publishers/community/generic.py
```diff
@@ -163,3 +164,89 @@ def _recursive_enumerate_fields(structure, output_reference, path=''):
...
+                # to the object that contains it
+                fringe.extend(next_item)
+            else:
+                # It's a leaf node, or it's some strange object that doesn't belong here
+                pass
+
+        return publication
+
+    @staticmethod
+    def is_scalar_array(item):
```
chunyong-lin: Also missing docstring.

stream_alert/shared/description.py
```diff
@@ -67,22 +82,31 @@ def parse(cls, rule_description):
...
         field_lines = {}
 
         current_field = 'description'
         for token in tokens:
-            if not token or not token.strip():
+            if current_field not in field_lines:
+                field_lines[current_field] = []
+
+            if not token:
+                field_lines[current_field].append('')
```
chunyong-lin: Why do we append an empty string to a list?

publishers/community/generic.py
Ryxias: The `publish` is an abstract method inherited from the parent class, `AlertPublisher`.

stream_alert/shared/description.py
Ryxias: I added a comment to make it clearer why; we need to preserve empty lines so that we know how to recognize when there is a true (markdown-style) line break

PR #925, opened 2019-03-28
[Publishers] Adds support for images on Pagerduty v2

Ryxias:
to: @chunyong-lin or @ryandeivert 

## Background

Nice to have

## Changes

* Add code support for images on pagerduty-v2. v1 was already supported.
* Adds a new publisher, `AttachImage` that attaches images. It's meant to be subclassed.
* (MInor) adds support for `component`, `group`, and `class` on Pagerduty v2, which aren't really major features but they're nice to have.

## Testing

CI + Tested on Staging


Ryxias:
We don't actually upload the image; we simply provide a URL. PagerDuty validates the URL and generates `<img src=...` tags on their end

PR #935, opened 2019-05-03
Make PagerDuty output idempotent; adds support for responder requests

Ryxias:
to: @ryandeivert or @chunyong-lin 

## Background

The `pagerduty-incident` code used to have a serious bug where, due to the way it would sequence multiple write methods non-atomically, it could result in a situation where a Pagerduty alert is sent but the output fails. This results in the alert retrying, creating *another* Pagerduty alert, then failing again... repeating ad infinitum.

## Changes

This PR introduces two concepts:

### Reduced number of write calls

PagerDuty's code used to do the following:

* `POST /incidents` (Create an incident container)
* `POST /events/enqueue` (Create an alert)
* `GET /incidents?dedup_key=???` (Find the alert-incident)
* `PUT /incidents/#/merge?incident_id=???` (Merge the alert-incident into the incident container)

I found that this was redundant; the workflow is optimized now:

* `POST /events/enqueue` (Create an alert)
* `GET /incidents?dedup_key=???` (Find the alert's incident)
* `PUT /incidents` (Modify the alert's incident)

Fewer API calls means the code is cleaner and easier to understand. Also it only has `POST` method, which reduces the number of non-idempotent write API calls.

### Idempotency

PagerDuty's `pagerduty-incident` output is now idempotent. The `POST /events/enqueue` API call
now leverages a `dedup_key` that is equal to the **Alert ID**.

The usage of a `dedup_key` ensures that any unique StreamAlert alert will only ever create a single PagerDuty alert. Subsequent retries will cause PagerDuty to return the previously created alert.

### Responder Requests

I added a new PagerDuty feature that allows the Alert to leverage [Responder Requests](https://api-reference.pagerduty.com/#!/Incidents/post_incidents_id_responder_requests). This is a neat feature that allows you to invite people **other** than the assignee to join in the PagerDuty.

It shows up like this on the PD UI.

image

All members of the response team get all notification (push notification/SMS/etc) related to the Alert **_except_** escalation notices. These stay with the assignee.

The awesome benefit here is that we can have PagerDuty alerts stick to a single escalation policy, but attach more watchers.

You can add responder requests to an alert using `context`:

```
context={
  responders=['derek1@email.com', 'derek2@email.com' ...],
  responder_message='This message shows up on their request',
}
```

## Testing

CI, Plus I tried
it on Stage A lot. I think 
it bothered people

### DIFF
--- a/stream_alert/alert_processor/outputs/pagerduty.py
+++ b/stream_alert/alert_processor/outputs/pagerduty.py
@@ -114,10 +114,13 @@ class EventsV2DataProvider(object):
             'routing_key': routing_key,
             'event_action': 'trigger',
 
-            # Beware of providing this; when this is provided, even if empty string, this will
-            # cause the dedup_key to be bound to the ALERT, not the incident. The implication
-            # is that the incident will no longer be searchable with incident_key=dedup_key
-            # 'dedup_key': '',
+            # Passing a dedup_key will ensure that only one event is ever created. Any subsequent
+            # request with the same dedup_key + routing_key + event_action will simply return
+            # the original result.
+            # Once the alert is resolved, the dedup_key can be re-used.
+            # https://v2.developer.pagerduty.com/docs/events-api-v2#alert-de-duplication
+            'dedup_key': alert.alert_id,
+            #'incident_key': alert.alert_id,
             'payload': {
                 'summary': summary,
                 'source': alert.log_source,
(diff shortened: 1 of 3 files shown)


stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -610,70 +623,131 @@ def run(self, alert, descriptor):
...
+                    error = '[{}] Failed to request a responder ({}) on incident ({})'.format(
+                        self._output.__service__,
+                        responder_email,
+                        incident_id
+                    )
+                    LOGGER.error(error)
+                    errors.append(error)
+
+        # Add a note to the incident
+        note = self._add_incident_note(incident_id, publication, rule_context)
```
chunyong-lin: where do we add this `note` to?

Ryxias: This note is a small string of text that is added to the PagerDuty incident. It shows up on their UI

stream_alert/alert_processor/outputs/pagerduty.py
```diff
@@ -610,70 +623,131 @@ def run(self, alert, descriptor):
...
+        # In the fateful event the alert gets stuck ANYWAY, the easiest solution is to destroy the
+        # associated record on the DynamoDB table.
+        errors = []
+
+        # Add responder requests
+        responders = rule_context.get('responders', [])
+        if responders and not isinstance(responders, list):
+            responders = [responders]
+
+        if responders:
```
ryandeivert: are responders optional? meaning should we return False `if not responder` or is that okay?

Ryxias: Responders are optional; it's a new feature so I need to maintain reverse compatibility.

PR #936, opened 2019-05-08
Logsource fleet

gavinelder:
to: @ryandeivert 
related to:
resolves:

## Background

Reason for the change

With the release of (Kolide/Fleet 2.1.0)[https://github.com/kolide/fleet/releases/tag/2.1.0] it now supports query output to Firehose. 

This enables a more centralised way of managing OSquery logs and results using Fleet.

## Changes

Adds fleet log type as it does not match existing OSquery formats defined in the current `logs.json`

Add a basic query & record for integration tests and future example.

~~Added decorators as this is not really well documented.~~

## Testing

Tested in production , deploying fleet writing to firehose storing to s3 and then being picked up by StreamAlerts.


conf/logs.json
```diff
@@ -1811,6 +1811,20 @@
       ]
     }
   },
+  "fleet:results":{
```
chunyong-lin: Do you open to rename the schema to `fleet:audit` to be more descriptive? The new feature in fleet 2.1.0 is to end audit logs.

rules/community/fleet/fleet_bad_action.py
```diff
@@ -0,0 +1,26 @@
...
+def fleet_bad_action(rec):
+    """
+    author:       gavinelder
+    description:  Alert when a user carries out a bad action.
+    reference:    https://
+    playbook:     (a) Reach out to the user who made the modification and confirm intent.
+                  (b) Link appropriate Jira ticket.
+    """
+    return (
+        '1' in rec['columns']['bad_action']
```
chunyong-lin: Do you mean `'1' == rec['columns']['bad_action']`? The value of `rec['columns']['bad_action']` is a string, not a list.

rules/community/fleet/fleet_bad_action.py
```diff
@@ -0,0 +1,26 @@
+from stream_alert.shared.rule import rule
+from stream_alert.shared.rule import disable
+from publishers.community.slack.slack_layout import Summary, AttachRuleInfo \
+        , AttachPublication, AttachFullRecord
+
+@disable
```
chunyong-lin: Can you add a comment about to remove `disable` decorator if user want to use this rule?

rules/community/fleet/fleet_bad_action.py
chunyong-lin: I would prefer to use `get` to retrieve keys in `columns` field.
```
return (
    '1' == rec['columns'].get('bad_action', '1') 
)
```

rules/community/fleet/fleet_bad_action.py
```diff
@@ -0,0 +1,26 @@
...
+@rule(
+    logs=['fleet:results'],
+    publishers=[
```
Ryxias: Mostly unrelated to the gist of your PR but had to chime in here: Publishers do not work on most of our default outputs. We wanted publishers to be mostly for human-readable presentation purposes, but force all default output stuff (for example, things that get dropped into Firehose-Athena) to have an unchanging format.

These publishers are slack-specific, so unless there's an `outputs=['slack:...']` then this will not do anything. If you choose to go with slack, I also suggest using the output-specific notation:

```
SLACK_PUBLISHER_BUNDLE = [
  Summary,
  AttachRuleInfo,
  AttachPublication,
  AttachFullRecord,
]

... 
@rule( 
  ...
  publishers={
    'slack:_descriptor_': SLACK_PUBLISHER_BUNDLE
  }
)
```

gavinelder: Agreed I didn't want to include a slack output so removed this section completly

rules/community/fleet/fleet_bad_action.py
gavinelder: I have removed this and will raise another PR with some docs as I think it's probably going to be more fitting.

conf/logs.json
gavinelder: True however this is specifically looking at the output of `filesystem_result_log_file` which is the result of the query. 

There is a secondary kinesis output of `filesystem_status_log_file` which is not used in this example as it contains basic processing info and nothing to alert on. 

I have no strong preference on this however I felt `results` in this case matched closer to the Kolide output name.

rules/community/fleet/fleet_bad_action.py
gavinelder: Ack and complete.

rules/community/fleet/fleet_bad_action.py
gavinelder: Ignore comment about PR this was intended for decorators.

rules/community/fleet/fleet_bad_action.py
```diff
@@ -0,0 +1,20 @@
...
+# Remove the @disable decorator to use this rule in a live deployment
+@disable
+@rule(
+    logs=['fleet:results']
+)
+def fleet_bad_action(rec):
+    """
+    author:       gavinelder
+    description:  Alert when a user carries out a bad action.
+    reference:    https://
```
chunyong-lin: Do you have the reference link?

gavinelder: Sorry can you provide more detail reference to the example fleet query ?

Let me know what exactly you would like added

chunyong-lin: I am thinking maybe the reference about each field in the fleet logs. But I just checked the PR  (https://github.com/kolide/fleet/pull/2022/files) which supporting fleet logs sent to firehose, I don't see documentation related to log schema.
If no reference, you may just remove this line, or use `N/A`.

gavinelder: Ack, Normally use reference for run-books internally.

PR #937, opened 2019-05-13
[apps] add Intercom app integration

catpham:
cc: @patrickod  

## Background

[Intercom](https://github.com/intercom) recently made V1.2 of our API stable. V1.2 introduces a new [**Activity Logs** endpoint](https://developers.intercom.com/building-apps/docs/api-changelog#section-view-activity-logs-for-admins-through-our-api). This endpoint makes it possible to get a log of activities taken by all admins in an app via a GET request. 

We (@catpham on behalf of Intercom Security team) are opening this PR to add an Intercom app integration to Streamalert. We think this would be a useful integration for customers who want to automate and further facilitate their Intercom app activity auditing and alarming.

## Changes

This PR has 3 commits:

1. Adding the Intercom app integration
2. Adding unit tests for the Intercom app integration
3. Updating the README to include instructions on how to use the Intercom app integration

## Testing

We've added unit tests. We've also been using the [**Activity Logs** endpoint](https://developers.intercom.com/building-apps/docs/api-changelog#section-view-activity-logs-for-admins-through-our-api) with the Intercom app integration in our private fork of [Airbnb's Streamalert](https://github.com/airbnb/streamalert) for the past 6 weeks.

### DIFF
--- a/stream_alert/apps/_apps/README.rst
+++ b/stream_alert/apps/_apps/README.rst
@@ -8,11 +8,11 @@ To obtain the access key and access key secret, an authorized user of the Aliyun
 Additionly, the user for whom the access key was created must have sufficient privileges to make use of ActionTrail; follow the directions on the `Grant ActionTrail permissions to RAM users <https://www.alibabacloud.com/help/doc-detail/28818.htm>`_ page.
 
 How to set up the intercom app
-###########################
+##############################
 
-The Intercom API requires an access token. Get an access token by following `these instructions <https://developers.intercom.com/building-apps/docs/authorization#section-how-to-get-an-access-token>`.
+The Intercom API requires an access token. Get an access token by following `these instructions <https://developers.intercom.com/building-apps/docs/authorization#section-how-to-get-an-access-token>`_.
 
-To specify an API version, follow `these instructions <https://developers.intercom.com/building-apps/docs/api-versioning>` to do so through Intercom's Developer Hub. 
+To specify an API version, follow `these instructions <https://developers.intercom.com/building-apps/docs/api-versioning>`_ to do so through Intercom's Developer Hub. 
 The default will be the latest stable version. The Intercom app works on versions 1.2 or later. 
 
 How to set up the slack app
--- a/stream_alert/apps/_apps/README.rst
+++ b/stream_alert/apps/_apps/README.rst
@@ -7,6 +7,14 @@ To obtain the access key and access key secret, an authorized user of the Aliyun
 
 Additionly, the user for whom the access key was created must have sufficient privileges to make use of ActionTrail; follow the directions on the `Grant ActionTrail permissions to RAM users <https://www.alibabacloud.com/help/doc-detail/28818.htm>`_ page.
 
+How to set up the intercom app
+###########################
+
+The Intercom API requires an access token. Get an access token by following `these instructions <https://developers.intercom.com/building-apps/docs/authorization#section-how-to-get-an-access-token>`.
+
+To specify an API version, follow `these instructions <https://developers.intercom.com/building-apps/docs/api-versioning>` to do so through Intercom's Developer Hub. 
+The default will be the latest stable version. The Intercom app works on versions 1.2 or later. 
+
 How to set up the slack app
 ###########################
 
--- a/stream_alert/apps/_apps/intercom.py
+++ b/stream_alert/apps/_apps/intercom.py
@@ -1,0 +1,103 @@
+import time
+import re
+
+import requests
+
+from . import AppIntegration, StreamAlertApp, get_logger
+
+
+LOGGER = get_logger(__name__)
+
+
+@StreamAlertApp
+class IntercomApp(AppIntegration):
+    """Intercom StreamAlert app"""
+    _INTERCOM_LOGS_URL = 'https://api.intercom.io/admins/activity_logs'
+
+    def __init__(self, event, config):
+        super(IntercomApp, self).__init__(event, config)
+        self._next_page = None
+
+    @classmethod
+    def _type(cls):
+        return 'admin_activity_logs'
+
+    @classmethod
+    def service(cls):
+        return 'intercom'
+
+    @classmethod
+    def _required_auth_info(cls):
+        return {
+            'token': {
+                'description': 'the OAuth token for this Intercom app',
+                'format': re.compile(r'.*')
+            }
+        }
+
+    def _sleep_seconds(self):
+        """Return the number of seconds this polling function should sleep for
+        between requests to avoid failed requests. Intercom API allows for a default of 500 requests
+        every minute, distributed over 10 seconds periods. This means for a default rate limit of
+        500 per minute, you can send a maximum of 83 operations per 10 second period.
+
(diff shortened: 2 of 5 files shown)


catpham:
Is there an auto fix option for the linter?

ryandeivert:
hi @catpham thanks for the contribution 😄 😄 

I apologize for just seeing your question now - sadly there is not an 'auto fix' option, but we do suggest using `yapf` to help with formatting/linting issues.

To format in place, you can use:

`yapf --style google -i path/to/file.py`

stream_alert/apps/_apps/README.rst
```diff
@@ -7,6 +7,14 @@ To obtain the access key and access key secret, an authorized user of the Aliyun
 
 Additionly, the user for whom the access key was created must have sufficient privileges to make use of ActionTrail; follow the directions on the `Grant ActionTrail permissions to RAM users <https://www.alibabacloud.com/help/doc-detail/28818.htm>`_ page.
 
+How to set up the intercom app
+###########################
```
ryandeivert: RST requires that the header extend all the way to the end of the preceding line. just add `###` here please :)

stream_alert/apps/_apps/README.rst
```diff
@@ -7,6 +7,14 @@ To obtain the access key and access key secret, an authorized user of the Aliyun
...
+
+The Intercom API requires an access token. Get an access token by following `these instructions <https://developers.intercom.com/building-apps/docs/authorization#section-how-to-get-an-access-token>`.
```
ryandeivert: The link is not valid RST syntax (you can see this by going to `View File` in the github UI and observing the content rendered improperly in the preview):

image

All you need is a trailing `_` after the closing tick (`), like (this won't render because GH only supports markdown):

```
This is a `link <https://www.google.com`_.
```

You can reference some of the other link in this file if need-be :).

stream_alert/apps/_apps/README.rst
```diff
@@ -7,6 +7,14 @@ To obtain the access key and access key secret, an authorized user of the Aliyun
...
+
+To specify an API version, follow `these instructions <https://developers.intercom.com/building-apps/docs/api-versioning>` to do so through Intercom's Developer Hub.
```
ryandeivert: This link is also missing a trailing underscore, breaking the rendering (similar to aforementioned issue).

stream_alert/apps/_apps/intercom.py
```diff
@@ -0,0 +1,103 @@
...
+    @classmethod
+    def service(cls):
+        return 'intercom'
+
+    @classmethod
+    def _required_auth_info(cls):
+        return {
+            'token': {
+                'description': 'the OAuth token for this Intercom app',
+                'format': re.compile(r'.*')
```
ryandeivert: is there no static length or anything that could be used for the token to try to detect invalid input? a lot of services at least have a consistent length for their tokens, some even use a static prefix that can be helpful to validate.

stream_alert/apps/_apps/intercom.py
```diff
@@ -0,0 +1,103 @@
...
+        every minute, distributed over 10 seconds periods. This means for a default rate limit of
+        500 per minute, you can send a maximum of 83 operations per 10 second period.
+
+        Resource(s):
+            https://developers.intercom.com/intercom-api-reference/reference#rate-limiting
+
+        Returns:
+            int: Number of seconds that this function should sleep for between requests
+        """
+        return 10
```
ryandeivert: given your docstring description of their rate limiting, it doesn't sound like this should actually be necessary 🤔 . I don't think the lambda will be able to do more than 83 requests a minute, so sleeping probably isn't necessary

stream_alert/apps/_apps/intercom.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # Results are paginated with a page url field provided that is used in subsequent queries.
+        # If this field exists, make a a query to the page url, and if not, make a fresh query to
+        # the default _INTERCOM_LOGS_URL with the required created_at_before and created_at_after
+        # parameters
+        if self._next_page:
+            params = None
+            url = self._next_page
+        else:
+            params = {'created_at_before': int(time.time()),
```
ryandeivert: is the `created_at_before` value in UTC? if so, it might be advisable to ensure you're always getting time in UTC (`time.time()` does local time and lambda runs in UTC, but enforcing UTC is usually safe when testing locally).

This seems clunky, but is the only way I know to get a consistent UTC timestamp in any locale in python 2:
```
import calendar
import time

current_time = int(calendar.timegm(time.gmtime()))
```

stream_alert/apps/_apps/intercom.py
```diff
@@ -0,0 +1,103 @@
...
+            activity
+            for activity
+            in response['activity_logs']
+            if int(activity['created_at']) > self._last_timestamp]
+
+        if not activities:
+            return False
+
+        # Save next page url if any for paginated results
+        if response['pages']['next'] is not None:
```
ryandeivert: is will the `next` key actually exist even when the value is `None` (ie - is this a non-optional key?)

If it's optional, you might want to do `response['pages'].get('next')` but I will defer to you since I do not know their API.

catpham: `next` is a non-optional key -- it'll always exist and it's value can be `None`

ryandeivert: awesome thanks for confirming!

stream_alert/apps/_apps/intercom.py
catpham: thanks for the suggestion!

stream_alert/apps/_apps/intercom.py
catpham: there doesn't seem to be a fixed length or prefix for the tokens. they are `base64` encoded though -- would a regex matching for a `base64` encoded string suffice?

catpham:
@ryandeivert thanks for the helpful review! Just pushed up a commit to address the comments and left a follow-up question re token validation 🙇🏻‍♀️

stream_alert/apps/_apps/intercom.py
ryandeivert: okay! all good to leave it as is :)

ryandeivert:
@catpham FYI I just changed the branch to merge this into to `release-2-3-0` instead of master (which would be for bug fixes only)

ryandeivert:
hi @catpham, it appears that few unit tests are now failing. can you PTAL :)

catpham:
@ryandeivert yes, on it!

stream_alert/apps/_apps/intercom.py
catpham: actually found the source code and there is a prefix :) just updated the validation regex to look to for a base64 string with prefix!

ryandeivert:
this will be included in release v2.3.0

catpham:
💃 thanks! any idea when v2.3.0 will be released?

ryandeivert:
probably not for a month or so, but you can always work off of this feature branch if you need it!

PR #945, opened 2019-06-13
Make Terraform version compatibility more explicit

geremyCohen:

## Background

Terrform 0.12.X required explicit relative paths (using a `./` prefix). See: https://github.com/poseidon/typhoon/pull/457. This breaks things, so we should be more explicit about required version.

## Changes

* Adding terraform required version note

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -8,7 +8,9 @@ Install Dependencies
 --------------------
 
 1. Install Python 2.7 and `pip <https://pip.pypa.io/en/stable/installing/>`_
-2. Install `Terraform <https://www.terraform.io/intro/getting-started/install.html>`_ v0.11.X:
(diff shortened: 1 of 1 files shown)


docs/source/getting-started.rst
```diff
@@ -8,7 +8,9 @@ Install Dependencies
 --------------------
 
 1. Install Python 2.7 and `pip <https://pip.pypa.io/en/stable/installing/>`_
-2. Install `Terraform <https://www.terraform.io/intro/getting-started/install.html>`_ v0.11.X:
+2. Install `Terraform <https://www.terraform.io/intro/getting-started/install.html>`_ v0.11.X: 
+
+**NOTE**: Terraform versions newer than 0.11.X are not currently supported.
```
ryandeivert: thanks for this! can you update the "note" to use RST syntax, like so:

```
.. note:: A note thing you want to note note
```

PR #946, opened 2019-06-13
make Terraform version compatibility more explicit

geremyCohen:

## Background

Terraform 0.12.X requires explicit relative paths (using a ./ prefix). See: poseidon/typhoon#457. This breaks things, so we should be more explicit about required version.

## Changes

* Adding terraform required version note


ryandeivert:
thanks for the contribution @geremyCohen !

Issue #885, opened 2019-01-22
bug: StreamAlert SQS Classified Logs Missing Prefix 


Ryxias:
## Background

Similar to #886 

During deployment on test account we noticed that the `streamalert_classified_logs` SQS queue is missing a prefix.

### Description

All SA resources require a prefix. Missing the prefix causes trouble when there are multiple SA deployments on a single AWS Account. Terraform will detect that a duplicate SQS resource already exists during deployment and error.

### Steps to Reproduce

Deploy SA to the staging account.

## Desired Change

Add the prefix to the SQS queue name.

Ryxias:
Leaving this for myself so I don't forget: 
If you use `python manage.py deploy --function classifier rule` you have to deploy **twice**. The reason for this is the SQS queue is created on the first run, but its URL is generated *during* creation, so Terraform doesn't know that the URL is changed and fails to update the classifier's `SQS_QUEUE_URL` environment variable.

Ryxias:
Changing this to a `3.0.0` release. There is no way to do this gracefully without introducing a breaking change.

PR #955, opened 2019-07-03
Add prefix to classifed logs queue

hkarumuri:
to:@airbnb/streamalert-maintainers
resolves: #885

## Background
One AWS account is unable to handle multiple instances of StreamAlert due to the name of the classified logs queue.

This issue was found with running ./manage.py init on a second instance of StreamAlert when there was already an instance set up on an AWS account. The specific error was this:
```Error: Error applying plan:

1 error occurred:
    * module.globals.module.classifier_queue.aws_sqs_queue.classifier_queue: 1 error occurred:
    * aws_sqs_queue.classifier_queue: Error creating SQS queue: QueueAlreadyExists: A queue already exists with the same name and a different value for attribute KmsMasterKeyId
    status code: 400, request id: 1e4d6ff7-2f84-5741-8efb-3a8c6d65b95a
```

## Changes
* Added the prefix to the streamalert_classified_log queue to remove a QueueAlreadyExists conflict

## Testing
After adding a prefix to the classifier queue, I initialized two instances of StreamAlert and ran the following command to test if it worked
`./manage.py test rules`

The 77 provided tests all passed, and here is the full output of the tests below
<details><summary>Full Output</summary>
<p>

```File: cloudwatch/cloudtrail_via_cloudwatch.json

Test #01: Pass

File: cloudwatch/rds_aurora_via_cloudwatch.json

Test #01: Pass
Test #02: Pass
Test #03: Pass

File: box/box_admin_events.json

Test #01: Pass

File: aliyun/aliyun_basic.json

Test #01: Pass
Test #02: Pass
Test #03: Pass
Test #04: Pass

File: mitre_attack/right_to_left_character.json

Test #01: Pass
Test #02: Pass
Test #03: Pass
Test #04: Pass

File: gsuite/gsuite_admin.json

Test #01: Pass

File: binaryalert/binaryalert_yara_match.json

Test #01: Pass
Test #02: Pass

File: onelogin/onelogin_events_assumed_role.json

Test #01: Pass
Test #02: Pass

File: cloudtrail/cloudtrail_critical_api_calls.json

Test #01: Pass
Test #02: Pass
Test #03: Pass
Test #04: Pass
Test #05: Pass
Test #06: Pass
Test #07: Pass
Test #08: Pass

File: cloudtrail/cloudtrail_ec2_image_creation.json

Test #01: Pass
Test #02: Pass

File: cloudtrail/cloudtrail_mfa_policy_abuse_attempt.json

Test #01: Pass
Test #02: Pass
Test #03: Pass
Test #04: Pass

File: cloudtrail/cloudtrail_network_acl_ingress_anywhere.json

Test #01: Pass
Test #02: Pass

File: cloudtrail/cloudtrail_put_bucket_acl.json

Test #01: Pass
Test #02: Pass

File: cloudtrail/cloudtrail_put_object_acl_public.json

Test #01: Pass
Test #02: Pass
Test #03: Pass
Test #04: Pass

File: cloudtrail/cloudtrail_root_account_usage.json

Test #01: Pass
Test #02: Pass

File: cloudtrail/cloudtrail_security_group_ingress_anywhere.json

Test #01: Pass
Test #02: Pass

File: github/github_disable_dismiss_stale_pull_request_approvals.json

Test #01: Pass
Test #02: Pass

File: github/github_disable_protect_this_branch.json

Test #01: Pass
Test #02: Pass

File: github/github_disable_required_pull_request_reviews.json

Test #01: Pass
Test #02: Pass

File: github/github_disable_required_status_checks.json

Test #01: Pass
Test #02: Pass

File: github/github_disable_two_factor_requirement_org.json

Test #01: Pass
Test #02: Pass

File: github/github_disable_two_factor_requirement_user.json

Test #01: Pass
Test #02: Pass

File: github/github_oauth_application_create.json

Test #01: Pass
Test #02: Pass

File: github/github_site_admin_action.json

Test #01: Pass
Test #02: Pass

File: github/github_site_admin_user_promotion.json

Test #01: Pass
Test #02: Pass

File: slack/slack_basic.json

Test #01: Pass
Test #02: Pass
Test #03: Pass

File: guardduty/guard_duty_all.json

Test #01: Pass

File: duo/duo_anonymous_ip_failure.json

Test #01: Pass
Test #02: Pass

File: duo/duo_bypass_code_create_non_auto_generated.json

Test #01: Pass
Test #02: Pass

File: duo/duo_bypass_code_create_non_expiring.json

Test #01: Pass
Test #02: Pass

File: duo/duo_bypass_code_create_unlimited_use.json

Test #01: Pass
Test #02: Pass

File: duo/duo_fraud.json

Test #01: Pass
Test #02: Pass

Summary:

Total Tests: 77
Pass: 77
Fail: 0
```

</p>
</details>

### DIFF
--- a/terraform/modules/tf_stream_alert_globals/classifier_queue/sqs.tf
+++ b/terraform/modules/tf_stream_alert_globals/classifier_queue/sqs.tf
@@ -1,28 +1,10 @@
 // SQS Queue: Send logs from the Classifier to the SQS queue
-# resource "aws_sqs_queue" "classifier_queue_prefixed" {
-#   count = "${var.use_prefix == "true" ? 1 : 0}"
-#   name = "${var.prefix}_streamalert_classified_logs"
-
-#   # The amount of time messages are hidden after being received from a consumer
-#   # Default this to 2 seconds longer than the maximum AWS Lambda duration
-#   visibility_timeout_seconds = "${var.rules_engine_timeout + 2}"
-
-#   # Enable queue encryption of messages in the queue
-#   kms_master_key_id = "${aws_kms_key.sqs_sse.arn}"
-
-#   tags {
-#     Name = "StreamAlert"
-#   }
-# }
-
 locals {
   classifier_prefix = "${var.prefix}_"
 }
 
 resource "aws_sqs_queue" "classifier_queue" {
-  # count = "${var.use_prefix == "true" ?  : 1}"
   name = "${var.classifier_use_prefix == "true" ? local.classifier_prefix : ""}streamalert_classified_logs"
-  # name = "streamalert_classified_logs"
 
   # The amount of time messages are hidden after being received from a consumer
   # Default this to 2 seconds longer than the maximum AWS Lambda duration
--- a/terraform/modules/tf_stream_alert_globals/classifier_queue/sqs.tf
+++ b/terraform/modules/tf_stream_alert_globals/classifier_queue/sqs.tf
@@ -1,6 +1,6 @@
 // SQS Queue: Send logs from the Classifier to the SQS queue
 resource "aws_sqs_queue" "classifier_queue" {
-  name = "streamalert_classified_logs"
+  name = "${var.prefix}_streamalert_classified_logs"
 
   # The amount of time messages are hidden after being received from a consumer
   # Default this to 2 seconds longer than the maximum AWS Lambda duration
--- a/terraform/modules/tf_stream_alert_globals/classifier_queue/sqs.tf
+++ b/terraform/modules/tf_stream_alert_globals/classifier_queue/sqs.tf
@@ -1,6 +1,28 @@
 // SQS Queue: Send logs from the Classifier to the SQS queue
+# resource "aws_sqs_queue" "classifier_queue_prefixed" {
+#   count = "${var.use_prefix == "true" ? 1 : 0}"
+#   name = "${var.prefix}_streamalert_classified_logs"
+
+#   # The amount of time messages are hidden after being received from a consumer
+#   # Default this to 2 seconds longer than the maximum AWS Lambda duration
+#   visibility_timeout_seconds = "${var.rules_engine_timeout + 2}"
+
+#   # Enable queue encryption of messages in the queue
+#   kms_master_key_id = "${aws_kms_key.sqs_sse.arn}"
+
+#   tags {
+#     Name = "StreamAlert"
+#   }
+# }
+
+locals {
+  classifier_prefix = "${var.prefix}_"
+}
+
 resource "aws_sqs_queue" "classifier_queue" {
-  name = "${var.prefix}_streamalert_classified_logs"
+  # count = "${var.use_prefix == "true" ?  : 1}"
+  name = "${var.classifier_use_prefix == "true" ? local.classifier_prefix : ""}streamalert_classified_logs"
+  # name = "streamalert_classified_logs"
 
   # The amount of time messages are hidden after being received from a consumer
   # Default this to 2 seconds longer than the maximum AWS Lambda duration
--- a/terraform/modules/tf_stream_alert_globals/variables.tf
+++ b/terraform/modules/tf_stream_alert_globals/variables.tf
@@ -2,6 +2,10 @@ variable "account_id" {}
 
 variable "prefix" {}
 
+variable "classifier_use_prefix" {
+   description = "Allows support for multiple instances of StreamAlert on one account"
+}
+
 variable "region" {}
 
 variable "kms_key_arn" {}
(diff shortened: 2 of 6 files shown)


ryandeivert:
hello @hkarumuri and thank you for the PR!! you're on the right track here, but unfortunately this change is currently a (very) breaking change and will require a bit more work.

What we have discussed is actually having a configurable 'queue' url and having the `classifier_queue` terraform resource be optional.. meaning if a queue is specified in the config, use it, otherwise create a queue to use. I'm happy to point out the other changes that must take place if you'd like, or would you like to take a whack at it?

hkarumuri:
hello @ryandeivert! I'd like to go ahead and take a whack at it. Details on why this change causes StreamAlert to break would be very helpful. thanks!

ryandeivert:
This is a breaking change for current deployments because it will cause the queue resource to get destroyed/recreated with the new name (you cannot rename in place to my knowledge)

This means that any messages in the queue will be lost, and there could be gaps in processing.

We could theoretically have a 'upgrade guide', but this is probably more work than just writing the config logic.

hkarumuri:
Ah ok. Is it possible to make the classifier_queue a shared resource?

ryandeivert:
@hkarumuri can you elaborate on what you mean by a shared resource..? shared in what way?

Ryxias:
Hi @hkarumuri 

As @ryandeivert mentioned, we have to be aware of existing consumers of StreamAlert and how they will complete the migration without losing in-flight SQS frames. Destroying the old SQS queue with frames in-flight can pose a potential security risk.

Instead, there will have to be a more involved migration process. It will definitely look _something_ like this:

1. Create a 2nd SQS queue, correctly named with a prefix
2. Create a 2nd Rules Engine mapping to trigger from this new SQS queue
3. _**Build & deploy this commit**_
4. Direct all Classified logs to the new SQS queue
5. _**Deploy this commit**_
6. Wait until the old SQS queue drains naturally
7. Remove the original Rules Engine event source mapping
8. Remove the old SQS queue
9. _**Build & deploy this commit, which will destroy the old resources**_
10. Write an upgrade document detailing the process

hkarumuri:
Here are the changes made: 
- The solution I attempted was to provide a conditional classifier – the global.json file in the conf directory now has a "classifier_use_prefix" that can be set to true or false. 

- sqs.tf was modified to use a ternary on whether the prefix should be used based on what is located in global.json. 

- The main.tf and generate.py was modified in order to allow were made to allow the new variable to be read into terraform.

terraform/modules/tf_stream_alert_globals/classifier_queue/sqs.tf
```diff
@@ -1,6 +1,10 @@
 // SQS Queue: Send logs from the Classifier to the SQS queue
+locals {
+  classifier_prefix = "${var.prefix}_"
```
Ryxias: I would either put the ternary operator here, or omit the use of this local altogether.

terraform/modules/tf_stream_alert_globals/variables.tf
```diff
@@ -2,6 +2,10 @@ variable "account_id" {}
 
 variable "prefix" {}
 
+variable "classifier_use_prefix" {
```
Ryxias: I like the idea of using levers to accommodate the migration. This way you can perform the whole migration by only changing variables.

That said, this won't quite work, as Terraform will still destroy the old SQS resource while creating the new one. Take a look at my [previous comment](https://github.com/airbnb/streamalert/pull/955#issuecomment-509806439)

smiller171: It might be reasonable to just not upgrade an in-place environment and only add a prefix to new installations. We're defaulting to false here on the prefix so unless you explicitly decide to add a prefix to an existing environment this code is non-destructive. I'm not sure we understand the structure of your code well enough to build the logic to replace an existing queue correctly.

Ryxias: No worry, I got you guys. Going to punch in a fix to the prefix problem today/tomorrow. It's been bothering me for months too

Stay tuned...

smiller171: Thanks @Ryxias

Ryxias:
Hi @smiller171 and @hkarumuri please feel free to take a look at #958

Ryxias:
Hey @hkarumuri, I ended up coming up with a solution that is basically the same as yours over in #960. The only major difference is the "default" behavior, as well as where the configuration node is specified.

cc @smiller171

hkarumuri:
Hey that's sweet! If that's the case, I'm going to go ahead and close this out

PR #958, opened 2019-07-11
Migration Commit to add resource prefix to Classified logs SQS

Ryxias:
to: @ryandeivert 
related to: #955 
resolves: #885 

## Background

During our SQS migration some time late in 2018, we introduced the use of AWS SQS as a dispatch layer between the `classifier` functions and the `rules_engine` functions.

However, the SQS resource had a minor bug that caused to be missing a resource prefix (it's currently named `streamalert_classified_logs`).

This is not a problem on any AWS environment where there is only 1 deployment of StreamAlert. When you need to deploy 2 copies in parallel, however, this causes resource conflict issues.

## Changes

This is **THE FIRST COMMIT OF TWO** that will be involved in the migration process.

## Upgrade Process
Due to the stateful nature of AWS SQS, the "renaming" of the SQS queue cannot be done safely. This is because Terraform does not rename resources; it destroys the old one and creates a new one.

Doing so on a production environment where there are SQS records in-flight will permanently lose unprocessed records on the SQS, when it is destroyed.

This PR is intended to offer a migration path to allow users to upgrade StreamAlert 2.x.x —> 3.0.0 (fixing the prefix problem) without losing data.

If you don't care about data loss, then you can skip this migration completely.

### Step 1: Upgrade to latest 2.x.x StreamAlert
Not mandatory but this was tested against `master` so there's less chance for problems if you start the migration at 2.3.0+.

### Step 2: Check out **this specific commit**
This commit cannot be skipped without risking losing data.

### Step 3: Build SQS
`$ ./manage.py build -t globals`

This commit creates a second, identical SQS queue in StreamAlert that operates in parallel to the original `streamalert_classified_logs` queue. This new queue is named properly, incorporating the `$var.prefix`.

Initially after being built, this SQS queue will just sit as it is not connected to anything. That comes later...

### Step 4: Build other stuff
`$ ./manage.py build`

This goes ahead and builds all of the rest of the stuff, including attaching SQS to the `rules_engine` and the `classifier`.

However, it is notable that, despite having the new `SQS_QUEUE_URL` on the `classifier` Lambda environment variables, this only applies to the newest Lambda versions. The kinesis streams (and other cluster ingress points) are still attached to old versions of the `classifier` Lambda, which means their records will continue to go to the old SQS queue until...

### Step 5: Deploy the classifier
`$ ./manage.py deploy --function classifier`

From here on, all new records from the classifiers will be sent to the new SQS queue, whereas the old SQS queue will continue to trigger the `rules_engine`.

### Step 6: Wait until the old SQS Drains
Watch the number of "messages in flight" drain away in the old SQS queue. This should not take very long... probably.

### Step 7: Deploy the NEXT COMMIT 
Now go ahead and checkout #959 and build/deploy that. (The 2nd one is at #959)

## Testing

Painfully. Next time ideally with more alcohol

### DIFF
--- a/stream_alert_cli/terraform/classifier.py
+++ b/stream_alert_cli/terraform/classifier.py
@@ -69,8 +69,8 @@ def generate_classifier(cluster_name, cluster_dict, config):
         'function_role_id': '${{module.{}_lambda.role_id}}'.format(tf_module_prefix),
         'function_alias_arn': '${{module.{}_lambda.function_alias_arn}}'.format(tf_module_prefix),
         'function_name': '${{module.{}_lambda.function_name}}'.format(tf_module_prefix),
-        'classifier_sqs_queue_arn': '${module.globals.classifier_sqs_queue_arn}',
-        'classifier_sqs_queue_url': '${module.globals.classifier_sqs_queue_url}',
+        'classifier_sqs_queue_arn': '${module.globals.classifier_destination_sqs_queue_arn}',
+        'classifier_sqs_queue_url': '${module.globals.classifier_destination_sqs_queue_url}',
         'classifier_sqs_sse_kms_key_arn': '${module.globals.classifier_sqs_sse_kms_key_arn}',
     }
 
--- a/stream_alert_cli/terraform/classifier.py
+++ b/stream_alert_cli/terraform/classifier.py
@@ -95,7 +95,7 @@ def generate_classifier(cluster_name, cluster_dict, config):
         config,
         environment={
             'CLUSTER': cluster_name,
-            'SQS_QUEUE_URL': '${module.globals.classifier_sqs_queue_url}',
+            'SQS_QUEUE_URL': '${module.globals.classifier_destination_sqs_queue_url}',
         },
         tags={
             'Cluster': cluster_name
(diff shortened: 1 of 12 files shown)


stream_alert_cli/terraform/classifier.py
```diff
@@ -69,8 +69,8 @@ def generate_classifier(cluster_name, cluster_dict, config):
         'function_role_id': '${{module.{}_lambda.role_id}}'.format(tf_module_prefix),
         'function_alias_arn': '${{module.{}_lambda.function_alias_arn}}'.format(tf_module_prefix),
         'function_name': '${{module.{}_lambda.function_name}}'.format(tf_module_prefix),
-        'classifier_sqs_queue_arn': '${module.globals.classifier_sqs_queue_arn}',
-        'classifier_sqs_queue_url': '${module.globals.classifier_sqs_queue_url}',
```
Ryxias: The classifier terraform module was not using the URL. It was used in `SQS_QUEUE_URL` environment variable, which is below.

ryandeivert: thanks for the note

Ryxias:
After some discussion, we've concluded that there is an easier way to do this

smiller171:
> After some discussion, we've concluded that there is an easier way to do this

Are there more details available about the path you plan on taking?

Ryxias:
@smiller171 hi sorry

I've been meaning to put a pull request for the change out, but I've been distracted by other things this week; I'll try to get it sometime early next week

The approach we're going to take is simple: allow the configuration files to specify name the SQS queue. By default, the configuration will recommend a prefixed SQS resource name. But it will not explicitly rename a SQS url that already exists. This allow us to upgrade without causing data loss for StreamAlert users who currently have the misspelled `streamalert_classified_logs` SQS queue name, as they can explicitly state they will use that SQS queue. For StreamAlert users that do not care about data loss due to low volume, or for new setups of StreamAlert, they can adopt the default convention of using the prefixed SQS queue name.

PR #959, opened 2019-07-12
StreamAlert SQS missing prefixes migration cleanup

Ryxias:
to: @ryandeivert 
related to: #958 

## Background

Cleans up the code in the migration commit and destroys the SQS queue that is no longer needed.

## Changes

This is the **SECOND COMMIT OF TWO**. If you're here and haven't read #958 yet, YOU NEED TO GO OVER THERE TO UNDERSTAND WHAT'S GOING ON!!!!!!

In this PR, the old SQS resource that was improperly named without a prefix is deleted, along with other supporting resources.

### If you did the migration...
Doing `$ ./manage.py build` will destroy the old, now unused SQS Queue `streamalert_classified_logs`.

### If you did not do the migration...
**If you did not do the migration you are at risk of losing data. You've been warned!!!**

To do the deployment correctly **_you will need build twice AND deploy the classifier_**

```
$ ./manage.py build
$ ....
$ ./manage.py build
$ ....
$ ./manage.py deploy --function classifier
```

This has to do with the generation of SQS URLs and other lame stuff.

## Testing

TODO (haven't gotten to this yet).

### DIFF
--- a/stream_alert_cli/terraform/classifier.py
+++ b/stream_alert_cli/terraform/classifier.py
@@ -92,7 +92,7 @@ def generate_classifier(cluster_name, cluster_dict, config):
         config,
         environment={
             'CLUSTER': cluster_name,
-            'SQS_QUEUE_URL': '${module.globals.new_classifier_sqs_queue_url}',
+            'SQS_QUEUE_URL': '${module.globals.classifier_sqs_queue_url}',
         },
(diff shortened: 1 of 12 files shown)


Ryxias:
After some discussion we've concluded there's an easier way to do this

PR #960, opened 2019-07-30
Adds optional prefix to SQS URL

Ryxias:
to: @ryandeivert  @chunyong-lin 

related to: #955 
resolves: #885 

## Background
As explained in #885, there is a bug in StreamAlert where the classifier SQS resource does not properly adopt the `PREFIX`.

This problem is complicated by the fact that AWS SQS is stateful, and the resource cannot be renamed without destroying it.

## Changes
To address the issue, we adopted the suggestion in #955; allowing for the SQS resource name to be conditionally specified.

Now, users can add the StreamAlert prefix to _**NEW**_ classifier SQS deployments. This is configured on `global.json` through the parameter:

`globals.json`
```
  ...
  "infrastructure": {
    "classifier_sqs": {
      "use_prefix": null
    }
  },
  ..
```

The default value currently is `None`/`null`. This is not actually a valid value and will crash during `manage.py generate`. The error message instructs StreamAlert users to specify a value of `true` or `false`:

> (venv) ~/extrepos/streamalert-deploy (dw--sqs-url)$ ./manage.py generate
    raise MisconfigurationError(message)
stream_alert_cli.terraform.common.MisconfigurationError: [WARNING] As of StreamAlert 2.3.0+ you must specify the classifier_sqs.use_prefix parameter in global.json. For existing/legacy deployments, change this value to False. For new deployments, change this value to True. For more information, refer to https://github.com/airbnb/streamalert/pull/960.

For existing production deployments of StreamAlert, the value must be specified as `false`. This PR will not rename the existing resource.

For all new deployments of StreamAlert, the value should be specified as `true`. In a future 3.0.0+ release, prefixing will be the "default" behavior.

## Testing

Deployed to stage; following cases:

### Case 1: New deploy, `use_prefix` = true
Correctly creates a prefixed SQS

### Case 2: Old deploy, `use_prefix` not specified
Errors (correctly)

### Case 3: Old deploy, `use_prefix` = false
On stage it just errors because it resource conflicts with another StreamAlert's un-prefixed SQS (which is correct)

### Case 4: Old deploy, `use_prefix` = true
Completely renames the old SQS queue, destroying the old resource. (Correct)

### DIFF
--- a/conf/global.json
+++ b/conf/global.json
@@ -35,6 +35,9 @@
         "read_capacity": 20,
         "write_capacity": 5
       }
+    },
+    "classifier_sqs": {
+      "use_prefix": null
     }
   },
   "s3_access_logging": {
--- a/stream_alert_cli/terraform/common.py
+++ b/stream_alert_cli/terraform/common.py
@@ -48,4 +48,3 @@ def monitoring_topic_arn(config):
 
 class MisconfigurationError(ValueError):
     """This error is thrown when StreamAlert is misconfigured."""
-    pass
--- a/stream_alert_cli/terraform/common.py
+++ b/stream_alert_cli/terraform/common.py
@@ -44,3 +44,8 @@ def monitoring_topic_arn(config):
         account_id=config['global']['account']['aws_account_id'],
         topic=topic_name
     )
+
+
+class MisconfigurationError(ValueError):
+    """This error is thrown when StreamAlert is misconfigured."""
+    pass
--- a/stream_alert_cli/terraform/classifier.py
+++ b/stream_alert_cli/terraform/classifier.py
@@ -70,7 +70,6 @@ def generate_classifier(cluster_name, cluster_dict, config):
         'function_alias_arn': '${{module.{}_lambda.function_alias_arn}}'.format(tf_module_prefix),
         'function_name': '${{module.{}_lambda.function_name}}'.format(tf_module_prefix),
         'classifier_sqs_queue_arn': '${module.globals.classifier_sqs_queue_arn}',
-        'classifier_sqs_queue_url': '${module.globals.classifier_sqs_queue_url}',
         'classifier_sqs_sse_kms_key_arn': '${module.globals.classifier_sqs_sse_kms_key_arn}',
     }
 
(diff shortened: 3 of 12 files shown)


smiller171:
I wonder if the added friction here of not assuming a working value to begin with will break anyone's automated deployments? Can we make a call to Terraform to check whether it's already been deployed and set the initial value accordingly?

conf/global.json
```diff
@@ -35,6 +35,9 @@
         "read_capacity": 20,
         "write_capacity": 5
       }
+    },
+    "classifier_sqs": {
```
ryandeivert: we might want to actually put this in the `rules_engine` config in `lambda.json` instead

stream_alert_cli/terraform/common.py
```diff
@@ -44,3 +44,8 @@ def monitoring_topic_arn(config):
         account_id=config['global']['account']['aws_account_id'],
         topic=topic_name
     )
+
+
+class MisconfigurationError(ValueError):
+    """This error is thrown when StreamAlert is misconfigured."""
+    pass
```
ryandeivert: unnecessary pass here (not needed when there is a doc string)

conf/global.json
Ryxias: So I was thinking about that—I wasn't sure about the conventions on what goes in `lambda.json` and what goes in `globals.json`

I chose here because this resource isn't _exactly_ relegated to the rules engine; it also is connected to the classifier. An alternative was to put a configuration under the `classifier` AND the `rules_engine` but that would duplicate data, giving an opportunity to screw things up

ryandeivert: yeah I think this is fine since it also effects a 'global' resource. feel free to keep it

Ryxias:
@smiller171 the intention here is to force everyone upgrading to 2.3.0+ to make a conscious choice. It is intended to hard stop the deployment process until a developer reads the instructions and fixes it.

The reason for the friction is because I want to do both of the following:
1) Dissuade new adopters of StreamAlert from using a wrong "default" SQS name
2) Prevent accidental data loss from existing StreamAlert users

These two require opposing "default" behaviors. 

In 3.0.0+ I will change the default behavior to use the correct SQS name. As part of the major version upgrade, it will be expected that issues will occur in reverse-compatibility.

PR #965, opened 2019-08-12
Fixed logging output for failed classifications

blakemotl:
to: @Ryxias 
related to:
resolves:

## Background

When a record cannot be classified it is logged to CloudWatch. In the current state when printing out a payload, the code relies on the underlying data type to format itself as a string. However this can lead to inconsistent behavior with dictionaries especially when python thinks that there is unicode in the dictionary. This makes working with failed classifications a pain because the output varies. Sometimes it gets output as an actual JSON string, sometimes it gets output as a unicode python dictionary.

## Changes

* Updated payload_base to encode the record as a JSON string using json.dumps
* Updated tests to reflect the change in output(actual JSON string versus python representation of a dictionary)

## Testing

Passed unit and integration tests locally.

### DIFF
--- a/stream_alert/classifier/payload/payload_base.py
+++ b/stream_alert/classifier/payload/payload_base.py
@@ -68,7 +68,7 @@ class PayloadRecord(object):
         except (TypeError, ValueError):
             record_data = self._record_data
             invalid_records = self.invalid_records
-            LOGGER.warning('A PayloadRecord has records that is not serializable as JSON')
+            LOGGER.debug('A PayloadRecord has data that is not serializable as JSON')
 
         if not self:
             return '<{} valid:{}; raw record:{};>'.format(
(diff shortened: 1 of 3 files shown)


stream_alert/classifier/payload/payload_base.py
```diff
@@ -62,11 +62,19 @@ def __len__(self):
         )
 
     def __repr__(self):
+        try:
+            record_data = json.dumps(self._record_data)
+            invalid_records = json.dumps(self.invalid_records)
+        except (TypeError, ValueError):
+            record_data = self._record_data
+            invalid_records = self.invalid_records
+            LOGGER.warning('A PayloadRecord has records that is not serializable as JSON')
```
ryandeivert: typo?

PR #968, opened 2019-08-15
Fixed classifier log output if payload record is a string

blakemotl:
to: @chunyong-lin 
related to: https://github.com/airbnb/streamalert/pull/965
resolves:

## Background

The previous change fixed the case where the payload record stored its data as a dictionary, but did not account for the situation where the payload is valid JSON, but is stored as a string already. This caused some messy output for failed classifications.

## Changes

* Will now only attempt conversion if the payload data is not stored as a string.

## Testing

Passed unit/integration tests locally.

### DIFF
--- a/stream_alert/classifier/payload/payload_base.py
+++ b/stream_alert/classifier/payload/payload_base.py
@@ -64,15 +64,17 @@ class PayloadRecord(object):
     def __repr__(self):
         record_data = self._record_data
         invalid_records = self.invalid_records
+
         try:
-            if not isinstance(self._record_data, str):
-                record_data = json.dumps(self._record_data)
+            if isinstance(record_data, dict):
+                record_data = json.dumps(record_data)
         except (TypeError, ValueError):
             LOGGER.debug('A PayloadRecord has data that is not serializable as JSON')
 
         try:
-            if not isinstance(invalid_records, str):
-                invalid_records = json.dumps(self.invalid_records)
+            #Build a JSON string if data is a dict or non-empty list of dicts
+            if isinstance(invalid_records, dict) or (isinstance(invalid_records, list) and invalid_records and isinstance(invalid_records[0], dict)):
+                invalid_records = json.dumps(invalid_records)
         except (TypeError, ValueError):
             LOGGER.debug('A PayloadRecord has invalid records that are not serializable as JSON')
 
@@ -229,7 +231,7 @@ class StreamPayload(object):
             )
         raw_record = self.raw_record
         try:
-            if not isinstance(raw_record, str):
+            if isinstance(raw_record, dict):
                 raw_record = json.dumps(self.raw_record)
         except (TypeError, ValueError):
             LOGGER.debug('A StreamPayload has data that is not serializable as JSON')
--- a/stream_alert/classifier/payload/payload_base.py
+++ b/stream_alert/classifier/payload/payload_base.py
@@ -62,13 +62,19 @@ class PayloadRecord(object):
         )
 
     def __repr__(self):
+        record_data = self._record_data
+        invalid_records = self.invalid_records
         try:
-            record_data = json.dumps(self._record_data)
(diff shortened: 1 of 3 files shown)


ryandeivert:
@blakemotl / @chunyong-lin what is the status of this? should it be updated to be merged? or are we scrapping it?

chunyong-lin:
@ryandeivert I haven't reviewed this PR. @blakemotl let's fix the CI and I will review the PR when it is ready.

Ryxias:
Suggestion to rebase this onto `release-3-0-0` and release it there

blakemotl:
@chunyong-lin Fixed the CI and switched to release 3-0-0

stream_alert/classifier/payload/payload_base.py
```diff
@@ -62,14 +62,24 @@ def __len__(self):
         )
 
     def __repr__(self):
+        record_data = self._record_data
+        invalid_records = self.invalid_records
+
         try:
-            record_data = json.dumps(self._record_data)
-            invalid_records = json.dumps(self.invalid_records)
+            if isinstance(record_data, dict):
```
Ryxias: Seems like this should invert the `try` and `if`:

```
if isinstance(record_data, dict):
  try:
    record_data = json.dumps(self._record_data) 
    #....
```

stream_alert/classifier/payload/payload_base.py
```diff
@@ -62,14 +62,24 @@ def __len__(self):
...
+                record_data = json.dumps(record_data)
         except (TypeError, ValueError):
-            record_data = self._record_data
-            invalid_records = self.invalid_records
             LOGGER.debug('A PayloadRecord has data that is not serializable as JSON')
 
+        try:
+            #Build a JSON string if data is a dict or non-empty list of dicts
```
Ryxias: 1 space after `#` in comments

stream_alert/classifier/payload/payload_base.py
```diff
@@ -62,14 +62,24 @@ def __len__(self):
...
+            if (isinstance(invalid_records, dict) or
+                    (isinstance(invalid_records, list) and
+                     invalid_records and isinstance(invalid_records[0], dict))):
```
Ryxias: I would not do the `invalid_records[0]` check here and simply do `isinstance(invalid_records, list)`. Nested type checking is a really tough problem which isn't really worth solving.

blakemotl: I agree with you on this. However in testing I found some situations/test cases where the underlying data is a list of strings. Dumping this as a json string caused some weird encoding.

stream_alert/classifier/payload/payload_base.py
blakemotl: @Ryxias Do you know if you can setup pylint or IDE to detect or automatically add the space?

stream_alert/classifier/payload/payload_base.py
Ryxias: Can you show me the exact case that's causing the issue? It sounds like a good candidate for writing a unit test

ryandeivert:
@blakemotl is this ready for final review?

stream_alert/classifier/payload/payload_base.py
blakemotl: @Ryxias This test: test_repr_invalid_records
And you end up with [{\'key\': \'value\'}] instead of [{"key": "value"}] if you don't check one level down.

blakemotl:
@ryandeivert yes it is

stream_alert/classifier/payload/payload_base.py
blakemotl: @Ryxias In this case it is a list of dicts not a list of strings so it should be {"key": "value"} not {\'key\': \'value\'}. So, I DO want to encode as JSON if the object inside the list is a dict but not if its a string.

PR #969, opened 2019-08-15
[LookupTables] Phase 1 implementation of LookupTables rebuild

Ryxias:
to: @ryandeivert , @blakemotl , @chunyong-lin 
related to:
resolves:

## Background

LookupTables is making a comeback, this time with twice the gusto!

LookupTables was originally a stealth feature that was released for a very specific purpose in the Rules Engine. This PR expands LookupTables to be a publicly available feature, now functional on all Lambda Functions, instead of just on the Rules Engine.

LookupTables is intended to be a simple key value store that any StreamAlert lambda function can reach into for a variety of additional data:
* Metadata that is not included in the incoming telemetry
* Additional configuration that controls behavior (example: Gating new features, controlling Publishers, etc)
* Whitelists/Blacklists that are not hard-coded in Python Code
* (Future) Threat Intel

### Step 1) Configure
There is a new file, `conf/lookup_tables.json` that houses the configuration of LookupTables. This configuration allows you to define LookupTables, their backends, and their behavior.

### Step 2) Put Data
Throw data into AWS S3 or AWS DynamoDB!

### Step 3) Use!
The syntax is super easy. Simply do:

```
from stream_alert.shared.lookup_tables.core import LookupTables

my_value = LookupTables.get('my-table', 'my-key')
```

Furthermore, you can use this code anywhere in the application, not just in the Rules Engine!

### Redesigns LookupTables syntax
As above.

### LookupTables now supports DynamoDB as a backing mechanism
Documentation coming, but you can take a look at (THIS FILE) for examples on how to configure it.

## Breaking Changes!!!! ⚠️ 
Beware that in the original implementation of LookupTables, because the tables were loaded in entirety and returned as a fully-populated dict, performing dictionary iterations was possible:

```
map = LookupTables.get('my-table')
for k, v in map.iteritems():
  # ...
```

When supporting other databases such as DynamoDb, this behavior is equivalent to performing table scans. This is an anti-pattern and thus this access pattern is no longer supported. The reasoning is because this necessarily causes large amount of network I/O **_PLUS_** memory utilization burden on AWS Lambda functions, which increases costs significantly.

Existing systems that perform "table-scan-like" behavior on LookupTable will have to redesign the access pattern. More coming later ... when I get around to writing the documentation...

## Testing

See stickied comment below.


Ryxias:
STICKY #1

Ryxias:
STICKY COMMENT 2

stream_alert/shared/lookup_tables/core.py
```diff
@@ -0,0 +1,160 @@
...
+            table_name (str): The name of the lookup table desired
+
+        Returns:
+            LookupTable: An individual instance. Not to be mixed up with LookupTables (plural!)
+        """
+        if table_name in self._tables:
+            return self._tables[table_name]
+
+        LOGGER.error(
+            (
```
ryandeivert: this is gonna seem like a lie - but the parenthesis around the first are are not required:
```
LOGGER.error(
    'Nonexistent LookupTable \'%s\' referenced. Defaulting to null table. '  # <-- no comma here
    'Valid tables were (%s)',
    table_name,
    ', '.join(self._tables.keys())
)
```

just omit the comma at the end of the first arg's first line

stream_alert/shared/lookup_tables/core.py
```diff
@@ -0,0 +1,160 @@
...
+                'Nonexistent LookupTable \'%s\' referenced. Defaulting to null table. '
+                'Valid tables were (%s)'
+            ),
+            table_name,
+            ', '.join(self._tables.keys())
```
ryandeivert: recommendation: use `', '.joined(sorted(self._tables))` for consistent printing (`keys()` is non-deterministic)

stream_alert/shared/lookup_tables/core.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+        # FIXME (derek.wang) Would it be preferable to raise an exception instead?
+        #  In the current implementation if it returns a NullTable, the code could continue
+        #  executing with None values everywhere, which may cause undesirable behavior. Maybe
+        #  fail fast?
+        return self._null_table
+
+        # raise LookupTablesError('Nonexistent LookupTable {} referenced.'.format(table_name))
```
ryandeivert: you could theoretically raise this here, but beware that (relevant to how these are currently used) rule logic is wrapped in a bare except that will just log the exception anyway. logging here is probably fine for now

stream_alert/shared/lookup_tables/drivers.py
```diff
@@ -0,0 +1,146 @@
+from abc import abstractmethod, ABCMeta
+from stream_alert.shared.logger import get_logger
+from stream_alert.shared.lookup_tables.errors import LookupTablesConfigurationError
+
+LOGGER = get_logger(__name__)
+
+
+# pylint: disable=invalid-name
```
ryandeivert: why is this name invalid?

stream_alert/shared/lookup_tables/drivers.py
```diff
@@ -0,0 +1,146 @@
...
+        Configures and initializes this driver
+
+        Initialization is called exactly once, and is should always called BEFORE any other
+        interaction (get/set/commit) is made with the driver.
+        """
+
+    @abstractmethod
+    def commit(self):
+        """
+        Takes any changes and flushes them to remote storage.
```
ryandeivert: could be on one line (like `get` below)

stream_alert/shared/lookup_tables/drivers.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def construct_persistence_driver(table_configuration):
+    """
+    Constructs a raw, uninitialized PersistenceDriver from the given configuration.
+
+    :args
+        table_configuration (dict)
+
+    :return
```
ryandeivert: `:args` and `:return` switched to different format here

stream_alert/shared/lookup_tables/drivers.py
```diff
@@ -0,0 +1,146 @@
...
+        PersistenceDriver
+    """
+    import stream_alert.shared.lookup_tables.driver_dynamodb as driver_dynamodb
```
ryandeivert: why do these imports happen here? it's fine to do these at top of file even if unused

stream_alert/shared/lookup_tables/drivers.py
```diff
@@ -0,0 +1,146 @@
...
+        return EphemeralDriver(table_configuration)
+    else:
+        raise LookupTablesConfigurationError(
+            'Unrecognized driver name: {}'.format(driver_name)
+        )
+
+
+class EphemeralDriver(PersistenceDriver):
+    """
+    Ephemeral persistence driver
```
ryandeivert: lol `ephemeral` `persistence` 🤕

ryandeivert: also don't change it, it's funny

stream_alert/shared/lookup_tables/errors.py
```diff
@@ -0,0 +1,15 @@
+
+class LookupTablesError(RuntimeError):
+    pass
+
+
+class LookupTablesInitializationError(LookupTablesError):
+    pass
+
+
+class LookupTablesCommitError(LookupTablesError):
```
ryandeivert: some doc strings on what these different errors are for might be nice. up to you

stream_alert/shared/lookup_tables/table.py
```diff
@@ -0,0 +1,56 @@
+from stream_alert.shared.logger import get_logger
+
+LOGGER = get_logger(__name__)
+
+
+class LookupTable(dict):
```
ryandeivert: if these are just embedding a driver that handles the logic, can you remind me again why we need to subclass `dict` here?

stream_alert/shared/lookup_tables/driver_dynamodb.py
```diff
@@ -0,0 +1,194 @@
...
+        try:
+            boto_config = boto_helpers.default_config(timeout=10)
+            resource = boto3.resource('dynamodb', config=boto_config)
+            self._table = resource.Table(self._dynamo_db_table)
+            _ = self._table.table_arn  # This is only here to blow up on invalid tables
+        except ClientError as err:
+            message = (
+                'LookupTable ({}): Encountered error while connecting with DynamoDB: \'{}\''
+            ).format(self.id, err.response['Error']['Message'])
+            LOGGER.error(message)
```
ryandeivert: should be okay to just raise and not log here

stream_alert/shared/lookup_tables/driver_dynamodb.py
```diff
@@ -0,0 +1,194 @@
...
+        pass
+
+    def _reload_if_necessary(self, key):
+        """
+        Uses the "cache_refresh_minutes" option to determine whether or not the current LookupTable
+        should be re-fetched from S3.
+
+        If it needs a reload, this method will appropriately call reload.
+        """
+        last_load_time = self._dynamo_load_times.get(key, 0)
```
ryandeivert: neato

stream_alert/shared/lookup_tables/driver_dynamodb.py
```diff
@@ -0,0 +1,194 @@
...
+        else:
+            key_schema = {
+                self._dynamo_db_partition_key: key,
+            }
+
+        LOGGER.debug(
+            'LookupTable (%s): Loading key \'%s\' with schema (%s)',
+            self.id,
+            key,
+            json.dumps(key_schema)
```
ryandeivert: keep in mind `json.dumps` is actually fairly costly and can slow things down if this function is called a lot. it's also why we've guarded this other places: 

https://github.com/airbnb/streamalert/blob/0cc53101fd3e22a02bd9bbce9fddc6abcffe2d0d/stream_alert/classifier/parsers.py#L277

stream_alert/shared/lookup_tables/driver_dynamodb.py
```diff
@@ -0,0 +1,194 @@
...
+
+        try:
+            response = self._table.get_item(
+                Key=key_schema,
+
+                # It's not urgently vital to do consistent reads; we accept that for some time we
+                # may get out-of-date reads.
+                ConsistentRead=False,
+
+                # FIXME (derek.wang) Should be off for non-debug mode
```
ryandeivert: are you saying you only want to do `ReturnConsumedCapacity='TOTAL'` if debug is on? you could use the same "guard" mentioned above if so. then set your arts to `get_item` like so:

```python

args = {
  'Key': key_schema,
  'ConsistentRead': False
}

if LOGGER_DEBUG_ENABLED:
  args['ReturnConsumedCapacity'] = 'TOTAL'

response = self._table.get_item(**args)
```

stream_alert/shared/lookup_tables/driver_s3.py
```diff
@@ -0,0 +1,199 @@
...
+    def initialize(self):
+        LOGGER.info('LookupTable (%s): Running initialization routine', self.id)
+        self._reload()
+
+    def commit(self):
+        if not self._dirty:
+            LOGGER.warn('LookupTable (%s): Empty commit; no records dirtied', self.id)
+            return
+
+        raise NotImplementedError('Help derek you screwed me')
```
ryandeivert: :troll_dance:

stream_alert/shared/lookup_tables/driver_s3.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def get(self, key, default=None):
+        self._reload_if_necessary()
+
+        return self._s3_data.get(key, default)
+
+    def set(self, key, value):
+        self._s3_data.set(key, value)
```
ryandeivert: is there a unit test for this? `dict` (which I think `self._s3_data` is according to [this](https://github.com/airbnb/streamalert/pull/969/files#diff-df361a7400cf1aa32ce5675d89932844R59)) does not have a `set` attribute

stream_alert/shared/lookup_tables/driver_s3.py
```diff
@@ -0,0 +1,199 @@
...
+        self._compression = configuration.get('compression', False)
+        self._cache_refresh_minutes = configuration.get(
+            'cache_refresh_minutes',
+            self._DEFAULT_CACHE_REFRESH_MINUTES
+        )
+
+        self._last_load_time = None
+
+        self._s3_data = {}
+        self._dirty = False
```
ryandeivert: can you leave a comment as to what the dirty attribute is for

stream_alert/shared/lookup_tables/driver_s3.py
```diff
@@ -0,0 +1,199 @@
...
+                    'LookupTable (%s): Data is not compressed; defaulting to original payload',
+                    self.id
+                )
+        else:
+            LOGGER.debug('LookupTable (%s): File does not need decompression')
+
+        # Decode the data; right now we make the assumption that the data is always encoded
+        # as JSON.
+        try:
+            self._s3_data = json.loads(data)
```
ryandeivert: should we support other "formats" as well? like yaml or ini? idk if we should

stream_alert/shared/lookup_tables/core.py
Ryxias: dafuq

stream_alert/shared/lookup_tables/driver_s3.py
Ryxias: My bad; I updated it later but actually the `.set()` methods are not implemented yet (going to do in a future commit). Will update to `NotImplementedErr`

stream_alert/shared/lookup_tables/driver_s3.py
Ryxias: My bad; it's gonna be for using the `.set()` method in the future.... currently not used. Will remove until I get to that future commit

stream_alert/shared/lookup_tables/driver_s3.py
Ryxias: Definitely could support them!

stream_alert/shared/lookup_tables/drivers.py
Ryxias: Bad copypasta

stream_alert/shared/lookup_tables/drivers.py
Ryxias: There's a cyclical dependency between `drivers.py` (the abstract class, which also contacts the factory method) and `driver_***.py`.

The better way of doing it would be to put the `construct_persistence_river` in a separate module.

stream_alert/shared/lookup_tables/table.py
Ryxias: No reason; forgot to remove the `dict` after our discussion

stream_alert/shared/lookup_tables/driver_dynamodb.py
Ryxias: I refactored this significantly later. There's a new class called `DriverCache`

stream_alert/shared/lookup_tables/driver_dynamodb.py
Ryxias: Good to know

stream_alert/shared/lookup_tables/driver_s3.py
Ryxias: :trollface:

stream_alert/shared/lookup_tables/drivers.py
Ryxias: Wait... hold on it's actually valid; there's a property named `id` that I was accidentally catching.

I'll move onto the prop

stream_alert/shared/lookup_tables/driver_s3.py
```diff
@@ -0,0 +1,195 @@
...
+        return self._cache.get(key, default)
+
+    def set(self, key, value):
+        """
+        Under construction
+        FIXME (derek.wang)
+        """
+        self._dirty = True
+
+    def _reload_if_necessary(self):
```
chunyong-lin: This method is almost same in DynamoDB driver class. Can we move the common method(s) to base class `PersistenceDriver`?

stream_alert/shared/lookup_tables/driver_s3.py
```diff
@@ -0,0 +1,195 @@
...
+                )
+            )
+
+        except (ConnectTimeoutError, ReadTimeoutError):
+            # Catching ConnectTimeoutError and ReadTimeoutError from botocore
+            LOGGER.error(
+                'LookupTable (%s): Reading from S3 timed out',
+                self.id
+            )
+            raise LookupTablesInitializationError(
```
chunyong-lin: Should we fail the whole lambda because of LookupTables errors especially timeout. It is common to see timeout? [In current LookupTables code](https://github.com/airbnb/streamalert/blob/master/stream_alert/shared/lookup_tables.py#L71-L78), we log the errors without failing the whole lambda.

conf/lookup_tables.json
```diff
@@ -0,0 +1,5 @@
+{
+  "enabled": false,
+  "tables": {
+  }
```
chunyong-lin: Move to previous line, bc I am crazy.

PR #972, opened 2019-08-16
Bumped the slack request timeout to 10

blakemotl:
related to:
resolves:

## Background

The slack app has been failing consistently due to a read timeout. This also occurs locally when testing the endpoint. Additionally the timeout did not appear to change even if the count param was lowered. Therefore we need to bump the default timeout for the app. 10 was chosen to give it some headroom above 8.5 which is around when the timeout stopped happening in local testing. 

## Changes

* Changed the slack app default request timeout to 10 up from 6.05

## Testing

Integration and unit tests passed locally.

### DIFF
--- a/stream_alert/apps/_apps/slack.py
+++ b/stream_alert/apps/_apps/slack.py
@@ -35,7 +35,7 @@ class SlackApp(AppIntegration):
         contain details about your workspace's integrated apps
     """
 
-    _DEFAULT_REQUEST_TIMEOUT = 6.05
+    _DEFAULT_REQUEST_TIMEOUT = 10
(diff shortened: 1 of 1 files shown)


Ryxias:
I almost want to say:

1) Consider moving timeout into a configuration parameter so we can accommodate for different users' experiences with Slack API
2) Maybe test this internally-only to find the time out "sweet spot", then open a PR after you've found it

blakemotl:
@Ryxias I like the configuration idea but it seems like extra work for only a single value. Additionally, Ryan and I did find the sweet spot at 8.5 but agreed we should give it a little extra headroom, especially since retries don't help here because a timeout means we are getting 0 data back.

PR #983, opened 2019-08-28
bumping slack app timeout to 30 seconds

blakemotl:
to: @chunyong-lin, @ryandeivert 

## Background

The slack app was still timing out with a 10s api request timeout causing the app to constantly fall behind on data. Bumping the timeout to 30s resolved this issue.

## Changes

* Bumped default slack request timeout from 10s to 30s.

## Testing

Unit and integration tests passing locally.


ryandeivert:
@blakemotl pls merge when you have time

PR #984, opened 2019-08-29
LookupTables management via manage.py

Ryxias:

# Background

I've added several `manage.py` commands that allow LookupTables to be managed by the CLI.

### `manage.py lookup-tables describe-tables`
Shows all currently configured tables.

### `manage.py lookup-tables get`
Using the `--table` and `--key` options, queries the respective LookupTable and returns the value at the key.

### `manage.py lookup-tables set`
Using the `--table`, `--key` and `--value` options, it sets the value into the given LookupTable's key.

### DIFF
--- a/stream_alert/shared/lookup_tables/cache.py
+++ b/stream_alert/shared/lookup_tables/cache.py
@@ -131,6 +131,9 @@ class DriverCache:
         for key in keyvalue_data.keys():
             self._ttls[key] = ttl
 
+    def getall(self):
+        return self._data
+
 
 class DriverCacheClock:
     """
--- a/stream_alert/shared/lookup_tables/cache.py
+++ b/stream_alert/shared/lookup_tables/cache.py
@@ -1,5 +1,25 @@
+"""
+Copyright 2017-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+import copy
 from datetime import datetime, timedelta
 import random
+
+from stream_alert.shared.logger import get_logger
+
+LOGGER = get_logger(__name__)
 
 
 class DriverCache:
@@ -88,10 +108,7 @@ class DriverCache:
             value (mixed)
             ttl_minutes (int)
         """
-        self._data[key] = value
-        self._ttls[key] = self._clock.utcnow() + timedelta(minutes=ttl_minutes)
-
-        if 0 < self._maximum_key_count < len(self._data):
+        if 0 < self._maximum_key_count < len(self._data) + 1:
             # Cache is too full, evict a random record
             # The reason for RANDOM eviction is that for a LRU cache replacement policy,
             # Cache pollution can occur for Lambdas that experience extremely deterministic
@@ -99,6 +116,9 @@ class DriverCache:
             selected_key = random.choice(self._ttls.keys())
             del self._ttls[selected_key]
             del self._data[selected_key]
+
+        self._data[key] = value
+        self._ttls[key] = self._clock.utcnow() + timedelta(minutes=ttl_minutes)
 
     def set_blank(self, key, ttl_minutes):
         """
@@ -132,7 +152,7 @@ class DriverCache:
             self._ttls[key] = ttl
 
     def getall(self):
-        return self._data
+        return copy.copy(self._data)
 
 
 class DriverCacheClock:
--- a/stream_alert/shared/lookup_tables/driver_dynamodb.py
+++ b/stream_alert/shared/lookup_tables/driver_dynamodb.py
@@ -232,9 +232,9 @@ class DynamoDBDriver(PersistenceDriver):
             components = key.split(self._key_delimiter, 2)
             if len(components) != 2:
                 message = (
-                    'LookupTable (%s): Invalid key. The requested table requires a sort key, '
-                    'which the provided key \'%s\' does not provide, given the configured '
-                    'delimiter: (%s)'
+                    'LookupTable ({}): Invalid key. The requested table requires a sort key, '
+                    'which the provided key \'{}\' does not provide, given the configured '
+                    'delimiter: ({})'
                 ).format(self.id, key, self._key_delimiter)
                 raise LookupTablesInitializationError(message)
 
--- a/stream_alert/shared/lookup_tables/driver_dynamodb.py
+++ b/stream_alert/shared/lookup_tables/driver_dynamodb.py
@@ -70,6 +70,8 @@ class DynamoDBDriver(PersistenceDriver):
 
         self._table = None
 
+        self._dirty_rows = {}
+
     @property
     def driver_type(self):
         return self.TYPE_DYNAMODB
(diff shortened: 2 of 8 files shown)


stream_alert/shared/lookup_tables/driver_s3.py
```diff
@@ -120,6 +144,131 @@ def _reload(self):
...
+        # Decode the data; right now we make the assumption that the data is always encoded
+        # as JSON.
+        data = Encoding.json_decode(self, data)
+
+        self._cache.setall(data, self._cache_refresh_minutes)
+        self._cache.set(self._MAGIC_CACHE_TTL_KEY, True, self._cache_refresh_minutes)
+        LOGGER.info('LookupTable (%s): Successfully loaded', self.id)
+
+
+class Compression:
```
ryandeivert: up to you if you want to take this suggestion - but you could have the `S3Driver` (and other drivers that might use encoding/compression) use multi-inheritance for this, like:

```
class S3Driver(Compression, Encoding, PersistenceDriver):
    ...
```

so these static methods would be accessible from within the class itself. then again, it might be hard with the `id` stuff.. aside from passing in `self` as an arg

ryandeivert: EDIT: actually I lied.. using a mix-in like that, you would have to convert the methods from `staticmethod`s to instance methods, and `self.id` would have the `id` attribute accessible from `S3Driver`.. again your call! but I know you like mixins :P

stream_alert/shared/lookup_tables/driver_s3.py
```diff
@@ -61,7 +62,13 @@ def __init__(self, configuration):
 
         # Explicitly set timeout for S3 connection. The boto default timeout is 60 seconds.
         boto_config = boto_helpers.default_config(timeout=10)
-        self._s3_client = boto3.resource('s3', config=boto_config)
+
+        self._s3_adapter = S3Adapter(
+            self,
+            boto3.resource('s3', config=boto_config),
```
ryandeivert: see other comment in `S3Adapter` - but I _think_ you could just create the `Bucket` resource right here instead of passing in the `resource` and `bucket_name` as args here.. like:

```
self._s3_adapter = S3Adapter(
    self,
    boto3.resource('s3', config=boto_config).Bucket(self._s3_bucket),
    self._s3_key
)
```

stream_alert/shared/lookup_tables/driver_s3.py
```diff
@@ -120,6 +144,131 @@ def _reload(self):
...
+                'LookupTable (%s): Failed to json decode data', driver.id
+            )
+
+
+class S3Adapter:
+    """Adapter class that manages uploading data to and downloading data from AWS S3"""
+
+    def __init__(self, driver, boto_s3_client, s3_bucket, s3_key):
+        self._driver = driver
+        self._s3_client = boto_s3_client
```
ryandeivert: if I'm reading this right, the `boto_s3_client` here is actually an `S3.ServiceResource`. instead of storing this, couldn't you store the `Bucket` resource directly (which is created/used in the `upload` method)?

stream_alert/shared/lookup_tables/driver_s3.py
ryandeivert: also - could even use `S3Adapter` as a mixin along with these? unsure, since after further review it sorta seems like you're moving _away_ from this approach?

stream_alert_cli/lookup_tables/handler.py
```diff
@@ -0,0 +1,187 @@
...
+    def _setup_set_subparser(cls, subparsers):
+        """
+        Set subcommand:
+
+        $ ./manage.py lookup-tables set -t [table] -k [key] -v [value]
+        """
+        set_parser = generate_subparser(
+            subparsers,
+            'set',
+            description='Validate defined log schemas using integration test files',
```
ryandeivert: I think maybe a copy/paste description here

stream_alert_cli/lookup_tables/handler.py
```diff
@@ -0,0 +1,187 @@
...
+            required=True,
+            default=None
+        )
+
+    @classmethod
+    def _setup_set_subparser(cls, subparsers):
+        """
+        Set subcommand:
+
+        $ ./manage.py lookup-tables set -t [table] -k [key] -v [value]
```
ryandeivert: note - you can use an `epilog` to actually help with providing a good example to the user

stream_alert_cli/lookup_tables/handler.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # pylint: disable=protected-access
+    @staticmethod
+    def _describe_tables_handler(config):
+        LOGGER.info('==== LookupTables; Describe Tables ====\n')
+
+        lookup_tables = LookupTables.get_instance(config=config)
+
+        LOGGER.info('%d Tables:\n', len(lookup_tables._tables))
+        for _, table in lookup_tables._tables.items():
```
ryandeivert: you're welcome to use `print` statements when providing feedback to the user if you'd like (similar to what we do for output during tests). instead you'll have everything prefixed with logger stuff. your call really

ryandeivert: as an aside, it's also much easier to ~handle~ control formatting, etc with print = str.format than with logger output

stream_alert_cli/lookup_tables/handler.py
```diff
@@ -0,0 +1,187 @@
...
+            LOGGER.info(' Table Name: %s', table.table_name)
+            LOGGER.info(' Driver Id: %s', table.driver_id)
+            LOGGER.info(' Driver Type: %s\n', table.driver_type)
+
+    @staticmethod
+    def _get_handler(options, config):
+        """
+
```
ryandeivert: pls fill out docstring or remove if you don't plan to :)

stream_alert_cli/lookup_tables/handler.py
```diff
@@ -0,0 +1,187 @@
...
+        LOGGER.error('Unhandled lookup-tables subcommand %s', subcommand)
+
+    # pylint: disable=protected-access
+    @staticmethod
+    def _describe_tables_handler(config):
+        LOGGER.info('==== LookupTables; Describe Tables ====\n')
+
+        lookup_tables = LookupTables.get_instance(config=config)
+
+        LOGGER.info('%d Tables:\n', len(lookup_tables._tables))
```
ryandeivert: instead of having to `pylint: disable` here, you could also add an instance `@property` that would let you return a read-only reference to the [`_tables` attribute](https://github.com/airbnb/streamalert/blob/dw--lookup-managepy/stream_alert/shared/lookup_tables/core.py#L75). or consider making the `_tables` unprotected if mutation isn't a concern..?

stream_alert_cli/lookup_tables/handler.py
```diff
@@ -0,0 +1,187 @@
...
+
+        table_name = options.table
+        key = options.key
+
+        LOGGER.info('==== LookupTables; Get Key ====')
+
+        LookupTables.get_instance(config=config)
+
+        LOGGER.info('  Table: %s', table_name)
+        LOGGER.info('  Key:   %s', key)
```
ryandeivert: same comment as print vs logger

stream_alert/shared/lookup_tables/cache.py
```diff
@@ -131,6 +131,9 @@ def setall(self, keyvalue_data, ttl_minutes):
         for key in keyvalue_data.keys():
             self._ttls[key] = ttl
 
+    def getall(self):
```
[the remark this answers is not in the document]
Ryxias: Good catch

stream_alert/shared/lookup_tables/driver_dynamodb.py
```diff
@@ -94,17 +96,47 @@ def initialize(self):
...
+                }
+
+                if LOGGER_DEBUG_ENABLED:
+                    put_item_args['ReturnConsumedCapacity'] = 'TOTAL'
+
+                # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services
+                #       /dynamodb.html#DynamoDB.Table.put_item
+                self._table.put_item(**put_item_args)
+
+            except (ConnectTimeoutError, ReadTimeoutError):
```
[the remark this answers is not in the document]
Ryxias: ACK, another good catch

stream_alert/shared/lookup_tables/driver_s3.py
Ryxias: There was no good reason I used static methods. My idea with the `Compression` class was to have another layer where we could support multiple types of compression (gzip being one), that the user could configure. In this way a mixin wouldn't work and we'd have to resort to composition... but static methods don't make this obvious

stream_alert/shared/lookup_tables/driver_s3.py
Ryxias: I'm using the `Object` resource in the `download` method so this wouldn't work. I could refactor it to use `Bucket` and `download_fileobj` but I wanted the code to be as similar to the original LookupTables code (.. where possible) to reduce chance of scalability issues

ryandeivert: the `Bucket` resource has an `Object` method that returns an `Object` resource: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Bucket.Object

stream_alert_cli/lookup_tables/handler.py
Ryxias: 🤦‍♂

stream_alert_cli/lookup_tables/handler.py
Ryxias: I wrote it here because I didn't really want `_tables` to be a publicly accessible element on the LookupTablesCore's API. I didn't want other people to use it at all in (for example) the rules engine, read-only or otherwise

Issue #677, opened 2018-04-11
split `manage.py` for testing and management


ryandeivert:
## Background

The `manage.py` script is becoming too large and we should think about splitting it out into it's two core functionalities: management and testing

The 'testing' part of the current `manage.py` has to do a ton of mocking, etc that the management portion does not need. This change would simplify the flow a lot.

## Desired Change

Split the `manage.py` script out so that only 'management' functionality lives within `manage.py` itself and a secondary testing script would house all of the test functionality.

The current idea would be to also update the `stream_alert_cli` module to be something like `streamalert_mgmt` that would include all functions/helpers for management. A new module should be created like `streamalert_test` that would contain all context for mocking and testing.

ryandeivert:
I don't envision getting much of a gain out of this but will leave open for now to reference

PR #987, opened 2019-08-30
[DO NOT REVIEW[ Refactors Manage.py

Ryxias:
related to: #677

## Background

While building out lookup tables, I noticed that `manage.py` is complex and fragmented in its functionality.

This PR refactors manage.py to make it easier to add/manage commands. It brings the subparsers closer to the handlers, which reduces the number of places a developer has to look to add/remove/modify commands.

## Changes

* New `CliCommand` abstract class that all commands inherit from.


Ryxias:
This was done before Python3... so I'm abandoning this

PR #988, opened 2019-08-30
fixing trash logging in cloudwatch logs

ryandeivert:
to: @Ryxias or @chunyong-lin 

## Background

I discovered that tracebacks in the python3 Lambda environment log in a terrible format to CloudWatch Logs. One line gets logged per-entry, instead of a uniform stack trace.

image

## Changes

* Adding a custom formatter that replaces new lines with carriage returns.
* Applying formatting to applicable handlers (inside and outside of Lambda env).

## Testing

Adding unit tests for new formatter logic.

### DIFF
--- a/stream_alert/shared/logger.py
+++ b/stream_alert/shared/logger.py
@@ -16,6 +16,44 @@ limitations under the License.
 import logging
 import os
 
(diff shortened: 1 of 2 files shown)


ryandeivert:
Note: this was testing in a standalone lambda to determine printing looks better now

PR #996, opened 2019-09-04
Automatically generates AWS IAM Policies for LookupTables

Ryxias:

## Changes

* The `manage.py generate` process now will automatically generate IAM Policies for all of the `classifiers`, `rule_engine`, `alert_processor`, and `alert_merger` to access LookupTables. It uses the `lookup_tables.json` configurations to determine which resources to grant access to.

## Testing

A long, time consuming process on Staging environment

### DIFF
--- a/stream_alert_cli/terraform/generate.py
+++ b/stream_alert_cli/terraform/generate.py
@@ -389,13 +389,7 @@ def terraform_generate_handler(config, init=False, check_tf=True, check_creds=Tr
 
     # Setup the main.tf.json file
     LOGGER.debug('Generating cluster file: main.tf.json')
-    with open('terraform/main.tf.json', 'w') as tf_file:
-        json.dump(
-            generate_main(config, init=init),
-            tf_file,
-            indent=2,
-            sort_keys=True
-        )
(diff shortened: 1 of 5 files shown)


stream_alert_cli/terraform/generate.py
```diff
@@ -486,9 +486,60 @@ def terraform_generate_handler(config, init=False, check_tf=True, check_creds=Tr
...
+                'account_id': config['global']['account']['aws_account_id'],
+                'region': config['global']['account']['region'],
+                'prefix': config['global']['account']['prefix'],
+                'dynamodb_tables': dynamodb_tables,
+                's3_buckets': s3_buckets,
+                'roles': roles,
+            }
+        }
+    }
+    with open(tf_file_name, 'w') as tf_file:
```
ryandeivert: note to self: we should refactor this tf generation crap so there is a central facility for this logic.. instead of a bunch of `json.dump(...)` nonsense lying around

ryandeivert: lol no need to address now. I was just confused reviewing this as to why you had to do the writing out of the file within this function.. then realize _every_ generation function is like this

Issue #998, opened 2019-09-04
slack app throws JSONDecodeError


chunyong-lin:
## Background
Occasionally, slack app may throw out `JSONDecodeError` after upgrade StreamAlert to python3. Note: the issue only exists in release-3-0-0 branch as I know.
```
[ERROR] JSONDecodeError: Expecting value: line 1 column 1 (char 0)
Traceback (most recent call last):
  File "/var/task/stream_alert/apps/main.py", line 30, in handler
    StreamAlertApp.get_app(event['app_type'])(event, context).gather()
  File "/var/task/stream_alert/apps/app_base.py", line 384, in gather
    while (((self._gather() * self._POLL_BUFFER_MULTIPLIER) + self._sleep_seconds()) <
  File "/var/task/stream_alert/apps/app_base.py", line 41, in _wrapper
    func(*args, **kwargs)
  File "/var/task/stream_alert/apps/app_base.py", line 353, in _gather
    logs = self._gather_logs()
  File "/var/task/stream_alert/apps/_apps/slack.py", line 116, in _gather_logs
    success, response = self._make_post_request(url, headers, data, False)
  File "/var/task/stream_alert/apps/app_base.py", line 57, in _wrapper
    return func(*args, **kwargs)
  File "/var/task/stream_alert/apps/app_base.py", line 338, in _make_post_request
    return self._check_http_response(response), response.json()
  File "/var/task/requests/models.py", line 892, in json
    return complexjson.loads(self.text, **kwargs)
  File "/var/lang/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/var/lang/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/var/lang/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
```

## Desired Change
Find out the root cause and handle the `JSONDecodeError`.

chunyong-lin:
This issue was triggered when slack api endpoint returns status code `504` (Gateway Timeout).

chunyong-lin:
[...]

PR #999, opened 2019-09-05
[apps] Handle JSONDecoderError

chunyong-lin:
related to: #974 
resolves: #998 

## Background
See the details in issue #998 

## Changes

* Handle `JSONDecoderError` by returning `False, None` without calling to `.json()` method in `response` object
* Add unit test case

## Testing
* Added following unit test case to reproduce `JSONDecoderError` exception we see in issue #998.
```
@patch('requests.sessions.Session.request')
    def test_make_post_request_gateway_timeout(self, requests_mock):
        """App Integration - Make Post Request when gateway timed out"""
        # set the return value to an empty response
        # the reponse's content defaults to None, which returns an empty str
        requests_mock.return_value = requests.Response()

        # need to set status_code in the response
        requests_mock.return_value.status_code = 504
        assert_raises(JSONDecodeError, self._app._make_post_request, '', None, None, False)
```
* After we reproduce and well catch the issue, we add current unit test case to test the exception handling flow.

### DIFF
--- a/stream_alert/apps/app_base.py
+++ b/stream_alert/apps/app_base.py
@@ -322,8 +322,10 @@ class AppIntegration(metaclass=ABCMeta):
         """Method for returning the json loaded response for this POST request
 
         Returns:
-            tuple (bool, dict): False if the was an error performing the request,
-                and the dictionary loaded from the json response
+            tuple (bool, dict|None): The first return value will be False if there
+                was an error performing the request.
+                The second return value will be None if JSONDecodeError raised,
+                otherwise it will be the dictionary loaded from the json response.
         """
         LOGGER.debug('[%s] Making POST request on poll #%d', self, self._poll_count)
 
--- a/stream_alert/apps/app_base.py
+++ b/stream_alert/apps/app_base.py
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 from abc import ABCMeta, abstractmethod, abstractproperty
+from json import JSONDecodeError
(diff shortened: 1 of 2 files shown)


stream_alert/apps/app_base.py
```diff
@@ -335,7 +336,15 @@ def _make_post_request(self, full_url, headers, data, is_json=True):
...
-        return self._check_http_response(response), response.json()
+        try:
+            return self._check_http_response(response), response.json()
+        except JSONDecodeError:
+            # https://github.com/airbnb/streamalert/issues/998
+            # When response returns Gateway Timeout with status_code 504, the response
+            # object will return empty string and raises JSONDecoderError a when .json() refers to.
+            # See https://github.com/psf/requests/blob/v2.22.0/requests/models.py#L853
+            # Instead of raising exception, we can just return False, None
+            return False, None
```
ryandeivert: there are a lot of subclasses that depend on this `_make_post_request` function. is there any concern on returning `None` here..? should this return an empty dict (`{}`) instead?

ryandeivert: I just searched quickly and it looks like this change would be safe

ryandeivert: please just update the docstring to include `None` as a possible return value

chunyong-lin: Yeah, I searched around, looks the callers will return earlier when the first return value is `False` and they won't check second return value eventually. This change would be safe. Will update the comment.

Issue #1000, opened 2019-09-05
python3 throws TypeError: '<' not supported between instances of 'str' and 'int'


chunyong-lin:
## Background
After convert StreamAlert to use python3 after #974, we discover a bug that python3 will throws `TypeError: '<' not supported between instances of 'str' and 'int'`, while python2 will just return `False`. 🌝 

The full traceback is
```
[ERROR] TypeError: '<' not supported between instances of 'str' and 'int'
Traceback (most recent call last):
  File "/var/task/stream_alert/rule_promotion/main.py", line 21, in handler
    RulePromoter().run(event.get('send_digest', False))
  File "/var/task/stream_alert/rule_promotion/promoter.py", line 126, in run
    publisher.publish(list(self._staging_stats.values()))
  File "/var/task/stream_alert/rule_promotion/publisher.py", line 122, in publish
    self._publish_message(stats)
  File "/var/task/stream_alert/rule_promotion/publisher.py", line 104, in _publish_message
    Message=self._format_digest(stats),
  File "/var/task/stream_alert/rule_promotion/publisher.py", line 69, in _format_digest
    return '\n\n'.join(str(stat) for stat in sorted(stats, reverse=True))
  File "/var/task/stream_alert/rule_promotion/statistic.py", line 49, in __lt__
    return self.alert_count < other.alert_count
```

The root cause is the default value of `self.alert_count` is `unknown`, [code](https://github.com/airbnb/streamalert/blob/release-3-0-0/stream_alert/rule_promotion/statistic.py#L21). Python3 throws TypeError when `unknown` compares with an integer.

### Steps to Reproduce
Did a small verification
* Use python2
```
python
Python 2.7.16 (default, Apr 12 2019, 15:32:40)
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 'unknown' < 10
False
>>>
```

* Use python3
```
python
Python 3.7.4 (default, Jul  9 2019, 18:13:23)
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 'unknown' < 10
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'str' and 'int'
>>>
```
## Desired Change
Change the default value of `self.alert_count` to `-1`.

PR #1001, opened 2019-09-05
[rule_promotion] Change default value of alert_count to -1

chunyong-lin:
related to: #974 
resolves: #1000 

## Background
See [issue #1000 python3 throws TypeError: '<' not supported between instances of 'str' and 'int'](https://github.com/airbnb/streamalert/issues/1000) for the details and root cause.

## Changes

* Change default value of alert_count from a string to an integer `-1`.

## Testing
See the reproduce step in issue #1000

### DIFF
--- a/stream_alert/rule_promotion/statistic.py
+++ b/stream_alert/rule_promotion/statistic.py
@@ -18,7 +18,7 @@ limitations under the License.
 class StagingStatistic:
     """Store information on generated alerts."""
 
-    _ALERT_COUNT_UNKOWN = 'unknown'
+    _ALERT_COUNT_UNKOWN = -1
 
     _COUNT_QUERY_TEMPLATE = (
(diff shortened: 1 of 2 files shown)


ryandeivert:
the more appropriate fix would be changing the `__lt__` function to something like:

```python
    def __lt__(self, other):
        """Statistic should be ordered by their alert count."""
        if other.alert_count == self._ALERT_COUNT_UNKOWN:
            return False
        if self.alert_count == self._ALERT_COUNT_UNKOWN:
            return True
        return self.alert_count < other.alert_count
```

ryandeivert:
alternatively, if you want to use `-1` as the default value - you can update the `__str__` method:

https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/stream_alert/rule_promotion/statistic.py#L89-L128

chunyong-lin:
@ryandeivert PTAL. I took your 2nd suggestion and added unit test to cover the situation.

PR #980, opened 2019-08-23
tweaking .gitignore file slightly for venv and IDE files

ryandeivert:
to: @chunyong-lin 
cc: @airbnb/streamalert-maintainers

## Background

A `venv` directory should only be ignored if it's at the root of the repo.

## Changes

* Updating `.gitignore` to only ignore `venv` at root of repo.
* Removing IDE/environment-specific ignored items.
  * **These should be added to your [_global_ `.gitignore`](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore), not the local one**

PR #1002, opened 2019-09-05
adding proper vagrant ignore

ryandeivert:
to: @chunyong-lin 

### Changes

* ignoring `/.vagrant` instead of `/vagrant`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@ terraform/*.tf.json
 
 # Vagrant VM artifacts
 ubuntu*.log
-/vagrant
+/.vagrant
 
 # Coverage artifacts
 /htmlcov


.gitignore
```diff
@@ -23,7 +23,7 @@ terraform/*.tf.json
 
 # Vagrant VM artifacts
 ubuntu*.log
-/vagrant
+/.vagrant
```
chunyong-lin: This is in my .gitignore_global file and it works well. Maybe it doesn't matter to have leading `/`.
```
.vagrant/
```

ryandeivert: the leading `/` does matter.. otherwise we could have another situation like this:

https://github.com/airbnb/streamalert/pull/980/files

PR #1004, opened 2019-09-05
LookupTables documentation

Ryxias:
to: @ryandeivert or @chunyong-lin or @blakemotl 

In addition to RST documentation I fixed some broken tests

### DIFF
--- a/docs/source/lookup-tables.rst
+++ b/docs/source/lookup-tables.rst
@@ -1,0 +1,273 @@
+Lookup Tables
+=============
+
+Overview
+--------
+LookupTables is a framework for injecting additional data into your StreamAlert lambda functions. LookupTables
+offers a unified key-value interface into a set of backend storage solutions, allowing StreamAlert lambda functions
+to use state from outside of the raw telemetry that they receive.
+
+With LookupTables, you can hydrate alerting data, add statefulness to alerts, scalable pull down remote data, rapidly
+tune rule logic, and much more!
+
+
+How do LookupTables work?
+`````````````````````````
+LookupTables is a unified Python interface into backend data storage mechanisms. The two currently supported storage
+solutions are AWS S3 and AWS DynamoDb.
+
+LookupTables makes these storage solutions available to StreamAlert's lambda functions. It is available on all
+classifiers, the rules engine, the alert merger, and the alert processor.
+
+
+Usage
+-----
+Pulling keys from LookupTables is very easy!
+
+.. code-block:: python
+
+  from stream_alert.shared.lookup_tables.core import LookupTables
+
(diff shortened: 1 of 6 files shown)


stream_alert/shared/lookup_tables/driver_s3.py
```diff
@@ -195,13 +201,20 @@ def gz_decompress(driver, data):
                 'LookupTable (%s): Data is not compressed; defaulting to original payload',
                 driver.id
             )
-        return data
+        return data.decode()
```
Ryxias: FYI; this is a problem that exists on live right now due to `zlib` on Python3 not accepting strings. It only accepts Bytes-like.

Ryxias:
@ryandeivert or @chunyong-lin could you please take a look at the `.decode()` and `.encode()` usage in the `Compression` class? It's not clear to me whether this is reverse-compatible with the existing lookuptables, but I couldn't get it to work properly after the Python3 upgrade.

Ryxias:
Oh I'm stupid; it's because `s3_object.get('Body').read()` returns a bytes-like object..

docs/source/lookup-tables.rst
```diff
@@ -0,0 +1,273 @@
...
+  available. Passing ``false`` allows eventually consistent reads, which can greatly improve performance.
+* ``key_delimiter`` — (str) (Optional) When accessing keys in a dynamodb LookupTable that uses both a ``partition_key`` and a
+  ``sort_key``, the syntax of the final key is ``{partition_key}{delimiter}{sort_key}``. The default delimiter is a
+  colon (``:``), but this parameter can be provided to offer a different delimiter.
+* ``cache_refresh_minutes`` — (int) Number of minutes to cache each individual key.
+* ``cache_maximum_key_count`` — (int) Maximum number of keys to cache on this LookupTable. Once the cache is full, keys
+  will be evicted on a random-selection basis.
+
+
+The nitty gritty
```
ryandeivert: title case again for header

docs/source/lookup-tables.rst
```diff
@@ -0,0 +1,273 @@
...
+
+1) Table Name — The name of the LookupTable. This is specified in the config (below).
+2) Key — A key on the given LookupTable.
+3) Default Value — If the key is not found, this value will be returned instead. Notably, *if the key is empty*
+   the empty value will be returned, **NOT** this default value.
+
+
+Configuration
+-------------
+LookupTables is configured via a single file, ``conf/lookup_tables.json``.
```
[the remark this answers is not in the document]
Ryxias: LookupTables is the name of the system and thus is it not plural

PR #1010, opened 2019-09-13
Accepts JSON from LookupTables CLI, adds `list-add` command

Ryxias:

## Background

I designed LookupTables' drivers to be able to store any complex JSON-serializable type...

Sadly the CLI command as-is could not do anything other than string. Whooops!

## Changes

* Now there's a new option: `--json`. It accepts the input `--value` as a JSON encoded string.
* I added a new subcommand, `list-add`. 
* I restructured some of the subcommand code

### DIFF
--- a/stream_alert/__init__.py
+++ b/stream_alert/__init__.py
@@ -1,2 +1,2 @@
 """StreamAlert version."""
-__version__ = '2.3.0'
+__version__ = '3.0.0'
--- a/stream_alert_cli/lookup_tables/handler.py
+++ b/stream_alert_cli/lookup_tables/handler.py
@@ -29,11 +29,7 @@ class LookupTablesCommand(CLICommand):
     @classmethod
     def setup_subparser(cls, subparser):
         # FIXME (derek.wang) Refactor this into a more robust command-nesting framework
-        subcommands = cls._subcommands()
-        set_parser_epilog(
-            subparser,
-            epilog=(
-                '''\
+        template = '''\
 Available Sub-Commands:
 
 {}
@@ -41,7 +37,12 @@ Available Sub-Commands:
 Examples:
 
     manage.py lookup-tables [describe-tables|get|set]
-'''.format('\n'.join([
+'''
+        subcommands = cls._subcommands()
+        set_parser_epilog(
+            subparser,
+            epilog=(
+                template.format('\n'.join([
                     '\t{command: <{pad}}{description}'.format(
                         command=command,
                         pad=30,
--- a/stream_alert_cli/lookup_tables/handler.py
+++ b/stream_alert_cli/lookup_tables/handler.py
@@ -13,6 +13,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import json
+
 from stream_alert.shared.logger import get_logger
 from stream_alert.shared.lookup_tables.core import LookupTables
 from stream_alert_cli.utils import CLICommand, generate_subparser, set_parser_epilog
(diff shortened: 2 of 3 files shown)


stream_alert/__init__.py
```diff
@@ -1,2 +1,2 @@
 """StreamAlert version."""
-__version__ = '2.3.0'
+__version__ = '3.0.0'
```
chunyong-lin: I believe we don't upgrade version until we cut the release.

stream_alert_cli/lookup_tables/handler.py
```diff
@@ -21,33 +24,149 @@
...
             epilog=(
                 '''\
+Available Sub-Commands:
+
+{}
+
+Examples:
+
+    manage.py lookup-tables [describe-tables|get|set]
+                '''.format('\n'.join([
```
ryandeivert: do you want these trailing spaces `                ` here ..?

stream_alert_cli/lookup_tables/handler.py
```diff
@@ -21,33 +24,149 @@
...
+                    )
+                    for command, subcommand
+                    in subcommands.items()
+                ]))
+            )
+        )
+
+        lookup_tables_subparsers = subparser.add_subparsers()
+
+        for _, subcommand in subcommands.items():
```
ryandeivert: just use `.values` if you don't need the key

stream_alert_cli/lookup_tables/handler.py
```diff
@@ -21,33 +24,149 @@
...
+            required=True,
+            default=None
+        )
+
+        set_parser.add_argument(
+            '-k',
+            '--key',
+            help='Key to modify on the LookupTable',
+            required=True,
+            default=None
```
ryandeivert: reasoning for `default=None` vs the default for this type?

stream_alert_cli/lookup_tables/handler.py
```diff
@@ -64,10 +183,27 @@ def _setup_describe_tables_subparser(cls, subparsers):
...
+    # pylint: disable=protected-access
     @classmethod
-    def _setup_get_subparser(cls, subparsers):
+    def handler(cls, options, config):
+        print('==== LookupTables; Describe Tables ====\n')
+
+        lookup_tables = LookupTables.get_instance(config=config)
+
+        print('{} Tables:\n'.format(len(lookup_tables._tables)))
+        for _, table in lookup_tables._tables.items():
```
ryandeivert: suggestion: implement a `__str__` or `__repr__` function on the lookup tables class(es) that can just handle this for you. that way it could also be "neatly" printed elsewhere

stream_alert/__init__.py
ryandeivert: ... this is fine. this isn't being cut yet and is a WIP branch

stream_alert_cli/lookup_tables/handler.py
Ryxias: _Bad_ copypasta

stream_alert_cli/lookup_tables/handler.py
Ryxias: I think probably somewhere in my brain I mixed up a string and a heredoc and  ... yea

stream_alert/__init__.py
Ryxias: Yea I figured we could just bump it cuz it's going on `release-3-0-0` and we don't **_really_** use the version for anything important, if I'm not mistaken

stream_alert_cli/lookup_tables/handler.py
Ryxias: I'll add a FIXME and cut a new branch for this improvement. I've been thinking of how to refactor the command classes into something that nests more easily

PR #1017, opened 2019-09-18
[LookupTables] Adds new JSON file import CLI command

Ryxias:
to: @ryandeivert @chunyong-lin @blakemotl @fusionrace 

## Background

Per request from @fusionrace, LookupTables can now import data from a JSON file.

## Changes

* Adds a new lookup tables CLI command: `lookup-tables set-from-json-file`
* Adds two options to `list-add`: `--sort` and `--unique`
* Refactors some magic code into its own class with disclaimer

## How to use

There is now a subcommand under the `lookup-tables` command, called `set-from-json-file`

```
$ ./manage.py lookup-tables --help
usage: ./manage.py lookup-tables [-h]
                                 {describe-tables,get,set,list-add,set-from-json-file}
                                 ...

Describe and manage your LookupTables

positional arguments:
  {describe-tables,get,set,list-add,set-from-json-file}

optional arguments:
  -h, --help            show this help message and exit

Available Sub-Commands:

	describe-tables               Show information about all currently configured LookupTables
	get                           Retrieve a key from an existing LookupTable
	set                           Update the key of an existing LookupTable
	list-add                      Add a value to a list-typed LookupTables key
	set-from-json-file            NotImplemented

Examples:

    manage.py lookup-tables [describe-tables|get|set]
```

`set-from-json-file` takes 3 arguments: `--table`, `--key`, and `--file`.

```
usage: ./manage.py lookup-tables set-from-json-file [-h] -t TABLE -k KEY -f
                                                    FILE

Pushes the contents of a given json file into the LookupTable key

optional arguments:
  -h, --help            show this help message and exit
  -t TABLE, --table TABLE
                        Name of the LookupTable
  -k KEY, --key KEY     Key to modify on the LookupTable
  -f FILE, --file FILE  Path to the json file, relative to the current working
                        directory

Examples:

    manage.py lookup-tables set-from-json-file -t [table] -k [key] -f [path/to/file.json]
```

It takes the entire JSON blob in the given `.json` file and puts it into the LookupTable key as a dict/list.

### DIFF
--- a/streamalert/shared/lookup_tables/utils.py
+++ b/streamalert/shared/lookup_tables/utils.py
@@ -1,0 +1,70 @@
+"""
+Copyright 2017-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+from streamalert.shared.lookup_tables.drivers import PersistenceDriver
+
+
+# pylint: disable=protected-access
+class LookupTablesMagic:
+    """Namespace full of magic methods that dig around the public interface of LookupTables.
+
+    These methods are not on the public interface by design to prevent these access patterns from
+    being utilized in "normal" Lambda code.
+    """
+
+    @staticmethod
+    def get_all_table_data(table):
+        """
+        Return all of the data in the given lookup table as a dict. Only works with S3, and you
+        should DEFINITELY AVOID USING THIS.
+
+        Params:
+            - table (LookupTable)
+
+        Returns:
+            dict
+        """
+        if table.driver_type != PersistenceDriver.TYPE_S3:
+            raise RuntimeError("Cannot use lookup_table helper on non-S3 table.")
+
+        # Make a single dummy call to force the table to initialize
+        table.get('dummy', None)
+
+        # Do some black magic tomfoolery
+        return table._driver._cache._data
+
+    @staticmethod
+    def set_table_value(table, key, new_value):
+        """Set a value into a LookupTable and then immediately commit it.
+
+        Params:
+            - table (LookupTable)
+            - key (str)
+            - new_value (str|int|list|dict|mixed)
+        """
+        table._driver.set(key, new_value)
+        table._driver.commit()
+
+    @staticmethod
+    def get_all_tables(lookup_tables_core):
+        """Returns all lookup tables, keyed by their names
+
+        Params:
+        - lookup_tables_core (LookupTablesCore)
+
+        Returns:
+             dict[str, LookupTable]
+        """
+        return lookup_tables_core._tables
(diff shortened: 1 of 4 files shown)


streamalert/shared/lookup_tables/utils.py
```diff
@@ -0,0 +1,70 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+from streamalert.shared.lookup_tables.drivers import PersistenceDriver
+
+
+# pylint: disable=protected-access
```
ryandeivert: `# pylint: disable=...` comments apply to the entire _scope_ where they're added. please only apply this where you need to (preferably on the exact line only)

streamalert_cli/lookup_tables/handler.py
```diff
@@ -80,9 +81,85 @@ def _subcommands(cls):
...
+            required=True
+        )
+
+    @classmethod
+    def handler(cls, options, config):
+        print('==== LookupTables; Set from JSON File ====')
+
+        table_name = options.table
+        key = options.key
+        file = options.file
```
ryandeivert: don't use `file` as a var name if you can avoid it. also, I don't think you need to re-ref these from the `options` named tuple..

streamalert_cli/lookup_tables/handler.py
```diff
@@ -80,9 +81,85 @@ def _subcommands(cls):
...
+
+        table = core.table(table_name)
+
+        old_value = table.get(key)
+
+        with open(file, "r") as json_file_fp:
+            new_value = json.load(json_file_fp)
+
+        print('  Value: {} --> {}'.format(
+            json.dumps(old_value, indent=2),
```
ryandeivert: `sort_keys=True` might be nice here

streamalert_cli/lookup_tables/handler.py
Ryxias: Sorry; force of habit. I was de-refing them to be explicit about what keys are expected from `options`, but I realize that the `setup_subparser` does exactly that but better; I will remove

streamalert_cli/lookup_tables/handler.py
```diff
@@ -152,12 +242,16 @@ def handler(cls, options, config):
 
         new_value = copy.copy(old_value)
         new_value.append(options.value)
-        sorted(new_value)
+
+        if options.unique:
+            new_value = list(dict.fromkeys(new_value))
```
ryandeivert: I've never used `dict.fromkeys(...)`.. why not just use `set()`?

streamalert/shared/lookup_tables/utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # Do some black magic tomfoolery
+        return table._driver._cache._data
+
+    @staticmethod
+    def set_table_value(table, key, new_value):
+        """Set a value into a LookupTable and then immediately commit it.
+
+        Params:
+            - table (LookupTable)
```
ryandeivert: inconsistent `Params` (vs `Args`) format here (and other places)

Ryxias: Is it pythonic to use `params` or `args`? My IDE says I'm supposed to use `:params:` but I'm not sure which way to go

Ryxias: I think I will go with `Args`... since I see it in more places in the codebase (100+ usages instead of 61).

ryandeivert:
@Ryxias we try to abide by the Google Python Style guide. example of the usage here: http://google.github.io/styleguide/pyguide.html#doc-function-raises

PR #1015, opened 2019-09-18
updates to terraform for better namespacing, consistency, tagging

ryandeivert:
to: @chunyong-lin / @Ryxias 
cc: @airbnb/streamalert-maintainers
resolves: #886 
resolves: #1011 

## Background

Continuing on #1013 and #1014

## Changes

- Fixing role name in tf_kinesis_firehose_setup module, updating var name and tests
- Fixing role name in tf_cloudwatch module, updating var name and tests
- Fixing role name in tf_kinesis_streams module, updating var name and tests
- Fixing role name in tf_flow_logs module, updating tests
- Fixing role name in tf_cloudtrail module
- Fixing tf_* module references in `README.md`s
- Role policy name updates for consistency, etc, etc
- Updating to `tf_threat_intel_downloader` module, removing unused variables
- Adding tags to resources that support them
- Adding fix for #886
- General killing of all `stream_alert` usage in terraform
- The `aws_iam_roles` that are created will now all be within a `/streamalert/` path by default, shortening the whole name.

### Notes

- While there will be effectively no actual _changes_ to infrastructure because of the above changes, and **lot** of resources will likely be destroyed/recreated because of the naming changes. This will probably result in a lot of `forces new resource` warnings in terraform.
- Some role names are updated with these changes. If there are any out-of-band places where these role names are hard-coded, they will also need to be updated./

## Testing

- Updates to unit tests for above changes where appropriate.
- Adding unit test for #886 change

Issue #1011, opened 2019-09-13
[bug] roles missing prefix


ryandeivert:
### Description

https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/terraform/modules/tf_stream_alert_flow_logs/iam.tf#L5

https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/terraform/modules/tf_stream_alert_cloudwatch/iam.tf#L4

https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/terraform/modules/tf_stream_alert_kinesis_firehose_setup/iam.tf#L3

https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/terraform/modules/tf_stream_alert_kinesis_firehose_setup/iam.tf#L23

https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/terraform/modules/tf_stream_alert_kinesis_firehose_setup/iam.tf#L65

Also, these should be made consistent - `stream_alert` vs `streamalert`

Make sure no others are misnamed.

ryandeivert:
similar to #886

ryandeivert:
Fixed with #1015 & #1022

PR #1022, opened 2019-09-20
adding prefix to firehoses and other small updates

ryandeivert:
to: @Ryxias / @blakemotl 

## Background

Building more on #1011, #1013, #1014, #1015, #1019

## Changes

- removing athena default db name to enforce that one is provided
- adding default value for metric to tf_metric_filters module
- prefixing firehoses created for dat and adding support in dependent classes
- adding missing test docstring
- making prefix for unit testing consistent

## Testing

Updates to all unit tests (mostly regarding the consistent prefix usage)


ryandeivert:
it will force a new resource - I will have go be delicate about about this is applied, yes. I'm not going to merge yet because I'm unsure of how this should happen

ryandeivert:
actually I think I need these change to do the migration like I would like. merging now

PR #1024, opened 2019-09-20
fixing missing tf var and adding support for optional firehose prefixing

ryandeivert:
to: @Ryxias 

## Changes

* fixing bug introduced in #1022
* adding optional prefixing for firehose

## Testing

Updates to unit tests.


streamalert/classifier/clients/firehose.py
```diff
@@ -61,7 +61,10 @@ class FirehoseClient:
     _ENABLED_LOGS = dict()
 
     def __init__(self, prefix, firehose_config=None, log_sources=None):
-        self._prefix = prefix
+        self._prefix = (
+            '{}_'.format(prefix)
+            if firehose_config and firehose_config.get('use_prefix') else ''
```
Ryxias: Opinion:

```
'{}_'.format(prefix)
if firehose_config and firehose_config.get('use_prefix')
else ''
```

Putting else on newline makes it more clearly distinguished from the `if` statement

ryandeivert: 👍

PR #1027, opened 2019-10-01
making change for proper formatting of kinesis stream names

ryandeivert:
to: @Ryxias / @chunyong-lin 

## Background

Kinesis stream names have a nasty underscore in `stream_alert`.

## Changes

* Updating the default kinesis stream name to use `streamalert` instead of `stream_alert`.
* Removing the unnecessary (redundant) `_kinesis` suffix from the default stream name.
* Adding support for supplying a custom stream name, effectively overridden by the above.

## Testing

* Updating all unit tests for above changes.


ryandeivert:
@chunyong-lin / @Ryxias this is ready for review - can you PTAL?

streamalert_cli/terraform/monitoring.py
```diff
@@ -72,7 +72,7 @@ def generate_monitoring(cluster_name, cluster_dict, config):
 
     if monitoring_config.get('kinesis_alarms_enabled', True):
         cluster_dict['module']['cloudwatch_monitoring_{}'.format(cluster_name)].update({
-            'kinesis_stream': '{}_{}_stream_alert_kinesis'.format(prefix, cluster_name),
+            'kinesis_stream': '${{module.kinesis_{}.stream_name}}'.format(cluster_name),
```
chunyong-lin: Should we keep `prefix` here?

ryandeivert: @chunyong-lin this does effectively still use the prefix - except it uses the output from here:

https://github.com/airbnb/streamalert/pull/1027/files/9f738ef03c371190e9540fc0cdecd04c9df10fd5#diff-517d65d9de13bb32b0329474e9b11627R6

which gets its stream name from here:

https://github.com/airbnb/streamalert/pull/1027/files/9f738ef03c371190e9540fc0cdecd04c9df10fd5#diff-916958fe1956202328ac3949bb5ea39cR33

ryandeivert: this change avoids defining the "stream name" multiple places.. potentially missing one if the other changes

PR #1023, opened 2019-09-20
[rules] Detect use of PutAccountPublicAccessBlock

0xdabbad00:
Resolves #1016

Checks for `PutAccountPublicAccessBlock` the same way as `PutBucketPublicAccessBlock`.

PR #1029, opened 2019-10-02
Critical API call detection fixes

0xdabbad00:
If you make changes to S3 public block access, then all 4 fields associated with that call will be filled in.  If you make the call via the API though, then you might only have one or two fields changed and present.  This fixes the logic with that.

This also adds detection for the Organization calls as these can have far reaching impacts, such as an SCP that disables API calls, a change to an SCP that weakens it, or an account leaving an Organization.


0xdabbad00:
PR #1023 is related, although I did't base this one on that.

ryandeivert:
hey @0xdabbad00 can you rebase from `release-3-0-0` so we can get this updated/merged

0xdabbad00:
@ryandeivert I won't have time to do this for a while unfortunately.

ryandeivert:
@0xdabbad00 okay no worries - I'm fixing it up for you. will merge soon :)

PR #1030, opened 2019-10-03
AlertMerger query optimizations

Ryxias:
to: @ryandeivert or @chunyong-lin or @blakemotl 

## Background

We found a couple bugs in the alert merger that caused performance issues.

## Changes

* Fixes a bug in `_get_alerts()` (which I renamed to `_alert_generator()` where a generator was being immediately materialized, greatly increasing up-front memory consumption
* Sets a limit of 5000 alerts to be processed by the Alert Merger, for a single invocation on any single rule_name. This prevents OOM errors when there are too many alerts on the table. This case also includes a warning.
* Changes the up-front consistent scan operation to a generator to alleviate the need for scanning upfront. Removes high consistency requirement.

## Testing

Guess I'll find out on production? _**Fingers crossed**_

### DIFF
--- a/streamalert/alert_merger/main.py
+++ b/streamalert/alert_merger/main.py
@@ -163,8 +163,7 @@ class AlertMerger:
         merged_alerts = []  # List of newly created merge alerts
         alerts_to_delete = []  # List of alerts which can be deleted
 
-        # TODO: Find a way to avoid a full table scan just to get rule names
-        for rule_name in self.table.rule_names():
+        for rule_name in self.table.rule_names_generator():
             merge_enabled_alerts = []
             for alert in self._alert_generator(rule_name):
                 if alert.remaining_outputs:
--- a/streamalert/alert_merger/main.py
+++ b/streamalert/alert_merger/main.py
@@ -87,16 +87,15 @@ class AlertMerger:
         To limit memory consumption, the generator yields a maximum number of alerts, defined
         by self._alert_generator_limit.
         """
-        alert_count = 0
-        for record in self.table.get_alert_records(rule_name, self.alert_proc_timeout):
+        generator = self.table.get_alert_records(rule_name, self.alert_proc_timeout)
+        for idx, record in enumerate(generator, start=1):
             try:
                 yield Alert.create_from_dynamo_record(record)
-                alert_count += 1
             except AlertCreationError:
                 LOGGER.exception('Invalid alert record %s', record)
                 continue
 
-            if alert_count >= self._alert_generator_limit:
+            if idx >= self._alert_generator_limit:
                 LOGGER.warning(
                     'Alert Merger reached alert limit of %d for rule "%s"',
                     self._alert_generator_limit,
--- a/streamalert/alert_merger/main.py
+++ b/streamalert/alert_merger/main.py
@@ -60,6 +60,10 @@ class AlertMerger:
     # Set the max payload size to slightly under that to account for the rest of the message.
     MAX_LAMBDA_PAYLOAD_SIZE = 126000
 
+    # The maximum number of alerts that are loaded into memory for a single rule, during a single
+    # loop in the Alert Merger.
+    ALERT_GENERATOR_DEFAULT_LIMIT = 5000
+
     @classmethod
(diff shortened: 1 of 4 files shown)


streamalert/alert_merger/main.py
```diff
@@ -73,17 +77,32 @@ def __init__(self):
...
+
+    def _alert_generator(self, rule_name):
+        """
+        Returns a generator that yields Alert instances triggered from the given rule name.
 
+        To limit memory consumption, the generator yields a maximum number of alerts, defined
+        by self._alert_generator_limit.
+        """
+        alert_count = 0
         for record in self.table.get_alert_records(rule_name, self.alert_proc_timeout):
```
ryandeivert: omit the `alert_count` and use something like the following:
```python
for idx, record in enumerate(self.table.get_alert_records(rule_name, self.alert_proc_timeout), start=1):
    ...
    if idx >= self. _alert_generator_limit:
        ...
```

streamalert/shared/alert_table.py
```diff
@@ -68,6 +72,36 @@ def rule_names(self):
...
+    def rule_names_generator(self):
+        """Returns a generator that yields unique rule_names of items on the table
+
+        Each unique name is yielded only once. Additionally, because the names materialized over
+        several paginated calls, it is not 100% guaranteed to return every possible rule_name on
+        the Alert Table; if there are other operations that are writing items to the DynamoDB table,
+        it is possible for certain names to get skipped.
+
+        Returns:
+            generator
```
ryandeivert: please include the type that the generator will yield

streamalert/shared/alert_table.py
```diff
@@ -59,6 +59,10 @@ def _paginate(func, func_kwargs):
     def rule_names(self):
         """Find all of the distinct rule names in the table.
 
+        @deprecated
```
ryandeivert: is this still used internally? I'd say, if not, just 86 it

Ryxias: It's used in a variety of places; as far as I can tell it's used in a lot of tests. My worry was it was used in a bunch of public methods, which other customers might be using... that said, I think it's a good idea cuz using this method is just not a good idea. I'll hunt down the references in a separate PR

Ryxias: Separate note: "just 86 it" 🤣 

I watch too much food TV

streamalert/alert_merger/main.py
Ryxias: (I'll test this on my own but regardless—) 
Does `enumerate` preserve the generator? It won't materialize the generator values up-front like `list`, right?

ryandeivert: @Ryxias yep it will preserve the generator underneath

ryandeivert: @Ryxias 
```python
def gen():
    for i in range(10):
        print('yielding', i)
        yield i


def main():
    iter = enumerate(gen(), start=1)
    print('created enumerator', iter)
    for i, v in iter:
        print('index: {}; value: {}'.format(i, v))


if __name__ == '__main__':
    main()
```

Output:
```
created enumerator <enumerate object at 0x10ddd1e60>
yielding 0
index: 1; value: 0
yielding 1
index: 2; value: 1
yielding 2
index: 3; value: 2
yielding 3
index: 4; value: 3
yielding 4
index: 5; value: 4
yielding 5
index: 6; value: 5
yielding 6
index: 7; value: 6
yielding 7
index: 8; value: 7
yielding 8
index: 9; value: 8
yielding 9
index: 10; value: 9
```

streamalert/shared/alert_table.py
ryandeivert: YOU or I watch too much!? I worked in a restaurant 😛

Ryxias: I'm just surprised I understood the kitchen jargon
Too much Gordon Ramsey on Youtube
Yelling at people
ITS R AW
☠️

ryandeivert: LOL

PR #1034, opened 2019-10-07
updating the flow logs module to not create unnecessary resources

ryandeivert:
to: @Ryxias or @chunyong-lin 

## Background

During some refactoring, I discovered that the `tf_flow_logs` module is implemented in a way that creates unnecessary resources.

## Changes

* Updating the module to use two submodules.
* One module creates all resources necessary for VPC Flog Log ingestion (with support for both the "internal" account and for cross-account ingestion)
* The second module will actually create the necessary resources for doing flow logs in the deployment account (for vpcs, enis, and subnets) along with the CloudWatch Logs Group. Previously, a CloudWatch Logs Group and other resources were made without actually needing to be.
* Fixing `Stream Alert` ridiculousness in places.

* Caveat - the 'multi-region' support for this module was implemented poorly before, and has been completely removed for now. We need to properly implement aliased providers for regions instead of the way it was previously done.

## Testing

Updating unit tests. The change has also been deployed and tested.


streamalert_cli/terraform/flow_logs.py
```diff
@@ -30,35 +32,56 @@ def generate_flow_logs(cluster_name, cluster_dict, config):
...
-            'source': 'modules/tf_flow_logs',
+    # If 'vpcs', 'subnets', or 'enis' is defined within the config, we should create
+    # flow logs for these values
+    create_internal_flow_logs = any(
+        modules['flow_logs'].get(flow_log_type)
+        for flow_log_type in DEFAULT_FLOW_LOG_TYPES
+    )
+    if create_internal_flow_logs:
+        flow_logs_settings = {
+            'source': 'modules/tf_flow_logs/modules/internal',
```
Ryxias: omg...........


🤦‍♂

Ryxias: If only I knew about this

terraform/modules/tf_flow_logs/README.md
```diff
@@ -1,27 +1,99 @@
...
+[via a CloudWatch Logs Subscription Filter](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CrossAccountSubscriptions.html).
 
-## Components
-Creates the following:
-* A CloudWatch log group to store the flow logs.
-* A subscription filter delivering the logs to the specified destination Kinesis Stream.
-* Enables flow logs for VPC resources specified in `var.targets`.
-* IAM Roles and Policies to allow flow logs to be delivered.
+## Submodules
+This folder contains the following two submodules: `defualt` and `internal`
```
Ryxias: `defualt` —> `default`

ryandeivert: damn good catch

PR #1039, opened 2019-10-14
fix setting of rules engine environment variable

ryandeivert:
to: @chunyong-lin 

## Background

The `STREAMALERT_TRACK_RULE_STATS` is currently getting set even if the rules engine config contains something like:

`"log_rule_statistics": false`

which is bad.

## Changes

* Updating the environment variable inclusion to look for a truthy setting.

## Testing

N/A

### DIFF
--- a/streamalert_cli/terraform/rules_engine.py
+++ b/streamalert_cli/terraform/rules_engine.py
@@ -52,7 +52,7 @@ def generate_rules_engine(config):
         'STREAMALERT_PREFIX': prefix,
     }
 
-    if 'log_rule_statistics' in config['lambda']['rules_engine_config']:
+    if config['lambda']['rules_engine_config'].get('log_rule_statistics'):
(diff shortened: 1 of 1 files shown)


streamalert_cli/terraform/rules_engine.py
```diff
@@ -52,7 +52,7 @@ def generate_rules_engine(config):
         'STREAMALERT_PREFIX': prefix,
     }
 
-    if 'log_rule_statistics' in config['lambda']['rules_engine_config']:
+    if config['lambda']['rules_engine_config'].get('log_rule_statistics'):
```
chunyong-lin: Probably we can implement validation against `log_rule_statistics` parameter in the `rules_engine_config` and only allow boolean values. But for now, LGTM.

ryandeivert: we don't do any validation of these lambda settings now. I'm not about to write a validator just for this :P

PR #1041, opened 2019-10-17
[terraform] update cloudwatch and flow logs terraform module to reduce redundancy

ryandeivert:
to: @chunyong-lin  / @Ryxias 

## Background

The `tf_cloudwatch` and `tf_flow_logs` modules are somewhat redundant, and implemented poorly. They both embed their own `provider` block, which terraform [recommends against](https://www.terraform.io/docs/configuration/modules.html#providers-within-modules). The best practice is to have providers in the root module that can be referenced in the submodules via their `alias`. Otherwise, this can create state issues when the module changes/is renamed/etc.

## Changes

- adding aliased terraform providers for regions
- updating `tf_flow_logs` module to remove redundancy
- adding a `tf_cloudwatch_logs_destination` module to replace `tf_cloudwatch` module
- removing replaced `tf_cloudwatch` module
- updating infinitedict function to support initial value
- updating terraform generation code for aforementioned changes
- removing legacy test because we ain't living in the past
- updates to documentation

## Notes on Config Changes

- The config for the `cloudwatch` module is now named `cloudwatch_destinations` to be more explicit/clear.
- The config settings within `cloudwatch_destinations` now uses _inclusion_ logic for regions vs exclusion logic (`regions` vs the old `exluded_regions`).
- The `tf_flow_logs` module should no longer be supplied any `cross_account_ids`. This variable should be supplied directly to the `cloudwatch_destinations`, since that is all that is needed to support cross account flow logs.

## Testing

- fixing tests for default terraform settings
- other updates for better, more encompassing unit tests


streamalert_cli/terraform/cloudwatch.py
```diff
@@ -18,7 +18,32 @@
 LOGGER = get_logger(__name__)
 
 
-def generate_cloudwatch(cluster_name, cluster_dict, config):
+def generate_cloudwatch_destinations(cluster_name, cluster_dict, config):
```
chunyong-lin: Can you add a comment to describe what is this function doing?

streamalert_cli/terraform/cloudwatch.py
```diff
@@ -18,7 +18,32 @@
...
+            'CloudWatch destinations must be enabled for at '
+            'least one region in the \'%s\' cluster',
+            cluster_name
+        )
+        return False
+
+    return _generate(cluster_name, cluster_dict, config, account_ids, regions)
+
+
+def generate_cloudwatch_destinations_internal(cluster_name, cluster_dict, config):
```
chunyong-lin: A short comment here will be very helpful what is the different between this function and `generate_cloudwatch_destinations` function.

docs/source/clusters.rst
```diff
@@ -216,16 +213,14 @@ Example: CloudWatch Logs Cluster
   {
     "id": "cloudwatch-logs-example",
     "modules": {
-      "cloudwatch": {
+      "cloudwatch_destinations": {
```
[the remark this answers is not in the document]
ryandeivert: good comment! I can make that change

ryandeivert:
@chunyong-lin I added a commit addressing PR feedback: https://github.com/airbnb/streamalert/pull/1041/commits/7e6b113d866448b9d747743306c01f2d8b0a1303

Issue #1044, opened 2019-10-23
[bug][test] AttributeError: 'Namespace' object has no attribute 'stats' when run classifier test


chunyong-lin:
## Background
PR #1019 introduced `stats` argument from cli command line, but is not always exist. It exists in rule test parser, but not classifier parser. When I run `python manage.py test classifier` it throws error `AttributeError: 'Namespace' object has no attribute 'stats'`. 
### Description
I did print out the value of options when it passed to test handler, it doesn't has `stats` attribute. 
```
Namespace(command='test', debug=False, files=set(), files_dir='tests/integration/rules/', quiet=False, rules=set(), subcommand='classifier', verbose=False, **{'test subcommand': 'classifier'})
```

The full traceback for future reference
```
python manage.py test classifier                 

Traceback (most recent call last):
  File "manage.py", line 116, in <module>
    main()
  File "manage.py", line 112, in main
    sys.exit(not cli_runner(options))
  File "/Users/xxxxx/xxx/streamalert/streamalert_cli/runner.py", line 70, in cli_runner
    result = cmds[args.command](args)
  File "/Users/xxxxx/xxx/streamalert/streamalert_cli/runner.py", line 126, in <lambda>
    command: lambda opts, cmd=cli_command: cmd.handler(opts, config)
  File "/Users/xxxxx/xxx/streamalert/streamalert_cli/test/handler.py", line 198, in handler
    result = result and TestRunner(options, config).run()
  File "/Users/xxxxx/xxx/streamalert/streamalert_cli/test/handler.py", line 237, in __init__
    if options.stats:
AttributeError: 'Namespace' object has no attribute 'stats'
```
### Steps to Reproduce
Run
```
python manage.py test classifier
```
## Desired Change
* Handle `AttributeError` when refers to `stats` attribute in classifier testing.
* Add classifier test to CI `.travis.yml` to catch such failure in the future.

chunyong-lin:
This issue is fixed in PR #1045 😄

PR #1045, opened 2019-10-23
[cli] Handle when AttributeError when test classifier

chunyong-lin:
to: @Ryxias or @ryandeivert 
related to: #1019 
resolves: #1044

## Background
PR #1019 introduced `stats` argument from cli command line, but is not always exist. It exists in rule test parser, but not classifier parser. When I run `python manage.py test classifier` it throws error `AttributeError: 'Namespace' object has no attribute 'stats'`. 
### Description
I did print out the value of options when it passed to test handler, it doesn't has `stats` attribute. 
```
Namespace(command='test', debug=False, files=set(), files_dir='tests/integration/rules/', quiet=False, rules=set(), subcommand='classifier', verbose=False, **{'test subcommand': 'classifier'})
```

The full traceback for future reference
```
python manage.py test classifier                 

  File "manage.py", line 116, in <module>
    main()
  File "manage.py", line 112, in main
    sys.exit(not cli_runner(options))
  File "/Users/xxxxx/xxx/streamalert/streamalert_cli/runner.py", line 70, in cli_runner
    result = cmds[args.command](args)
  File "/Users/xxxxx/xxx/streamalert/streamalert_cli/runner.py", line 126, in <lambda>
    command: lambda opts, cmd=cli_command: cmd.handler(opts, config)
  File "/Users/xxxxx/xxx/streamalert/streamalert_cli/test/handler.py", line 198, in handler
    result = result and TestRunner(options, config).run()
  File "/Users/xxxxx/xxx/streamalert/streamalert_cli/test/handler.py", line 237, in __init__
    if options.stats:
AttributeError: 'Namespace' object has no attribute 'stats'
```

## Changes

* Catch `AttributeError` and doing nothing if `stats` attribute doesn't exist.
* Add `classifier test` to CI to catch similar failure in the future.

## Testing
Manually ran
```
./manage.py test classifier
```

### DIFF
--- a/streamalert_cli/test/handler.py
+++ b/streamalert_cli/test/handler.py
@@ -234,13 +234,9 @@ class TestRunner:
             'AWS_ACCOUNT_ID': self._config['global']['account']['aws_account_id'],
             'ALERTS_TABLE': '{}_streamalert_alerts'.format(prefix),
         }
-        try:
-            if options.stats:
-                env['STREAMALERT_TRACK_RULE_STATS'] = '1'
-        except AttributeError:
-            # "stats" attribute only exists when doing integration test on rules.
-            # We do nothing for other test options, e.g. classifier
-            pass
+
+        if 'stats' in options and options.stats:
+            env['STREAMALERT_TRACK_RULE_STATS'] = '1'
 
         patch.dict(
             os.environ,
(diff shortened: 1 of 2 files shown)


streamalert_cli/test/handler.py
```diff
@@ -234,8 +234,13 @@ def __init__(self, options, config):
             'AWS_ACCOUNT_ID': self._config['global']['account']['aws_account_id'],
             'ALERTS_TABLE': '{}_streamalert_alerts'.format(prefix),
         }
-        if options.stats:
-            env['STREAMALERT_TRACK_RULE_STATS'] = '1'
+        try:
+            if options.stats:
```
ryandeivert: please remove the `try` block and just make this:

```python
if 'stats' in options and options.stats:
    ...
```

Issue #1036, opened 2019-10-10
Gsuite Ggroups "applicationName" value "groups" is not allowed


Ryxias:
## Background

StreamAlert app for gsuite errors.

### Stack Trace

```
[ERROR] TypeError: Parameter "applicationName" value "groups" is not an allowed value 
in "['access_transparency', 'admin', 'calendar', 'drive', 'gcp', 'gplus', 'group', 'groups_enterprise', 
'login', 'mobile', 'rules', 'token', 'user_accounts']" 
Traceback (most recent call last): 
File "/var/task/streamalert/apps/main.py", line 30, in handler StreamAlertApp.get_app(event['app_type'])(event, context).gather() 
File "/var/task/streamalert/apps/app_base.py", line 395, in gather while (((self._gather() * self._POLL_BUFFER_MULTIPLIER) + self._sleep_seconds()) < 
File "/var/task/streamalert/apps/app_base.py", line 42, in _wrapper func(*args, **kwargs) 
File "/var/task/streamalert/apps/app_base.py", line 364, in _gather logs = self._gather_logs() 
File "/var/task/streamalert/apps/_apps/gsuite.py", line 143, in _gather_logs pageToken=self._next_page_token 
File "/var/task/googleapiclient/discovery.py", line 764, in method (name, value, str(enums)))
```

ryandeivert:
resolved with #1046

Issue #888, opened 2019-01-22
apps: update gsuite app to use new version of google-api-python-client


ryandeivert:
### Description

Google updated the [`google-api-python-client`](https://github.com/googleapis/google-api-python-client) to change/remove some legacy stuff related to oath. This module was "supposed" to be in maintenance mode, but they made significant changes in the past year.

## Desired Change

Update the client to the newest version to get rid of some old oauth dependency cruft.

ryandeivert:
resolved with #1046

PR #1046, opened 2019-10-31
update gsuite apps for changes to gsuite api

ryandeivert:
to: @chunyong-lin 
resolves: #1036 
resolves: #888

## Background

The GSuite API changed, and a few `applicationName` values were deprecated/updated:

https://developers.google.com/admin-sdk/reports/v1/reference/activities/watch#parameters
https://stackoverflow.com/a/58282106

## Changes

* Updating gsuite reports apps in include:
  * `access_transparency`
  * `group` (vs old `groups`)
  * `groups_enterprise`
  * `user_accounts`
* Removing deprecated `saml` app.

## Testing

Updates to unit tests.


ryandeivert:
@chunyong-lin I actually just updated the client lib here too - see new commits

PR #1049, opened 2019-11-13
fix bugs with updated dependencies and gsuite app

ryandeivert:
to: @chunyong-lin 

## Changes

* Updating the precompiled box dependencies to reflect the new 2.6.1 version
* Fixing bug with the 'group' vs 'groups' endpoint that google can't make their mind up on.

## Testing

Updates to unit tests


ryandeivert:
cc @chunyong-lin

Issue #1026, opened 2019-10-01
[bug] terraform s3_events module proper s3 event notification support


ryandeivert:
## Background

The [current s3 events module](https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/terraform/modules/tf_stream_alert_s3_events/main.tf#L16) only supports one s3 event notification due to the way the way the `aws_s3_bucket_notification` resource is implemented.

Terraform version 0.12.x introduces the concept of for loops, and we will be able to properly support multiple event notifications per-lambda function.

Reference: https://www.hashicorp.com/blog/hashicorp-terraform-0-12-preview-for-and-for-each

## Desired Change

Once #947 merges, we should update this module.

We should also update (reimplement) the logic for the [generation](https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/stream_alert_cli/terraform/s3_events.py#L45-L67) so it does not create multiple [role policies](https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/terraform/modules/tf_stream_alert_s3_events/main.tf#L27) per bucket/function.

Notifications should be handled separately from policy creation. The config for this could look like:

```json
    "s3_events": {
      "bucket_name": [
        {
          "filter_prefix": "AWSLogs/account_id_01/",
          "enable_events": false
        },
        {
          "filter_prefix": "AWSLogs/account_id_02/",
          "enable_events": false
        }
      ]
    }
```

ryandeivert:
complete with #1051

PR #1051, opened 2019-11-15
fixing the tf_s3_events module to properly handle multiple lambda_function blocks

ryandeivert:
to: @chunyong-lin / @Ryxias 
resolves: #1026 

## Background

See #1026

## Changes

* not sorting keys in config, since it is not necessary
* fixing typo in tf description
* removing s3 events legacy garbage for config
* updating tf_s3_events module for proper handling of more than one filter per bucket
* updating tf_s3_events module generation code and tests
* updating documentation to reflect changes to tf_s3_events module

## Testing

Updates to unit tests. I also init'd this module by itself with tf vars to ensure it was working as expected.

### DIFF
--- a/docs/source/clusters.rst
+++ b/docs/source/clusters.rst
@@ -613,16 +613,18 @@ Example: S3 Events Cluster
     {
       "id": "s3-events-example",
       "modules": {
-        "s3_events": [
-          {
-            "bucket_id": "bucket-1",
-            "enable_events": true
-          },
-          {
-            "bucket_id": "bucket-2",
-            "enable_events": true
-          }
-        ],
+        "s3_events": {
+          "bucket_name_01": [
+            {
+              "filter_prefix": "AWSLogs/1234",
+              "filter_suffix": ".log"
+            },
+            {
+              "filter_prefix": "AWSLogs/5678"
+            }
+          ],
+          "bucket_name_02": []
+        },
         "stream_alert": {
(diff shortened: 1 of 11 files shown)


docs/source/clusters.rst
```diff
@@ -636,17 +638,19 @@ Example: S3 Events Cluster
       "region": "us-east-1"
     }
 
-This configures the two buckets to notify the classifier function in this cluster when new objects
-arrive in the bucket, and authorizes the classifier to download objects from either bucket.
+This configures the two buckets (``bucket_name_01`` and ``bucket_name_02``) to notify the classifier
+function in this cluster when new objects arrive in the bucket at the specified (optional) prefix(es),
+provided the objects have the specified (optional) suffix(es). Additionally, this will authorize the
```
chunyong-lin: Is it `AND` or `OR` logic if we have both prefix(es) and suffix(es) configured for a bucket?

ryandeivert: I think it's `and`.. but they're both optional

https://www.terraform.io/docs/providers/aws/r/s3_bucket_notification.html

chunyong-lin: Aha, you are right. It will be `AND` logic. There is an example in aws doc page https://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html.

PR #1035, opened 2019-10-10
[terraform] upgrade Terraform to 0.12.9

scoders:
to: @ryandeivert
cc: @airbnb/streamalert-maintainers
resolves: #947 

## Background
Update streamalert's terraform modules to conform to the latest terraform specifications

## Changes
- Lots of syntax changes to terraform modules introduced by the 0.12 major release 
- Upgraded unit tests to expect explicit relative paths
- Fixed a bug in `streamalert_cli/test/handler.py` while getting non-existent dict key

## Testing
All of the following tests pass:
`tests/scripts/unit_tests.sh`
```
	Ran 1106 tests in 11.219s
	OK
```
`python manage.py test rules`
`python manage.py test classifier`
`python manage.py test live`
```
	Summary:
	Total Tests: 103
	Pass: 103
	Fail: 0
```

PR #1052, opened 2019-11-15
final updates for terraform 12 support

ryandeivert:
to: @chunyong-lin / @Ryxias 
related to: #1035, #1051 

## Changes

* This PR is just the combination of #1035 and #1051


conf/sources.json
```diff
@@ -0,0 +1,69 @@
+{
```
chunyong-lin: We had migrated `sources` configure to each cluster. We should remove this file.

ryandeivert: done!

PR #1053, opened 2019-11-20
fixing small bugs with terraform 12 code

ryandeivert:
to: @chunyong-lin 
related to: #1035 , #1052 

## Changes

* Fixing documentation update that was missed.
* Adding fix for potential invalid terraform module name (eg - a bucket name contains a `.`)
* Fixing issue where no `lambda_function` block would be defined for s3 events where no `filter_prefix` or `filter_suffix` parameters were set.
* Removing unnecessary prefixing from role policy/lambda permission resources in `tf_s3_events` module.
* Removing duplicative variable declaration in `tf_kinesis_events` module.
* Updating `# TF-UPGRADE-TODO` block to use local arn list directly as list.

## Testing

Updating unit tests.

### DIFF
--- a/terraform/modules/tf_s3_events/main.tf
+++ b/terraform/modules/tf_s3_events/main.tf
@@ -1,6 +1,10 @@
+locals {
+  sanitized_bucket_name = replace(var.bucket_name, "/[^a-zA-Z0-9_-]/", "_")
+}
+
 // Lambda Permission: Allow S3 Event Notifications to invoke Lambda
 resource "aws_lambda_permission" "allow_bucket" {
-  statement_id  = "${var.prefix}_${var.cluster}_InvokeFromS3Bucket_${var.bucket_name}"
+  statement_id  = "${var.prefix}_${var.cluster}_InvokeFromS3Bucket_${local.sanitized_bucket_name}"
   action        = "lambda:InvokeFunction"
   function_name = var.lambda_function_name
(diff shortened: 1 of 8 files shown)


terraform/modules/tf_s3_events/main.tf
```diff
@@ -26,7 +36,7 @@ resource "aws_s3_bucket_notification" "bucket_notification" {
 
 // IAM Policy: Allow Lambda to GetObjects from S3
 resource "aws_iam_role_policy" "lambda_s3_permission" {
-  name   = "${var.prefix}_${var.cluster}_S3GetObjects_${var.bucket_name}"
+  name   = "S3GetObjects_${var.bucket_name}"
```
chunyong-lin: Why don't we also use `local.sanitized_bucket_name` instead of `var.bucket_name`?

ryandeivert: because the `name` attribute for these resources support special characters (eg - characters that are NOT in the set `[a-zA-Z0-9_-]`).. while the `statement_id` attribute about only supports those characters: https://docs.aws.amazon.com/lambda/latest/dg/API_AddPermission.html#SSS-AddPermission-request-StatementId

chunyong-lin: Aha, this is annoy that we have to be very accurate to know which resource name needs to be sanitized 😒 Thanks for the explanation!

Issue #1040, opened 2019-10-15
[apps] box admin app: requests.exceptions.ReadTimeout


ryandeivert:
## Background

```
Traceback (most recent call last):
File "/var/task/urllib3/connectionpool.py", line 387, in _make_request
six.raise_from(e, None)
File "<string>", line 2, in raise_from
File "/var/task/urllib3/connectionpool.py", line 383, in _make_request
httplib_response = conn.getresponse()
File "/var/lang/lib/python3.7/http/client.py", line 1336, in getresponse
response.begin()
File "/var/lang/lib/python3.7/http/client.py", line 306, in begin
version, status, reason = self._read_status()
File "/var/lang/lib/python3.7/http/client.py", line 267, in _read_status
line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
File "/var/lang/lib/python3.7/socket.py", line 589, in readinto
return self._sock.recv_into(b)
File "/var/lang/lib/python3.7/ssl.py", line 1071, in recv_into
return self.read(nbytes, buffer)
File "/var/lang/lib/python3.7/ssl.py", line 929, in read
return self._sslobj.read(len, buffer)
socket.timeout: The read operation timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File "/var/task/requests/adapters.py", line 449, in send
timeout=timeout
File "/var/task/urllib3/connectionpool.py", line 641, in urlopen
_stacktrace=sys.exc_info()[2])
File "/var/task/urllib3/util/retry.py", line 368, in increment
raise six.reraise(type(error), error, _stacktrace)
File "/var/task/urllib3/packages/six.py", line 686, in reraise
raise value
File "/var/task/urllib3/connectionpool.py", line 603, in urlopen
chunked=chunked)
File "/var/task/urllib3/connectionpool.py", line 389, in _make_request
self._raise_timeout(err=e, url=url, timeout_value=read_timeout)
File "/var/task/urllib3/connectionpool.py", line 307, in _raise_timeout
raise ReadTimeoutError(self, url, "Read timed out. (read timeout=%s)" % timeout_value)
urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='api.box.com', port=443): Read timed out. (read timeout=3.05)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File "/var/task/streamalert/apps/app_base.py", line 58, in _wrapper
return func(*args, **kwargs)
[...]

PR #1057, opened 2019-11-26
adding fix for box app request timeout (#1040)

ryandeivert:
to: @chunyong-lin 
resolves: #1040 

## Background

See #1040 

## Changes

* Adding a change to detect a requests timeout and retry with increased timeout. 
* Once 5 attempts have been made, the original timeout exception will still be raised.

## Testing

Updating unit test to for retry failure/success cases.

### DIFF
--- a/streamalert/apps/_apps/box.py
+++ b/streamalert/apps/_apps/box.py
@@ -18,7 +18,7 @@ import json
 from boxsdk import Client, JWTAuth
 from boxsdk.exception import BoxException
 from boxsdk.object.events import EnterpriseEventsStreamType
-from requests.exceptions import ConnectionError as requestsConnectionError
+import requests
 
 from . import AppIntegration, StreamAlertApp, get_logger, safe_timeout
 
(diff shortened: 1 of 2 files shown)


streamalert/apps/_apps/box.py
```diff
@@ -30,6 +30,7 @@
 class BoxApp(AppIntegration):
     """BoxApp integration"""
     _MAX_CHUNK_SIZE = 500
+    _MAX_RETRY_COUNT = 5
```
chunyong-lin: Can we reduce the value to `3`? With `_DEFAULT_REQUEST_TIMEOUT` set to 3.05s, and the timeout will be 96.64 seconds. Most SA apps set lambda timeout to 40 or 60 seconds and the lambda will be timed out before it reaches max retry count `5`.

ryandeivert: with the logic in place, it will actually only _retry_ 4 times: `retry_count == 5`

this means the max will be 48.8 seconds. the 40 to 60 seconds things is something we have adopted internally but that's not really a "requirement" per-se

Issue #1054, opened 2019-11-22
[improvement] Lock terraform remote state on deployment


blakemotl:
## Background
Currently if multiple people attempt to deploy to the same Streamalert deployment, there is nothing stopping them. This can result in a broken state file since two users are reading/writing from it at the same time. Terraform has native support for remote state locking. In the case of the s3 backend this involves using a dynamodb table.

## Desired Change

1. Add terraform state locking support to Streamalert by leveraging dyanmodb.
2. Update CLI to manage the dynamodb table.

PR #1059, opened 2019-11-27
[terraform] support remote state file locking

blakemotl:
resolves: [#1054](https://github.com/airbnb/streamalert/issues/1054)

## Background
Currently if multiple people attempt to deploy to the same Streamalert deployment, there is nothing stopping them. This can result in a broken state file since two users are reading/writing from it at the same time. Terraform has native support for remote state locking. In the case of the s3 backend this involves using a DynamoDB table.

This implementation attempts to create and destroy the table automatically as part of the cli process. It bypasses terraform because once locking is enabled the table needs to be created and instead uses boto3. I tested to see if terraform was smart enough to recognize there was a module to create the locking table and execute that before attempting to lock the state, but it does not.

## Changes

* Added a new helper function to manage creation and destruction of the DynamoDB table used for s3 state locking as per terraform's [docs](https://www.terraform.io/docs/backends/types/s3.html) 
* Setup the CLI to automatically create the table on init and destroy the table on destroy.
* Updated docs to cover this new feature.
* Added tests for table creation/destruction.

## Testing
Unit and integration tests passing locally.
Functionally tested in a staging environment.

### DIFF
--- a/conf/global.json
+++ b/conf/global.json
@@ -47,7 +47,7 @@
   },
   "terraform": {
     "create_bucket": true,
-    "lock_state_file": true,
+    "lock_state_file": false,
     "tfstate_bucket": "PREFIX_GOES_HERE.streamalert.terraform.state",
     "tfstate_s3_key": "stream_alert_state/terraform.tfstate"
   }
--- a/conf/global.json
+++ b/conf/global.json
@@ -47,6 +47,7 @@
   },
   "terraform": {
     "create_bucket": true,
+    "lock_state_file": true,
     "tfstate_bucket": "PREFIX_GOES_HERE.streamalert.terraform.state",
     "tfstate_s3_key": "stream_alert_state/terraform.tfstate"
   }
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -95,7 +95,17 @@ Deploy
   ./manage.py configure aws_account_id 111111111111  # Replace with your 12-digit AWS account ID
   ./manage.py configure prefix NAME                  # Choose a unique name prefix (alphanumeric characters only)
 
-2. Build the StreamAlert infrastructure for the first time:
+2. Configure terraform state locking (optional):
+
+  Streamalert supports terraforms native ability to lock the remote s3 state file whenever a user is planning and applying terraform configuration.
+  This is to prevent multiple users from deploying Streamalert at the same time potentially resulting in a broken state.
+  To enable this feature this you must first create a DynamoDB table named ``<prefix>_streamalert_terraform_state_lock`` with a primary key named ``LockID``.
+  Streamalert will automatically create the proper IAM permissions so that terraform can read and write to this table.
+  See `terraform's documentation <https://www.terraform.io/docs/backends/types/s3.html>`_ for more information.
+  
+  Once you have created this table, open ``conf/global.json``, find the ``terraform`` key and change ``lock_state_file`` to ``true``.
+
+3. Build the StreamAlert infrastructure for the first time:
 
 .. code-block:: bash
 
@@ -106,7 +116,7 @@ There will be multiple Terraform prompts, type "yes" at each one to continue.
 .. note:: You only need to ``./manage.py init`` once for any given StreamAlert deployment,
    although it is safe to run again if necessary.
 
-3. At this point, StreamAlert is up and running! You can, for example, see the S3 buckets
+4. At this point, StreamAlert is up and running! You can, for example, see the S3 buckets
 that were automatically created:
 
 .. code-block:: bash
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -87,6 +87,10 @@ with permissions for at least the following services:
 
 Deploy
 ------
+.. note:: Streamalert supports Terraform's native ability to lock the remote s3 state file whenever a user is planning and applying Terraform configuration.
+    This is to prevent multiple users from deploying StreamAlert at the same time potentially resulting in a broken state.
+    StreamAlert will automatically create and destroy this table via the command line interface.
+    See `Terraform's documentation <https://www.terraform.io/docs/backends/types/s3.html>`_ for more information.
 
 1. Set basic StreamAlert configuration options:
 
@@ -95,15 +99,7 @@ Deploy
   ./manage.py configure aws_account_id 111111111111  # Replace with your 12-digit AWS account ID
   ./manage.py configure prefix NAME                  # Choose a unique name prefix (alphanumeric characters only)
 
-2. Configure terraform state locking:
-
-  Streamalert supports terraforms native ability to lock the remote s3 state file whenever a user is planning and applying terraform configuration.
-  This is to prevent multiple users from deploying Streamalert at the same time potentially resulting in a broken state.
-  Streamalert will automatically create and destroy this table via the command line interface;
-  however, you may need to add the appropriate permissions (GetItem, PutItem and DeleteItem) to an IAM role to allow other users to deploy.
-  See `terraform's documentation <https://www.terraform.io/docs/backends/types/s3.html>`_ for more information.
-
-3. Build the StreamAlert infrastructure for the first time:
+2. Build the StreamAlert infrastructure for the first time:
 
 .. code-block:: bash
 
@@ -114,7 +110,7 @@ There will be multiple Terraform prompts, type "yes" at each one to continue.
 .. note:: You only need to ``./manage.py init`` once for any given StreamAlert deployment,
    although it is safe to run again if necessary.
 
-4. At this point, StreamAlert is up and running! You can, for example, see the S3 buckets
+3. At this point, StreamAlert is up and running! You can, for example, see the S3 buckets
 that were automatically created:
 
 .. code-block:: bash
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -95,15 +95,13 @@ Deploy
   ./manage.py configure aws_account_id 111111111111  # Replace with your 12-digit AWS account ID
   ./manage.py configure prefix NAME                  # Choose a unique name prefix (alphanumeric characters only)
 
-2. Configure terraform state locking (optional):
+2. Configure terraform state locking:
 
   Streamalert supports terraforms native ability to lock the remote s3 state file whenever a user is planning and applying terraform configuration.
   This is to prevent multiple users from deploying Streamalert at the same time potentially resulting in a broken state.
-  To enable this feature this you must first create a DynamoDB table named ``<prefix>_streamalert_terraform_state_lock`` with a primary key named ``LockID``.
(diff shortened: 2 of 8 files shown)


streamalert_cli/terraform/generate.py
```diff
@@ -167,6 +167,9 @@ def generate_main(config, init=False):
             'key': config['global']['terraform']['tfstate_s3_key'],
             'region': config['global']['account']['region'],
             'encrypt': True,
+            'dynamodb_table': "{}_streamalert_terraform_state_lock".format(
```
ryandeivert: single quotes around format string here please

streamalert_cli/terraform/helpers.py
```diff
@@ -24,3 +29,33 @@ def terraform_check():
                        'your $PATH:\n'
                        '\t$ export PATH=$PATH:/usr/local/terraform/bin')
     return run_command(['terraform', 'version'], error_message=prereqs_message, quiet=True)
+
+def terraform_state_lock(action, region, table):
+    ddb_client = boto3.client('dynamodb', region_name=region)
```
ryandeivert: boto3 clients should utilize [this](https://github.com/airbnb/streamalert/blob/2746509ccf4513e882623fcd7ea83f1b5b29e6d7/streamalert/shared/helpers/boto.py#L28) helper whenever possible for the config.

[See here for an example](https://github.com/airbnb/streamalert/blob/2746509ccf4513e882623fcd7ea83f1b5b29e6d7/streamalert/shared/helpers/aws_api_client.py#L47).

streamalert_cli/terraform/helpers.py
```diff
@@ -24,3 +29,33 @@ def terraform_check():
...
+                    'AttributeType': 'S'
+                }],
+                TableName=table,
+                KeySchema=[{
+                    'AttributeName': 'LockID',
+                    'KeyType': 'HASH'
+                }],
+                BillingMode='PAY_PER_REQUEST'
+            )
+            wait = True
```
ryandeivert: there are utility functions in boto3 that implement this for you. See: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Waiter.TableExists

you don't have to implement this yourself here.

streamalert_cli/terraform/helpers.py
```diff
@@ -24,3 +29,33 @@ def terraform_check():
...
+            while wait:
+                desc_resp = ddb_client.describe_table(TableName=table)
+                if desc_resp['Table']['TableStatus'] == 'ACTIVE':
+                    wait = False
+                else:
+                    time.sleep(1)
+    elif action == 'destroy':
+        if table in ddb_tables['TableNames']:
+            LOGGER.info('Destroying terraform state locking DynamoDB table')
+            ddb_client.delete_table(TableName=table)
```
ryandeivert: you should also use the [`table_not_exsits`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html#DynamoDB.Waiter.TableNotExists) waiter when deleting the table

streamalert_cli/terraform/helpers.py
```diff
@@ -24,3 +29,33 @@ def terraform_check():
                        'your $PATH:\n'
                        '\t$ export PATH=$PATH:/usr/local/terraform/bin')
     return run_command(['terraform', 'version'], error_message=prereqs_message, quiet=True)
+
+def terraform_state_lock(action, region, table):
+    ddb_client = boto3.client('dynamodb', region_name=region)
+    ddb_tables = ddb_client.list_tables()
+    if action == 'create':
+        if table not in ddb_tables['TableNames']:
```
ryandeivert: I'm a fan of avoiding large nested blocks, so returning early here would make this cleaner:
```
if table in ddb_tables['TableNames']:
    return True  # table already exists
```

streamalert_cli/terraform/helpers.py
ryandeivert: something like:

```
ddb_client.create_table(
    AttributeDefinitions=[{
        'AttributeName': 'LockID',
        'AttributeType': 'S'
    }],
    TableName=table,
    KeySchema=[{
        'AttributeName': 'LockID',
        'KeyType': 'HASH'
    }],
    BillingMode='PAY_PER_REQUEST'
)

waiter = ddb_client.get_waiter('table_exists')
waiter.wait(TableName=table)
```

streamalert_cli/terraform/helpers.py
```diff
@@ -24,3 +29,33 @@ def terraform_check():
                        'your $PATH:\n'
                        '\t$ export PATH=$PATH:/usr/local/terraform/bin')
     return run_command(['terraform', 'version'], error_message=prereqs_message, quiet=True)
+
+def terraform_state_lock(action, region, table):
```
ryandeivert: since this function appears to be specifically related to creation and deletion of the dynamo db table, maybe pick a more appropriate/descriptive function name?

docs/source/getting-started.rst
```diff
@@ -95,7 +95,15 @@ Deploy
   ./manage.py configure aws_account_id 111111111111  # Replace with your 12-digit AWS account ID
   ./manage.py configure prefix NAME                  # Choose a unique name prefix (alphanumeric characters only)
 
-2. Build the StreamAlert infrastructure for the first time:
+2. Configure terraform state locking:
```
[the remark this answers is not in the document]
chunyong-lin: I think @ryandeivert  has a good point. Is the terraform state locking mandatory or optional?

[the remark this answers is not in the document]
ryandeivert: I suggested that we _not_ make it configurable, since I could not come up with an argument as to why anyone would explicitly not want it. do you have an opinion @chunyong-lin ?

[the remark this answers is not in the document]
blakemotl: I left it as configure because there is still some configuration needed on there end, I.E adding the proper permissions to whatever role the users that need to deploy are in. This is definitely a mandatory feature.

conf/global.json
```diff
@@ -47,6 +47,7 @@
   },
   "terraform": {
     "create_bucket": true,
+    "lock_state_file": false,
```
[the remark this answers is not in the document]
blakemotl: @ryandeivert yeah thats exactly what happened 🤦‍♂

streamalert_cli/terraform/helpers.py
blakemotl: Ohhh very good to know

docs/source/getting-started.rst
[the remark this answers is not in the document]
ryandeivert: I'm of the opinion that there is a LOT of permissions a user needs to have in order to even deploy streamalert (all the actions/api calls that terraform makes under the hood require the user to have all applicable permissions...). I don't think we should be highlighting this here, when we make no mention of the super set of permissions needed for deployment in general

eventually, we should publish a least-privileged IAM role template that people can use, but we already mention the user should have DynamoDB permissions in general. if they don't have the right privileges, they'll get an `AccessDenied` exception and be able to address on their end. we shouldn't try to account for this yet

streamalert_cli/terraform/helpers.py
blakemotl: I thought about doing this. I decided not to because technically it doesn't need to wait for the table to be destroyed since this is one of the last steps and is not required to continue execution. Plus I didn't like the idea of it hanging if the table takes a while to delete. However now that I think about it again, terraform does this and this potentially makes the user experience better so I will add it.

ryandeivert: my reasoning is that the user likely would not want a rogue/orphaned table hanging around without their knowledge. if the deletion fails and the waiter doesn't succeed, at least an exception will be raise so they know to go an remove manually

docs/source/getting-started.rst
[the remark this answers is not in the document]
blakemotl: @ryandeivert Ah yeah that makes sense. DynamoDB is already in that list. Since there is no action for the user to take, I'm going to make this a note in the getting started section and remove it from deployment steps.

streamalert_cli/terraform/helpers.py
blakemotl: I think I'm just going to break it out into two functions. It will make the logic cleaner and easier to name the function

ryandeivert: sgtm

docs/source/getting-started.rst
```diff
@@ -87,6 +87,10 @@ with permissions for at least the following services:
 
 Deploy
 ------
+.. note:: Streamalert supports Terraform's native ability to lock the remote s3 state file whenever a user is planning and applying Terraform configuration.
```
ryandeivert: `Streamalert` --> `StreamAlert`

streamalert_cli/terraform/handlers.py
```diff
@@ -23,7 +23,11 @@
 from streamalert_cli.helpers import check_credentials, continue_prompt, run_command, tf_runner
 from streamalert_cli.manage_lambda.deploy import deploy
 from streamalert_cli.terraform.generate import terraform_generate_handler
-from streamalert_cli.terraform.helpers import terraform_check, terraform_state_lock
+from streamalert_cli.terraform.helpers import (
+    terraform_check,
```
ryandeivert: nit: order imports alphabetically pls

docs/source/getting-started.rst
blakemotl: 🤦‍♂ I double checked and still missed one

streamalert_cli/terraform/helpers.py
```diff
@@ -30,32 +31,45 @@ def terraform_check():
...
+        BillingMode='PAY_PER_REQUEST'
+    )
+    waiter = ddb_client.get_waiter('table_exists')
+    waiter.wait(TableName=table)
+
+def destroy_tf_state_lock_ddb_table(region, table):
+    """Destroy the DynamoDB table for terraform remote state locking
+
+    Args:
+        region: The AWS region to destroy the table in
```
ryandeivert: provide a type for these in parenthesis please - like:

```
region (str): The AWS ...
```

streamalert_cli/terraform/helpers.py
```diff
@@ -30,32 +31,45 @@ def terraform_check():
                        '\t$ export PATH=$PATH:/usr/local/terraform/bin')
     return run_command(['terraform', 'version'], error_message=prereqs_message, quiet=True)
 
-def terraform_state_lock(action, region, table):
-    ddb_client = boto3.client('dynamodb', region_name=region)
+def create_tf_state_lock_ddb_table(region, table):
+    """Create the DynamoDB table for terraform remote state locking
+
+    Args:
+        region: The AWS region to create the table in
```
ryandeivert: same as below comment

blakemotl:
Thanks for the feedback @ryandeivert. I fixed your last comments. I will merge this today but sync it on Monday.

PR #1063, opened 2019-12-03
raise exception when config sources are misconfigured

ryandeivert:
to: @chunyong-lin / @blakemotl 

## Background

A misconfigured source configuration will cause the classifiers to fail to process data. Currently, an error is logged and the function just exits. This should fail more verbosely so the user knows something is wrong.

## Changes

* Adding a new `ConfigError` class (that inherits from a new `StreamAlertError` class)
* Raising `ConfigError` exceptions if there is a sources misconfiguration.

## Testing

Updating unit tests to check that exceptions are now raised.

### DIFF
--- a/streamalert/shared/exceptions.py
+++ b/streamalert/shared/exceptions.py
@@ -1,0 +1,23 @@
+"""
+Copyright 2017-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+
+class StreamAlertError(Exception):
+    """Base streamalert exception for inheritance"""
+
+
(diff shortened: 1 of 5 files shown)


streamalert/shared/exceptions.py
```diff
@@ -0,0 +1,23 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+
+class StreamAlertError(Exception):
```
Ryxias: 👍

Ryxias: I'm a big of marker interfaces!!!1`1!!1

Not necessarily saying this is better, but is it possible to do it with `StreamAlertError` being some sort of mixin that takes advantage of Python's multiple inheritance?

Like, something like

```
class StreamAlertError: # <-- Does not extend Exception
  pass

class StreamAlertRuntimeError(RuntimeError, StreamAlertError):  # <-- inherits from both
  pass

class StreamAlertTypeError(TypeError, StreamAlertError):  # <-- etc
  pass
```

I'm not 100% certain this is in any way pythonic or better or worse, just something that I've struggled with a lot in the past when I **_didn't_** have multiple inheritance and I really wished I did.

ryandeivert: what's the benefit in the base exception not inheriting from `Exception`? type inference would fail and you cannot `raise` a `StreamAlertError` directly if you do not

Issue #963, opened 2019-08-08
[Enhancement] Able to alter existing table when statement reaches query API limits


chunyong-lin:
## Background
It is pretty common that there will have new keys added to an event schema over the time. User will have to add these new keys as optional keys to the schema, alter the Athena table to reflect new keys and allow historical data to use new schema as well.

StreamAlert provides command line to alter the table based on the new schema by running following command.
```
manage.py athena rebuild-partitions  --bucket s3.bucket.name  --table-name my_athena_table
```

However, [the maximum allowed query string length in Athena API call is 262144 bytes](https://docs.aws.amazon.com/athena/latest/ug/service-limits.html). In [current implementation](https://github.com/airbnb/streamalert/blob/master/stream_alert_cli/athena/handler.py#L137), StreamAlert will save the whole query into a txt file if the length exceeds the limit. And then the user has to run the query from Athena console manually. It is not ideal.

## Desired Change
Slice query string based on the length limit and loop through the query strings to alter table.

PR #1067, opened 2019-12-04
[cli] Allow to rebuild partitions when the statement length exceeds the limit

chunyong-lin:
related to:
resolves: #963 

## Background
It is pretty common that there will have new keys added to an log schema over the time.  StreamAlert users will have to add these new keys as optional keys to the schema. If the users use Athena to search historical data, it is required to rebuild the table for the log schema which has new key added.

StreamAlert provides command line to alter the table based on the new schema by running following command.
```
manage.py athena rebuild-partitions  --bucket s3.bucket.name  --table-name my_athena_table
```

However, [the maximum allowed query string length in Athena API call is 262144 bytes](https://docs.aws.amazon.com/athena/latest/ug/service-limits.html). In current implementation, StreamAlert will save the whole query into a txt file if the length exceeds the limit. And the user has to run the query from Athena console manually. It is not ideal.
https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/stream_alert_cli/athena/handler.py#L137-L145

## Changes

* Break the `ALTER TABLE` statement for rebuilding partitions into a list of string under the Athena statement length limit.
* Add test cases for the code change.

## Testing
* Unit test cases added.
* Tested in staging environment.

### DIFF
--- a/streamalert_cli/athena/helpers.py
+++ b/streamalert_cli/athena/helpers.py
@@ -36,23 +36,28 @@ SCHEMA_TYPE_MAPPING = {
 # Athena query statement length limit
 MAX_QUERY_LENGTH = 262144
 
-def add_partition_statement(partitions, bucket, table_name):
-    """Generate ALTER TABLE commands from existing partitions.
+def add_partition_statements(partitions, bucket, table_name):
+    """Generate ALTER TABLE commands from existing partitions. It wil yield Athena
+    statement string(s), the length of each string should be less than Athena query
+    statement length limit, 262144 bytes.
+    https://docs.aws.amazon.com/athena/latest/ug/service-limits.html
 
     Args:
         partitions (set): The unique set of partitions gathered from Athena
         bucket (str): The bucket name
         table_name (str): The name of the Athena table
 
-    Returns:
-        list[str]: The ALTER TABLE statements to add the new partitions
+    Yields:
+        string: The ALTER TABLE statements to add the new partitions
     """
+    # Each add partition statement starting with "ALTER TABLE"
+    initial_statement = 'ALTER TABLE {} ADD IF NOT EXISTS'.format(table_name)
+    initial_statement_len = len(initial_statement)
 
-    # monitor statement length, if it is less than 0, start a new statement.
-    # return a list of str
-    results = []
-    statement = ['ALTER TABLE {} ADD IF NOT EXISTS'.format(table_name)]
-    statement_len = MAX_QUERY_LENGTH - len(statement[0])
+    # The statement will be stored in a list of string format before join into a string
+    statement = [initial_statement]
+    statement_len = initial_statement_len
+
     fmt_values = {
         'bucket': bucket,
         'table_name': table_name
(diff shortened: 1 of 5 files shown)


streamalert_cli/athena/helpers.py
```diff
@@ -33,6 +33,8 @@
     list: 'array<string>'
 }
 
+# Athena query statement length limit
+MAX_QUERY_LENGTH = 262144
 
 def add_partition_statement(partitions, bucket, table_name):
```
ryandeivert: this should probably be renamed to `add_partition_statements` (plural)

streamalert_cli/athena/helpers.py
```diff
@@ -43,9 +45,14 @@ def add_partition_statement(partitions, bucket, table_name):
         table_name (str): The name of the Athena table
 
     Returns:
-        str: The ALTER TABLE statement to add the new partitions
+        list[str]: The ALTER TABLE statements to add the new partitions
     """
-    statements = ['ALTER TABLE {} ADD IF NOT EXISTS'.format(table_name)]
+
```
ryandeivert: remove empty line here

streamalert_cli/athena/helpers.py
```diff
@@ -43,9 +45,14 @@ def add_partition_statement(partitions, bucket, table_name):
...
+    # monitor statement length, if it is less than 0, start a new statement.
+    # return a list of str
+    results = []
```
ryandeivert: store the initial value outside the list so it can be referenced without recreating:

```
initial_value = 'ALTER TABLE {} ADD IF NOT EXISTS'.format(table_name)
statement = [initial_value]
```

then reference `initial_value` below on line 72

streamalert_cli/athena/helpers.py
```diff
@@ -57,10 +64,21 @@ def add_partition_statement(partitions, bucket, table_name):
             continue
 
         fmt_values.update(parts.groupdict())
+        partition_stmt = PARTITION_STMT.format(**fmt_values)
+
+        # If statement will be reset if it is about to reach the MAX_QUERY_LENGTH
+        if statement_len - len(partition_stmt) - len(statement) < 0:
+            results.append(' '.join(statement))
```
ryandeivert: you could make this a generator and have this `yield ' '.join(statement)` instead of storing them in a list that is returned

streamalert_cli/athena/helpers.py
```diff
@@ -57,10 +64,21 @@ def add_partition_statement(partitions, bucket, table_name):
             continue
 
         fmt_values.update(parts.groupdict())
+        partition_stmt = PARTITION_STMT.format(**fmt_values)
+
+        # If statement will be reset if it is about to reach the MAX_QUERY_LENGTH
+        if statement_len - len(partition_stmt) - len(statement) < 0:
```
ryandeivert: store the value of `len(partition_stmt)`  to avoid calculating twice (L75 below)

ryandeivert: I like the idea of counting _up_ to the ceiling of max size, not down to the floor. imho the way this logic is now makes it harder to reason about.. meaning, could you do something like:

```python

init_stmt_len = len(statement[0])
statement_len = init_stmt_len

...
partition_stmt = PARTITION_STMT.format(**fmt_values)
partition_stmt_len = len(partition_stmt)

if statement_len + partition_stmt_len >= MAX_QUERY_LENGTH:
    # do the append/yield
    ...
    # reset the initial value
    statement_len = init_stmt_len

statement_len += partition_stmt_len

...
```

streamalert_cli/athena/helpers.py
chunyong-lin: Great idea!!

streamalert_cli/athena/helpers.py
chunyong-lin: Aha, I get your point and it makes sense. Will make a change.

chunyong-lin:
@ryandeivert I have addressed your comments. PTAL! Thanks!

streamalert_cli/athena/helpers.py
ryandeivert: NEW LINE!!!

PR #1070, opened 2019-12-12
Don't throw an exception if no partitions are added

blakemotl:

## Background
There is a small bug in the Athena partition refresh function where if enough consecutive s3 event notifications are for files and folders that are invalid, the function will throw an exception and fall into a loop where it keeps trying to process these events. An example of this happening, is the creation of s3 placeholder files that end in _$folder$, which fail the regex check.

## Changes
* Remove the exception thrown when add_partitions adds no partitions(it already logs this anyway)
* Log when a placeholder file is processed by the function.
* Updates tests to account for these changes.

## Testing
Passing unit/integration tests.

### DIFF
--- a/streamalert/athena_partition_refresh/main.py
+++ b/streamalert/athena_partition_refresh/main.py
@@ -118,7 +118,7 @@ class AthenaRefresher:
                         break
 
                 if not match:
-                    LOGGER.error('The key %s does not match any regex, skipping', key)
+                    LOGGER.warning('The key %s does not match any regex, skipping', key)
                     continue
 
                 # Get the path to the objects in S3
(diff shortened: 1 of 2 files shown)


streamalert/athena_partition_refresh/main.py
```diff
@@ -205,17 +203,16 @@ def run(self, event):
                 # Account for special characters in the S3 object key
                 # Example: Usage of '=' in the key name
                 object_key = urllib.parse.unquote_plus(s3_rec['s3']['object']['key']).encode()
+                if object_key.endswith(b'_$folder$'):
```
ryandeivert: why the `b` prefix here? shouldn't this be a string not bytes?

blakemotl: Encode returns a byte array and the argument to endswith has to match that type.

Issue #1055, opened 2019-11-23
migrate threat_intel_downloader function to tf_lambda module


ryandeivert:
## Background

The `threat_intel_downloader` lambda function is not currently using the `tf_lambda` terraform module we introduced some time back. The [current implementation](https://github.com/airbnb/streamalert/tree/4b034c93a8af32772f6cf39771dc71562f10ee51/terraform/modules/tf_threat_intel_downloader) could be drastically simplified by using the [`tf_lambda` module](https://github.com/airbnb/streamalert/tree/4b034c93a8af32772f6cf39771dc71562f10ee51/terraform/modules/tf_lambda).

## Desired Change

Update the [`threat_intel_downloader.py`](https://github.com/airbnb/streamalert/blob/4b034c93a8af32772f6cf39771dc71562f10ee51/streamalert_cli/terraform/threat_intel_downloader.py#L20) generation code to use the `tf_lambda` module.

Update the [tf_threat_intel_downloader](https://github.com/airbnb/streamalert/blob/4b034c93a8af32772f6cf39771dc71562f10ee51/terraform/modules/tf_threat_intel_downloader/iam.tf#L1) module to only create other resources/IAM permissions not included in the `tf_lamdba` module.

[Example](https://github.com/airbnb/streamalert/blob/4b034c93a8af32772f6cf39771dc71562f10ee51/streamalert_cli/terraform/rules_engine.py#L22) terraform generation code using the `tf_lambda` module

[Example](https://github.com/airbnb/streamalert/blob/4b034c93a8af32772f6cf39771dc71562f10ee51/streamalert_cli/terraform/rules_engine.py#L22) of iam/etc configuration needed in addition to the `tf_lambda` module code.

ryandeivert:
this is complete with #1074

PR #1074, opened 2020-01-09
threat_intel_downloader module now uses tf_lambda module

GarretReece:
to: @ryandeivert 
resolves: #1055 

## Background

The threat_intel_downloader module creates a lambda as part of its instantiation, but does it directly in its own module.

## Changes

* the thread_intel_downloader module now uses the tf_lambda module for lambda creation

## Testing

Unit tests all pass. The threat_intel_downloader was enabled via the `conf/` configuration on an extant streamalert instance, and then `./manage.py deploy threat_intel_downloader` was run. The resulting terraform files that were generated were visually inspected to make sure they matched the desired output, and the aws console was used to see that the lambda was successfully created and deployed. `./manage.py destroy` was then run and the aws console was used to see that the lambda was destroyed.


ryandeivert:
hey @GarretReece as an FYI I changed the base branch for this to `release-3-1-0` after some discussion internally here. bug fixes + doc updates are still fine against the `release-3-0-0` branch but we're going to freeze it for other improvements / new features

PR #1076, opened 2020-01-09
Docs/general update

jack1902:
to: @chunyong-lin
related to:
resolves:

## Background
Reason for the change

- Documentation was slightly out of sync with what steps are required to deploy StreamAlert
- Outputs no longer required the step of altering the cli, as the choices are dynamically set when invoked

## Changes

* Bumped Terraform version to 12.9 as discussed on slack
* changed git clone command to point to release-3-0-0 
* removed step no longer required when configuring outputs (i noticed this when i started work on a separate branch to implement Microsoft Teams)

## Testing

1. Ran `make html`
2. Opened the documentation in my browser and verified visually that the changes took affect

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -8,14 +8,14 @@ Install Dependencies
 --------------------
 
 1. Install Python 3.7 and `pip <https://pip.pypa.io/en/stable/installing/>`_
-2. Install `Terraform <https://www.terraform.io/intro/getting-started/install.html>`_ v0.11.X:
+2. Install `Terraform <https://www.terraform.io/intro/getting-started/install.html>`_ >= v0.12.9:
 
 .. code-block:: bash
 
   brew install terraform  # MacOS Homebrew
-  terraform --version     # Must be v0.11.X
+  terraform --version     # Must be >= v0.12.9
 
-.. note:: Terraform versions >= 0.12.X are not currently supported.
+.. note:: Terraform versions <= 0.12.8 are not supported.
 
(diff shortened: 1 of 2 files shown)


docs/source/getting-started.rst
```diff
@@ -8,14 +8,14 @@ Install Dependencies
...
+2. Install `Terraform <https://www.terraform.io/intro/getting-started/install.html>`_ >= v0.12.9:
 
 .. code-block:: bash
 
   brew install terraform  # MacOS Homebrew
-  terraform --version     # Must be v0.11.X
+  terraform --version     # Must be >= v0.12.9
 
-.. note:: Terraform versions >= 0.12.X are not currently supported.
+.. note:: Terraform versions <= 0.12.8 are not supported.
```
chunyong-lin: I would say `Terraform versions lower than 0.12 are not supported. Recommend to install terraform version 0.12.9 or up.`

docs/source/getting-started.rst
```diff
@@ -38,7 +38,7 @@ Download StreamAlert
 
 .. code-block:: bash
 
-  git clone --branch stable https://github.com/airbnb/streamalert.git
+  git clone --branch release-3-0-0 https://github.com/airbnb/streamalert.git
```
chunyong-lin: We always recommend to checkout `stable` branch. The release branch is for development only and it will be tagged as stable when we cut the release.

jack1902: ok, let me role this back :)

PR #1077, opened 2020-01-09
[schema] trendmicro:malwareevent

jack1902:
to: @chunyong-lin 
related to:
resolves:

## Background

Reason for the change: Added an additional schema for trendmicro malware events

## Changes

* Added an additional schema
* Added tests for the schema
* renamed default sns topic to be log specific (i ran into errors when listing binaryalert and trendmicro under the same sns topic)

## Testing
* `./manage.py test rules` - This passed all tests
* I have also been using this exact schema in Production workloads with this rule and it has triggered whenever a malware event from Trend came in.


streamalert/classifier/parsers.py
```diff
@@ -241,6 +241,10 @@ def _key_check(cls, record, schema, optionals=None, is_envelope=False):
         if not schema:
             return True
 
+        # Expect the record is a dict. Return False (schema doesn't match) if it is not.
+        if not isinstance(record, dict):
```
ryandeivert: I'd prefer if this, and the tests related to it, go into another PR if that's okay :)

chunyong-lin: It is okay to move it to a separated PR. But we will have to create an separate SNS example topic for `trendmicro` in this PR. Or rebase this PR after the fix merged.
I am okay with both approach. @jack1902 your call.

jack1902: As this will fix both issues the quickest, I'd argue to merge it. Going through the trouble of removing a commit to rebase and carry out other work could introduce issues if things are forgotten. This will be the simplest way of resolving both at the same time

chunyong-lin: Jack and me chatted over slack, we agree to move this to a separated PR! I will open an issue for this as well.

jack1902:
rebased off of release-3-0-0 to include the fix for the `classifier`

Issue #1078, opened 2020-01-10
[output] Microsoft Teams


jack1902:
## Background

NOTE: Before filing this issue, please consider the following:

Have you tried pinging us on Slack?
https://streamalert.herokuapp.com/

Are you on the latest version of StreamAlert? yes, this is a feature i would like for 3-1-0

### Description
Where i work, we use Microsoft Teams for IM. Therefor, i would like the ability to integrate Microsoft Teams as an output for Streamalert. I have a branch with some work commited currently, but i wanted to see if the community would like this feature?

Current work: [jack1902:output/teams](https://github.com/jack1902/streamalert/blob/output/teams/streamalert/alert_processor/outputs/teams.py)

## Desired Change
The ability to use Microsoft Teams (teams) as an output for alerts within StreamAlert

High level overview of the desired change or outcome.
- Alerts can be sent to teams

jack1902:
Current output looks like this (the `...` are just redacted information on the incoming record): 
<img width="1115" alt="Screen Shot 2020-01-10 at 11 34 19" src="https://user-images.githubusercontent.com/39212456/72158596-85310900-33b2-11ea-9c34-561f04d272fd.png">

jack1902:
Closing as work is now merged

PR #1079, opened 2020-01-10
[outputs] add Microsoft Teams as an alerting output

jack1902:
to: @chunyong-lin 
related to: #1078
resolves: #1078

## Background

We use Microsoft Teams instead of Slack, so I wanted to have the ability to use SA and Teams together 😄

## Changes

* Implemented `TeamsOutput` (very very similar to the `SlackOutput`)
* Added Tests
* Updated documentation for outputs.rst

### Code
* I implemented the output and tested it against an Incoming Webhook on our Teams Instance
* wrote tests in a style similar to the `SlackOutput` tests
* Ran tests and now have 100% Code Coverage for this new addition
* Ran `pylint` against files

### Documentation
* ran `make html` and visually checked addition in browser

### DIFF
--- a/conf/lambda.json
+++ b/conf/lambda.json
@@ -44,6 +44,9 @@
         "threshold": 0
       }
     },
+    "third_party_libraries": [
+      "pymsteams"
+    ],
     "timeout": 60,
     "vpc_config": {
       "security_group_ids": [],
--- a/conf/lambda.json
+++ b/conf/lambda.json
@@ -44,9 +44,6 @@
         "threshold": 0
       }
     },
-    "third_party_libraries": [
-      "pymsteams"
-    ],
     "timeout": 60,
     "vpc_config": {
       "security_group_ids": [],
--- a/docs/source/outputs.rst
+++ b/docs/source/outputs.rst
@@ -42,7 +42,7 @@ Adding a new configuration for a currently supported service is handled using ``
     - ``<SERVICE_NAME>`` above should be one of the following supported service identifiers:
       ``aws-cloudwatch-log``, ``aws-firehose``, ``aws-lambda``, ``aws-s3``, ``aws-sns``, ``aws-sqs``,
       ``carbonblack``, ``github``, ``jira``, ``komand``, ``pagerduty``, ``pagerduty-incident``,
-      ``pagerduty-v2``, ``phantom``, ``slack``
+      ``pagerduty-v2``, ``phantom``, ``slack``, ``teams``
 
 For example:
  - ``python manage.py output slack``
--- a/docs/source/outputs.rst
+++ b/docs/source/outputs.rst
@@ -22,6 +22,7 @@ Out of the box, StreamAlert supports:
 * **PagerDuty**
 * **Phantom**
 * **Slack**
+* **Microsoft Teams**
 
 StreamAlert can be extended to support any API. Creating a new output to send alerts to is easily accomplished through inheritance from the ``StreamOutputBase`` class. More on that in the `Adding Support for New Services`_ section below.
 
--- a/streamalert/alert_processor/outputs/teams.py
+++ b/streamalert/alert_processor/outputs/teams.py
@@ -1,0 +1,217 @@
+"""
+Copyright 2017-present, Airbnb Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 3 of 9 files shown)


conf/lambda.json
```diff
@@ -44,6 +44,9 @@
         "threshold": 0
       }
     },
+    "third_party_libraries": [
```
ryandeivert: this is confusing, but this instance of `third_party_libraries` is more for if you have custom stuff that is not part of the 'core' of streamalert. since you're adding this as an output to the core, this should actually go here: https://github.com/airbnb/streamalert/blob/b87f7769c1e93f74f7072d92da1961012f928ecb/streamalert_cli/manage_lambda/package.py#L229

and pin the package name/version here:
https://github.com/airbnb/streamalert/blob/b87f7769c1e93f74f7072d92da1961012f928ecb/streamalert_cli/manage_lambda/package.py#L45

streamalert/alert_processor/outputs/teams.py
```diff
@@ -0,0 +1,217 @@
...
+    @classmethod
+    def get_user_defined_properties(cls):
+        """Get properties that must be assigned by the user when configuring a new Microsoft Teams
+        output.  This should be sensitive or unique information for this use-case that needs
+        to come from the user.
+
+        Every output should return a dict that contains a 'descriptor' with a description of the
+        integration being configured.
+
+        Microsoft Teams also requires a user provided 'webhook' url that is comprised of the Teams
```
Ryxias: `that is composed of the Team's` OR `that comprises the Team's`

conf/lambda.json
jack1902: awesome thanks, i didn't know where to place it, will update now :D

streamalert/alert_processor/outputs/teams.py
```diff
@@ -0,0 +1,217 @@
...
+            ]
+        )
+
+    @classmethod
+    def _format_message(cls, alert, publication, webhook_url):
+        """Format the message to be sent to Teams
+
+        Args:
+            alert (Alert): The alert
+            publication (dict): Alert relevant to the triggered rule
```
chunyong-lin: `webhook_url` should be a string, right? Can you add the information about it here as well?

streamalert/alert_processor/outputs/teams.py
```diff
@@ -0,0 +1,217 @@
...
+
+                    Record:
+                      key   value
+                      key   value
+                      ...
+        """
+        # Presentation defaults
+        default_title = "StreamAlert Rule Triggered: {}".format(alert.rule_name)
+        default_description = alert.rule_description
+        default_color = "E81123"  # Red in Hexstring format
```
chunyong-lin: wow, colorful!!

streamalert/alert_processor/outputs/teams.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Args:
+            alert (Alert): Alert instance which triggered a rule
+            descriptor (str): Output descriptor
+
+        Returns:
+            bool: True if alert was sent successfully, False otherwise
+        """
+        creds = self._load_creds(descriptor)
+        if not creds:
```
chunyong-lin: I would log an error message if no creds provided.

streamalert/alert_processor/outputs/teams.py
jack1902: Good spot, will add this to the `__docstring__'

streamalert/alert_processor/outputs/teams.py
jack1902: I copied the styling from: https://github.com/airbnb/streamalert/blob/b87f7769c1e93f74f7072d92da1961012f928ecb/streamalert/alert_processor/outputs/slack.py#L446

As this returns `True`/`False`

jack1902:
Having been using this in production for a little while, i have realised very quickly that i need to add an `alert` section.

# what does this contain?

- `alert_id` - This is crucial, as i want people to talk about alerts based on their id, to prevent duplication when alerts are sent out to multiple outputs
- `rule_name` - Ensure the rule name is always present

Will update the tests and push this change

docs/source/outputs.rst
```diff
@@ -22,6 +22,7 @@ Out of the box, StreamAlert supports:
 * **PagerDuty**
 * **Phantom**
 * **Slack**
+* **Microsoft Teams**
```
chunyong-lin: Can you also update service type `teams` to line 41 on this page?

jack1902: consider it done :)

Issue #1084, opened 2020-01-14
[Bug][Classifier] TypeError: unhashable type: 'dict' when processing raw event encapsulated in a string


chunyong-lin:
## Background
The PR #1077 surfaces a bug in our Parser that StreamAlert would throw an exception `TypeError: unhashable type: 'dict'` when parsing `TrendMicro` schema because the schema is strange!!!

Okay, the root cause is `TrendMicro` events are a list of dict and encapsulated in string. The parser for this type of events will be json with `json_path` configured in the schema conf file. We will hit the bug if `TrendMicro` events goes to same data source where contains other events won't require `json_path`.

TL;DR, the issue can be reproduced by two approaches.
1. Adding a unit test
```
def test_parse_record_not_dict_mismatch(self):
    """JSONParser - Parse record not in dict type and doesn't match schema"""
    options = {
        'schema': {
            'key': 'string'
        },
        'parser': 'json'
    }
    record_data = "[{\"key\": \"value\"}]"
    parser = JSONParser(options)
    assert_equal(parser.parse(record_data), False)
```
2. Verify via python interpreter
```
python 
>>> set({'a': 1})
{'a'}
>>> set([{'a': 1}])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'dict'
>>>
```

The full traceback is similar to
```
Traceback (most recent call last):
  File "manage.py", line 116, in <module>
    main()
  File "manage.py", line 112, in main
    sys.exit(not cli_runner(options))
  File "/Users/SECRET_DIR_PATH_BALABALA/streamalert_cli/runner.py", line 70, in cli_runner
    result = cmds[args.command](args)
  File "/Users/SECRET_DIR_PATH_BALABALA/streamalert_cli/runner.py", line 126, in <lambda>
    command: lambda opts, cmd=cli_command: cmd.handler(opts, config)
  File "/Users/SECRET_DIR_PATH_BALABALA/streamalert_cli/test/handler.py", line 198, in handler
    result = result and TestRunner(options, config).run()
  File "/Users/SECRET_DIR_PATH_BALABALA/streamalert_cli/test/handler.py", line 372, in run
    classifier_result = self._run_classification(event)
  File "/Users/SECRET_DIR_PATH_BALABALA/streamalert_cli/test/handler.py", line 251, in _run_classification
    return _classifier.run(records=[record])
[...]

PR #1085, opened 2020-01-14
Fix a parser bug when processing raw event encapsulated in a string

jack1902:
to: @chunyong-lin 
related to: 
resolves: #1084

## Background

Please see #1084 

## Testing

* ran unit_test and pylint

### DIFF
--- a/tests/unit/streamalert/classifier/test_parsers_json.py
+++ b/tests/unit/streamalert/classifier/test_parsers_json.py
@@ -1091,3 +1091,35 @@ class TestJSONParser:
         parser = JSONParser(options)
         result = parser._extract_via_json_regex_key(record_data)
         assert_equal(result, False)
+
+    def test_parse_record_not_dict_mismatch(self):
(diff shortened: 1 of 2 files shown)


tests/unit/streamalert/classifier/test_parsers_json.py
```diff
@@ -1091,3 +1091,32 @@ def test_extract_via_json_regex_key_no_key(self):
...
+            },
+            'parser': 'json',
+            'configuration': {
+                'json_path': "[*]"
+            }
+        }
+        record_data = "[{\"key\": \"value\"}]"
+
+        parser = JSONParser(options)
+        assert_equal(parser.parse(record_data), True)
```
ryandeivert: can we also add a check for the actual value of the parsed records here? I want to ensure this is for sure doing what we'd expect.

see: https://github.com/airbnb/streamalert/blob/b6c7960e937e1657c0bca15f5c4068cd66418504/tests/unit/streamalert/classifier/test_parsers_json.py#L88-L89

PR #1088, opened 2020-01-15
move alerts table creation to after the creation of other infrastructure

GarretReece:
to: @ryandeivert 
related to:
resolves:

## Background

Invocation of `./manage.py init` fails with an error when attempting to create the `alerts` table, as the database does not yet exist.

## Changes

* The `alerts` table is created after all other infrastructure

## Testing

* Unit tests all pass. A basic initialization and deployment of a minimal streamalert instance was performed, and the aws console was used to verify that the `alerts` table was created.

### DIFF
--- a/streamalert_cli/terraform/handlers.py
+++ b/streamalert_cli/terraform/handlers.py
@@ -110,13 +110,14 @@ class TerraformInitCommand(CLICommand):
 
         deploy(functions, config)
 
-        # we need to manually create the streamalerts table since terraform does not support this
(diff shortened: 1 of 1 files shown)


jack1902:
Hey @GarretReece Just opened a PR for something that seems similar. The issue i was getting was based around the `rules_staging`. I have a fix which allows me to deploy everytime with no issues. Although you are referring to the `alerts` table which i only had the error pop up when the `rules_engine` errored out

ryandeivert:
I think this has been address via another means

Issue #1083, opened 2020-01-14
[release-3-0-0] Error when running terraform init


jack1902:
## Background

Currently trying to initialize the infrastructure through `terraform` via `./manage.py init` and i am running into the following error

### Description
stacktrace: [error.txt](https://github.com/airbnb/streamalert/files/4060240/error.txt)
Copy and paste the error here for the convenience.
```
Error: Invalid count argument

  on modules/tf_rules_engine/iam.tf line 27, in resource "aws_iam_role_policy" "read_rules_table":
  27:   count  = var.rules_table_arn == "" ? 0 : 1

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.

[ERROR 2020-01-14 16:25:46,960 (streamalert_cli.helpers:68)]: An error occurred while running: terraform apply -refresh=true -auto-approve=false -target=module.classifier_prod_lambda -target=module.alert_merger_lambda -target=module.alert_processor_lambda -target=module.rules_engine_iam -target=module.streamalert_athena -target=module.classifier_prod_iam -target=module.alert_merger_iam -target=module.alert_processor_iam -target=module.rules_engine_lambda
['terraform', 'apply', '-refresh=true', '-auto-approve=false', '-target=module.classifier_prod_lambda', '-target=module.alert_merger_lambda', '-target=module.alert_processor_lambda', '-target=module.rules_engine_iam', '-target=module.streamalert_athena', '-target=module.classifier_prod_iam', '-target=module.alert_merger_iam', '-target=module.alert_processor_iam', '-target=module.rules_engine_lambda']
Traceback (most recent call last):
  File "./manage.py", line 116, in <module>
    main()
  File "./manage.py", line 112, in main
    sys.exit(not cli_runner(options))
  File "/Users/USER/personal/github/streamalert/streamalert_cli/runner.py", line 70, in cli_runner
    result = cmds[args.command](args)
  File "/Users/USER/personal/github/streamalert/streamalert_cli/runner.py", line 126, in <lambda>
    command: lambda opts, cmd=cli_command: cmd.handler(opts, config)
  File "/Users/USER/personal/github/streamalert/streamalert_cli/terraform/handlers.py", line 116, in handler
    create_table('alerts', alerts_bucket, config)
  File "/Users/USER/personal/github/streamalert/streamalert_cli/athena/handler.py", line 367, in create_table
[...]

PR #1089, opened 2020-01-15
[terraform] Implemented a fix for the count error

jack1902:
to: @chunyong-lin 
related to:
resolves: #1083 
# Description

Running `terraform init` failed on `terraform` count

## Changes

* Updated `streamalert_cli` to pass `bool` to `tf_rules_engine` instead of using the arn for the count (`bool` is known before apply)
* Updated test case to include new key

## Testing

* Ran `tests/scripts/unit_tests.sh`
* Ran `tests/scripts/pylint.sh`
* Changed to the `tf_rules_engine` directory and ran `terraform fmt`
* Did a full deploy from scratch and it works!

### DIFF
--- a/streamalert_cli/terraform/rules_engine.py
+++ b/streamalert_cli/terraform/rules_engine.py
@@ -42,6 +42,9 @@ def generate_rules_engine(config):
         'threat_intel_enabled': config.get('threat_intel', {}).get('enabled'),
         'dynamodb_table_name': config.get('threat_intel', {}).get('dynamodb_table_name'),
         'rules_table_arn': '${module.globals.rules_table_arn}',
+        'enable_rule_staging': config['global']['infrastructure']['rule_staging'].get(
+            'enabled', False
+        ),
         'classifier_sqs_queue_arn': '${module.globals.classifier_sqs_queue_arn}',
         'classifier_sqs_sse_kms_key_arn': '${module.globals.classifier_sqs_sse_kms_key_arn}',
         'sqs_record_batch_size': min(config.get('sqs_record_batch_size', 10), 10)
--- a/terraform/modules/tf_rules_engine/iam.tf
+++ b/terraform/modules/tf_rules_engine/iam.tf
@@ -24,14 +24,14 @@ data "aws_iam_policy_document" "read_threat_intel_table" {
 
 // Allow the Rules Engine to read the rules table
 resource "aws_iam_role_policy" "read_rules_table" {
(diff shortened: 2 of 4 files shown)


terraform/modules/tf_rules_engine/iam.tf
```diff
@@ -24,14 +24,14 @@ data "aws_iam_policy_document" "read_threat_intel_table" {
 
 // Allow the Rules Engine to read the rules table
 resource "aws_iam_role_policy" "read_rules_table" {
-  count  = var.rules_table_arn == "" ? 0 : 1
+  count  = var.enable_rule_staging ? 1 : 0
```
chunyong-lin: We probably don't need to define a new flag.  `var.rules_table_arn` is one of the outputs defined in `tf_globals` module and it is default value is empty string `""`
https://github.com/airbnb/streamalert/blob/ff3aa2183c88e6907a794919aeae573d65eee433/terraform/modules/tf_globals/output.tf#L1-L3  

Then this variable is referenced in rules_engine tf code here
https://github.com/airbnb/streamalert/blob/ff3aa2183c88e6907a794919aeae573d65eee433/streamalert_cli/terraform/rules_engine.py#L44

Let me see can I find the root cause.

streamalert_cli/terraform/rules_engine.py
```diff
@@ -42,6 +42,9 @@ def generate_rules_engine(config):
         'threat_intel_enabled': config.get('threat_intel', {}).get('enabled'),
         'dynamodb_table_name': config.get('threat_intel', {}).get('dynamodb_table_name'),
         'rules_table_arn': '${module.globals.rules_table_arn}',
+        'enable_rule_staging': config['global']['infrastructure']['rule_staging'].get(
```
chunyong-lin: For naming consistency, I would use `rule_staging_enabled`. See `threat_intel_enabled` naming at line 42.

ryandeivert: I hate to contradict you here, but I actually like the convention of `enable_*` better.. I don't like the tense-y version of `*_enabled` much

ryandeivert: by tense-y I mean past test like enable**d**

chunyong-lin: cat_bff

jack1902: This is a copy of a variable that already exists in the terraform modules. I could change it but then I need to change it across the board

PR #949, opened 2019-06-17
[bug] fixing logic for cleaning up old tf.json files

geremyCohen:
to: @airbnb/streamalert-maintainers

## Background

System was looking for TF files by wrong extension.  

## Changes

* The file extension for the TF state files is modified from .tf --> .tf.json.

PR #1093, opened 2020-01-16
[terraform] fixed #1048 issue by reverting #1060 and fixed #1047

### LINKED ISSUE #1047: [bug] "unknown module referenced" errors when re-init StreamAlert
## Background
During a new deployment of StreamAlert based on branch `release-3-0-0` to stage environment, re-run `python manage.py init` if first init was failed and it would be interrupted with error messages
```
Error: module 'metric_filters_Classifier_FailedParses_PROD_AGGREGATE': unknown module referenced: classifier_prod_lambda

Error: module 'metric_filters_Classifier_SQSFailedRecords_PROD_AGGREGATE': unknown module referenced: classifier_prod_lambda

Error: module 'metric_filters_Classifier_FirehoseRecordsSent_PROD_AGGREGATE': unknown module referenced: classifier_prod_lambda
(...)
```

### Description
My investigation is that terraform configuration file `terraform/prod.tf.json` for `prod` cluster (in my case) is deleted but it has been saved in terraform state file locally. So terraform complaint the reference is not found.

### Steps to Reproduce
```
python manage.py configure aws_account_id 11112222
python manage.py configure prefix MY_PREFIX
python manage.py init
[If bad luck, the first init failed or force trigger the init to fail using our imagination]
python manage.py init
```
We will see the error messages mentioned above.

## Desired Change
Don't delete terraform configuration files for clusters when re-run `python manage.py terraform init`.
One potential fix is to set `init` to `False` in the handler of `TerraformDestroyCommand` class.
https://github.com/airbnb/streamalert/blob/b9a1b6c642ce320525af73abc33680b67e0535ce/streamalert_cli/terraform/handlers.py#L227

### LINKED ISSUE #1048: [bug] S3 bucket for terraform remote state file got deleted before destroy completed
## Background
When deploy StreamAlert `release-3-0-0` branch to stage environment, I found that the S3 bucket for terraform remote state file got deleted before destroy completed, so that terraform throws error message
```
Failed to save state: failed to upload state: NoSuchBucket: The specified bucket does not exist
	status code: 404 ...
```

### Steps to Reproduce
```
python manage.py destroy
```
## Desired Change
Either delete S3 bucket for terraform backend in the end of resource destroy or terraform only uses local state file and not try to read remote state file again during destroy.

jack1902:
TL;DR fixing #1047 and #1048 

to: @chunyong-lin 
related to: #1060 #1047 #1048 
resolves: #1048 #1047

## Background

The fix introduced in #1060 to fix #1047 caused #1048 . Having deployed numerous times recently, this is how i have managed to remove both issues at the same time.

## Changes

* changing back to `init=True` so that the backend on destroy is configured as local resolves: #1048 
* Changing which files are removed by the `terraform_cli` to include the metric_filters resolves #1047 (this is because the `metric_filters.tf.json`

### Issues this introduces

Whilst this fixes two issues, it does introduce a new one: 

```Error: Error deleting CloudWatch Log Metric Filter: ResourceNotFoundException: The specified log group does not exist.```

Above is a snippet from: [error.txt](https://github.com/airbnb/streamalert/files/4071882/terraform.txt)

This one feels much more related to Terraform itself, rather than anything StreamAlert does though

## Testing

- `tests/scripts/unit_tests.sh`
- `tests/scripts/pylint.sh`
- `tests/scripts/test_the_docs.sh`
- Deployed numerous times, required #1089 to be able to deploy consistently and then ran `./manage.py destroy` which produced the issue above

### DIFF
--- a/streamalert_cli/terraform/generate.py
+++ b/streamalert_cli/terraform/generate.py
@@ -362,17 +362,13 @@ def generate_cluster(config, cluster_name):
     return cluster_dict
 
 
-def cleanup_old_tf_files(config):
-    """Cleanup old .tf files, these are now .tf.json files per Hashicorp best practices"""
-    files_for_removal = set(config.clusters()).union({'athena', 'main'})
+def cleanup_old_tf_files():
+    """
+    Cleanup old *.tf.json files
+    """
     for terraform_file in os.listdir('terraform'):
-        if terraform_file == 'variables.tf':
-            continue
-
         if fnmatch(terraform_file, '*.tf.json'):
-            # Allow to retain misc files in the terraform/ directory
-            if terraform_file.split('.')[0] in files_for_removal:
-                os.remove(os.path.join('terraform', terraform_file))
+            os.remove(os.path.join('terraform', terraform_file))
 
 
 class TerraformGenerateCommand(CLICommand):
@@ -405,7 +401,7 @@ def terraform_generate_handler(config, init=False, check_tf=True, check_creds=Tr
     if check_tf and not terraform_check():
         return False
 
-    cleanup_old_tf_files(config)
+    cleanup_old_tf_files()
 
     # Setup the main.tf.json file
     LOGGER.debug('Generating cluster file: main.tf.json')
--- a/streamalert_cli/terraform/handlers.py
+++ b/streamalert_cli/terraform/handlers.py
@@ -233,7 +233,7 @@ class TerraformDestroyCommand(CLICommand):
         # Migrate back to local state so Terraform can successfully
         # destroy the S3 bucket used by the backend.
         # Do not check for terraform or aws creds again since these were checked above
-        if not terraform_generate_handler(config=config, init=False, check_tf=False,
+        if not terraform_generate_handler(config=config, init=True, check_tf=False,
                                           check_creds=False):
             return False
 

### CODE REVIEW

**Comment by chunyong-lin** [comment]
Thanks @jack1902 !!! Great job on this PR! You pick up where I fail to address in-time and we are following the sun to fix issues!!!
I have deployed StreamAlert (release-3-0-0 branch) in my testing account, I will test it with your fix and keep you updated.

File: streamalert_cli/terraform/generate.py [hunk 1]
```diff
@@ -363,8 +363,13 @@ def generate_cluster(config, cluster_name):
...
 def cleanup_old_tf_files(config):
-    """Cleanup old .tf files, these are now .tf.json files per Hashicorp best practices"""
-    files_for_removal = set(config.clusters()).union({'athena', 'main'})
+    """
+    Cleanup old .tf files, these are now .tf.json files per Hashicorp best practices
+
+    These are the files left behind
+    ['rules_engine.tf.json', 'alert_processor.tf.json', 'alert_merger.tf.json']
+    """
+    files_for_removal = set(config.clusters()).union({'athena', 'main', 'metric_filters'})
```
ryandeivert: I'll be honest - I don't know why we aren't just removing _all_ `*.tf.json` files here. do you know @chunyong-lin ?

ryandeivert: this might be an oversight or something that was never changed when we started moving to more tf.json config files.

chunyong-lin: I also struggle to understand these few lines.

jack1902: So my guess is that when everything was `*.tf`, you wanted to preserve the provider.tf which was being tracked by `git`.

Now that every important terraform file produced is `*.tf.json` we could remove the if statement and see if that works :)?

ryandeivert: yeah - I think I support this and anything that ends in `.tf.json` should be removed. the `providers.tf` file is fairly new (last few months) so that wouldn't explain this

jack1902: consider it shipped!

I've removed the code that wasn't required anymore as well

ryandeivert: @chunyong-lin I think this is good for me - I'll let you have final say!

chunyong-lin: It makes lot of sense! Found the history about `*.tf` vs. `*.tf.json` transition. The reason we are not just removing all `*.tf.json` files was because we used `*.tf` to store terraform configuration files before 2018. During that time, we need to keep some internal/customized `*.tf` file(s). 
Then we switched to use `*.tf.json` files... 
TL;DR, few interesting commit history
* https://github.com/airbnb/streamalert/pull/949/files
* https://github.com/airbnb/streamalert/commit/64611e8865875d35b4c3263a0afd525cc1078be5
* https://github.com/airbnb/streamalert/commit/96fd3c3bdd34dbc1a532e4c4f6b768ffe98a25cb

Issue #1075, opened 2020-01-09
[Feature Request] Dynamic Routing of Alerts to Output


jack1902:
## Background
I would like to configure one output on a rule but have it send to different destinations based on something inside the alert. (Eg, account "a" slack channel a, account "b" slack channel "b")
NOTE: Before filing this issue, please consider the following:

Have you tried pinging us on Slack?
https://streamalert.herokuapp.com/

Are you on the latest version of StreamAlert?

### Description
I would like to write one rule, and have that route the alert to an output based on information within the alert. 
The current way i am thinking of implementing this is to use the context field and have a placeholder output, such as "slack:placeholder". 

1. The rule will use a lookup table (new feature in release 3-0-0) to use the account_id on the record and find out which team owns that account (i currently work with alot of AWS Accounts).
2. Use this team name as the descriptor (i plan to add each team as an output, to store the secrets relevant for that output in the s3 bucket) and carry on the output.

This would be amazing, as using matchers is relevant most of the time but not in a case when you want a rule to trigger regardless of the account, but only notify the team that owns it. (I don't like the idea of notifying people who don't need to be notified)

Ryxias:
This is a great feature idea. We also have the (upcoming) requirement of dynamic routing based upon AWS Accounts.

What do you think of the following 2 solutions:


# Solution 1, `dynamic_output=`
When writing a rule, you can now specify a new `dynamic_outputs=` option:

```
@rule(
  logs=['blah'],
  outputs=['default-output'].
  dynamic_outputs=[dynamic_output_function],
  # ..
)
def blahblahblah(record):
  # ... zzzz

def dynamic_output_function(record, context):
  account_id = record.get('recipientaccountid')
  if account_id = 1234:
    return 'slack:integration1'
  
  return None
```

You can pass `dynamic_outputs` an array of functions. Each function accepts a record & context. It returns a string corresponding to an output, or `None`.

Then in the rules_engine.py code we run this function whenever there is a rule match to generate the output. The implementation sounds pretty easy—we'd just need to change how `RulesEngine._configure_outputs()` is implemented.

# Solution2, overloaded `outputs=`
Same as solution 1, except instead of a separate option, we overload the existing `outputs=` option to respect both strings and functions. Tradeoff here is it may be a bit more confusing to read?

```
@rule(
  logs=['blah'],
  outputs=['slack:default', dynamic_output_function].
  # ..
)
def blahblahblah(record):
  # ... zzzz

[...]

PR #1095, opened 2020-01-21
[core] Added dynamic_outputs to Rule

jack1902:
to: @ryandeivert @Ryxias 
related to:
resolves: #1075 

## Background

Wanted the ability to dynamically configure outputs based on information in the record

## Changes

* Now possible to pass dynamic_outputs to the @rule decorator and have
outputs be dynamically configured based on information in the record.
* Added documentation

## Testing

- Ran `tests/scripts/pylint.sh`
- Ran `tests/scripts/unit_tests.sh`
- Added additional tests to cover the new code
- Ran `tests/scripts/test_the_docs.sh` and verified visually the new documentation changes

### DIFF
--- a/docs/source/dynamic-outputs.rst
+++ b/docs/source/dynamic-outputs.rst
@@ -1,0 +1,129 @@
+Dynamic Outputs
+===============
+
+Prerequisites
+-------------
+
+* Any output assigned must be added with ``python manage.py output``
+* ``functions`` must return ``None``, ``str`` or ``List[str]`` which maps to an output configured with the above.
+* Only pass ``context`` if the ``rule`` sets context.
+
+Overview
+--------
+
+Adds the ability to have custom logic run to define an ``output`` or ``outputs`` based on information within the ``record``.
+For information on supported outputs and how to add support for additional outputs, see `outputs <outputs.html>`_
+
+As can be seen by the examples below, they are easy to configure, but add a very useful feature to StreamAlert. 
+
+- StreamAlert sends to all outputs defined within a rules ``outputs=[]`` and ``dynamic_outputs=[]`` when sending ``Alerts``.
+- It is also possible to pass ``context`` to the ``dynamic_function`` if the ``rule`` sets it.
+
+.. note::
+  Any ``output`` passed must be configured with ``./manage.py output -h``
+
+
+Dynamic Outputs, Simple
+-----------------------
+
+The below code block is considered a simple ``dynamic_output`` function, because the outputs are dynamically configured, but the information used still lives within the code. It also:
+
+ - allows you to maintain a static list of information inside your code
+ - will return the outputs relevant to the team who "own" the account
+ - ``Alerts`` are sent to the ``aws-sns:security`` output aswell as those returned by the function
+
+.. code-block:: python
+
+  from streamalert.shared.rule import rule
+
+  def map_account_to_team(record):
+      teams = {
+        "team_a": {"accounts": ["123", "456", ...], "outputs": ["aws-sns:team_a"]},
+        "team_b": {"accounts": ["789", ...], "outputs": ["aws-sns:team_b", "slack:team_b"]},
+      }
+
+      account_id = record.get('recipientaccountid')
+
+      for team in teams:
+          if account_id in team["accounts"]:
+              return team["outputs"]
+      # None is guarded against by StreamAlert
+
+  @rule(
+    logs=['cloudwatch:events'],
+    req_subkeys={
+      'detail': ['userIdentity', 'eventType']
+    },
+    outputs=["aws-sns:security"],
+    dynamic_outputs=[map_account_to_team]
+  )
+  def cloudtrail_root_account_usage(rec):
+    # Rule logic
+
+
+Dynamic Outputs, With LookupTables
+----------------------------------
+
+With the simple addition of a `lookup-table <lookup-tables.html>`_ you can take a rule like ``cloudtrail_root_account_usage`` and configure it as such:
+
+.. code-block:: python
+
+  from streamalert.shared.rule import rule
+  from streamalert.shared.lookup_tables.core import LookupTables
+
+  def dynamic_output_with_context(record, context): # pass context only if the rule added context
+      account_id = context["account_id"]
+ 
+      return LookupTables.get(
+        'my_lookup_table',
+        'aws-account-owner:{}'.format(account_id), 
+        None
+      ) # potentially returns [aws-sns:team_a]
+
+  @rule(
+    logs=['cloudwatch:events'],
+    outputs=["aws-sns:security],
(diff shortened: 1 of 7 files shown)


streamalert/rules_engine/rules_engine.py
```diff
@@ -213,15 +213,82 @@ def _rule_analysis(self, payload, rule):
...
+            set: unique set of outputs, only required outputs if the rule is staged
+        """
         # Check if the rule is staged and, if so, only use the required alert outputs
         if rule.is_staged(self._rule_table):
             all_outputs = self._required_outputs_set
         else:  # Otherwise, combine the required alert outputs with the ones for this rule
             all_outputs = self._required_outputs_set.union(rule.outputs_set)
+            if rule.dynamic_outputs:
+                dynamic_outputs = self._configure_dynamic_outputs(record, rule)
+                all_outputs = all_outputs.union(dynamic_outputs)
```
Ryxias: Nitpick; I recommend eliminating duplicates from `all_outputs`. Maybe return a Python set from `dynamic_outputs`, or if we use a list, we could dedupe with something like 

```
all_outputs = list(set(all_outputs) | set(dynamic_outputs))
```

or something

streamalert/rules_engine/rules_engine.py
```diff
@@ -213,15 +213,82 @@ def _rule_analysis(self, payload, rule):
...
+        """
+        try:
+            outputs = dynamic_output(*args_list)
+        except Exception:  # pylint: disable=broad-except
+            LOGGER.error(
+                "Exception when calling dynamic_output %s for rule %s",
+                dynamic_output.__name__, rule_name
+            )
+        else:
+            return outputs if isinstance(outputs, list) else [outputs]
```
Ryxias: In the case `dynamic_output()` returns `None`, this will actually return a list of a single `None`, which I don't think we want

streamalert/rules_engine/rules_engine.py
```diff
@@ -213,15 +213,82 @@ def _rule_analysis(self, payload, rule):
...
+            args_list.append(rule.context)
+
+        additional_outputs = []
+        for dynamic_output in rule.dynamic_outputs_set:
+            outputs = cls._call_dynamic_output(dynamic_output, rule.name, args_list)
+            if not outputs:
+                continue
+
+            for output in outputs:
+                if ":" not in output:
```
Ryxias: Minor nitpick: Recommend putting this logic into a function, maybe call it `_is_valid_output_name()` or something, to make it clear why this logic is here.

tests/unit/streamalert/rules_engine/test_rules_engine.py
```diff
@@ -346,6 +349,102 @@ def test_rule_analysis_with_publishers(self):
 
             assert_equal(result is not None, True)
 
+    # --- Tests for _configure_outputs()
```
Ryxias: Nice, great tests 👍

docs/source/dynamic-outputs.rst
```diff
@@ -0,0 +1,87 @@
...
+===============
+
+Overview
+--------
+
+Adds the ability to have custom logic run to define an ``output`` based on information within the ``record``.
+For information on supported outputs and how to add support for additional outputs, see `outputs <outputs.html>`_
+
+As can be seen by the examples below, they are easy to configure, but add a very useful feature to StreamAlert. With the simple addition of a
+`lookup-table <lookup-tables.html>`_ you can take a rule like ``cloudtrail_root_account_usage`` and configure it as such:
```
Ryxias: nitpick 
Using LookupTables alongside Dynamic Outputs is a great (and _intended_) way of using it. However, maybe include LookupTables as a part 2 in the documentation? Because you **have** to use Dynamic Outputs with LookupTables. You can use a simple switch statement too. 

This would be to reduce the (perceived) barrier to entry, maybe increase the chance for users to adopt Dynamic Outputs without thinking _"oh man, I have to configure LookupTables too?"_ Just my 2cents

jack1902: Will update and push tomorrow (running on UK time) but thanks for the feedback!

streamalert/rules_engine/rules_engine.py
jack1902: Great point!

streamalert/rules_engine/rules_engine.py
jack1902: I did struggle with how to add the outputs bit here, wasn't sure what to call it. Since I need to take three sets of outputs and create one unique set (`required_outputs`, `rule.output_set` and `rule.dynamic_outputs`) prehaps there's a cleaner way of combining like you mentioned?

streamalert/rules_engine/rules_engine.py
jack1902: i'll potentially set something like:

## My preferred option
```python
else:
    if outputs:
       # return outputs otherwise just default to None
        return outputs if isinstance(outputs, list) else [outputs]
```

or 

## second option (less clean, alot of return statements)
```python
else:
    if not outputs:
        # Return if outputs are None
        return
    return outputs if isinstance(outputs, list) else [outputs]
```

streamalert/rules_engine/rules_engine.py
Ryxias: If each of the inputs is a `set` you can use a three way `union`

```
set1.union(set2, set3)
```

The thing is I'm pretty sure not all of the three of them are `set`s.

streamalert/rules_engine/rules_engine.py
Ryxias: I'm a fan of the 2nd (I like early returns).

Also, it should return empty array, ya? To be consistent with the advertised return type in the docblock

streamalert/rules_engine/rules_engine.py
jack1902: take another look at how i configured `_configure_outputs` i personally think this new implementation is better as its not trying to do what i had before. Its going through each `output_source` and then adding the outputs to a set. This made testing this much easier!

docs/source/rules.rst
```diff
@@ -252,10 +253,16 @@ but those keys can be nested anywhere in the record structure.
 outputs
 ~~~~~~~
 
-Defines the alert destination if the return value of a rule is ``True``.
+``outputs`` defines the alert destination if the return value of a rule is ``True``.
```
ryandeivert: Please rephrase this (and below) to say "The ``outputs`` keyword argument defines ..."

and "The ``dynamic_outputs`` keyword argument can ...`

streamalert/rules_engine/rules_engine.py
ryandeivert: also a fan of this version - good change :)

Ryxias:
Thank you again @jack1902 for your tireless contributions!

PR #1102, opened 2020-01-27
[rule] Fix cloudtrail_public_resources

jack1902:
related to:
resolves: 

## Background

Lambda error when this rule was evaluated. There is a `.get()` on a `string`. Setting default as `{}` so that `.get()` works

## Changes

* default `''` -> `{}`

## Testing

* Ran `tests/scripts/rule_test.sh`
* Ran `tests/scripts/unit_tests.sh`


0xdabbad00:
Oof. Thanks for finding and fixing @jack1902

chunyong-lin:
Good catch!

PR #1105, opened 2020-01-29
[core] Added trendmicro normalized_types

jack1902:
Signed-off-by: jack1902 <39212456+jack1902@users.noreply.github.com>

to: @chunyong-lin 
related to:
resolves:

## Background

Added normalized types, mainly using the `sourceAccount` within a `dynamic_output` function 

## Changes

* Added `normalized_types` for trendmicro

## Testing

- ran `tests/scripts/unit_tests.sh`

### DIFF
--- a/conf/normalized_types.json
+++ b/conf/normalized_types.json
@@ -243,5 +243,10 @@
       "srcuser",
       "dstuser"
     ]
+  },
+  "trendmicro": {
+    "sourceAccount": [
+      "HostOwnerID"
+    ]
   }
 }


conf/normalized_types.json
```diff
@@ -243,5 +243,13 @@
       "srcuser",
       "dstuser"
     ]
+  },
+  "trendmicro": {
+    "instanceID": [
```
chunyong-lin: Do you want to rename it to `hostId` to be more generic? The key name convention is following Common Event Format white-paper, and you can find [here](https://kc.mcafee.com/resources/sites/MCAFEE/content/live/CORP_KNOWLEDGEBASE/78000/KB78712/en_US/CEF_White_Paper_20100722.pdf).

jack1902: That's fine if we use it for things like cloudtrail to map `instance_id` to hostId too?

chunyong-lin: yeah, sounds good to me.

jack1902: Actually removed the `instanceId`, as this is not something tracked by the other schemas. This information is nested within the records so not sure normalization covers these keys. `sourceAccount` is defined on others so that makes sense to keep

Issue #1104, opened 2020-01-29
[feature] Ability to send one Alert then Merge


jack1902:
## Background

I discussed the below with @Ryxias on slack. Creating this issue to track the idea

### Description

I would like the ability to be able to configure a `rule` so that when it triggers an `Alert` it sends it immediately to the configured outputs, but then proceeds to merge the next set of `Alerts` according to the `merge_by_keys` and `merge_window_mins`. 

### Why?

Simply put, the reason why i would like this feature is to get the security team moving on an `Alert` straight away, without waiting for it to be merged before they get notified. Then get additional `Alerts` that are merged to reduce the noise caused by a `rule`.

## Desired Change

A `rule` such as:

```
@rule(
    logs=['cloudtrail:events', 'cloudwatch:events'],  # Rule applies to these 2 schemas
    merge_by_keys=['useragent'],  # Merge alerts with the same 'useragent' key-value pair
    merge_window_mins=5,  # Merge alerts every 5 minutes
    outputs=['pagerduty:csirt', 'slack:security']  # Send alerts to these 2 outputs
)
def cloudtrail_root_account_usage(record):
    """Page security team for any usage of AWS root account"""
    return (
        get_first_key(record, 'userIdentity', {}).get('type') == 'Root' and
        not get_first_key(record, 'invokedBy') and
        get_first_key(record, 'eventType') != 'AwsServiceEvent'
    )
```

Where to send a `Alert` immediately, but then `merge` according to the settings before sending out a merged `Alert`.

- Also it could be possible to set this as a default behaviour, but have the ability to disable this per `rule`.

chunyong-lin:
Good proposal. To add little history here,  before alert_merger feature implemented, an alert would be generated immediately after the event matched to a rule.
* Old pipeline: events -> giant rules engine (includes classifier) -> alert processor -> outputs
* Current pipeline: event -> classifier -> rules engine -> alert merger -> alert processor -> outputs

So, we probably can just send the event/alert to alert processor if the matching rule doesn't contain merger keys, a.k.a skip alert merger.

jack1902:
# Additional explanation

```python
@rule(
  logs=['your-schema'],
  merge_by_keys=['alpha', 'beta'],
  merge_window_mins=30,
  merge_send_first_alert=True # Send the first Alert Then merge
)
def merged_rule(record):
    return True


record_1 = {
    "alpha": "foo",
    "beta": "bar"
    "other_key": "This record is sent to outputs straight away"
[...]

PR #1106, opened 2020-02-04
[core] Adding ability to send initial alert in merge_group

jack1902:
TL;DR I will add more tests etc but this is just an initial attempt, what do you think?

related to: #1104 
resolves: #1104 

## Background

Wanted the ability to send the first `Alert` when merging alerts. This allows the outputs to be informed immediately, but not flooded (hence merging) without waiting for the merging to take place before outputs are notified.

## Changes

* Added kwargs to `Rule`: 
    `merge_send_initial`: State if the first `Alert` in a `merge_group` should be sent immediately
* Added kwargs to `Alert`: 
    `merge_initial_sent` - track if the initial_alert has been sent
    `merge_send_initial` - Passed in from `Rule` but states if the first `Alert` in a group should be sent immediately
* `alert_merger` now creates a `merged_alert` from this `initial_alert` and updates the `initial_alert context` to state it `relates_to_alert: MERGED_ALERT_ID` This allows for the final merged alert to state it relates to the `alert` you received 30 mins ago (configured with `merge_window_mins`). This is picked up by the `alert_processor` as if its any other alert.

## Testing

- Updated unit_tests but want input on if we should be using mocks here. Since i am reliant on the `alerts_table` to track state i think i should place records into the table and then verify they exist as expected
- Ran `tests/scripts/pylint.sh`

### DIFF
--- a/streamalert/alert_merger/main.py
+++ b/streamalert/alert_merger/main.py
@@ -111,23 +111,30 @@ class AlertMerger:
             alerts (list): List of Alert instances with defined merge configuration.
 
         Returns:
-            list<AlertMergeGroup>: Each returned merge group has the following properties:
-                (1) The oldest alert is older than its merge window (i.e. should be sent now), AND
-                (2) All alerts in the merge group fit within a single merge window, AND
-                (3) All alerts in the merge group have the same values for all of their merge keys.
+            list(tuple(initial_alert, <AlertMergeGroup>): Each tuple has the following properties:
+              initial_alert (<Alert>): This is the first alert within a given merge_group
+              <AlertMergeGroup>:
+                  (1) The oldest alert is older than its merge window (i.e. should be sent now), AND
+                  (2) All alerts in the merge group fit within a single merge window, AND
+                  (3) All alerts in the merge group have the same values for all of their merge
+                      keys.
 
-            Alerts which are too recent to fit in any merge group are excluded from the results.
+            Alerts which are too recent to fit in any merge group are not yet merged, but passed
+            back as the initial_alert.
         """
         merge_groups = []
 
         for alert in sorted(alerts):
             # Iterate over alerts (in order of creation) and try to add them to each merge group.
-            if not any(group.add(alert) for group in merge_groups):
+            if not any(group.add(alert) for _, group in merge_groups):
                 # The alert doesn't fit in any merge group - try creating a new one.
                 if datetime.utcnow() < alert.created + alert.merge_window:
                     # This alert is too recent - no other alerts can be merged. Stop here.
+                    # Ensure the initial alert is present to be sent and merged
+                    merge_groups.append((alert, None))
                     break
-                merge_groups.append(AlertMergeGroup(alert))
+
+                merge_groups.append((alert, AlertMergeGroup(alert)))
 
         return merge_groups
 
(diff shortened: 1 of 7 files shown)


jack1902:
an alternative for testing things which reach out to dynamo and require state (it is already being mocked):

## Add data to the table within aws_mock (`tests/unit/helpers/aws_mocks.py`)

```python
def add_data_to_table(table_name, data):
    """Add data to a table
    Args:
        table_name (str): The name of the table
        data (list): List of dynamo compatible data eg alert.dynamo_record
    """
    table = boto3.resource('dynamodb', region_name='us-east-1').Table(table_name)
    with table.batch_writer() as batch:
        for datum in data:
            batch.put_item(Item=datum)

```
Call this function to add information into the table from within a test, and don't patch the call within `AlertTable`. Then within a test:

```python
alerts = [Alert('valid_alert'), Alert('valid_alert')]

add_data_to_table(_ALERT_TABLE, [alert.dynamo_record for alert in alerts])
self._merger._merge_groups(alerts)

# Add some tests here to pull data from the mocked table and verify it is what expected
```

jack1902:
@Ryxias glad you asked, so essentially if the `merge_send_initial` is set then it will send one alert as if it's a normal alert but will be merged into the merged alert sent later on. This is because i wanted to have this initial alert still part of the merge group but send it as soon as possible before it is merged in order to notify outputs.

As can be seen on the issue referenced, others have done this work outside of SA

Ryxias:
That's sensible; pretty smart to merge the initial alert into the merged payload. 

What about when many alerts come in as a stream? Let's say the merge window is 30 minutes and we get 2 hours of alerts. Will we get:

00:01 — Initial alert 1
00:30 — Merged alert 1 (+ alerts from 00:01 - 00:30)
00:31 — Initial alert 2
01:00 — Merged alert 2 (+ alerts from 00:31 - 01:00)
01:01 — Initial alert 3
... etc

?

jack1902:
@Ryxias This is what i expect to happen with the `merge_groups` as long as the alerts are grouped correctly by the `merge_groups` logic, i am just inserting the initial in there whilst tracking if its been sent across the `alert_merger` and `alert_processor`.

I want to put tests in place for this but feel it more of an integration test rather than a unit test since i need to setup the `alert_table` mock and put an alert in there, merge it and then send it and then run `merge` again to verify it won't get sent again

Ryxias:
I think this is on the right track.

What about when only 1 alert is received? Does it sent 1 (original) alert + 1 (merged with only 1 alert) alert, 30 minutes later?

I'd also like to see how publishers would look like. I'm still really iffy on the fact that 2 different alerts are dispatched, one merged and one unmerged. A possible alternative is to have the original alert be merged (with only 1 entry), so the format of the two are consistent. That way both of the alerts adopt the same format, and there becomes less confusion around what the publishers should do.

jack1902:
@Ryxias yes I agree on the format, it should merge the single alert into a merged alert, so that the format is consistent between the first alert and the merged one later on.

Additionally, I think that in the case of a single alert, the merged alert that would be sent at the end of the `merge_window` should be dropped. Since this alert states `initial` this should be fairly easy to implement

jack1902:
I've updated the issue this would have closed. Due to issues with alerts which the merger pulls not always being the same, it's not easy to track the first alert within a `merge_group`. Without changing how dispatching and merging works I don't see how else to do this feature.

PR #1112, opened 2020-02-05
Vagrant AWS cli and env vars

jack1902:
to: @chunyong-lin 
related to:
resolves:

## Background

Using vagrant on a windows machine, i have AWS creds setup through assumed_roles so have to pass `AWS_*` variables through as they only last for a short period of time

## Changes

* Added `awscli` to `apt-get install` - This fixes the missing `awscli` when connecting to the machine
* Updated the Vagrant file to pass variables through when runninng `vagrant ssh` this includes all the `AWS_*` variables. This required an update of the `sshd_config`

## Testing

- `vagrant provision && vagrant up && vagrant ssh` then on the box i ran `aws s3 ls | grep streamalert` to see a list of s3 buckets on the machine

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -33,9 +33,6 @@ def configure_python(machine, version)
 end
 
 STREAMALERT_SHELL_ENV = %{
-export AWS_DEFAULT_REGION='#{ENV.fetch('SA_AWS_DEFAULT_REGION', 'Your region here!')}'
-export AWS_ACCESS_KEY_ID='#{ENV.fetch('SA_AWS_ACCESS_KEY_ID', 'Your access key ID here!')}'
-export AWS_SECRET_ACCESS_KEY='#{ENV.fetch('SA_AWS_SECRET_ACCESS_KEY', 'Your secret access key here!')}'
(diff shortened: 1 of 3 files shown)


Vagrantfile
```diff
@@ -114,4 +111,11 @@ Vagrant.configure(2) do |config|
     configure_streamalert(py3)
     final_message(py3)
   end
+
+  config.ssh.forward_env = [
+    'AWS_DEFAULT_REGION',
+    'AWS_ACCESS_KEY_ID',
```
chunyong-lin: Is it possible forward `AWS_PROFILE`? I use `AWS_PROFILE` so I don't export my `AWS_ACCESS_KEY_ID ` and `AWS_SECRET_ACCESS_KEY ` to the shell env. But the real credential is saved in ~/.aws/ directory. So I'm not clear is `AWS_PROFILE` forwarding working in the VM.

jack1902: to forward `AWS_PROFILE` we would need to mount your `~/.aws/` into the VM, which i would prefer not to do. I expose my credentials by using [aws-vault](https://github.com/99designs/aws-vault). i then can do things like `aws-vault exec NAME -- vagrant ssh` which will set my env_vars accordingly and support assuming roles

Issue #1086, opened 2020-01-14
[conf] remove need to configure the prefix in multiple places


ryandeivert:
## Background

Currently, there is a command to configure the 'prefix', which replaces a preset value in a handful of places. This can be prone to issues if one of these is missed.

## Desired Change

Store the prefix in _only_ one [place](https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/conf/global.json#L5). Every other place that the prefix is used, like [here](https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/conf/global.json#L4), should be programmatically interpolated instead of being hard-coded. We should still support using a 'custom' (un-prefix, user-provided) value (like the kms key alias, for example) if the user desires.

jack1902:
@ryandeivert can this issue be closed now?

ryandeivert:
fixed with #1114

PR #1114, opened 2020-02-07
[docs/misc] documentation overhaul, config format changes, removing periods from bucket names

ryandeivert:
to: @chunyong-lin / @Ryxias  
resolves: #1086

## Background

On the road up to release of v3, the documentation is wildly out of date and there are other nonsensical issues that need to be addressed.

## Changes

- Major config formatting changes to remove the usage of the `streamalert` module from cluster config (for the most part).. because why TF was that a thing in the first place. "Modules" should be optional, with the `streamalert` module contains the non-optional `classifier_config`. Thus, `classifier_config` has been migrated to a top-level config item in the `<cluster-name>.json` config files.
- All buckets previously hard-coded bucket names are now configurable, with the exception of the `<prefix>-streamalert-secrets` bucket. This is largely because 1) this bucket will be small/easily "migrate-able" and 2) this bucket will be going to hell soon. Fixed buckets:
  - terraform-state (small enough to not need backwards compat for, but support still added)
  - s3-logging
  - athena-results
  - streamalerts
  - streamalert-data
- removal of ridiculous `prefix` formatting trash. The prefix is now stored once and interpolated everywhere it is needed (resolves: #1086)
- complete overhaul of documentation, including an entire restructure/format change, along with actual documentation of config settings (particularly related to global config settings)
  - includes major consistency updates to align with [reST suggestions](https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#headings)
- updating secrets bucket name to remove periods
- updating s3-logging bucket name to remove periods
- updating athena-results bucket name to remove periods
- updating streamalerts bucket name to remove periods
- updating terraform-state bucket name to remove periods
- updating streamalert-data bucket name to remove periods
- fixing misc places that were missed regarding periods in bucket names

### Context on S3 Bucket Name Change

While periods (`.`) are technically allowed in [bucket names](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html), they prevent enabling [S3 Transfer Acceleration](https://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html). We do not currently use the [`acceleration_status`](https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#acceleration_status) setting on our buckets, but we could in the future and should be compliant with this features requirements.

## Testing

Updates to all unit tests in response to above changes.


conf/global.json
```diff
@@ -43,12 +43,12 @@
   },
   "s3_access_logging": {
     "create_bucket": true,
-    "logging_bucket": "PREFIX_GOES_HERE.streamalert.s3-logging"
+    "logging_bucket": "PREFIX_GOES_HERE-streamalert-s3-logging"
```
Ryxias: Random and this may be handled already; according to the Transfer acceleration docs, the S3 bucket names need to confirm to DNS naming conventions, so underscores can't be used right?

ryandeivert: yes underscores are not allowed - do you see somewhere that they're being used?

ryandeivert: I think we're already compliant with that restriction

Ryxias: Ah
I'm stupid
`PREFIX_GOES_HERE` gets replaced

ryandeivert: oh yeah I also want to fix that .. https://github.com/airbnb/streamalert/issues/1086

ryandeivert:
@chunyong-lin I'm not sure if it helps with firehose tbh - that depends how the firehose logic on the AWS backend is implemented

It's worth noting that this only helps if "virtual style requests" are used. See the section at the bottom of [this page](https://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html#transfer-acceleration-requirements).

> Transfer Acceleration is only supported on virtual style requests. For more information about virtual style requests, see Making Requests Using the REST API.

The [boto3](https://boto3.amazonaws.com/v1/documentation/api/1.9.42/guide/s3.html) docs suggest you can set the style to `virtual` or `path`, but `auto` is used (not sure what "auth" really means for us though)

ryandeivert:
I was also thinking this _might_ speed up athena under the hood, but I can't find anything that confirms that at all (so I'm suspecting it would not).

S3 Transfer Acceleration is also significantly more expensive:

image

ryandeivert:
I actually think we should do this. I can't come up with reasons _not_ to. marking ready for review @chunyong-lin

conf/global.json
chunyong-lin: We probably also need to sanity check the prefix, LOL. Maybe someone use period in the prefix then it breaks the S3 Transfer acceleration support.

ryandeivert:
@chunyong-lin I'll be making some config changes to support overriding defaults :)

conf/global.json
ryandeivert: I think we already have something that checks for invalid characters in the prefix (underscores in particular). I'll check for it and updates accordingly

ryandeivert:
okay @chunyong-lin / @Ryxias I know this has ballooned a bit, and taken on a completely different form, but this is now ready for official review

conf/global.json
ryandeivert: this is now configurable

jack1902:
@ryandeivert any possibility to prefix commits with the `[core]` tags etc that are suggested by the `CONTRIBUTING` File?

ryandeivert:
These all will get squashed on merge so there’s really no point...

terraform/modules/tf_flow_logs/variables.tf
```diff
@@ -7,7 +7,7 @@ variable "cluster" {
 }
 
 variable "cloudwatch_logs_destination_arn" {
-  type = "string"
+  type = string
```
jack1902: nice catch 👍

terraform/modules/tf_globals/alerts_firehose/main.tf
```diff
@@ -1,11 +1,15 @@
+locals {
+  stream_name = "${var.prefix}_streamalert_alert_delivery"
+}
+
 // AWS Firehose Stream for Alerts to S3
 resource "aws_kinesis_firehose_delivery_stream" "streamalerts" {
-  name        = "${var.prefix}_streamalert_alert_delivery"
+  name        = local.stream_name
```
jack1902: `locals` are game changers! 👍

terraform/modules/tf_globals/alerts_firehose/main.tf
```diff
@@ -24,3 +28,12 @@ resource "aws_kinesis_firehose_delivery_stream" "streamalerts" {
   }
 }
 
+// CloudWatch Log Group: Firehose
+resource "aws_cloudwatch_log_group" "firehose" {
+  name              = "/aws/kinesisfirehose/${local.stream_name}"
+  retention_in_days = var.cloudwatch_log_retention
+
+  tags = {
+    Name = "StreamAlert"
```
jack1902: should this be `Application` ?

tests/unit/streamalert_cli/terraform/test_generate.py
```diff
@@ -531,7 +559,7 @@ def test_generate_cloudtrail_cloudwatch_logs_and_s3(self):
                 'prefix': 'unit-test',
                 'region': 'us-west-1',
                 's3_bucket_name': 'unit-test-advanced-streamalert-cloudtrail',
-                's3_logging_bucket': 'unit-test.streamalert.s3-logging',
+                's3_logging_bucket': 'unit-test-streamalert-s3-logging',
```
jack1902: could these be pulled out to `CONSTANTS` so that these can be updated in one place in future?

conf/clusters/prod.json
```diff
@@ -48,43 +76,12 @@
...
-        "timeout": 60,
-        "vpc_config": {
-          "security_group_ids": [],
-          "subnet_ids": []
-        }
-      },
-      "enable_threat_intel": false
     }
-  },
-  "region": "us-east-1"
```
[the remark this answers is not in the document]
ryandeivert: as it turns out, this particular `region` value was literally only used by the `kinesis` module. therefore, clusters have never really been "regional" (not sure what the intent of this value was originally anyway, tbh). in any case, I have modified the kinesis module to use the `global` region by default, and allow it to be overridden with a `region` value set within the `kinesis` module itself. see: https://github.com/airbnb/streamalert/pull/1114/files#diff-916958fe1956202328ac3949bb5ea39cL41

docs/source/outputs.rst
```diff
@@ -73,85 +85,96 @@ Adding support for a new service involves five steps:
...
+  that the value should be encrypted and stored in Amazon S3.
+  Default is: ``False``
+
+
+Strategy
+========
 A common strategy that has been found to be effective:
 
-* Write your rule, only list Slack as an output
+* Write your rule and only designate Slack as an output
```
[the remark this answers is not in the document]
ryandeivert: changing this to `Write your rule and only designate a notification-style service, such as Slack, as an output`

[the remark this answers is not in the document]
ryandeivert: slack was suggested here originally for turning since it's very easy to "ignore"

docs/source/testing.rst
```diff
@@ -1,13 +1,14 @@
+#####
 Tests
```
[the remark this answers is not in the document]
ryandeivert: the filename/name in the toc tree is really irrelevant (not actually used anywhere) but I did update this to appease you ;P

docs/source/troubleshooting.rst
```diff
@@ -1,18 +1,19 @@
+###############
 Troubleshooting
-===============
+###############
 
+***************
 Kinesis Streams
```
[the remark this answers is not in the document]
ryandeivert: I've updated these and the other references that inaccurately reference `AWS Kinesis...` (it's `Amazon Kinesis`)

streamalert/shared/config.py
```diff
@@ -255,12 +294,25 @@ def _validate_config(config):
...
+                if config_name in cluster_attrs:
+                    continue
+
+                error = "'{}' is missing in the '{}' cluster".format(config_name, cluster_name)
+
+                modules = cluster_attrs.get('modules', {})
+                old_format = None
+                for key in {'streamalert', 'stream_alert'}:
+                    if key in modules:
+                        old_format = key
```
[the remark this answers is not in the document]
ryandeivert: then it will just error for them twice in a row until they get it right ;)

not going to try to address every edge case when user error is in play

streamalert_cli/athena/handler.py
```diff
@@ -140,8 +142,10 @@ def _add_default_athena_args(athena_parser):
         """Adds the default required arguments for athena subcommands (bucket and table)"""
         athena_parser.add_argument(
             '-b', '--bucket',
-            help='Name of the S3 bucket where log data is located',
-            required=True
+            help=(
+                'Name of the S3 bucket where log data is located. If not supplied, default will '
+                'be "<prefix>-streamalert-data"'
```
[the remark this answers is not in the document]
ryandeivert: very ugly - good find. fixed

terraform/modules/tf_globals/alerts_firehose/main.tf
ryandeivert: we standardized on `Name` here (for whatever reason) so updating here means updating 100+ other places

tests/unit/streamalert_cli/terraform/test_generate.py
ryandeivert: IMHO constants should rarely (never?) be used in unit tests, so going to leave as is

docs/source/outputs.rst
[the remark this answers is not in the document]
jack1902: Same goes for Teams but I understand where your going with this

ryandeivert:
@jack1902 addressed your comments FYI

jack1902:
@ryandeivert is this good to merge or are there additional bits inbound?

streamalert_cli/terraform/kinesis_streams.py
```diff
@@ -35,10 +36,11 @@ def generate_kinesis_streams(cluster_name, cluster_dict, config):
         '{}_{}_streamalert'.format(prefix, cluster_name)
     )
 
-    cluster_dict['module']['kinesis_{}'.format(cluster_name)] = {
+    module_name = 'kinesis_{}'.format(cluster_name)
+    cluster_dict['module'][module_name] = {
         'source': './modules/tf_kinesis_streams',
         'account_id': config['global']['account']['aws_account_id'],
-        'region': config['clusters'][cluster_name]['region'],
+        'region': kinesis_module.get('region', config['global']['account']['region']),
```
chunyong-lin: This change will force all kinesis to be in the same region, although kinesis is per cluster basis. I think we should have `region` setting in cluster conf files, under `kinesis` module.

tests/unit/conf/clusters/advanced.json
```diff
@@ -36,7 +64,12 @@
         "shard_level_metrics": [
           "IncomingBytes"
         ],
-        "shards": 1
+        "shards": 1,
+        "terraform_outputs": [
+          "username",
```
chunyong-lin: I would output `arn` instead of `username`. The outputs we need for kinesis data stream are kinesis data stream name which can get it from the `arn`, `access_key_id` and `secret_key`.

streamalert_cli/terraform/kinesis_streams.py
ryandeivert: that is what this does now - I will update the docs to reflect the region setting in the kinesis module

ryandeivert: I've pushed a commit to address this and remove the region setting from this module. as discussed, this will actually break the lambda event source mapping

tests/unit/conf/clusters/advanced.json
ryandeivert: updated!

docs/source/datasources.rst
```diff
@@ -1,25 +1,26 @@
...
-* AWS Kinesis Streams
-* AWS SNS
+* Amazon Kinesis Data Streams
+* Amazon Simple Notification Service (SNS)
 
 These services above can accept data from:
 
 * Log Forwarding Agents
 * Custom Applications
 * AWS CloudTrail
```
chunyong-lin: Also replace `AWS` with `Amazon` here, since you replace most of them. 🤷‍♀

tests/unit/conf/clusters/advanced.json
```diff
@@ -36,7 +64,12 @@
         "shard_level_metrics": [
           "IncomingBytes"
         ],
-        "shards": 1
+        "shards": 1,
+        "terraform_outputs": [
+          "user_arn",
```
chunyong-lin: I should've checked the outputs in `tf_kinesis_streams` module in my first comment on this line. There is no `user_arn` output. We can output either `arn` or `stream_name`.
https://github.com/airbnb/streamalert/blob/03b7d160b76271aea96804f5a98128c90ab7688e/terraform/modules/tf_kinesis_streams/outputs.tf#L1-L7

ryandeivert: @chunyong-lin - in the same commit as the change you're referencing, I added this:

https://github.com/airbnb/streamalert/pull/1114/commits/b62b0ece2d2bcd4b447e9e12fe35c551cf90fc83#diff-517d65d9de13bb32b0329474e9b11627R13-R15

docs/source/datasources.rst
ryandeivert: sadly, they (Amazon/AWS) use different terminology for different services, and I have no idea why... but, all the changes I made were to match what they refer to them as:

## `AWS CloudTrail`
image



## Versus `Amazon S3`
image

PR #1115, opened 2020-02-10
[core] Support to store data in Parquet format and create Athena tables by Terraform

chunyong-lin:
to: @blakemotl @ryandeivert @Ryxias 
related to: #874 #1143 
resolves:

## Background
PR #874 is to convert logs and alerts data into parquet format to improve historical searching performance against StreamAlert data based on AWS S3 and Athena services. It also removes athena partition lambda function, however, during rebasing, I found out we still need this lambda function to add new partitions to the glue catalog. So I add this lambda function back with proper changes to process data in parquet format. I will initialize a discussion about update partition issue later.

#874 is based on v2 branch which is way too old and the PR was sent a year ago in python2 and terraform v0.11 🤷‍♀ It is very time consuming to rebase that PR and resolve every single conflict. So I kinda copy&pasta the changes in #874 to this new PR based on release-3-1-0 branch and add more features to support backward compatible.

This is a rabbit hole, while digging this hole, I added code to create Athena tables by terraform, yay!

## Changes
* `Copy&pasta` most code related to save classified data in parquet format via Kinesis firehose.
* Keep `athena_partition_refresh` lambda function and update the regex to process data in parquet format and new s3 file prefix.
* Address couple bugs in original PR #874.
  * create `module.streamalert_athena` database in init stage. The database is required for alerts table.
  * use `open_x_json_ser_de` deserializer to be more resilient with log schemas that have nested JSON in kinesis firehose configure. This bug was discovered when an alert was triggered during test in staging environment.
* Add backward compatible support for JSON format, see the discussion in #1143 
* Update `Historical Search` documenation
* When user choose to store alerts/data in `parquet` format, it will also change to use terraform [aws_glue_catalog_table](https://www.terraform.io/docs/providers/aws/r/glue_catalog_table.html) resource to create Athena tables for alerts and all data tables. We are getting more benefit by this change
  * We don't need to re-create table anymore when a new key field added
  * Reduce the complexity when creating Athena tables. Historically, we use Athena DDL query to create tables.
* If user choose to continue using `json` format, the infrastructure remains same and it will automatically create Athena tables when run `python manage.py build`.

## Testing
* Unit test is passed
* Deployed the new code to staging AWS account and feed in osquery data. Test `classifier`, `rules_engine`, `alert processor` and `athena_partition_refresh` lambda function pipelines. Make sure new alerts and osquery data can be searched in Athena.

## TODO
Will add migration instruction in the doc for existing StreamAlert deployment in another PR.


streamalert/shared/__init__.py
```diff
@@ -8,3 +8,5 @@
 THREAT_INTEL_DOWNLOADER_NAME = 'threat_intel_downloader'
 
 CLUSTERED_FUNCTIONS = {CLASSIFIER_FUNCTION_NAME}
+
+STREAMALERT_DATABASE = '{}_streamalert'
```
Ryxias: I would rather this CONST be a function, e.g.

```
def get_streamalert_database_name(prefix):
  return '{}_streamalert'.format(prefix)
```

Ryxias: Or maybe `get_default_streamalert_database_name()`

terraform/modules/tf_globals/alerts_firehose/main.tf
```diff
@@ -1,16 +1,36 @@
...
+        serializer {
+          parquet_ser_de {}
+        }
+      }
+      schema_configuration {
+        database_name = var.alerts_db_name
+        role_arn      = aws_iam_role.firehose.arn
+        table_name    = "alerts"
+      }
+    }
```
Ryxias: Neat.

chunyong-lin:
@Ryxias interesting point. I think it is not a bad idea. But instead of pop-in-per-schema, I am thinking provide choice for historical data format, choice between `json` or `parquet`, a.k.a backward compatible.

The reason I don't want to go `pop-in-per-schema` is it will make the data inconsistent for post process because some s3 prefix contains data in json format, some in parquet. What do you think?

streamalert/shared/utils.py
```diff
@@ -143,3 +143,20 @@ def get_keys(data, search_key, max_matches=-1):
...
     return results
+
+def get_database_name(config):
+    """Get the name of the athena database using the current config settings
+    Args:
+        config (CLIConfig): Loaded StreamAlert config
+    Returns:
+        str: The name of the athena database
+    """
+    prefix = config['global']['account']['prefix']
```
ryandeivert: ha I was just thinking about doing something similar for this, and applied similar logic to other areas. good change

streamalert/shared/utils.py
```diff
@@ -143,3 +143,20 @@ def get_keys(data, search_key, max_matches=-1):
...
+    athena_config = config['lambda'].get('athena_partition_refresh_config')
+    if athena_config:
```
ryandeivert: suggestion:
```python
athena_config = config['lambda'].get('athena_partition_refresh_config', {})
return athena_config.get('database_name', '{}_streamalert'.format(prefix))
```

chunyong-lin:
@blakemotl @ryandeivert @Ryxias This PR is ready for 2nd round of review. 🙇‍♀

streamalert/athena_partition_refresh/main.py
```diff
@@ -55,13 +65,19 @@ def __init__(self):
         config = load_config(include={'lambda.json', 'global.json'})
         prefix = config['global']['account']['prefix']
         athena_config = config['lambda']['athena_partition_refresh_config']
+        store_format = get_data_store_format(config)
+
+        if store_format == 'parquet':
+            self._alerts_regex = self.ALERTS_REGEX_PARQUET
+            self._data_regex = self.DATA_REGEX_PARQUET
+
+        else:
```
Ryxias: Maybe 

```
if = 'parquet':
  ...
elif = 'json':
  ..
else:
  Raise
```

streamalert_cli/athena/handler.py
```diff
@@ -32,10 +32,12 @@
 
 CREATE_TABLE_STATEMENT = ('CREATE EXTERNAL TABLE {table_name} ({schema}) '
                           'PARTITIONED BY (dt string) '
-                          'ROW FORMAT SERDE \'org.openx.data.jsonserde.JsonSerDe\' '
-                          'WITH SERDEPROPERTIES (\'ignore.malformed.json\' = \'true\') '
+                          '{store_format} '
                           'LOCATION \'s3://{bucket}/{table_name}/\'')
+STORE_FORMAT = ('ROW FORMAT SERDE \'org.openx.data.jsonserde.JsonSerDe\' '
```
Ryxias: `STORE_FORMAT_JSON`

streamalert_cli/athena/helpers.py
```diff
@@ -89,11 +97,13 @@ def add_partition_statements(partitions, bucket, table_name):
     yield ' '.join(statement)
 
 
-def logs_schema_to_athena_schema(log_schema):
+def logs_schema_to_athena_schema(log_schema, ddl_statement=True):
```
Ryxias: This is a minor nitpick:

Whenever I have optional parameters that control the "mode" of a function, I prefer to do something like this:


```
def logs_schema_to_athena_schema(log_schema):
  return _internal_function(log_schema, False)

def logs_schema_to_athena_ddl(log_schema):
  return _internal_function(log_schema, True)

# This method is private AND the "optional" statement is not optional, so there is no "default behavior"
def _internal_function(args, ddl_statement):  
  # ....
```

This allows you to avoid using default args, which can be really annoying

Ryxias: It also allows you to search for the named function usages more easily in your codebase

streamalert_cli/athena/helpers.py
```diff
@@ -134,3 +147,98 @@ def unique_values_from_query(query_result):
...
+
+    return formated_schema
+
+def generate_alerts_table_schema():
+    """Generate the schema for alerts table in terraform by using a fake alert
+
+    Returns:
+        athena_schema (dict): Equivalent Athena schema used for generating create table statement
+    """
+    alert = Alert('temp_rule_name', {}, {})
```
Ryxias: 🤔

streamalert_cli/manage_lambda/package.py
```diff
@@ -274,6 +274,7 @@ class AthenaPackage(LambdaPackage):
         'streamalert/shared'
     }
     package_name = 'athena_partition_refresh'
+    package_libs = {'netaddr'}
```
Ryxias: Do you really need `netaddr`? I dont see it used in the partition refresh function

streamalert_cli/terraform/common.py
```diff
@@ -25,6 +25,10 @@ class InvalidClusterName(Exception):
     """Exception for invalid cluster names"""
 
 
+class MisconfigurationError(ValueError):
```
Ryxias: Nice 👍

Ryxias: Maybe put this into `streamalert.shared` or `streamalet.shared.exceptions` or something

streamalert_cli/terraform/firehose.py
```diff
@@ -46,32 +50,35 @@ def generate_firehose(logging_bucket, main_dict, config):
...
+
+    db_name = get_database_name(config)
 
     # Add the Delivery Streams individually
     for log_stream_name, log_type_name in enabled_logs.items():
         module_dict = {
             'source': './modules/tf_kinesis_firehose_delivery_stream',
             'buffer_size': (
-                config['global']['infrastructure']['firehose'].get('buffer_size', 64)
+                firehose_conf.get('buffer_size', 128)
```
Ryxias: Any particular reason to change the default?

Ryxias: Also, (not related to your code) I would like to eventually move away from these inline defaults. I would like more instances eventually where a missing configuration simply raises the `MisconfigurationError` that you added

streamalert_cli/terraform/generate.py
```diff
@@ -591,6 +594,23 @@ def generate_global_lambda_settings(config, config_name, generate_func, tf_tmp_f
...
+        # Raise MisconfigurationError when user doesn't explicitly set `store_format`
+        # in `athena_partition_refresh_config` in conf/lambda.json when upgrade to v3.1.0.
+        store_format = get_data_store_format(config)
+
+        if not store_format or store_format not in ('parquet', 'json'):
+            message = (
+                '[WARNING] '
+                'It is required to explicitly set "store_format" for '
+                'athena_partition_refresh_config in "conf/lambda.json" when upgrading to v3.1.0. '
+                'Availble value is "parquet" and "json". For more information, refer to '
```
Ryxias: **_Available values are "parquet" and "json"._**

terraform/modules/tf_globals/alerts_firehose/main.tf
```diff
@@ -1,28 +1,80 @@
...
+      cloudwatch_logging_options {
+        enabled         = true
+        log_group_name  = aws_cloudwatch_log_group.firehose.name
+        log_stream_name = "S3Delivery"
+      }
+    }
+  }
+
+  // AWS Firehose Stream for Alerts to S3 and saved in Parquet format
+  dynamic "extended_s3_configuration" {
```
Ryxias: I ran into this problem during Normalization v2. Be very careful because changing from `s3` —> `extended_s3` will destroy your existing Firehose. Because Firehose is stateful it can cause a small amount of data loss. Just FYI to consider for the migration process

Ryxias:
Let's try to get this merged and deployed soon.

terraform/modules/tf_globals/alerts_firehose/main.tf
chunyong-lin: Good catch!!!

streamalert/athena_partition_refresh/main.py
chunyong-lin: It is not necessary to have additional check here. I add a new helper to return `None` and trigger `MisConfigurationError` exception if the value neither `parquet` nor `json`.
```
def get_data_store_format(config):
    """Get the data store format using the current config settings
    Args:
        config (CLIConfig): Loaded StreamAlert config
    Returns:
        str: The data store format either "parquet" or "json"
    """
    athena_config = config['lambda'].get('athena_partition_refresh_config', {})

    return athena_config.get('store_format', None)
```

conf/lambda.json
```diff
@@ -52,6 +52,7 @@
   },
   "athena_partition_refresh_config": {
     "concurrency_limit": 10,
+    "store_format": null,
```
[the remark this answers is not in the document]
chunyong-lin: The `store` word comes from Athena create table syntax
```
CREATE [EXTERNAL] TABLE [IF NOT EXISTS]
 [db_name.]table_name [(col_name data_type [COMMENT col_comment] [, ...] )]
 [COMMENT table_comment]
 [PARTITIONED BY (col_name data_type [COMMENT col_comment], ...)]
 [ROW FORMAT row_format]
 [STORED AS file_format] 
 [WITH SERDEPROPERTIES (...)] ]
 [LOCATION 's3_loc']
 [TBLPROPERTIES ( ['has_encrypted_data'='true | false',] ['classification'='aws_glue_classification',] property_name=property_value [, ...] ) ]
```
Parameter `[STORED AS file_format]` description is  `Specifies the file format for table data. If omitted, TEXTFILE is the default. Options for file_format are: ... `
https://docs.aws.amazon.com/athena/latest/ug/create-table.html

I can change it to `file_format`. Thoughts? @Ryxias

streamalert_cli/athena/helpers.py
chunyong-lin: I am sorry I am afraid I will leave this "jankiness" for now. I don't want to have too much changes in one PR.

streamalert_cli/terraform/firehose.py
chunyong-lin: Checked, we have default value `5` in `terraform/modules/tf_kinesis_firehose_delivery_stream/variables.tf` and `128` value in `conf/global.json`. So take Ryan's suggestion, I remove the default value here.

streamalert_cli/athena/helpers.py
chunyong-lin: Personally, I don't want to add two more methods which can be simplified by adding a default argument.

Issue #1118, opened 2020-02-11
[feature] Ability to configure outputs using a .json file


jack1902:
## Background

I would like the ability to add multiple `outputs` to StreamAlert in one go with a `.json` file instead of having to continuously  use `./manage.py output SERVICE`

### Description

Have the ability to configure numerous `outputs` via a `.json` file.

- this might require a re-write of how `./manage.py output` is written so that `SERVICE` becomes optional. This would allow the following options (Assuming #1117 is merged):

- `--edit` - Allow pre-existing outputs to be merged
- `--service SERVICE` This makes SERVICE optional, as currently it is always required
- `--file FILE_NAME.json` the name of the file to read output configuration options to save
- `--generate_skeleton` Would create a `.json` file like the below depending on `--service` being passed to create a skeleton for that one service or create a skeleton for all `outputs` in StreamAlert.

#### Example skeleton (`--generate_skeleton --service aws-ses`)
```json
{
    "aws-ses": [
        {"descriptor": "", "from_email": "", "to_emails": ""}
    ]
}
```

This would use the returned value for each outputs `get_user_defined_properties` method to generate the skeleton relevant to that `output`. This would be generated using the command `./manage.py output SERVICE --generate_skeleton` and save it to a file. 

- It should also be possible to not specify a service so that you can configure numerous outputs across numerous services in one go

#### Example (for use with `--file`)
```json
{
    "aws-ses": [
        {"descriptor": "example", "from_email": "alerts@test.com", "to_emails": "example@test.com"}
    ]
}
```

The above will configure an `output` for the service `aws-ses` using the values within the list. Each element of the list will be an `output` to configure for the service.

## Desired Change

- Ability to configure multiple `outputs` for multiple services using a `.json` file.
- Ability to generate a skeleton for all `outputs` or a specific `output` using a command like `./manage.py output --generate_skeleton` or `./manage.py output aws-ses --generate_skeleton` 
- Assuming #1117  were to be merged, this feature should continue to work for `.json` files

jack1902:
There is the option of breaking `outputs` out into additional sub-commands so that you can do things like:

- `./manage.py outputs configure ...` - would support `--file`, `--generate_skeleton` and `--service`
- `./manage.py outputs update ...` - would support `--file`, `--generate_skeleton` and `--service` (update is used to update existing outputs, instead of `--edit`)
[...]

Issue #940, opened 2019-05-20
[improvement] Add ability to over-write existing outputs


0xdabbad00:
## Background

Mentioned on Slack: https://streamalert.slack.com/archives/C3BHE2Z0S/p1558386389107200
Are you on the latest version of StreamAlert? Yes

### Description

If you configure a Slack webhook, and want to change it, then trying to rerun `./manage.py output slack` with the same name will result in an error.  Currently, you have to manually edit `conf/outputs.json` to remove the output and then retry that command.  It would be nice to have a way to over-write or delete the existing output.

### Steps to Reproduce

Run `./manage.py output slack` and configure an output.
Rerun the same command with the same output name and a new webhook URL and you'll get an error.

## Desired Change

Provide a `--override` to delete the current output and add the new one.

Ryxias:
👍

VinodKumarLogan:
Hi @0xdabbad00 @Ryxias , I'm interested in working on this feature. Can I please work on it and submit a PR?

jack1902:
@0xdabbad00 Will add a PR for this feature for `release-3-1-0`

0xdabbad00:
@jack1902 I'm not sure if you meant to say that you will be adding this feature, but just to clarify, I unfortunately do not have the time to add it.

jack1902:
@0xdabbad00 yeah sorry for the confusion. Coffee hadn't kicked in

jack1902:
@0xdabbad00 I believe this issue can be closed at the `output` command has now got some new features :)

0xdabbad00:
Thanks @jack1902 !

PR #1120, opened 2020-02-12
[cli] Added additional commands to outputs

jack1902:
to: @0xdabbad00 @chunyong-lin @ryandeivert 
related to: #1118 #940
resolves: #1118 #940

## Background

Wanted the ability to do the following when configuring `outputs` via the CLI:
- set a single output
- set multiple outputs via a file
- get an outputs configuration
- update previously configured url (e.g. add an additional email to `aws-ses`)

## Changes

Added the following subcommands to `./mannage.py outputs`:
- `set` - Set one output, using user_input (pass `--update` to overwrite
existing outputs)

- `get` - Get configured outputs for a service (includes creds) optionally pass `--decriptors` to only pull certain descriptor secrets for the service

- `set-from-file` - Set numerous outputs via a json file. Can set multiple services and descriptors (pass `--update` to overwrite existing outputs)

- `generate-skeleton` - Use to create a skeleton JSON file to be used with `set-from-file`

- `list` - lists out currently configured outputs according to the `conf/outputs.json` file

### QOL changes

* Moved the core_logic of `user_input` to a seperate method so that i can re-use it in cases where the user doesn't need to be asked for input on the command line

* Made it possible to pass `help` to the `generate_subparser` function. This allows for `help` messages to be passed through to argparse. This then works very nicely with commands like `./manage.py outputs -h` and `./manage.py outputs get -h`

* Made the log message in `output_exists` optional, as i didn't want to have it logged in some cases

## Testing

* Ran `./tests/scripts/pylint.sh` - Fixed errors and set `# pylint: disable=protected-access` on one function as previously, this method was never used outside of the `alert_processor`
* Ran `./tests/scripts/test_the_docs.sh` - I understand that documentation requires some love but want opinions first before commiting it to docs

#### FYI

As an FYI, for the `streamalert_cli` a really cool plugin for `sphinx` is called `sphinx-argparse` which can generate documentation for the cli by using argparse. I think this would be a very useful addition to the documentation overall and prevent the CLI falling behind. A cool example of `sphinx-argparse` in use: https://rdk.readthedocs.io/en/latest/references.html

### DIFF
--- a/streamalert_cli/outputs/handler.py
+++ b/streamalert_cli/outputs/handler.py
@@ -54,19 +54,20 @@ class OutputCommand(CLICommand):
             'set': OutputSetSubCommand,
             'set-from-file': OutputSetFromFileSubCommand,
             'generate-skeleton': OutputGenerateSkeletonSubCommand,
-            'get': OutputGetSubCommand
+            'get': OutputGetSubCommand,
+            'list': OutputListSubCommand
         }
 
 
 class OutputSharedMethods:
     @staticmethod
-    def save_credentials(service, config, props):
+    def save_credentials(service, config, properties):
         """Save credentials for the provided service
 
         Args:
             service (str): The name of the service the output belongs too
             config (StreamAlert.config): The configuration of StreamAlert
-            props (OrderedDict): Contains various OutputProperty items
+            properties (OrderedDict): Contains various OutputProperty items
         Returns:
             bool: False if errors occurred, True otherwise
         """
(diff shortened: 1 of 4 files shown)


streamalert_cli/outputs/handler.py
```diff
@@ -13,43 +13,60 @@
...
+
+        LOGGER.error('Unhandled output subcommand %s', options.subcommand)
+
+    @staticmethod
+    def _subcommands():
+        return {
+            'set': OutputSetSubCommand,
+            'set-from-file': OutputSetFromFileSubCommand,
+            'generate-skeleton': OutputGenerateSkeletonSubCommand,
+            'get': OutputGetSubCommand
```
chunyong-lin: It would be helpful if you can add `list` command to list current existing outputs. Just output names. So users don't need to dig into outputs.json file.

streamalert_cli/outputs/handler.py
```diff
@@ -13,43 +13,60 @@
...
+            return subcommands[options.subcommand].handler(options, config)
+
+        LOGGER.error('Unhandled output subcommand %s', options.subcommand)
+
+    @staticmethod
+    def _subcommands():
+        return {
+            'set': OutputSetSubCommand,
+            'set-from-file': OutputSetFromFileSubCommand,
+            'generate-skeleton': OutputGenerateSkeletonSubCommand,
```
chunyong-lin: This great!! +1000000

Ryxias:
This looks very cool. Is it ready for review?

streamalert_cli/outputs/handler.py
```diff
@@ -13,43 +13,60 @@
...
+
+class OutputSharedMethods:
+    @staticmethod
+    def save_credentials(service, config, props):
+        """Save credentials for the provided service
 
+        Args:
+            service (str): The name of the service the output belongs too
+            config (StreamAlert.config): The configuration of StreamAlert
+            props (OrderedDict): Contains various OutputProperty items
```
chunyong-lin: Totally optional, I kinda like the variable name to be more descriptive. I know it will be longer, can we use `properties` or `params`?

streamalert_cli/outputs/handler.py
```diff
@@ -78,25 +180,271 @@ def handler(cls, options, config):
...
+    @classmethod
+    def handler(cls, options, config):
+        """Configure multiple outputs for multiple services
+        Args:
+            options (argparse.Namespace): Basically a namedtuple with the service setting
+            config (StreamAlert.config): The configuration of StreamAlert
+        Returns:
+            bool: False if errors occurred, True otherwise
+        """
+        with open(options.file, "r") as json_file_fp:
```
chunyong-lin: Use single quote `'r'` to be consistent.

streamalert_cli/outputs/handler.py
```diff
@@ -78,25 +180,271 @@ def handler(cls, options, config):
...
+            file_contents = json.load(json_file_fp)
```
chunyong-lin: Probably we should do exception handling here because the file may have syntax error or not a json file. Exit if the file is not a json format with error message.

jack1902:
@Ryxias yes for the code itself, I need to add documentation hence the draft PR. Happy for feedback on the code itself though

streamalert_cli/outputs/handler.py
jack1902: True! Will add :D

streamalert_cli/outputs/handler.py
jack1902: Sorry, force of habit as I use `black` all the time which uses these. Will update

streamalert_cli/outputs/handler.py
jack1902: This would make a useful helper method that's for sure!

jack1902:
@chunyong-lin in regards to adding unit tests for the cli. I'm hesitant to add to it as i would prefer to redo the cli from scratch so it is consistent and can be used to complement the documentation using `sphinx-argparse` I've identified a few bugs here and there which feel easier to resolve with a re-write

jack1902:
Having spoken to @ryandeivert on slack. I am closing this PR to raise a new one against `release-3-0-0` as this would be extremely useful to have in `release-3-0-0`

jack1902:
This was closed in favour of #1138

PR #1122, opened 2020-02-12
Fix bug with default value for firehose use_prefix

Ryxias:

## Background

If you don't provide an explicit value for `use_prefix`, the default values do not match between the Classifier Terraform and the Python firehose client.


Ryxias:
@here I rebased it; it's ready for review

ryandeivert:
lgtm thank youuu

PR #1123, opened 2020-02-12
[core] Moved the secret_store from S3 to SSM

jack1902:
to: @ryandeivert @Ryxias 
related to: #240
resolves: #240

## Background

Using S3 as a `secret_store` is no longer desirable. Using SSM Parameter store makes visibility into what secrets are set much easier and requires less setup by StreamAlert

## Changes

* Implemented SSMDriver (a bit of `copy&pasta` from the S3Driver)
* Updated unit_tests so they pass when using this new driver
* Removed secrets bucket creation
* Added `ssm:GetParameter` permissions to the `alert_processor`

## Testing

* ran `./tests/scripts/pylint.sh` - currently there are warnings (i aim to clean this code up but want to ensure this initial commit is good)
* ran `./tests/scripts/unit_tests.sh` - All created tests now pass, i need to add some for SSMDriver

### DIFF
--- a/scripts/migrate_from_s3_to_ssm.py
+++ b/scripts/migrate_from_s3_to_ssm.py
@@ -1,0 +1,158 @@
+#!/usr/bin/env python
+import sys
+import json
+from io import BytesIO
+import boto3
+from botocore.exceptions import ClientError
+
+FILE_NAME = 'ssm_ready_secrets.json'
+
+
+def decrypt_secrets(ciphertext, region):
+    """Decrypts the given ciphertext using AWS KMS
+
+    Args:
+        ciphertext (str): The raw, encrypted data to be decrypted
+        region (str): AWS region
+
+    Returns:
+        string: The decrypted plaintext
+
+    Raises:
+        ClientError
+    """
+    client = boto3.client('kms', region_name=region)
+    try:
+        response = client.decrypt(CiphertextBlob=ciphertext)
+    except ClientError:
+        print('An error occurred during KMS decryption')
+        raise
+    else:
+        return response['Plaintext']
+
+
+def pull_from_s3(bucket, key):
+    """Pull file contents from S3 into memory
+
+    Args:
+        bucket (boto3.resources.factory.s3.Bucket): The S3 Bucket resource
+        key (str): The key to pull
+
+    Returns:
+        None: An error occured
+        ciphertext: The encrypted contents of the file
+
+    Raises:
+        ClientError
+    """
+    io = BytesIO()
+
+    try:
+        bucket.download_fileobj(key, io)
+    except ClientError:
+        print(f"Error fetching {key} from s3")
+        raise
+    else:
+        return io.getvalue()
+
+
+def get_output_secrets(bucket, key, region):
+    """Pull the secrets from s3
+
+    Args:
+        bucket (boto3.resources.factory.s3.Bucket): The S3 Bucket resource
+        key (str): The key to download and decrypt
+    Returns:
+        None: An error occured
+        dict: The secrets from s3
+    """
+    try:
+        encrypted_secrets = pull_from_s3(bucket, key)
+        secrets = decrypt_secrets(encrypted_secrets, region)
+    except ClientError:
+        pass
+    else:
+        return json.loads(secrets)
+
+
+def save_to_fs(secrets):
+    """Save the secrets to the local FileSystem
+
+    Args:
+        secrets (dict): Secrets ready to be taken by set-from-file subcommand
+    """
+
+    saved = False
+
+    try:
+        with open(FILE_NAME, "w") as _fp:
+            json.dump(secrets, _fp)
+    except Exception:
+        print("Error saving secrets to FileSystem")
+    else:
+        print(f"Secrets saved to {FILE_NAME}")
+        saved = True
+
+    return saved
+
+
+if __name__ == "__main__":
+    import os
+    sys.path.append(os.path.realpath('.'))
+
+    from streamalert.alert_processor.outputs.output_base import (
+        StreamAlertOutput
+    )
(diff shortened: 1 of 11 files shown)


Ryxias:
The code looks solid. 

How would the migration process look? Is there going to be a script? You might need to keep some of the code around to be able to pull from S3 and save into SSM

jack1902:
@Ryxias You raise an interesting point. How should migration scripts be handled. I personally prefer creating a one time script which contains EVERYTHING.

This removes redundant code from the code base and allows for migration scripts to be dropped in future PRs.

What are your thoughts on this? The code removed to pull from S3 can easily be put into a one time script and create a local .JSON file for use with the seperate PR to `set-from-file`

ryandeivert:
we could have a subcommand of the `outputs` cli command temporarily that does this, with the caveat (`TODO`) that this will be going away in a future release. I could be convinced otherwise, though

jack1902:
> we could have a subcommand of the `outputs` cli command temporarily that does this, with the caveat (`TODO`) that this will be going away in a future release. I could be convinced otherwise, though

Consider a script in `scripts/` its stand alone ready to be removed. The reason this is better is because the script will currently be on the same release as the move from s3 to ssm. We could move this commit to `release-3-0-0` and then have a caveat on moving from `release-3-0-0` to `release-3-1-0` state, run this script?

jack1902:
Also, @ryandeivert as i can see you have a HUGE PR open currently, that makes a TONNE of changes to the documentation. i would prefer to create a seperate issue to "update" the documentation to reflect these changes after `release-3-1-0` is rebased off of `release-3-0-0` otherwise we will hit merge conflict city and no one wants that nightmare

streamalert/shared/helpers/aws_api_client.py
```diff
@@ -182,3 +183,71 @@ def download_fileobj(file_handle, bucket, key, region):
...
+            region (str): AWS region
+            with_decryption (bool): Should decryption be attempted via this call
+
+        Returns:
+            str: The parameter either encrypted or not
+
+        Raises:
+            ClientError
+        """
+        client = boto3.client('ssm', config=default_config(region=region))
```
Ryxias: Some boto3 clients maintain some sort of "internal state". For example, the `athena` one (I think?) performs API calls when you first instantiate it. Does SSM use this?

Alternative pattern with this class could be something like:

```
class AwsSsmClient:

  def __init__(config):
    self._client = boto3.client('ssm', config)

  def get_parameter(parameter_name, with_decryption=True):
    # ...
```

scripts/migrate_from_s3_to_ssm.py
```diff
@@ -0,0 +1,158 @@
...
+        print("Error saving secrets to FileSystem")
+    else:
+        print(f"Secrets saved to {FILE_NAME}")
+        saved = True
+
+    return saved
+
+
+if __name__ == "__main__":
+    import os
```
Ryxias: Is there any particular reason the `imports` are here instead of at the top of the file?

scripts/migrate_from_s3_to_ssm.py
```diff
@@ -0,0 +1,158 @@
...
+    config = CLIConfig()
+    output_config = config.get("outputs")
+    if not output_config:
+        print("No Outputs configured")
+        sys.exit(1)
+
+    region = config["global"]["account"]["region"]
+    prefix = config["global"]["account"]["prefix"]
+
+    bucket_name = f"{prefix}.streamalert.secrets"
```
Ryxias: Subtle thing: @ryandeivert is working on changing the default bucket to include dashes instead of periods. If that PR goes into 3.1.0 you may have to change this to pull from config instead of hardcoding here

scripts/migrate_from_s3_to_ssm.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if service_outputs:
+            # Add the service secrets into the secrets dict
+            secrets[service] = service_outputs
+
+    if not save_to_fs(secrets):
+        # An error occured so exit with code 1
+        sys.exit(1)
+
+    print(
```
Ryxias: This is nice 👍

scripts/migrate_from_s3_to_ssm.py
jack1902: Just cleanliness and the fact these are only relevant if name... Since I have to append streamalert to path in order to import it due to the location of the script

scripts/migrate_from_s3_to_ssm.py
jack1902: Nice and easy to change if this is the case

streamalert/shared/helpers/aws_api_client.py
jack1902: Since this is a one time throw away script I'd argue this is good enough as is. Since after the migration is complete this can be thrown away

jack1902:
I'll pull down release-3-1-0 tomorrow and rebase this work and resolve conflicts

jack1902:
@Ryxias The migration script wouldn't be required if we were to have this in `release-3-1-0` as the migration from 2 -> 3 essentially means starting a fresh. I am happy to close this PR and re-open it against `release-3-0-0` just wanting to confirm if this will be ok @ryandeivert and @Ryxias ?

streamalert/shared/helpers/aws_api_client.py
jack1902: The previous one had this pattern since everywhere it was called was via the `staticmethod`. i could instantiate this Class within the callers init so it is cached etc between invocations but the parameter/secret that is pulled back is already cached so this seems like complexity for complexities sake

jack1902:
Closed in favour of #1142

Issue #1124, opened 2020-02-13
[bug] dynamic_outputs and publishers not playing nicely


jack1902:
## Background

A rule with `dynamic_outputs` and `publishers` currently end up with no `publishers` due to how the `publishers` are added to an `Alert`

### Description

A basic rule configuration:
```python
from streamalert.shared.rule import rule

def get_outputs(record):
    return "slack:test"

@rule(
  logs=['cloudwatch:events'],
  dynamic_outputs=[get_outputs],
  publishers={"slack": basic_publisher}
)
def anything_cloudwatch(*_):
    return True
```

With the above `rule` it has `publishers` configured. The expected behavior is for the `dynamic_outputs` function to configure the `outputs` for the `Alert` and then these `outputs` will have the `publisher` run against them.

What actually happens:
- `dynamic_outputs` - Configures the `outputs` to be passed to the `Alert` Class
- `RulesEngine._configure_publishers()` - uses the `outputs_set` present on the `rule`. the above doesn't touch this, hence why the `publishers` are left blank

### Steps to Reproduce

- Create a rule similar to the above
- Place `breakpoint()` into `Alert.__init__`
- run `./manage.py test live -f SINGLE_EVENT` which triggers the rule
- Observe the `kwargs` being passed. `publishers` will be `{}`

## Desired Change

This doesn't happen (Will push a PR to fix, just documenting the issue here)

jack1902:
Closing as fix has been merged

PR #1125, opened 2020-02-13
[core] Fixed bug with dynamic_outputs and publishers

jack1902:
to: @Ryxias
related to: #1124
resolves: #1124

## Background

`publishers` and `outputs` weren't being correctly configured

## Changes

* outputs are generated and then passed into `_configure_publishers` instead of being used on the `rule`
* Updated unit_tests to reflect this

## Testing

- ran `tests/scripts/pylint.sh`
- ran `tests/scripts/unit_tests.sh` - Fixed errors that occurred due to the change

### DIFF
--- a/streamalert/rules_engine/rules_engine.py
+++ b/streamalert/rules_engine/rules_engine.py
@@ -192,15 +192,18 @@ class RulesEngine:
         if not rule_result:
             return
 
+        # Define the outputs
+        outputs = self._configure_outputs(payload['record'], rule)
+
         alert = Alert(
-            rule.name, payload['record'], self._configure_outputs(payload['record'], rule),
+            rule.name, payload['record'], outputs,
             cluster=payload['cluster'],
             context=rule.context,
             log_source=payload['log_schema_type'],
             log_type=payload['data_type'],
             merge_by_keys=rule.merge_by_keys,
             merge_window=timedelta(minutes=rule.merge_window_mins),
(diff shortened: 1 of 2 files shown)


streamalert/rules_engine/rules_engine.py
```diff
@@ -346,18 +349,22 @@ def _configure_publishers(cls, rule):
 
         Args:
             rule (Rule): The rule to create publishers for
+            requested_outputs (set): A set containing the outputs
 
         Returns:
             dict: Maps string outputs names to lists of strings of their fully qualified publishers
         """
-        requested_outputs = rule.outputs_set
```
Ryxias: Neat, good catch

streamalert/rules_engine/rules_engine.py
```diff
@@ -346,18 +349,22 @@ def _configure_publishers(cls, rule):
...
         requested_publishers = rule.publishers
         if not requested_publishers:
             return None
 
         configured_publishers = {}
 
         for output in requested_outputs:
+            if output == "aws-firehose:alerts":
```
Ryxias: I actually think you can omit this. That way no hard coded "skipped" outputs that we have to hunt down

jack1902: This only skips on the `publisher`, since this change, impacted the publishers there were thrown back. `aws-firehose:alerts` came back with `[]` which seemed pointless as it didn't need a `publisher`

PR #1127, opened 2020-02-13
[rules] Add community rule to detect ssh login activity based on osquery events

chunyong-lin:
to: @ryandeivert @Ryxias 
related to:
resolves:

## Background
This is a contribution to StreamAlert community rules. StreamAlert can process osquery events, and I am surprised we don't have community rules based on osquery events available. This rule is to detect ssh login activity to a linux host based on [osquery last table](https://osquery.io/schema/4.1.2#last).

This rule also assumes we use default osquery pack shipped with osquery package located at `/usr/share/osquery/packs/incident-response.conf` on the linux host. 

## Changes

* Add a community rule with two matchers for osquery events.
* Add two testing events.
* ~Per @ryandeivert 's request, this PR also move the matchers out of  rules' scope.~ Will send a separate PR to address this request since we will also need to pack rule_engine with matchers in new location.

## Testing
Add two testing events and the rule testing is passed.


blakemotl:
@chunyong-lin I have just one comment. Can you add a test with event type login set to something other than 7?

chunyong-lin:
@jack1902 very good point! I checked and you are totally right. We haven't packed the `matchers` with `rules_engine` yet. I will remove the matcher change from this PR to a new PR and address the packaging issue in `rules_engine` as well.
https://github.com/airbnb/streamalert/blob/282ec1bf8590840888097bec8bf551b8b562f210/streamalert_cli/manage_lambda/package.py#L202-L215

chunyong-lin:
@blakemotl done!

PR #1128, opened 2020-02-13
Ports over internal StreamQuery code; integrates into manage.py

Ryxias:
to: @ryandeivert @blakemotl @chunyong-lin 
cc: @airbnb/streamalert-maintainers
addresses: #1116 

## Background
After 8 months of being an internal-only feature, we're happy to finally open source StreamAlert scheduled queries, originally dubbed "StreamQuery".

This feature is inspired by scheduled queries in other historic searching products like Kibana, SumoLogic, etc. The big idea:

> Run SQL statements on a schedule, on data that drops in Athena, and pass the results back to StreamAlert for automated analysis.

You can more on how to configure it (etc) [in the documentation](https://github.com/airbnb/streamalert/blob/a0ede76bcf39165bc1d6d6819951944187485c7d/docs/source/scheduled-queries.rst).

PR #1130, opened 2020-02-19
Refactors StreamQuery a little to "hide" the ServiceContainer

Ryxias:
to: @ryandeivert @blakemotl @chunyong-lin 

## Background

Was following up PR #1128 with some refactoring

## Changes

* The `ServiceContainer` is a controversial point which is rather hard to refactor out of StreamQuery at this time. I decided to "hide it" with a layer of [David Wheeler's dreaded *indirection*](https://en.wikipedia.org/wiki/Indirection). Guess his aphorism holds true. This new layer is the [`ApplicationServices` class](https://github.com/airbnb/streamalert/pull/1130/files#diff-c7c50c03b935b677633e8db443006aaeR40-R67), which is intended to hide the Service Container and its usage through the services.py file
* I greatly expanded the doc blocks available in the Lambda function handler. I think this was important to make it easier to figure out "how" StreamQuery works
* Removed `FileWritingStateManager` as we no longer use that
* Improved test coverage

## Why is the ServiceContainer hard to refactor?

The ServiceContainer implements a classical [Inversion of Control container](https://www.martinfowler.com/articles/injection.html), and one of the extremely key elements to it is singleton services, particularly the `StateManager`.

Since it's not really Pythonic to do IoC, the **_right_** way to refactor this is to extract the `StateManager` into an object, not a service, and to pass it around as the application state.

This isn't a _huge_ refactor but it's probably a bit too risky for now. I'll do it later

## Testing

Deployed it to staging and observed that it's working properly

### DIFF
--- a/streamalert/scheduled_queries/config/services.py
+++ b/streamalert/scheduled_queries/config/services.py
@@ -88,51 +88,51 @@ def configure_container(container):
     container.register(ServiceDefinition('boto3_kinesis_client', _make_boto3_kinesis_client))
 
 
-def _make_command_processor(_container):
+def _make_command_processor(container):
     return CommandProcessor(
-        logger=_container.get('logger'),
-        kinesis=_container.get('streamalert_forwarder'),
-        state_manager=_container.get('state_manager'),
-        manager_factory=_container.get('query_pack_manager_factory')
+        logger=container.get('logger'),
+        kinesis=container.get('streamalert_forwarder'),
+        state_manager=container.get('state_manager'),
+        manager_factory=container.get('query_pack_manager_factory')
     )
 
 
-def _make_logger(_container):
-    logger = logging.getLogger(_container.get_parameter('command_name'))
-    logger.setLevel(_container.get_parameter('log_level').upper())
+def _make_logger(container):
+    logger = logging.getLogger(container.get_parameter('command_name'))
+    logger.setLevel(container.get_parameter('log_level').upper())
     logging.basicConfig(
         format='%(name)s [%(levelname)s]: [%(module)s.%(funcName)s] %(message)s'
     )
     return logger
 
 
-def _make_kinesis(_container):
+def _make_kinesis(container):
     return KinesisClient(
-        logger=_container.get('logger'),
-        client=_container.get('boto3_kinesis_client'),
-        kinesis_stream=_container.get_parameter('kinesis_stream')
+        logger=container.get('logger'),
+        client=container.get('boto3_kinesis_client'),
+        kinesis_stream=container.get_parameter('kinesis_stream')
     )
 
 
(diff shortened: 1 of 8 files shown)


streamalert/scheduled_queries/config/services.py
```diff
@@ -20,137 +20,182 @@
...
+    def create_step_function_state_manager(self):
+        return StepFunctionStateManager(
+            self.state_manager,
+            self.logger,
+            self.clock
+        )
+
+
 # pylint: disable=too-many-statements
 def configure_container(container):
```
ryandeivert: maybe a dumb question - since this seems pretty tied to the `ServiceContainer`, why not just have this be a `classmethod` on that class and not a static function?

streamalert/scheduled_queries/config/services.py
```diff
@@ -20,137 +20,182 @@
...
+    container.register(ServiceDefinition('query_parameter_generator', _make_param_generator))
+    container.register(ServiceDefinition('query_pack_repository', _make_query_pack_repo))
+    container.register(ServiceDefinition('query_pack_manager_factory', _make_query_pack_factory))
+    container.register(ServiceDefinition('query_pack_execution_context', _make_execution_context))
+    container.register(ServiceDefinition('clock', _make_clock))
+    container.register(ServiceDefinition('boto3_athena_client', _make_boto3_athena_client))
+    container.register(ServiceDefinition('boto3_kinesis_client', _make_boto3_kinesis_client))
+
+
+def _make_command_processor(_container):
```
ryandeivert: is there a reason for the underscore prefix on these arg names? I'm unsure where this convention comes from or signifies?

streamalert/scheduled_queries/config/services.py
Ryxias: The ServiceContainer is meant to be a reusable component. It's a write-once, use-anywhere class. Each application can then configure its own container, as each application will have different services and configurations available to it

streamalert/scheduled_queries/config/services.py
Ryxias: No reason anymore. They used to be nested methods but I pulled them out to global namespace and didn't remember to remove the underscores

streamalert/scheduled_queries/config/services.py
ryandeivert: ah got it - thanks for clarifying!

streamalert/scheduled_queries/config/services.py
Ryxias: Fixed https://github.com/airbnb/streamalert/pull/1130/commits/db7d0f1cb5162140f42d9dd28b700739d0e1f107

PR #1131, opened 2020-02-19
[terraform] switching to state lock table being managed by terraform

blakemotl:
to: @ryandeivert 

## Background
I originally wrote remote state file locking to use boto 3 to manage the table since terraform is not smart enough to create the locking table if it doesn't exist before locking the state. However, we have faced this problem in StreamAlert before with the remote state bucket. So I moved management of the state locking table over to the local terraform backend which is where the remote state bucket is also created. Now it gets created and destroyed by terraform as it should.

## Changes

* Remove old helper functions for table creation and deletion
* Add dynamodb table definition to the generate logic.

## Testing
Deployed and destroyed a fresh copy of StreamAlert.


streamalert_cli/terraform/generate.py
```diff
@@ -72,7 +72,6 @@ def _terraform_defaults(region):
         },
     })
 
-
```
jack1902: Shouldn't this newline be there? Two new lines between functions?

blakemotl:
It’s built in the init phase

streamalert_cli/terraform/generate.py
ryandeivert: he's correct - there should be 2 new lines here

blakemotl: This was definitely unintentional. Thanks for the catch!

PR #1134, opened 2020-02-19
[misc] updating authors/contributors, copyright updates, docs project name fix

ryandeivert:

## Changes

* Updating copyright name throughout license blocks to be stylized properly
* Updating AUTHORS.rst to be up-to-date with reality. Maintainers are now sorted alphabetically by last name.
* Fixing stylized project name in streamalert.io docs

## Testing

N/A


docs/source/apps.rst
```diff
@@ -76,7 +76,7 @@ To get help configuring a new App, use:
 
 .. code-block:: bash
 
-  $ python manage.py app new --help
+  python manage.py app new --help
```
chunyong-lin: Whatsup!!! Why do you want to remove `$ ` that was added in previous PR #1114 couple days ago?

Ryxias: IMO, I think in this case we can get rid of it. The `$` is intended to signal that a code block is specifically **_bash_**, but here we already have the `code-block:: bash` declaration. 

Additionally... `$` isn't really all that relevant anymore, since macos ZSH uses `%`............

ryandeivert: yeah I got ride of it for this very reason - shell prompts are arbitrary so putting `$` in here seems unnecessary and more of a "personal" preference. I as also unable to find any documented uses of `.. code-block:: bash` that used prompts

ryandeivert:
@Ryxias yes I need to update logos one more time please stand by and I will merge in a bit

ryandeivert:
@Ryxias just pushed new images, merging after CI finishes

PR #1138, opened 2020-02-20
[cli] Added additional commands to outputs

jack1902:
to: @0xdabbad00 @chunyong-lin @ryandeivert 
related to: #1118 #940
resolves: #1118 #940

## Background

Wanted the ability to do the following when configuring `outputs` via the CLI:
- set a single output
- set multiple outputs via a file
- get an outputs configuration
- update previously configured url (e.g. add an additional email to `aws-ses`)

## Changes

Added the following subcommands to `./manage.py outputs`:
- `set` - Set one output, using user_input (pass `--update` to overwrite
existing outputs)

- `get` - Get configured outputs for a service (includes creds) optionally pass `--descriptors` to only pull certain descriptor secrets for the service

- `set-from-file` - Set numerous outputs via a json file. Can set multiple services and descriptors (pass `--update` to overwrite existing outputs)

- `generate-skeleton` - Use to create a skeleton JSON file to be used with `set-from-file`

- `list` - lists out currently configured outputs according to the `conf/outputs.json` file

### QOL changes

* Moved the core_logic of `user_input` to a separate method so that i can re-use it in cases where the user doesn't need to be asked for input on the command line

* Made it possible to pass `help` to the `generate_subparser` function. This allows for `help` messages to be passed through to argparse. This then works very nicely with commands like `./manage.py outputs -h` and `./manage.py outputs get -h`

* Made the log message in `output_exists` optional, as i didn't want to have it logged in some cases

## Testing

* Ran `./tests/scripts/pylint.sh` - Fixed errors and set `# pylint: disable=protected-access` on one function as previously, this method was never used outside of the `alert_processor`
* Ran `./tests/scripts/test_the_docs.sh` - I understand that documentation requires some love but want opinions first before committing it to docs

#### FYI

As an FYI, for the `streamalert_cli` a really cool plugin for `sphinx` is called `sphinx-argparse` which can generate documentation for the cli by using argparse. I think this would be a very useful addition to the documentation overall and prevent the CLI falling behind. A cool example of `sphinx-argparse` in use: https://rdk.readthedocs.io/en/latest/references.html

### DIFF
--- a/streamalert_cli/helpers.py
+++ b/streamalert_cli/helpers.py
@@ -176,30 +176,9 @@ def user_input(requested_info, mask, input_restrictions):
 
     if not mask:
         while not response:
-            response = input(prompt) # nosec
+            response = input(prompt)  # nosec
 
-        # Restrict having spaces or colons in items (applies to things like
-        # descriptors, etc)
-        if isinstance(input_restrictions, re.Pattern):
-            valid_response = input_restrictions.match(response)
-            if not valid_response:
-                LOGGER.error('The supplied input should match the following '
-                             'regular expression: %s', input_restrictions.pattern)
-        elif callable(input_restrictions):
-            # Functions can be passed here to perform complex validation of input
-            # Transform the response with the validating function
-            response = input_restrictions(response)
-            valid_response = response is not None and response is not False
-            if not valid_response:
-                LOGGER.error('The supplied input failed to pass the validation '
-                             'function: %s', input_restrictions.__doc__)
-        else:
-            valid_response = not any(x in input_restrictions for x in response)
-            if not valid_response:
-                restrictions = ', '.join(
-                    '\'{}\''.format(restriction) for restriction in input_restrictions)
-                LOGGER.error('The supplied input should not contain any of the following: %s',
-                             restrictions)
+        valid_response = response_is_valid(response, input_restrictions)
 
         if not valid_response:
             return user_input(requested_info, mask, input_restrictions)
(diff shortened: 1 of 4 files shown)


streamalert_cli/outputs/handler.py
```diff
@@ -71,32 +176,336 @@ def handler(cls, options, config):
             return False
 
         # get dictionary of OutputProperty items to be used for user prompting
-        props = output.get_user_defined_properties()
+        properties = output.get_user_defined_properties()
```
Ryxias: Big fan of renaming `props` to `properties`. I'm a reactJS developer so `props` triggers me

streamalert_cli/outputs/handler.py
```diff
@@ -71,32 +176,336 @@ def handler(cls, options, config):
...
+            if not response_is_valid(value, prop.input_restrictions):
+                # Error messages are handled by response_is_valid
+                break
+
+            properties[name] = prop._replace(value=value)
+        else:
+            return properties
+
+
+class OutputGenerateSkeletonSubCommand(CLICommand):
```
Ryxias: I am a big fan of this command.

streamalert_cli/helpers.py
```diff
@@ -210,6 +189,41 @@ def user_input(requested_info, mask, input_restrictions):
     return response
 
 
+def response_is_valid(response, input_restrictions):
```
Ryxias: Refactoring this = 💯

PR #1140, opened 2020-02-20
[rules] Added AWS Config Compliance and Remediation Rules

jack1902:
to: @ryandeivert 
related to:
resolves:

## Background

Added Rules to trigger alerts on AWS Config NON_COMPLIANT and when Config Remediation takes place

## Changes

* Added rule to Alert on AWS Config `NON_COMPLIANT`
* Added rule to Alert on AWS Config Remediation (this is extremely useful to notify people they got auto-remediated so they don't go crazy when things just suddenly change on them)

## Testing

* Added the `.json` file to run `./manage test rules` Rules engine passes :)
* Ran `./tests/scripts/pylint.sh` Fixed errors

### DIFF
--- a/rules/community/cloudtrail/cloudtrail_aws_config.py
+++ b/rules/community/cloudtrail/cloudtrail_aws_config.py
@@ -1,0 +1,47 @@
+"""Alert on AWS Config"""
+from rules.matchers.matchers import AwsConfigMatcher
+from streamalert.shared.rule import rule
+
+
+# Populate this list to alert on specific Config Rules, otherwise all rules will be in-scope
+# Also consider the use of Lookup-Tables
+RULES_TO_ALERT_ON = []
+
+
+@rule(logs=["cloudtrail:events"], matchers=[AwsConfigMatcher.is_config_compliance])
+def config_compliance(record):
+    """
+    author:         jack (jack1902)
+    description:    Alert on AWS Config Complaince Change events of NON_COMPLIANT
+    testing:        From the Config page (https://console.aws.amazon.com/config/home)
+                    ensure recording is turned on. And you have a basic rule you can
+                    trigger as compliant or non-compliant.
+    """
+
+    non_compliance_present = any(
+        evaluation["complianceType"] == "NON_COMPLIANT"
+        for evaluation in record["requestParameters"]["evaluations"]
+    )
+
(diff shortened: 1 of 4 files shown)


jack1902:
Alternatively, this could be placed under `rules/community/config/` with names like `compliance.py` and `remediation.py` this then makes one rule per file under the `service` within AWS. This would make it easier to break down test files to trigger ONE rule instead of multiple since the `test` files currently map to the `rules` folder.

rules/matchers/matchers.py
```diff
@@ -51,3 +51,38 @@ def user_login(cls, rec):
             int(rec['columns']['type']) == cls._EVENT_TYPE_LOGIN and
             (rec['columns']['username'] not in cls._RUNLEVELS)
         )
+
+
+class ConfigMatcher:
```
Ryxias: Suggestion, rename to `AwsConfigMatcher` to be slightly more verbose

rules/community/cloudtrail/cloudtrail_aws_config.py
```diff
@@ -0,0 +1,46 @@
+"""Alert on AWS Config"""
+from rules.matchers.matchers import ConfigMatcher
+from streamalert.shared.rule import rule
+
+
+# Populate this list to alert on specific Config Rules, otherwise all rules will be in-scope
+RULES_TO_ALERT_ON = []
```
Ryxias: Segue, but this would be a great place to use LookupTables. We would use it internally here, but sadly on the open source repo, LookupTables is an "optional" feature so this wouldn't really work out-of-the-box.

Ryxias: No changes requested, just an observation

jack1902: Very useful for sure. Just a way of notifying the user of this possibility

rules/matchers/matchers.py
jack1902: Will do 👌

rules/community/cloudtrail/cloudtrail_aws_config.py
jack1902: I could leave a comment saying could use lookup tables as a hint to the user?

PR #1142, opened 2020-02-20
[core] Moved the secret_store from S3 to SSM

jack1902:
to: @ryandeivert @Ryxias 
related to: #240
resolves: #240

## Background

Using S3 as a `secret_store` is no longer desirable. Using SSM Parameter store makes visibility into what secrets are set much easier and requires less setup by StreamAlert

## Changes

* Implemented SSMDriver (a bit of `copy&pasta` from the S3Driver)
* Updated unit_tests so they pass when using this new driver
* Removed secrets bucket creation
* Added `ssm:GetParameter` permissions to the `alert_processor`

## Testing

* ran `./tests/scripts/pylint.sh` - currently there are warnings (i aim to clean this code up but want to ensure this initial commit is good)
* ran `./tests/scripts/unit_tests.sh` - All created tests now pass, i need to add some for SSMDriver


streamalert/shared/helpers/aws_api_client.py
```diff
@@ -182,3 +183,71 @@ def download_fileobj(file_handle, bucket, key, region):
...
+            region (str): AWS region
+            with_decryption (bool): Should decryption be attempted via this call
+
+        Returns:
+            str: The parameter either encrypted or not
+
+        Raises:
+            ClientError
+        """
+        client = boto3.client('ssm', config=default_config(region=region))
```
Ryxias: Did you get a chance to verify if the boto3 ssm client has internal side effects upon instantiation?

jack1902: Nope, but like i mentioned on the previous PR. This isn't going to be used alot since once it is used to pull param `x` this will be cached across invocations so it won't call out to the client

Ryxias: 🆗

Ryxias:
@jack1902 I've been merging 3.0.0 PRs into `feature-3-0-0` so you will have to rebase this one before we can merge

jack1902:
@ryxias rebase on `feature-3-0-0`

Ryxias:
Per a conversation, I will help Jack rebase this

Question is... how?

Ryxias:
Looks like I got it yay

Ryxias:
For the record, the old migration script can still be found here: https://github.com/jack1902/streamalert/blob/fad19f32400196e04967d917991ddcc88fb5d689/scripts/migrate_from_s3_to_ssm.py

streamalert/alert_processor/outputs/credentials/provider.py
```diff
@@ -122,7 +124,7 @@ def save_credentials(self, descriptor, kms_key_alias, props):
                  for (name, prop) in props.items() if prop.cred_requirement}
 
         credentials = Credentials(creds, False, self._region)
-        return self._core_driver.save_credentials_into_s3(descriptor, credentials, kms_key_alias)
+        return self._core_driver.save_credentials(descriptor, credentials, kms_key_alias)
```
ryandeivert: just so I'm understanding properly, we're going to continue using the old kms key for secrets for the `SecureString` option of ssm right? I think that is the case but just clarifying

streamalert/alert_processor/outputs/credentials/provider.py
```diff
@@ -331,134 +333,84 @@ def get_formatted_output_credentials_name(service_name, descriptor):
...
-        Returns:
-            string
-        """
-        return get_formatted_output_credentials_name(self._service_name, descriptor)
-
-    def get_s3_secrets_bucket(self):
-        """Returns an appropriate S3 bucket for all credentials relevant to this driver.
+    def get_parameter_prefix(self):
+        """Generate the parameter prefix from prefix and descriptor"""
+        return "{}_streamalert_secrets".format(self._prefix)
```
ryandeivert: can we update this to be something a little more specific to `outputs`? right now, apps also use ssm (even though they are not very well implemented to use `paths`..)

I'm thinking a namespacing convention like:

`/streamalert/outputs/<output_type>/<output_paramter_suffix>`

where the end result would be something like:

``/streamalert/outputs/slack/my-cool-output`

... might be ideal here, since SSM allows for getting params by "path".

Thoughts? @Ryxias ?

Ryxias: Agree wholeheartedly; but what about Prefix?

Seems like the namespace convention is:

`/[PREFIX]/[APPLICATION]/[COMPONENT]/[DRIVER_OR_TYPE]/[INSTANCE]`

streamalert/shared/helpers/aws_api_client.py
ryandeivert: hmm - thoughts on adding this as a class property that wouldn't ever need to be recreated for the like of the container?

like:

```python
class AwsSsm:
    SSM_CLIENT = None

    @classmethod
    get_parameter(cls, parameter_name, region, with_decryption=True):
        """
        ...
        """
        cls.SSM_CLIENT = cls.SSM_CLIENT or boto3.client('ssm', config=default_config(region=region))
        
        # Now use cls.SSM_CLIENT instead
        ...
```

FWIW: I don't think this is necessary for the `put_parameter` method since that should only be called when a human is involved

streamalert/alert_processor/outputs/credentials/provider.py
jack1902: Yes, so the secrets in SSM will be encrypted with the same KMS key at rest :)

terraform/modules/tf_alert_processor_iam/main.tf
```diff
@@ -62,11 +62,12 @@ data "aws_iam_policy_document" "output_secrets" {
   // Allow retrieving encrypted output secrets
   statement {
     effect    = "Allow"
-    actions   = ["s3:GetObject"]
-    resources = ["arn:aws:s3:::${var.prefix}-streamalert-secrets/*"]
+    actions   = ["ssm:GetParameter"]
+    resources = ["arn:aws:ssm:${var.region}:${var.account_id}:parameter/${var.prefix}_streamalert_secrets/*"]
```
ryandeivert: if the above "namespace" change is added, this would need changed for the proper path spec

Ryxias: I've got a PR in vpc-tf that would need to be updated too

streamalert/alert_processor/outputs/credentials/provider.py
jack1902: If you look at line 410, this is how the parameter name is generated. Sorry if this isn't clear

streamalert/alert_processor/outputs/credentials/provider.py
ryandeivert: okay - just checking if we were creating a new key or not for this

streamalert/alert_processor/outputs/credentials/provider.py
ryandeivert: sorry I didn't see your response and just edited for prefix. unsure tbh

ryandeivert: @jack1902 it was clear how it's generated - but I'm suggesting we use a different convention than that :)

Ryxias:
I will address @ryandeivert 's comments in a follow-up PR

PR #1147, opened 2020-02-20
Follow up PR to SSM migration

Ryxias:
to: @ryandeivert  @jack1902
related to: #1142

## Background

Addresses @ryandeivert 's comment: https://github.com/airbnb/streamalert/pull/1142#discussion_r382267969

## Testing

Didn't test yet

### DIFF
--- a/docs/source/outputs.rst
+++ b/docs/source/outputs.rst
@@ -53,15 +53,6 @@ For example:
 
   python manage.py output slack
 
-.. note::
-
-  If this is the first time you have configured new outputs via the cli, you may see this error for certain services:
-
-  `An error occurred while sending credentials to S3 for key '<SERVICE>/<KEY>' in bucket '<PREFIX>-streamalert-secrets': The specified bucket does not exist`.
-
-  If you encounter this error, first make sure you've followed the `Quick Start <getting-started.html#quick-start>`_ steps.
-  If you've already configured StreamAlert in the past, you may just have to run `python manage.py build`.
-  This ensures the S3 bucket used for storing encrypted secrets is created and only needs to be run once.
 
 The above command will then prompt the user for a ``descriptor`` to use for this configuration::
 
--- a/streamalert/alert_processor/outputs/credentials/provider.py
+++ b/streamalert/alert_processor/outputs/credentials/provider.py
@@ -423,7 +423,7 @@ class SSMDriver(CredentialsProvidingDriver):
         parameter_suffix = get_formatted_output_credentials_name(self._service_name, descriptor)
 
         # The leading forward slash character is intentional for parameters in a hierarchy
-        return '/{}/streamalert/outputs/{}'.format(self._prefix, parameter_suffix)
+        return "/{}/streamalert/outputs/{}".format(self._prefix, parameter_suffix)
 
 
 class LocalFileDriver(CredentialsProvidingDriver, FileDescriptorProvider, CredentialsCachingDriver):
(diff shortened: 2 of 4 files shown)


Ryxias:
NOTE: The KMS key alias still adopts the `{prefix}_streamalert_secrets` convention. I can change this, it's just a bit more work

tests/unit/streamalert/alert_processor/outputs/test_output_base.py
```diff
@@ -168,7 +168,7 @@ def test_check_http_response(self, mock_response):
     @mock_kms
     def test_load_creds(self):
         """OutputDispatcher - Load Credentials"""
-        param_name = '/{}_streamalert_secrets/{}'.format(
+        param_name = '/{}/streamalert/outputs/{}'.format(
             PREFIX, get_formatted_output_credentials_name('test_service', self._descriptor)
```
ryandeivert: I know this is annoying nit - but I don't like when other methods are called unnecessarily in a function that is testing something else that is very specific (`_load_creds` is being tested, `get_formatted_output_credentials_name` is called).. 

IMO this leads to confusion as to what the value of `param_name` actually is here and I'd personally rather hard-code the `param_name` value to what we expect it to be (otherwise I can't really tell what is being "used" for this value here..)

BUT.. maybe the actual problem here is that there aren't unit tests for `get_parameter_prefix` or `get_parameter_name` .. otherwise your above change would have broken other unit tests.

Ryxias: Ok, I will refactor this to be clearer

ryandeivert: sorry 😑 😢

docs/source/outputs.rst
```diff
@@ -53,15 +53,6 @@ For example:
 
   python manage.py output slack
 
-.. note::
```
ryandeivert: good catch

Ryxias:
@ryandeivert so I removed some of the "public" methods since they do not have tests, and I changed `get_parameter_name` to private. I also added a doc block.

And yeah, I completely agree we should not be calling `get_formatted_output_credentials_name` in a test... it's kind of like... a tautological test. You reverse engineer the code and then you test that the reverse-engineered code is the same as the engineered code...... 🤣

streamalert/alert_processor/outputs/credentials/provider.py
```diff
@@ -399,16 +399,28 @@ def save_credentials(self, descriptor, credentials, kms_key_alias):
...
 
-    def get_parameter_name(self, descriptor):
-        parameter_prefix = self.get_parameter_prefix()
+        Params:
+            descriptor (str): Descriptor of current output
+
+        Returns:
+            str
+        """
+        parameter_prefix = "{}/streamalert/outputs".format(self._prefix)
```
ryandeivert: instead of the aggressive `(!) NOTE`, couldn't we just do this below...

```python
parameter_suffix = get_formatted_output_credentials_name(self._service_name, descriptor)
return '/{}/streamalert/outputs/{}'.format(prefix, parameter_suffix)
```

ryandeivert: also double quotes (`"`) vs single quotes (`'`)

streamalert/alert_processor/outputs/credentials/provider.py
```diff
@@ -399,18 +399,31 @@ def save_credentials(self, descriptor, credentials, kms_key_alias):
...
-    def get_parameter_name(self, descriptor):
-        parameter_prefix = self.get_parameter_prefix()
+        Returns:
+            str
+        """
         parameter_suffix = get_formatted_output_credentials_name(self._service_name, descriptor)
-        return '/{}/{}'.format(parameter_prefix, parameter_suffix)
+
+        # The leading forward slash character is intentional for parameters in a hierarchy
+        return "/{}/streamalert/outputs/{}".format(self._prefix, parameter_suffix)
```
ryandeivert: you no fix quotes 😭

Ryxias: I FIXED IT A SWEAR

Issue #1159, opened 2020-02-24
init -b not pulling correct versions of provider plugins


jack1902:
### Description
When running `./manage.py init -b` it doesn't pull the correct versions of the provider plugins

### Steps to Reproduce

Within a deployed project run:

* `./manage.py clean` - Remove terraform files
* `./manage.py init -b` - Initiate the Backend of an existing deployment (observe terraform provider plugin versions)
* `./manage.py build` - Will download the correct versions

## Desired Change

The correct provider plugin versions are downloaded on `init -b`

jack1902:
If you place `import pdb; pdb.set_trace()` or `breakpoint()` into 
https://github.com/airbnb/streamalert/blob/34606595ba08a3b2050674b1c75e6887e4614d81/streamalert_cli/helpers.py#L63-L65

You will see you have no terraform files as of yet under `terraform/` with the information from

https://github.com/airbnb/streamalert/blob/34606595ba08a3b2050674b1c75e6887e4614d81/streamalert_cli/terraform/generate.py#L65

Since there are no terraform files when the init command is actually being called it falls back on the providers.tf to pull down the `aws` provider which doesn't have any `version` set. Once you run `build` the `main.tf.json` is generated with the snipper:

```json
"provider": {
    "aws": {
      "region": "REGION",
      "version": "~> 2.28.1"
    }
  },
```

This then errors out as this version is not currently present. Running `init -b` again will then download the correct version

ryandeivert:
fixed with #1166

PR #1166, opened 2020-02-25
[cli] fix backend initialization

jack1902:
to: @ryandeivert @Ryxias 
related to:
resolves: #1159

## Background

Previously, the init command had a '-b' flag. This only initialized a local backend but the description on the flag stated 'useful for refreshing a pre-existing deployment'. This flag now actually reflects
this

## Changes

* added additional function calls to '-b' flag to initialize the remote backend

This generates the `main.tf.json` file containing the `remote` backend instead of the previously `local` backend and downloads the relevant provider plugin versions

## Testing

Ran the following in order to verify success (pre-existing deployment):

- `./manage.py clean` - Remove ALL terraform configuration files
- `./manage.py init -b` - Initialize the `remote` backend and download the relevant provider plugins and their versions
- `./manage.py build` - Perform a build (shouldn't contain ANY infrastructure changes, as no new infra has been added) 

These commands work as expected and in my view have fixed the issue better than i had originally proposed in a comment on the original issue

Also ran the `unit_tests` and `pylint`

### DIFF
--- a/streamalert_cli/terraform/handlers.py
+++ b/streamalert_cli/terraform/handlers.py
@@ -64,7 +64,7 @@ class TerraformInitCommand(CLICommand):
 
         # Stop here if only initializing the backend
         if options.backend:
-            return cls._terraform_init_backend()
+            return cls._terraform_init_backend(config)
 
         LOGGER.info('Initializing StreamAlert')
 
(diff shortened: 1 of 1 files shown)


jack1902:
Moved the function to the `_terraform_init_backend` method (previously i had it under the `if options.backend` before i moving it

Issue #1157, opened 2020-02-23
Ability to mark Severity levels on a rule


jack1902:
### Description

Have the ability to state a `severity` on a rule. If Rule x is triggered this is considered a `high`, `medium` or a `low` (potential for `critical`?)

This would be added to the rule Decorator:

```python
@rule(
    logs='',
    severity='high' # Default to `medium`
)
def really_important_alert(rec):
    return True
```

#### Where could these severity levels be used?

I am thinking they could be used in:
* the `subject` field within email outputs
* potentially set the colour of a card in `teams`
* Quicksight dashboards based on Athena historical searches to see how many `high`, `medium` and `low` severity alerts have occurred over a given period 

## Desired Change

The ability to set `severity` levels on specific rules with a default of `medium` or something along these lines

PR #1171, opened 2020-02-26
[core] Added severity levels to rules (#1157)

jack1902:
related to: #1157 
resolves: #1157

## Background

Wanted the ability to mark severity levels on `rules` to help with prioritization of Alerts raised.

## Changes

* [core] Added severity levels to rules (#1157)
* [testing] Updated unit_tests accordingly
* [docs] Updated documentation for the rule so there is a description around severity levels

## Testing

* built the docs `./tests/scripts/test_the_docs.sh`
* ran pylint `./tests/scripts/pylint.sh`
* ran unit tests `./tests/scripts/unit_tests.sh`

### comments

- The meanings of severity levels is up for debate, mine are a place holder
- Should we hard code the severity levels to force the user to use pre-defined severities?
- Should we update outputs to use these severity levels or just leave it down to the user via their own publications?


docs/source/rules.rst
```diff
@@ -285,6 +286,20 @@ The following table provides an overview of each rule option, with more details
...
+:severity:
+
+  The ``severity`` keyword argument defines the severity level for this rule. The default is ``low``, but this can be set on a per-rule basis.
+
+  Severities, can be user defined but a recommended approach is to use the following severity levels:
+
+==================  ===========
+**Severity level**  **meaning**
+------------------  -----------
+``low``             ``Alert should be looked at when the user has time to investigate``
```
blakemotl: I would hesitate to use SLA's to define severity levels here since those will vary wildly across organizations.

Ryxias:
I think my big worry here is that StreamAlert would be prescribing a framework for severity rating. This may be easy for rules writing, but will very likely not cooperate well with reporting terminology. Severity in the wild has a huge variance in framework. Example: PagerDuty uses p0-5, there's also `low-med-high-critical` and `low-moderate-severe-critical-p0` and a lot of times severity can be coupled to an organizations case management product(s), which may not be enforceable in a consistent manner across different products and orgs.

blakemotl:
@Ryxias I partially agree with you, except I do think StreamAlert should support severity. It should just be configurable instead of hard coded. Let the user define what their severity levels are. If translation to an output (I.E pagerduty) is needed can't that be handled on a case by case basis in the publisher?

chunyong-lin:
In my opinion severity would be useful for the team to categorize their own work when there are multiple teams getting alerts from StreamAlert. I would prefer leave the severity to users via their own publications.

If there is only one team receiving the alerts from StreamAlert, the user may just naming the outputs with `*high-pri` or `*low-pri` to categorize the rules.

gavinelder:
I think this has already been suggested but it seems like it's best suited as an extension of publishers as such a rule would like the following?

```
@rule(
    logs=['logs:foo'],
    outputs=['slack:foo'],
    publishers=[
        Severity,           # include the Severity of an incident. 
    ]
)
def rule_example(rec):
    """
    author:       coresec
    description:  something alerted 
    reference:    dome doc
    playbook:     (a) Reach out to the user who made the modification and confirm intent.
                          (b) Link appropriate Jira ticket and close alert.
    severity:       1 
    """
```

blakemotl:
The benefit of adding severity to a rule instead of publishers, is that you get the ability to search alerts by severity in athena as a top level key and potentially build metrics regardless of what you publish or output to which I see a lot of value in. I just think it should be disabled by default and expose the ability to set what valid severities are in a config file when its enabled.

jack1902:
@blakemotl I agree completely with you statement. Having the ability to historically search on a user defined severity will be very useful. Off by default, but easy to turn on. There would be defaul values to give the user some guidance but they can change it to what they want. I'm glad this PR has sparked some debate on this!

jack1902:
@ryandeivert unsure what the appetite for this one is. It's a controversial issue as the levels of severity differ wildly within organisations

rsavjani:
we are deffo looking at needing to have sev levels to alert so we can easily send sev 1  - 3 to one output and rest to different output and also for audit purposes so we can see how many level 1 we had a week etc

jack1902:
I'm redoing the bits of this PR to re-open as a proper PR to finalize if this is a `yay` or `nay` feature

PR #1173, opened 2020-02-27
[Schema] OSquery optional TLK

gavinelder:
to:
related to:
resolves:

## Background

The following makes `log_type` an optional top-level key as this is only added by the [TLS logger](https://github.com/osquery/osquery/blob/d373d042606acda3cb46d6424cfe09be6939b554/plugins/logger/tls_logger.cpp#L100) This fixes classification for Filebeat and other non-TLS sources such as Kolide Fleet.  

## Changes

Add `log_type` an optional top-level key to the OSquery Schema.

## Testing

### DIFF
--- a/conf/schemas/osquery.json
+++ b/conf/schemas/osquery.json
@@ -18,6 +18,7 @@
       "optional_top_level_keys": [
         "decorations",
         "epoch",
+        "log_type",
         "counter"
       ]
     }
@@ -70,6 +71,7 @@
         "decorations",
         "epoch",
         "counter",
+        "log_type",
         "logNumericsAsNumbers",
         "numerics"
       ]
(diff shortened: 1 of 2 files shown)


conf/schemas/osquery.json
```diff
@@ -18,6 +18,7 @@
       "optional_top_level_keys": [
         "decorations",
         "epoch",
+        "log_type",
```
chunyong-lin: Is it `log_type` also added it `osquery:batch`?

gavinelder: Good question let me double-check , I don't use batch myself, so it may need included in the JSON

gavinelder:
@chunyong-lin All good now log_type is also added to batch, when using the tls logger but not part of the standard schema. I have added this to the batch schema.

Issue #1152, opened 2020-02-21
[improvement] migrate rule "integration" tests out of root tests directory


ryandeivert:
## Background

The rule "integration" tests structure is currently poorly designed, and the tests are embedded in the `tests` directory (which should be strictly for unit tests). It also makes is prohibitive difficult to create a standalone package/decouple rules+config/do anything in modular way.

## Desired Change

Any user rules should have "tests" defined alongside of them. These could follow the convention that most unit-test frameworks use where files are prefixed with `test_`. The testing framework would then only load files that begin with `test_` and end with `.json`.

### Proposed Design

```
rules/
└── community
    ├── rule_01.py
    └── test_rule_01.json
```

### Things to consider

* Does the structure of test files themselves need to change?
  * Do we still need the `source`, etc attributes in the test events? It seems like we 
* Should a given test event could still be able to apply to multiple rules? Or should a file be restricted to testing only rules that it "lives beside". I'm thinking the former, but could be convinced otherwise.
* How do we handle `validate_schema_only` test events? Should these be in a "special" folder like `classifier_tests" or handled elsewhere? (this seems like the most problematic part of this change)

ryandeivert:
fixed with #1179, #1189

PR #1179, opened 2020-03-05
refactor rule integration tests to be included with rules

ryandeivert:
to: @chunyong-lin / @Ryxias / @blakemotl 
resolves: #1152 

## Background

See #1152

## Changes

* Introducing a new `TestEventFile` and `TestEvent` class to encapsulate errors/status of test event processing.
* `TestResult` is now a subclass of `TestEvent`, as there is a lot of shared logic between the two.
* Changing the argparse action of `UniqueSetAction` to be `UniqueSortedListAction`, as it's more consistent with the purpose. It still creates a "unique set", just returns it in a sorted/deterministic format.
* Huge update to the `test_fixtures` logic for `lookup_tables` and `threat_intel`
  * `test_fixtures` are now embedded within the test event itself, under respective keys
* Changing the verbiage for test events that should only be used for classification testing from `validate_schema_only` --> `classify_only` since it seems more consistent.
* Updating how the CLI exiting works in `manage.py`

## Testing

* All rule tests are passing.

~NOTE: I want to add more (read: more than the current # of 0) unit tests for this, but wanted to get it up to take a look at.~
* I have added a handful of unit tests now.


manage.py
```diff
@@ -109,8 +109,8 @@ def main():
     options = parser.parse_args()
 
     # Exit with the result, which will be False if an error occurs, or True otherwise
-    sys.exit(not cli_runner(options))
+    return not cli_runner(options)
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
```
Ryxias: Nice

rules/community/duo_authentication/duo_lookup_tables_example.json
```diff
@@ -0,0 +1,47 @@
...
+      "username": "user.name@email.com"
+    },
+    "description": "Duo authentication log marked as failure as a result of 'Anonymous IP' that will create an alert",
+    "log": "duo:authentication",
+    "service": "streamalert_app",
+    "source": "prefix_cluster_duo_auth_sm-app-name_app",
+    "trigger_rules": [
+      "duo_lookup_tables_example"
+    ],
+    "test_fixtures": {
```
Ryxias: 💯

rules/community/cloudwatch_events/cloudtrail_critical_api_calls.json
```diff
@@ -846,6 +846,81 @@
     "source": "prefix.cluster.sample.bucket",
     "trigger_rules": [
       "cloudtrail_critical_api_calls"
-    ]
+    ],
+    "publisher_tests": {
```
Ryxias: I know I wrote this.... but 

Since you wrote the `lookuptable` and `threat_intel` sample rules, maybe should also split this into one of those `example_publisher_test` kind of rules? Up to you

ryandeivert:
hey @Ryxias can you look at the commit I added and lmk if that works:

https://github.com/airbnb/streamalert/pull/1179/commits/0bcdf9be155cd934bdbc22bbee5af965fc80beaf

rules/community/cloudwatch_events/cloudtrail_critical_api_calls.json
ryandeivert: made update for this

conf/clusters/prod.json
```diff
@@ -77,7 +77,7 @@
       ]
     }
   },
-  "enable_threat_intel": false,
+  "enable_threat_intel": true,
```
chunyong-lin: I may miss some information, can you remind me why do we want to enable threat intel by default?

ryandeivert: @chunyong-lin for the "integration" test fixtures for threat intel to work, this needs to be enabled. do you envision this being a problem? I can disable the rule if so

ryandeivert: just pushed commits that disable this rule by default, commenting as to why

Issue #1149, opened 2020-02-21
[publishers] decouple alert publishers and tests


ryandeivert:
## Background

If the current [publishers](https://github.com/airbnb/streamalert/tree/release-3-0-0/publishers) are deleted/modified, they will break current [unit tests](https://github.com/airbnb/streamalert/tree/release-3-0-0/tests/unit/streamalert/alert_processor/publishers). This should not be the case, and we need to come up with away to either:

1) Have some "default" publishers that are *not* in the root `publishers` directory. These could be placed in `streamalert/shared/publishers` (?) or something similar, and unit tests could be relocated to `tests/unit/streamalert/shared/...` to align with this change.
2) Keep the `publishers` directory in the root, but *migrate* unit tests to _within_ that directory, (instead of in the root `tests/unit` directory). Proposed structure:

```
publishers/
├── community
│   ├── pagerduty
│   └── slack
└── tests
    └── unit
        └── publishers
            └── community
                ├── pagerduty
                └── slack
```

Then the unit test script should be updated to test this directory also.

Additionally, the [hard-coded](https://github.com/airbnb/streamalert/blob/90c42f619325abe4984a8de850b9cd2ba6aaf316/streamalert/shared/publisher.py#L120) publishers directory should be updated to take user-configurable paths for importing publishers, similar to how rules and matchers current work.

Ryxias:
# Foreword

Taking a look at the publisher tests I wrote, I'm really beginning to question the usefulness of unit tests for those.

It seems like it would be preferable to have publisher **_integration tests_** instead. Like some mechanism for the return value of `compose_alert` would be. I also think it's (currently) really annoying to figure out how to set up LookupTables for integration tests.


# Proposal
No more unit tests for publishers. Unit tests become strictly used for **StreamAlert Framework Code**.

## Publisher Integration Tests
Enhance our ability to test Publishers. Add the following to the integration `.json` config:

```
"publisher_tests": [
  [
    "$.record.field", 
    "is",
    "aaaa"
  ],
  [
    "$.record.otherfield",
    "in",
    ["thing1", "thing2"]
  ],
  ...
],
```

It specifies an array of conditionals.

## LookupTables Mocks per-test
Allow setting up LookupTable mocks more easily. Add the following to the integration `.json` config:

```
"lookup_tables": {
  "table1": {
    "key1": "asdfalkd",
    "key2": "addssejpfiwjef",
  },
 ... 
}
```

ryandeivert:
[...]

PR #1181, opened 2020-03-09
Redesigns Publisher Testing

Ryxias:
resolves: #1149 

**NOTE: I'm merging this into `origin/ryandeivert-refactor-integration-tests`**

## Background

The first time I did Alert Publishers I didn't really think through how to do testing well. Taking another pass at this, allowing individuals to write jsonpath-based conditionals.

## Changes

There is now a new field in integration test JSON files: `publisher_tests`. It takes the following format:

```
"publisher_tests": {
    "[OUTPUT_DESCRIPTOR]": [
      [
        "[PATH]",
        "[CONDITIONAL]",
        [VALUE]
      ],
      ...
  },
  "[OUTPUT_DESCRIPTOR_2]": [ ... ]
}
```

* `OUTPUT_DESCRIPTOR`: Should be an output that the alert publishes to, e.g. `slack:sample-channel`
* `PATH`: A jmespath-compatible search path.
* `CONDITIONAL`: Currently supports `is` for equality matching, or `in` for substring/list contains
* `VALUE`: The value that the jsonpath value should match 

### Other Change: No more unit tests for community content

I didn't see the value of continuing to have unit tests for Publishers. I think in general, we should use unit tests for canonical StreamAlert code, and liberally use **integration tests** for other code.

## Testing

Currently on @ryandeivert's branch the Integration tests are broken so I can't test this right now

### DIFF
--- a/rules/community/cloudwatch_events/cloudtrail_critical_api_calls.json
+++ b/rules/community/cloudwatch_events/cloudtrail_critical_api_calls.json
@@ -846,6 +846,24 @@
     "source": "prefix.cluster.sample.bucket",
     "trigger_rules": [
       "cloudtrail_critical_api_calls"
-    ]
+    ],
+    "publisher_tests": {
+      "slack:sample-channel": [
+        [
+          "keys(@)",
+          "is",
+          [
+            "@slack.text",
+            "@slack.attachments",
+            "@slack._previous_publication"
+          ]
+        ],
+        [
+          "\"@slack.text\"",
+          "is",
+          "Rule triggeredsss"
+        ]
+      ]
+    }
   }
 ]
--- a/rules/community/cloudwatch_events/cloudtrail_critical_api_calls.json
+++ b/rules/community/cloudwatch_events/cloudtrail_critical_api_calls.json
@@ -861,7 +861,7 @@
         [
           "\"@slack.text\"",
           "is",
-          "Rule triggeredsss"
+          "Rule triggered"
         ]
       ]
     }
--- a/rules/community/cloudwatch_events/cloudtrail_critical_api_calls.json
+++ b/rules/community/cloudwatch_events/cloudtrail_critical_api_calls.json
@@ -849,15 +849,15 @@
     ],
     "publisher_tests": {
       "slack:sample-channel": [
-        [
-          "keys(@)",
-          "is",
-          [
+        {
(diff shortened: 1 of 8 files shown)


streamalert_cli/test/handler.py
```diff
@@ -372,10 +373,77 @@ def _process_test_file(self, test_file_path):
...
+                                    })
+                                    continue
+
+                                publisher_test_results.append({
+                                    'success': res,
+                                    'failure': None if res else (
+                                        'Item at path "{}" {} "{}",\nActual value: "{}"'.format(
+                                            jsonpath,
+                                            (
+                                                "should have been" if condition == 'is'
```
ryandeivert: mixed use of `"` and `'` here

rules/community/cloudwatch_events/cloudtrail_critical_api_calls.json
```diff
@@ -846,6 +846,24 @@
     "source": "prefix.cluster.sample.bucket",
     "trigger_rules": [
       "cloudtrail_critical_api_calls"
-    ]
+    ],
+    "publisher_tests": {
+      "slack:sample-channel": [
+        [
+          "keys(@)",
```
ryandeivert: can we document this a little more so people understand what this refers to

ryandeivert: also - what are your thoughts on making this list more explicit.. something like:

```json
{
  "publisher_tests": {
    "slack:sample-channel": [
      {
        "jmespath_descriptor": "keys(@)",
        "conditional": "is",
        "value": [
          "@slack.text",
          "@slack.attachments",
          "@slack._previous_publication"
        ]
      },
      {
        "jmespath_descriptor": "\"@slack.text\"",
        "conditional": "is",
        "value": "Rule triggered"
      }
    ]
  }
}
```

Ryxias: More explicit, more wordy to write, harder to read, since the "subject-verb-predicate" doesn't make sense anymore.

ryandeivert: but but but.. usability

Ryxias: Compromise: will support **_both_** syntaxes

Issue #1186, opened 2020-03-12
[bug] Sanitize log name if there have dots


chunyong-lin:
## Background
When we have `log:name.with.dots.bala` in the `conf/schemas/weird_logs.json` or `conf/logs.json`, the `create_table` and `generate_data_table_schema` methods will throw `KeyError`.

https://github.com/airbnb/streamalert/blob/f8a9da465f2f9d187cc8300a0f0e2e1f013c1e4a/streamalert_cli/athena/handler.py#L405
https://github.com/airbnb/streamalert/blob/f8a9da465f2f9d187cc8300a0f0e2e1f013c1e4a/streamalert_cli/athena/helpers.py#L213

It is still a mystery to me that how did we create carbonblack tables correctly before.

### Steps to Reproduce
* The issue can be reproduce when running `create_table` cli to create an Athena table
```
python manage.py athena create-table  --bucket nobody2020030420-streamalert-data  --table-name carbonblack_alert_status_updated

Traceback (most recent call last):
  File "manage.py", line 116, in <module>
    main()
  File "manage.py", line 112, in main
    sys.exit(not cli_runner(options))
  File "/Users/nobody/Work/ext_streamalert/streamalert_cli/runner.py", line 70, in cli_runner
    result = cmds[args.command](args)
  File "/Users/nobody/Work/ext_streamalert/streamalert_cli/runner.py", line 126, in <lambda>
    command: lambda opts, cmd=cli_command: cmd.handler(opts, config)
  File "/Users/nobody/Work/ext_streamalert/streamalert_cli/athena/handler.py", line 187, in handler
    options.schema_override
  File "/Users/nobody/Work/ext_streamalert/streamalert_cli/athena/handler.py", line 406, in create_table
    log_info = config['logs'][table.replace('_', ':', 1)]
KeyError: 'carbonblack:alert_status_updated'
```

* I also added unit test cases to reproduce this issue.

## Desired Change
* Sanitize log name by replacing `.` with `_`.
* Increase the unit test coverage for the code under `streamalert_cli/`

chunyong-lin:
This should be fixed.

PR #1188, opened 2020-03-12
Sanitize log names

chunyong-lin:
to: @ryandeivert @Ryxias @blakemotl 
related to:
resolves: #1186 

## Background
See the issue #1186 for more details.

## Changes

* Replace all `.` (dot) in the log names in `conf/schemas/*.json` or `conf/logs.json` with `_` (underscore). The change is made to `Config` class where all the conf files are loaded.
* Also update rule test method to replace `.` with `_` when inspect `log` source in the testing events. Eventually there are two carbonblack testing events with `log` set to `carbonblack:ingress.event.procstart` and `carbonblack:ingress_event_procstart`.

## Testing
* Add unit tests to reproduce the issue and test the code change.
* `python manage.py athena create-table  --bucket nobody2020030420-streamalert-data  --table-name carbonblack_alert_status_updated` works in staging environment with the fix.

### DIFF
--- a/streamalert/shared/config.py
+++ b/streamalert/shared/config.py
@@ -243,7 +243,7 @@ def _requires_sanitized_log_names(config):
         config (dict): loaded config from conf/ directory
 
     Returns:
-        Boolean: True if there are settings under "enabled_logs" in the "firehose" conf,
+        boolean: True if there are settings under "enabled_logs" in the "firehose" conf,
             otherwise False.
     """
     infra_config = config.get('global', {}).get('infrastructure')
@@ -259,6 +259,37 @@ def _requires_sanitized_log_names(config):
 
     return True
 
+def sanitize_key(key):
+    """Sanitize a key by replacing non-characters with '_' (underscore), except ':' (colon).
+
+    Args:
+        key (str): a string needs to be sanitized
+
+    Returns:
+        str: sanitized string
+    """
+    key_parts = key.split(':')
+    if len(key_parts) == 1:
+        # if no ':' in the key name, replace all special characters with '_'
+        # e.g.
+        # 'osquery_differential' -> 'osquery_differential'
+        # 'osquery_differntial.with.dots' -> 'osquery_differntial_with_dots'
+        sanitized_key = re.sub(SPECIAL_CHAR_REGEX, SPECIAL_CHAR_SUB, key)
+    elif len(key_parts) == 2:
+        # if there is a ':', replace the special chars in 2nd part and reconstruct
+        # sanitized key name
+        # e.g.
+        # 'carbonblack:alert.status.updated' -> 'carbonblack:alert_status_updated'
+        key_parts[1] = re.sub(SPECIAL_CHAR_REGEX, SPECIAL_CHAR_SUB, key_parts[1])
+        sanitized_key = ':'.join(key_parts)
+    else:
+        message = (
+            'Found offended log name "{}". Log name can only contain up to one colon. '
+            'Please check naming convention in conf/schemas/ or conf/logs.json'.format(key)
+        )
+        raise ConfigError(message)
+    return sanitized_key
+
 def _sanitize_log_names(config):
     """Sanitize the name of logs and replace the dots with underscores. For some reason,
     we have log name with dots in it in the conf/schemas/carbonblack.json or conf/logs.json.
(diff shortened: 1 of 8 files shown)


streamalert/shared/config.py
```diff
@@ -218,9 +218,37 @@ def load_config(conf_dir='conf/', exclude=None, include=None, validate=True):
     if validate:
         _validate_config(config)
 
+    if config.get('logs'):
+        config['logs'] = _sanitize_logs_name(config['logs'])
+
+    if (config.get('global')
```
ryandeivert: eeek conditional blocks like this make me cringe so hard. readability is so bad. can be break this up into multiple statements?

like:

```python
infra_config = config.get('global', {}).get('infrastructure')
if not infra_config:
    return config

if 'firehose' not in infra_config:
    return config

# ... and so on ...
```

Ryxias: Alternative way is to lump it into a function>

```
if self._requires_sanitized_log_names(etc):
  return config['global']....

return config
```

streamalert/shared/config.py
```diff
@@ -218,9 +218,37 @@ def load_config(conf_dir='conf/', exclude=None, include=None, validate=True):
...
+
+    Args:
+        config (dict): loaded config from conf/ directory
+
+    Returns:
+        new_config (dict): new config with sanitized keys
+    """
+    new_config = dict()
+    for key, _ in config.items():
+        sanitized_key = key.replace('.', '_')
```
ryandeivert: I think we should use that same regex here to do special char replacement: 
https://github.com/airbnb/streamalert/blob/319fc115d3cf42dea6df10f3a9d9ed7583671f7f/streamalert/classifier/clients/firehose.py#L299

Since that finds all *non-word* characters (except for the `_` which it allows):

https://www.rexegg.com/regex-quickstart.html

streamalert/shared/config.py
```diff
@@ -218,9 +218,37 @@ def load_config(conf_dir='conf/', exclude=None, include=None, validate=True):
...
+            and config['global']['infrastructure'].get('firehose')
+            and config['global']['infrastructure']['firehose'].get('enabled_logs')):
+        config['global']['infrastructure']['firehose']['enabled_logs'] = _sanitize_logs_name(
+            config['global']['infrastructure']['firehose']['enabled_logs']
+        )
+
     return config
 
 
+def _sanitize_logs_name(config):
```
ryandeivert: does theis `_sanitize_logs_name` function actually _replace_ the value within a person's config to be written back out to disk??

if so, I don't think we should do this... this could mess up rules, etc that use specific logs. I think and "transformation" for log name --> table name + firehose name should be done by _us_ under the hood

streamalert/shared/config.py
chunyong-lin: I would, `re.sub(cls.SPECIAL_CHAR_REGEX, cls.SPECIAL_CHAR_SUB, log_name)` was my first fix. But we have to `:` because firehose use the `:` to determine if we enable a set of logs or single log, e.g. `osquery` vs. `osquery:differential`.

streamalert/shared/config.py
chunyong-lin: I don't think `_sanitize_logs_name` change a person's config. In my staging environment, I don't see anything changed under `conf/`. 
Totally agree with you that the changes should be done by us under the hood. I had iterated few solutions and find current change is the best. But I may miss something. I will look at rule and classifier code again today.

streamalert/shared/config.py
ryandeivert: didn't we used to split on the `:` and the only worry about sanitizing the second half? everything _after_ the `:`, then concatenating that with the first half again?

ryandeivert: I worry cause someone _might_ use a different character that will also break this... like `-` ...

chunyong-lin: Aha, actually this is a great idea!!! Smart smart...

chunyong-lin:
Hi @ryandeivert , about your concern
> does theis _sanitize_logs_name function actually replace the value within a person's config to be written back out to disk??

> if so, I don't think we should do this... this could mess up rules, etc that use specific logs. I think and "transformation" for log name --> table name + firehose name should be done by us under the hood

I have checked the code and tested the new changes in the staging environment with a weird log name `cloudwatch:events.dots-test`, the person's config on the disk is not changed, classifier, rules engine, alert processor and athena partition lambda functions are working as expected. Can you PTAL my latest two commits which address the comments. Thank you thank you.

Issue #1190, opened 2020-03-13
[tf] The length of firehose name may exceed 64 characters with prefix enabled


chunyong-lin:
## Background

Newly discovered that with the prefix enabled for firehose, some firehose name will exceed 64 characters and terraform will throw `Error: "name" cannot be longer than 64 characters` error, e.g carbonblack_alert_watchlist_hit_feedsearch_bin
```
>>> s = 'cylin2020031200-streamalert-data-carbonblack_alert_watchlist_hit_feedsearch_bin'
>>> len(s)
79
```
Without prefix, the length of firehose name is right on the edge
```
>>> s = 'streamalert-data-carbonblack_alert_watchlist_hit_feedsearch_bin'
>>> len(s)
63
```
## Desired Change
Couple options:
1. @Ryxias has a creative idea that to hash (md5) the part of firehose name if the name exceeds the limit. Personally I like this idea.
2. Currently, firehose has `streamalert` as part of prefix, we may shorten it to `sa`, but it can not fix the problem completely. Users may hit the issue if have long prefix or long schema name.

chunyong-lin:
PR #1197 simplified the solution as well.

PR #1191, opened 2020-03-14
[core] Hash firehose stream name if it is too long

chunyong-lin:
related to: #1115 #1188 
resolves: #1190 

## Background
See issue #1190 

## Changes
* Update firehose stream name in `classifier` and pass the new stream name to `tf_kinesis_firehose_delivery_stream` module during terraform generate time. The firehose stream name is based on schema name.
  * This change will generate a new stream name if it's longer than 64 characters. In another words, most of the firehose stream name will remain the same.
  * For the stream name longer than 64 characters, the first 58 characters remains unchanged and the rest characters will hash to 8 characters by `hashlib.md5()`. `hashlib.md5()` returns 32 characters (128-bit) and we will only take the first 8 characters. There is low chance is too reach the stream name limit, so I don't worry the hash conflict.
* Remove `_streamalert_` (13 chars) from firehose stream name. Firehose stream name is restrict to the name length is no longer than 64 characters. We want the schema names more descriptive and it will be easily hit the limit especially when we add prefix.
* Use `var.glue_catalog_table_name` in the s3 file prefix instead of `var.log_name`, then latter may be changed if it is too long.

## Testing
* Add unit test cases to cover the code change.
* Tested in staging environment with the data matched to schemas both with short and extremely long schema name, e.g. `osquery_differential` and `cloudwatch:events.dots-test.crazy.long_name_yay_hahahahaha`
  * classifier can process data and send data to both firehose
  * rule's engine can trigger alerts from both types of data.
  * alert process can send alerts to firehose correctly
  * athena partition can partition both types of data
  * both types of data and alerts can be searchable in the Athena in parquet format.

### DIFF
--- a/streamalert/classifier/clients/firehose.py
+++ b/streamalert/classifier/clients/firehose.py
@@ -342,7 +342,7 @@ class FirehoseClient:
         # chars.
         pos = reserved_len - cls.FIREHOSE_NAME_HASH_LEN
         hash_part = log_stream_name[pos:]
-        hash_result = hashlib.md5(hash_part.encode()).hexdigest()
+        hash_result = hashlib.md5(hash_part.encode()).hexdigest() # nosec
 
         # combine the first part and first 8 chars of hash result together as new
         # stream name.
--- a/streamalert/classifier/clients/firehose.py
+++ b/streamalert/classifier/clients/firehose.py
@@ -53,7 +53,7 @@ class FirehoseClient:
     MAX_RECORD_SIZE = 1000 * 1000 - 2
 
     # Default firehose name format, should be formatted with deployment prefix
-    DEFAULT_FIREHOSE_FMT = '{}data_{}'
+    DEFAULT_FIREHOSE_FMT = '{}streamalert_{}'
 
     # Exception for which backoff operations should be performed
     EXCEPTIONS_TO_BACKOFF = (ClientError, BotocoreConnectionError, HTTPClientError)
@@ -62,9 +62,9 @@ class FirehoseClient:
     _ENABLED_LOGS = dict()
 
     # The max length of the firehose stream name is 64. For streamalert data firehose,
-    # we reserve 5 chars to have `data_` as part of prefix. Please refer to
+    # we reserve 12 chars to have `streamalert_` as part of prefix. Please refer to
     # terraform/modules/tf_kinesis_firehose_delivery_stream/main.tf
-    FIREHOSE_NAME_MAX_LEN = 59
+    FIREHOSE_NAME_MAX_LEN = 52
 
     FIREHOSE_NAME_HASH_LEN = 8
 
@@ -318,8 +318,9 @@ class FirehoseClient:
         longer than 64 characters
 
         Args:
-            prefix (str): TODO
-            log_stream_name (str): TODO
+            use_prefix (bool): Does apply prefix defined in conf/global.json to firehose stream name
+            prefix (str): The prefix defined in conf/global.json to firehose stream name
+            log_stream_name (str): The name of the log from conf/logs.json or conf/schemas/*.json
 
         Returns:
             str: compliant stream name
--- a/streamalert/classifier/clients/firehose.py
+++ b/streamalert/classifier/clients/firehose.py
@@ -313,8 +313,8 @@ class FirehoseClient:
         return re.sub(cls.SPECIAL_CHAR_REGEX, cls.SPECIAL_CHAR_SUB, log_name)
 
     @classmethod
-    def generate_firehose_stream_name(cls, use_prefix, prefix, log_stream_name):
-        """Generate stream name complaint to firehose naming restriction, no
+    def generate_firehose_suffix(cls, use_prefix, prefix, log_stream_name):
+        """Generate suffix of stream name complaint to firehose naming restriction, no
         longer than 64 characters
 
         Args:
@@ -323,7 +323,7 @@ class FirehoseClient:
             log_stream_name (str): The name of the log from conf/logs.json or conf/schemas/*.json
 
         Returns:
-            str: compliant stream name
+            str: suffix of stream name
         """
 
         reserved_len = cls.FIREHOSE_NAME_MAX_LEN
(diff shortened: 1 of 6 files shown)


streamalert/classifier/clients/firehose.py
```diff
@@ -52,19 +53,32 @@ class FirehoseClient:
     MAX_RECORD_SIZE = 1000 * 1000 - 2
 
     # Default firehose name format, should be formatted with deployment prefix
-    DEFAULT_FIREHOSE_FMT = '{}streamalert_data_{}'
+    DEFAULT_FIREHOSE_FMT = '{}data_{}'
```
ryandeivert: IMO I'd rather keep the `streamalert` portion of this and drop the `data` bit.. these are `Firehose Data Delivery Streams` and including `data` is pretty redundant. at least having `streamalert` in the name will help with things like filtering streams in console, etc

Ryxias: I mentioned this before, but I recommend explicitly stating the firehose names in the `conf/` directory, and explicitly populating them during the `generate` step. I imagine the workflow is like:

* Look in conf directory for explicit firehose name
* If no explicit name, use `use_prefix` and the log type
* During `manage.py generate`, explicitly populate into Terraform json files
* Instead of generating them using `locals {$prefix}_data_{$hash}`, just use the explicit name

chunyong-lin: @Ryxias quick question.
> Look in conf directory for explicit firehose name

Potentially, we may have hundreds of firehose up running due to one log schema using one firehose. The firehose conf will be quite long and add complexity to read all firehose names from the `conf/global.json`.  IMO, I would take @ryandeivert 's suggestion. Thoughts?

ryandeivert: I do like Derek's idea, I'm just unsure where this would live.. in `global.json`? how do we map from `logs.json` to this reliably... or does it live in `logs.json`?? any suggestions for _how_ this would actually look?

ryandeivert: I also **still** insist we drop `_data_`, even with the hashing route. really not sure why we're hung up on keeping it...

Ryxias: Doing it my way will not require a re-deploy of Firehoses, as you can simply specify the original Firehose name.

But if you **need** to redeploy the Firehoses anyway because of the Terraform upgrade from `s3` to `extended_s3`, then I suppose it's fine to do it your way.

blakemotl: I think having to manually specify all of the firehose names for each schema is going to take as much time if not more, than to just recreate the firehoses. Also v3 technically isn't "production" yet so I think a breaking change is ok.

chunyong-lin: @Ryxias you are right, the firehoses will be recreated when they are upgraded from `s3` to `extended_s3`.
Since @Ryxias 's idea is getting popular, we probably can have firehose custom name in the `conf/schems/*.json` because firehose name is tightly linked to schema name.

ryandeivert: @Ryxias again I'm not opposed to your way, I just want to hear more about "how" this will live in the config. does it go in logs.json or the firehose config of global.json?

Ryxias: We can have it in `conf/logs.json`, or `conf/schemas/*.json or `conf/firehose.json`. I don't really care either way. 

I think the important one is that it's OPTIONAL. In most cases, you **_should not explicitly specify_** your firehoses. But they should be explicitly stated somewhere. I don't care much about the `conf/****.json` one, but I am very strongly in favor of explicitly generating them in the `terraform/***.tf.json` files, instead of dynamically creating them inside of the the Terraform modules.

Ryxias: It would be mean your `manage.py generate` step would just have a new function that's like:


```
def get_firehose_name(log_type):
  # Explicitly stated
  if config['whatever'].get(log_type, {}).get('firehose_name', None):
    return config['whatever'].get(log_type, {}).get('firehose_name', None)

  # Generated
  use_prefix = config['global']['whatever']['use_prefix']

  firehose_name = '{}{}_{}'.format(
    use_prefix ? '{}_'.format(prefix),
    'streamalert_data' # Or whatever
    log_schema
  )

  if len(firehose_name) > 64:
    # then do some hashing stuff

  return firehose_name
```

ryandeivert: I'm drawing attention to the _where_ on this because I think we're forgetting how this is currently performed. data retention is _optional_ and does not need to be enabled.. but enabling means toggling it in `global.json`. having _some_ settings in global.json and others in logs.json is confusing.. but on the other hand, having the name for an optional feature embedded in logs.json and other settings in global.json is also confusing. we're sorta asking a lot of users here to mind-map all of this together

streamalert/classifier/clients/firehose.py
```diff
@@ -298,6 +312,46 @@ def firehose_log_name(cls, log_name):
         """
         return re.sub(cls.SPECIAL_CHAR_REGEX, cls.SPECIAL_CHAR_SUB, log_name)
 
+    @classmethod
+    def generate_firehose_stream_name(cls, use_prefix, prefix, log_stream_name):
```
Ryxias: This method's name seems to imply that it generates the full firehose name, but that's actually not true right? It generates the **prefix-less** firehose name.

I suggest renaming the function to `generate_firehose_suffix` reduce confusion

chunyong-lin: Good point! Yes, you are right, the return result doesn't include prefix. Will rename this method.

streamalert/classifier/clients/firehose.py
chunyong-lin: Wrap up this thread. I took Ryan's suggestion and addressed it in [this commit](https://github.com/airbnb/streamalert/pull/1191/commits/d4cd2d3493a4f0d883d7702f33de6c766c28fe71). We will support the custom firehose stream name later and I have open an issue #1193 to track the future work.

PR #1194, opened 2020-03-17
Move Terraform Files to Support Packaging

blakemotl:
to: @ryandeivert @Ryxias @chunyong-lin 

## Background
To support packaging of StreamAlert with setup.py the terraform files need to be shipping with the CLI.

## Changes

* Move terraform files to a new streamalert_cli/_infrastructure folder
* Clean up some tech debt related to the terraform folder location

## Testing
Fully deployed and destroyed StreamAlert in a test environment.


blakemotl:
I realized after submitting this that, some of my changes may not make sense. If we package StreamAlert, _infrastructure won't exist right? So we need another, potentially configurable location, to store generated terraform files. Thoughts?

chunyong-lin:
Do we consider to move this change to release-3-2-0 or a feature branch? It is a breaking change.

blakemotl:
We talked about making it 4.0 but I'm good with doing 3.2

ryandeivert:
@chunyong-lin what makes this a breaking change?

ryandeivert:
@blakemotl - the `_infrastructure` directory can be packaged along with the python files, like I described before. we'd leverage the `package_data` part of setuptools for this: https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files

chunyong-lin:
At the first glance, I thought it would change the way we deploy the different functions in StreamAlert. After looking at the file locations, looks it won't change the deployment.

ryandeivert:
@blakemotl please add labels and a milestone to this

ryandeivert:
> I realized after submitting this that, some of my changes may not make sense. If we package StreamAlert, _infrastructure won't exist right? So we need another, potentially configurable location, to store generated terraform files. Thoughts?

eventually we will use `pkg_resources` for accessing these packaged files.. something like this should get the path:

```python
pkg_resources.resource_filename(__name__, '_infrastructure')
```

we'll then have to copy them to tmp or something to work with

streamalert_cli/_infrastructure/main.tf.json
```diff
@@ -0,0 +1,203 @@
+{
```
ryandeivert: please remove this and update these lines to ignore these files:

https://github.com/airbnb/streamalert/blob/7d198a3273781f66465420e90886a3ce53ec7559/.gitignore#L8-L12

streamalert_cli/helpers.py
```diff
@@ -49,8 +50,7 @@ def run_command(runner_args, **kwargs):
     """
     default_error_message = "An error occurred while running: {}".format(' '.join(runner_args))
     error_message = kwargs.get('error_message', default_error_message)
-    default_cwd = 'terraform'
-    cwd = kwargs.get('cwd', default_cwd)
+    cwd = kwargs.get('cwd', DEFAULT_TERRAFORM_FILES_PATH)
```
ryandeivert: I'd remove this constant in favor of doing this programmatically with:

```python
default_cwd = os.path.join(os.path.dirname(os.path.realpath(__file__)), '_infrastructure')
```

streamalert_cli/manage_lambda/package.py
```diff
@@ -21,10 +21,8 @@
 
 from streamalert.shared.logger import get_logger
 from streamalert_cli.helpers import run_command
+from streamalert_cli.terraform.common import DEFAULT_TERRAFORM_FILES_PATH
 
-# Build .zip files in the top-level of the terraform directory
-THIS_DIRECTORY = os.path.dirname(os.path.realpath(__file__))
-BUILD_DIRECTORY = os.path.join(THIS_DIRECTORY, '..', '..', 'terraform')
```
ryandeivert: I think we should prob keep these and make it:

```python
THIS_DIRECTORY = os.path.dirname(os.path.realpath(__file__))
BUILD_DIRECTORY = os.path.join(THIS_DIRECTORY, '..', '_infrastructure')

File: terraform/providers.tf [hunk 4]
```diff
@@ -1,84 +0,0 @@
-// This root terraform file provides aliased providers for all known AWS regions
```
ryandeivert: this file is needed - please don't delete it1

PR #1209, opened 2020-03-25
Make scheduled_queries/ directory configurable from global.json

Ryxias:
to: @ryandeivert 

### Changes

It's now configurable under the node `scheduled_query_locations` in the `global.json` file

### DIFF
--- a/streamalert/scheduled_queries/query_packs/configuration.py
+++ b/streamalert/scheduled_queries/query_packs/configuration.py
@@ -15,8 +15,6 @@ limitations under the License.
 """
 from streamalert.shared.importer import import_folders
 
-PACKS_DIRECTORY = 'scheduled_queries/'
-
 
 class QueryPackConfiguration:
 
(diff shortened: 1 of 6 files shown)


streamalert/scheduled_queries/query_packs/configuration.py
```diff
@@ -112,5 +110,6 @@ def register(cls, config):
         cls.QUERY_PACKS[name] = config
 
     @classmethod
-    def load_packs(cls):
-        import_folders(PACKS_DIRECTORY)
+    def load_packs(cls, directories):
+        for dir in directories:
```
ryandeivert: `import_folders` takes an arbitrary # of args (using `*args`).. all you have to do here is use list expansion:

```python
def load_packs(cls, directories):
    import_folders(*directories)
```

ryandeivert:
@Ryxias would you also mind adding an update to docs? http://streamalert.io/config-global.html#general

PR #1215, opened 2020-03-31
official release of v3.1.1

ryandeivert:


Ryxias:
We aren't ever going to have these many back-to-back releases in quick succession ever again, are we?

Issue #1144, opened 2020-02-20
Migrate athena_partition_refresh lambda function to tf_lambda module


chunyong-lin:
related to: #1090

## Background
We will keep `athena_partition_refresh` lambda function forverrrrrrrr. So we need to migrate the function to use `tf_lambda` module to take advantage of code reuse and predefined alarms.

## Desired Change
Migrate the function to use [terraform/module/tf_lambda](https://github.com/airbnb/streamalert/tree/release-3-1-0/terraform/modules/tf_lambda)

ryandeivert:
complete with #1217

Issue #1090, opened 2020-01-16
Unstable Terraform configuration, keeps switching between two states.


dgrenner:
depends on: #1144 

## Background

I'm using the release-3-0-0 branch, and I am getting a conflict between Terraform resources (I do not know if the same thing happens on master). Every time I run .`/manage.py build` the configuration switches between two states. I get the same result with `terraform apply` - so the issue is not due to the python code changing the Terraform files.

### Description

The output from Terraform is:
```
# module.cloudwatch_monitoring_prod.aws_cloudwatch_metric_alarm.streamalert_lambda_throttles[0] will be updated in-place
  ~ resource "aws_cloudwatch_metric_alarm" "streamalert_lambda_throttles" {
        actions_enabled           = true
      ~ alarm_actions             = [
          + "arn:aws:sns:eu-north-1:123123123123:sr_streamalert_monitoring",
        ]
        ...
      ~ ok_actions                = [
          + "arn:aws:sns:eu-north-1:123123123123:sr_streamalert_monitoring",
        ]
      ~ period                    = 120 -> 300
        ...
      ~ tags                      = {
          - "Cluster" = "prod" -> null
            "Name"    = "StreamAlert"
        }
        ...
    }
```
If I perform the actions, then re-run the command I get:
```
  # module.classifier_prod_lambda.aws_cloudwatch_metric_alarm.lambda_throttles[0] will be updated in-place
  ~ resource "aws_cloudwatch_metric_alarm" "lambda_throttles" {
        actions_enabled           = true
      ~ alarm_actions             = [
          - "arn:aws:sns:eu-north-1:123123123123:sr_streamalert_monitoring",
        ]
        ...
      ~ ok_actions                = [
          - "arn:aws:sns:eu-north-1:123123123123:sr_streamalert_monitoring",
        ]
      ~ period                    = 300 -> 120
        ...
      ~ tags                      = {
          + "Cluster" = "prod"
            "Name"    = "StreamAlert"
        }
        ...
    }
```
So the state keeps switching between these two settings.
The same thing (with same changes to alarm_actions, ok_actions, period and tags) happens with
```
 # module.cloudwatch_monitoring_prod.aws_cloudwatch_metric_alarm.streamalert_lambda_invocation_errors[0] will be updated in-place
  ~ resource "aws_cloudwatch_metric_alarm" "streamalert_lambda_invocation_errors" {
```

[...]

PR #1217, opened 2020-04-01
migrating Athena function to use tf_lambda module

ryandeivert:
resolves: #1144 
resolves: #1090

## Background

Need to squash crap tech debt.

## Changes

* Renaming the Athena Partition Refresh function to simply "Athena Partitioner" .. cause what exactly is being refreshed here anyway..?
* Updating the `generate_athena` code to leverage the `tf_lambda` module.
* Updating the `tf_athena` module to be used for IAM/misc resources only
* Changing references to the config name from `athena_partition_refresh_config` --> `athena_partitioner_config`

## Testing

Updating all generation tests.


streamalert/shared/athena.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, database_name, results_bucket, results_prefix, region=None):
         if not results_bucket.startswith('s3://'):
             results_bucket = 's3://{}'.format(results_bucket)
 
-        # Produces athena_partition_refresh/YYYY/MM/DD S3 keys
+        # Produces athena_partitioner/YYYY/MM/DD S3 keys
```
Ryxias, line 62: This comment seems wrong? It looks like it's supposed to reference the athena query results bucket.

ryandeivert: yeah I'm unsure where this came from initial .. find/replaced here but will address

ryandeivert:
@Ryxias / @chunyong-lin I added a commit to address feedback: https://github.com/airbnb/streamalert/pull/1217/commits/6f7e1ffefb2b033f1a07e12f46d4ce34ea58de88

PR #1227, opened 2020-04-06
Implements a v2 Lambda Output with AssumeRole

Ryxias:
to: @ryandeivert @blakemotl @chunyong-lin 

## Background

We needed to be able to invoke Lambda across AWS accounts. The problem is the existing Lambda output seems to be.. wrong. It doesn't derive configs from the right place.

Instead of breaking reverse compatibility, I just added a v2 that is easier to set up and is more consistent with existing implementations. This v2 also has the ability to make `sts:AssumeRole` calls, in order to make Lambda invocations across accounts. However, it _assumes_ that you already set up the IAM Policies correctly.

## Testing

Tested on stage; built a test lambda and was able to invoke it.

### DIFF
--- a/conf/outputs.json
+++ b/conf/outputs.json
@@ -2,6 +2,9 @@
   "aws-lambda": {
     "sample-lambda": "function-name:qualifier"
   },
+  "aws-lambda-v2": [
+    "sample-lambda"
+  ],
   "aws-s3": {
     "bucket": "aws-s3-bucket"
   },
--- a/docs/source/outputs.rst
+++ b/docs/source/outputs.rst
@@ -158,7 +158,7 @@ The ``OutputProperty`` object used in ``get_user_defined_properties`` is a ``nam
 
 :cred_requirement:
   A ``boolean`` that indicates whether this value is required for API access with this service. Ultimately, setting this value to ``True`` indicates
-  that the value should be encrypted and stored in Amazon S3.
+  that the value should be encrypted and stored in Amazon Systems Manager.
   Default is: ``False``
 
 
--- a/docs/source/outputs.rst
+++ b/docs/source/outputs.rst
@@ -44,9 +44,9 @@ Adding a new configuration for a currently supported service is handled using ``
 .. note::
 
   ``<SERVICE_NAME>`` above should be one of the following supported service identifiers.
-  ``aws-cloudwatch-log``, ``aws-firehose``, ``aws-lambda``, ``aws-s3``, ``aws-sns``, ``aws-sqs``,
-  ``carbonblack``, ``github``, ``jira``, ``komand``, ``pagerduty``, ``pagerduty-incident``,
-  ``pagerduty-v2``, ``phantom``, ``slack``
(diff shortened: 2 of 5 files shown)


conf/outputs.json
```diff
@@ -2,6 +2,9 @@
   "aws-lambda": {
     "sample-lambda": "function-name:qualifier"
   },
+  "aws-lambda-v2": [
```
ryandeivert, line 5: IIRC, this is not actually how this info will be written to the `outputs.json` config file for `AWSOutput` types .. see here:

https://github.com/airbnb/streamalert/blob/41da6b5514b4949c04cab90aa85631fa833634d0/streamalert/alert_processor/outputs/aws.py#L50-L69

did you confirm that this actually works as expected with the `manage.py outputs new ... `command?

Ryxias: Oh wow, no this is totally not gonna work. Good catch. No I didn't confirm it as I'm still testing it, but it's good to know that it's definitely wrong. 

I think the easy fix is to just not inherit from this base class and just use the `OutputDispatcher` base class like any other sane output.

ryandeivert: agreed - we can trend away from the AWSOutput class usage for future aws outputs now that ssm is a thing

Ryxias:
I got this to work with Terraform configurations that look like:

```
# Module for cross-account assume role policies
resource "aws_iam_role_policy" "assume_role_policy_on_alert" {
  name   = "AssumeRoleOnTestLambdaAccount"
  role   = module.alert_processor_lambda.role_id
  policy = data.aws_iam_policy_document.allow_assume_role.json
}

data "aws_iam_policy_document" "allow_assume_role" {
  statement {
    effect = "Allow"

    actions = [
      "sts:AssumeRole"
    ]

    resources = ["arn:aws:iam::009715504418:role/service-role/derek_test_func-role-k29ynxjk"]
  }
}

resource "aws_iam_role" "test_role" {
  name   = "derek_test_func-role-k29ynxjk"
  path   = "/service-role/"
  assume_role_policy = data.aws_iam_policy_document.instance-assume-role-policy.json
}

data "aws_iam_policy_document" "instance-assume-role-policy" {
  statement {
    effect = "Allow"
    actions = ["sts:AssumeRole"]

    principals {
      type        = "Service"
      identifiers = ["lambda.amazonaws.com"]
    }
  }

  statement {
    effect = "Allow"
    actions = ["sts:AssumeRole"]

    principals {
      type = "AWS"
      # identifiers = ["arn:aws:iam::123456789012:role/streamalert/derek20200406_streamalert_alert_processor_role"]
      identifiers = ["arn:aws:sts::123456789012:assumed-role/derek20200406_streamalert_alert_processor_role/derek20200406_streamalert_alert_processor"]
    }
  }
}

resource "aws_iam_role_policy" "invoke_test_func" {
  name   = "AssumeRoleOnTestLambdaAccount"
  role   = aws_iam_role.test_role.id
  policy = data.aws_iam_policy_document.invoke_policy.json
}

data "aws_iam_policy_document" "invoke_policy" {
  statement {
    effect = "Allow"
    actions = [
      "lambda:InvokeFunction"
    ]
    resources = [
      "arn:aws:lambda:us-east-1:123456789012:function:derek_test_func"
    ]
  }
}

```

PR #1229, opened 2020-04-07
adding terraform references for some buckets

ryandeivert:

## Background

A race condition could cause errors like:

```
Error: Error putting S3 notification configuration: NoSuchBucket: The specified bucket does not exist
	status code: 404, request id: 5A5ACCB6EB4055B8, host id: vIlTAABEslxBZXQXmF5QRhFi6z4kQ0wW2YUeJdMLKjGpJn4u5irjwctlzAPPDIY1RKsktnKyh5s=
  on modules/tf_athena/main.tf line 100, in resource "aws_s3_bucket_notification" "bucket_notification":
 100: resource "aws_s3_bucket_notification" "bucket_notification" {
...
```

## Changes

* Using references to terraform resources/outputs for bucket names to enforce the creation of the resources before we try to use their "value".

## Testing

Updates to unit tests.

### DIFF
--- a/streamalert/shared/config.py
+++ b/streamalert/shared/config.py
@@ -117,8 +117,10 @@ def firehose_alerts_bucket(config):
 
 def athena_partition_buckets(config):
     """Get the buckets from default buckets and additionally configured ones
+
     Args:
         config (dict): The loaded config from the 'conf/' directory
+
     Returns:
         list: Bucket names for which Athena is enabled
     """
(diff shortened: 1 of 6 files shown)


ryandeivert:
@Ryxias a reference to a resource is an implicit `depends_on`.. actually, `depends_on` should rarely be needed at all (if ever in most of our cases). see [here](https://www.terraform.io/docs/configuration/resources.html#resource-dependencies):

> Most resource dependencies are handled automatically. Terraform analyses any expressions within a resource block to find references to other objects, and treats those references as implicit ordering requirements when creating, updating, or destroying resources. Since most resources with behavioral dependencies on other resources also refer to those resources' data, it's usually not necessary to manually specify dependencies between resources.

PR #1178, opened 2020-03-05
Working proof-of-concept for StreamAlert Normalization v2

Ryxias:
to: Nobody
cc: @airbnb/streamalert-maintainers

## Background

I got really interested in this personally and I just kinda hacked together something.

## THE BIG IDEA
Make `streamalert:normalization` searchable by dropping all Normalized data into a new Athena `artifacts` table.

## New components

* Firehose Data Transformation
* Firehose Data Transformer Lambda
* New Configuration (?)
* New Firehose for Artifacts (Not done)
* New Athena Table (Note done)

## Testing

The prototype works on stage. But it has a lot more to go. Configuration.... etc

PR #1230, opened 2020-04-08
Normalization reboot - Artifacts inventory

chunyong-lin:
related to: Mostly copy&Pasta @Ryxias 's [Working proof-of-concept for StreamAlert Normalization v2](https://github.com/airbnb/streamalert/pull/1178/commits/dc047d0898b19e07a85667e49da06ccd47e23a5e)
resolves:

## Background
Quote our intention to reboot Normalization feature from [the documentation page](https://github.com/airbnb/streamalert/blob/319bc614f99d28f27d0da488a978c4acf3f4c1f9/docs/source/normalization.rst).

> StreamAlert has an unannounced feature Data Normalization. In its current implementation, it extracts recognized field names from classified records, and saves them to a top-level key on the same record.
> 
> This is useful for rules, as they can be written to compare data fields against IoCs, such as IP Address, instead of writing one rule for each incoming data type. However, there are couple limitations we have identified as we use Normalization internally for a while.

Please look at [the documentation page](https://github.com/airbnb/streamalert/blob/319bc614f99d28f27d0da488a978c4acf3f4c1f9/docs/source/normalization.rst) for more information.

The feature implementation will be shipped by different PRs based on same feature branch `feature-artifact-extractor`
1. Implement Lambda code  based on current Normalization logic. (Current PR)
2. Add terraform resources, including a dedicated Firehose delivering artifacts to S3, enable firehose data transformation configuration on the Firehoses where the artifacts will be extracted from and required permissions.
3. Change normalization configuration logic to add additional information and more flexible settings. This will require to update classifier code during normalization process.
4. Maybe more PRs ... 🤷‍♀ 

## Changes

* Add a new Lambda Function `Artifact Extractor` to extract artifacts from records processed by classifiers.
* Add unit test cases
* Move `FirehoseClient` class to `streamalert/shared/` directory to reuse many goodies.
* Add Normalization doc. More doc to come in the future PRs including the setup and deployment instruction.

## Testing
* Add unit test cases to test the code change
* Note: this lambda has not tested in staging. It will be tested in staging in next PR.

## Appreciation
captain


docs/source/normalization.rst
```diff
@@ -0,0 +1,51 @@
+#############
```
Ryxias, line 1: This doc sounds more like a design doc than a OS usage documentation. Maybe save this for a `release-3-3-0` release notes?

streamalert/artifact_extractor/artifact_extractor.py
```diff
@@ -0,0 +1,238 @@
...
+            list: A list of Artifacts from a normalized record.
+        """
+        artifacts = []
+
+        if Normalizer.NORMALIZATION_KEY not in self._decoded_record:
+            # Return an empty list if the record doesn't have normalization information.
+            return artifacts
+
+        if not self._source_type:
+            # Return immediately if can not identify source_type. a.k.a do not extract artifacts.
```
Ryxias, line 109: Maybe log a warning?

streamalert/artifact_extractor/main.py
```diff
@@ -0,0 +1,67 @@
...
+            event['deliveryStreamArn']
+        ).run(event.get('records', []))
+    except Exception:
+        # FIXME: (Optional) Add retry for Timeout exceptions. If the Lambda function invocation
+        # fails because of a network timeout or the lambda invocation limit, Kinesis Data Firehose
+        # retries the invocation three times by default. If the invocation does not succeed, Kinesis
+        # Data Firehose then skips that batch of records. The skipped records are treated as
+        # unsuccessfully processed records.
+        # https://docs.aws.amazon.com/firehose/latest/dev/data-transformation.html
+        print(event)
```
Ryxias, line 65: Extraneous `print()`

chunyong-lin: Oooops 😆

docs/source/normalization.rst
chunyong-lin: LOL, yeah. It is on my todo list to add configuration and deployment instructions in following two PRs. Then this page won't look just like a design doc.

streamalert/artifact_extractor/artifact_extractor.py
chunyong-lin: 👍 It generates a warning message in `def _get_source_type(self):` method when parse the source type from the first place.

docs/source/normalization.rst
ryandeivert: agree w/ derek here - remove these and set aside for later. I'd focus on docs last for this feature because of the likelihood to change before being finalized

PR #1236, opened 2020-04-10
official release of v3.2.1

ryandeivert:
to :@airbnb/streamalert-maintainers


streamalert_cli/terraform/lambda_module.py
```diff
@@ -121,9 +124,12 @@ def generate_lambda(function_name, handler, lambda_config, config, environment=N
         'tags': tags or {},
     }
 
+    if kwargs.get('include_layers', False):
```
KevinHock, line 127: Super Nit: like line 131 `, False)` doesn't seem needed.

PR #1237, opened 2020-04-14
Normalization reboot - Add terraform resources

chunyong-lin:

related to: #1230 
resolves:

## Background
This PR is to add terraform resources for Normalization v2 including
* Create a new dedicated Firehose to deliver artifacts to s3 bucket if Artifact Extractor enabled
* Create an Athena table for artifacts if Artifact Extractor enabled
* Attach proper permissions to invoke Artifact Extractor lambda function for Firehoses if it is enabled
* Enable `processing_configuration` on the firehose where enables normalization configuration in its schema configuration field
* Add Artifact Extractor lambda terraform code and proper permissions.

## Testing
* Added unit test cases to cover the code change
* `python manage.py build` works as expected in staging for following cases
  1. With default conf, Artifact Extractor disabled, no resource changes
  2. Enable `Artifact Extractor` in `conf/lambda.json` and `firehose` in `conf/global.json`, `Artifact Extractor` lambda function, a new firehose and athena table for artifacts will be created.
  3. Based on case 2, add `normalization` dummy configuration to `osquery:differential` and `cloudwatch:cloudtrail` schemas, `processing_configuration` will be enabled to invoke `Artifact Extractor` lambda on `osquery_differential` and `cloudwatch_cloudtrail` FIrehoses. 

## Next Step
Refactor normalization configure and update classifier code related to normalization in next PR.

### DIFF
--- a/streamalert/artifact_extractor/artifact_extractor.py
+++ b/streamalert/artifact_extractor/artifact_extractor.py
@@ -42,11 +42,12 @@ class Artifact:
                 record processed by classifier. This field is very useful for cross reference back
                 to the original record in the future.
         """
-        self._function = kwargs.get('function', 'not_specified')
-        self._record_id = kwargs.get('record_id', 'RESERVED')
-        self._source_type = source_type
-        self._type = normalized_type
-        self._value = value
+        # Enforce all fields are strings in a Artifact to prevent type corruption in Parquet format
+        self._function = str(kwargs.get('function', 'not_specified'))
+        self._record_id = str(kwargs.get('record_id', 'RESERVED'))
+        self._source_type = str(source_type)
+        self._type = str(normalized_type)
+        self._value = str(value)
 
     @property
     def record(self):
--- a/streamalert/artifact_extractor/artifact_extractor.py
+++ b/streamalert/artifact_extractor/artifact_extractor.py
@@ -26,6 +26,8 @@ LOGGER = get_logger(__name__)
 class Artifact:
     """Encapsulation of a single Artifact that is extracted from an input record."""
 
+    RESERVED = 'RESERVED'
+
     def __init__(self, source_type, normalized_type, value, **kwargs):
         """Create a new Artifact based on normalized information
 
@@ -44,7 +46,7 @@ class Artifact:
         """
         # Enforce all fields are strings in a Artifact to prevent type corruption in Parquet format
         self._function = str(kwargs.get('function', 'not_specified'))
-        self._record_id = str(kwargs.get('record_id', 'RESERVED'))
+        self._record_id = str(kwargs.get('record_id', self.RESERVED))
         self._source_type = str(source_type)
         self._type = str(normalized_type)
         self._value = str(value)
--- a/streamalert/shared/config.py
+++ b/streamalert/shared/config.py
@@ -465,7 +465,7 @@ def artifact_extractor_enabled(config, log_name=None):
     # if log_name is empty, it means caller only want to know if artifact extractor lambda
     # function enabled or not, so return early.
     if not log_name:
-        return False
+        return True
     # FIXME: (cylin) pretty this line and add more comment
     log_config = config.get('logs', {}).get(log_name, {})
     return 'normalization' in log_config.get('configuration', {})
--- a/streamalert/shared/config.py
+++ b/streamalert/shared/config.py
@@ -497,6 +497,6 @@ def artifact_extractor_enabled(config, log_name=None):
     # function enabled or not, so return early.
     if not log_name:
         return True
-    
+
     log_config = config.get('logs', {}).get(log_name, {})
     return 'normalization' in log_config.get('configuration', {})
--- a/streamalert/shared/config.py
+++ b/streamalert/shared/config.py
@@ -445,14 +445,45 @@ def _validate_sources(cluster_name, data_sources, existing_sources):
 # FIXME (derek.wang) write a configuration validator for lookuptables (new one)
 
 def artifact_extractor_enabled(config, log_name=None):
-    """Determine if artifactor extractor enabled one a firehose
-    TODO: (cylin) docstring
+    """Validate if Artifactor Extractor enabled.
+    There are two cases need validate if Artifact Extractor enabled.
+    1. For deploy Artifact Extractor Lambda function.
+    2. To enable firehoses to invoke Artifact Extractor Lambda function.
+
+    For case 1, Artifact Extractor Lambda function will be created and deployed if both
+    "artifact_extractor_config" in conf/lambda.json and "firehose" in conf/global.json both enabled.
+
+    For case 2, in addition to above two conditions, a firehose will be setup to
+    "processing_configure" invoke Artifact Extractor lambda function if "normalization" is
+    configured in its schema configuration.
+    For example, the mapped firehose for following log type will setup:
+        {
+            "some_log_type": {
+                "schema": {
+                    "key1": "string",
+                    "key2": "string"
+                },
+                "parser": "json",
+                "configuration": {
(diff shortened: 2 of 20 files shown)


streamalert/artifact_extractor/artifact_extractor.py
```diff
@@ -42,11 +42,12 @@ def __init__(self, source_type, normalized_type, value, **kwargs):
...
                 to the original record in the future.
         """
-        self._function = kwargs.get('function', 'not_specified')
-        self._record_id = kwargs.get('record_id', 'RESERVED')
-        self._source_type = source_type
-        self._type = normalized_type
-        self._value = value
+        # Enforce all fields are strings in a Artifact to prevent type corruption in Parquet format
+        self._function = str(kwargs.get('function', 'not_specified'))
+        self._record_id = str(kwargs.get('record_id', 'RESERVED'))
```
ryandeivert, line 47: can this `RESERVED` string literal be moved to a class constant instead and referenced like `cls.RESERVED` or `self.RESERVED`? and add a comment as to its purpose? I'm not a big fan of string literals just floating around

chunyong-lin: 👍 will make the change!

streamalert/shared/firehose.py
```diff
@@ -334,6 +334,33 @@ def generate_firehose_name(cls, prefix, log_stream_name):
...
+        ).get('firehose_stream_name')
+
+        if stream_name:
+            # support custom firehose stream name of Artifacts. User should make sure the length of
+            # the custom firehose name is no longer than 64 chars, otherwise the firehose will be
+            # failed to create. StreamAlert is not responsible for checking for custom firehose name
+            # since it should not change custom settings.
+            return stream_name
+
+        return cls.generate_firehose_name(
```
ryandeivert, line 358: remove the above `if stream_name:` check just do:

```suggestion
        return stream_name or cls.generate_firehose_name(
```

streamalert_cli/_infrastructure/modules/tf_artifact_extractor/iam.tf
```diff
@@ -0,0 +1,109 @@
...
+    resources = [
+      var.function_alias_arn
+    ]
+  }
+}
+
+// IAM Role: Artifacts Firehose Delivery Stream permissions
+resource "aws_iam_role" "streamalert_kinesis_firehose" {
+  name               = "${var.prefix}_firehose_artifacts_delivery"
+  path               = "/streamalert/"
```
ryandeivert, line 28: thank you for paying attention :)

streamalert_cli/_infrastructure/modules/tf_artifact_extractor/main.tf
```diff
@@ -0,0 +1,91 @@
...
+          open_x_json_ser_de {}
+        }
+      }
+      output_format_configuration {
+        serializer {
+          parquet_ser_de {}
+        }
+      }
+      schema_configuration {
+        database_name = var.glue_catalog_db_name
```
ryandeivert, line 43: if you reference the resource that will be created, you do not need the `depends_on` block below:

```suggestion
        database_name = aws_glue_catalog_table.artifacts.database_name
```

try to reference attributes of resources whenever possible!! instead of using `var` references. it helps terraform understand dependencies

streamalert_cli/_infrastructure/modules/tf_artifact_extractor/main.tf
```diff
@@ -0,0 +1,91 @@
...
+        role_arn      = aws_iam_role.streamalert_kinesis_firehose.arn
+        table_name    = var.glue_catalog_table_name
```
ryandeivert, line 45: same goes with this -  reference the resource:

```suggestion
        table_name    = aws_glue_catalog_table.artifacts.name
```

streamalert/artifact_extractor/artifact_extractor.py
Ryxias: It's meant to be temporary. Honestly I would just make it `None` and put a `FIXME`/`TODO` there

chunyong-lin: I have comment about `record_id at line 41 https://github.com/airbnb/streamalert/blob/ebba5c51db122be5747def5fb4056076f01dd050/streamalert/artifact_extractor/artifact_extractor.py#L41-L43

streamalert/shared/firehose.py
chunyong-lin: Neat!

streamalert/artifact_extractor/artifact_extractor.py
ryandeivert: I noticed, but it doesn't really help when it's not near the actual thing it relates to. putting this as a class constant gives you the ability to at a relative comment right next to it

streamalert_cli/_infrastructure/modules/tf_artifact_extractor/iam.tf
chunyong-lin: 😄 copy&pasta from somewhere.

ryandeivert: if copy/pasta is right then that's not a bad thing :P

streamalert_cli/_infrastructure/modules/tf_artifact_extractor/main.tf
chunyong-lin: Nice nice, I didn't know we can get database_name by getting a reference from a table. Pro!! Will make the change.

ryandeivert: remember to remove the `depends_on` block once you make this change

streamalert/shared/config.py
```diff
@@ -443,3 +443,60 @@ def _validate_sources(cluster_name, data_sources, existing_sources):
             existing_sources.add(source)
 
 # FIXME (derek.wang) write a configuration validator for lookuptables (new one)
+
+def artifact_extractor_enabled(config, log_name=None):
```
Ryxias, line 447: I recommend two functions:

```
def artifact_extractor_enabled_for_log(config, log_name):
  return _artifact_extractor_enabled_helper(config, log_name=log_name)

def artifact_extractor_enabled(config):
  return _artifact_extractor_enabled_helper(config, log_name=None)
```

I'm not a fan of optional arguments that alter the control flow of functions. This is acceptable for private methods used to DRY out blocks of code but it's better to be explicit (through the function public interface) exactly what a function is intended for

streamalert_cli/_infrastructure/modules/tf_artifact_extractor/iam.tf
```diff
@@ -0,0 +1,109 @@
...
+      "kms:Decrypt",
+      "kms:GenerateDataKey*",
+    ]
+
+    resources = [var.kms_key_arn]
+  }
+}
+
+// IAM Policy: Interact with the Glue Catalog
+resource "aws_iam_role_policy" "streamalert_firehose_glue" {
```
Ryxias, line 91: Remind me ... why does the Firehose transform function need to interact with glue?

chunyong-lin: Firehose will save data (artifacts) in parquet for athena searching and it needs to know the schema which is saved in glue catalog.

streamalert/shared/config.py
chunyong-lin: 🐟 Okay, this is your 2nd request of changes when I use a optional arguments. I will address your concern this time.

PR #1239, opened 2020-04-27
Demisto Playbooks

Ryxias:
to: @blakemotl @ryandeivert @chunyong-lin 

## Background

Adds ability for Demisto to specify playbook + sample rule, publisher, and test.

## Testing

Kudos to @blakemotl for owning most of the testing here and discovering this playbook field. I just copy pasta'd his code.

### DIFF
--- a/publishers/sample/sample_demisto.py
+++ b/publishers/sample/sample_demisto.py
@@ -1,0 +1,55 @@
+"""
+Batch of example publishers usable with Demisto.
+"""
+from streamalert.shared.publisher import Register
+
+
+@Register
+def demisto_classification(alert, publication):
+    """
+    This publisher appropriately sets the demisto incident type and playbook.
+
+    It first looks into the alert's context for the "demisto" key, where individual rules can
+    explcitly specify the desired classification traits of the output alert.
+    """
+
+    # If a rule explicitly states Demisto information with the alert context, obey that
+    # The convention to follow is any key in this dict (example, "incident_type") is mapped
+    # directly onto the Demisto output magic keys (example, @demisto.incident_type)
+    if 'demisto' in alert.context:
+        for key, value in alert.context['demisto'].items():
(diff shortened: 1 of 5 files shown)


blakemotl:
Thanks for PRing this!

publishers/sample/sample_demisto.py
```diff
@@ -0,0 +1,55 @@
...
+    This publisher appropriately sets the demisto incident type and playbook.
+
+    It first looks into the alert's context for the "demisto" key, where individual rules can
+    explcitly specify the desired classification traits of the output alert.
+    """
+
+    # If a rule explicitly states Demisto information with the alert context, obey that
+    # The convention to follow is any key in this dict (example, "incident_type") is mapped
+    # directly onto the Demisto output magic keys (example, @demisto.incident_type)
+    if 'demisto' in alert.context:
```
chunyong-lin, line 19: We can use a const for `demisto` string which is referenced in multiple places. It is not only make code clean but also reduce the work if `demisto` renames itself in the near future.

publishers/sample/sample_demisto.py
```diff
@@ -0,0 +1,55 @@
...
+        for key, value in alert.context['demisto'].items():
+            output_key = '@demisto.{}'.format(key)
+            publication[output_key] = value
+
+        return publication
+
+    # If no context was explicitly declared, then we default to our global rules
+    for code in GLOBAL_CLASSIFIERS:
+        payload = code(alert)
+        if payload:
```
chunyong-lin, line 29: We can break the for loop if there is no payload. This way will make code look cleaner.
```
if not payload:
   break

for key, value in payload:
     output_key = '@demisto.{}'.format(key)
     publication[output_key] = value

     return publication

```

Ryxias:
@chunyong-lin FYI a lot of the code here are sample publishers/tests. They're here to demonstrate how to write publishers and tests.

PR #1238, opened 2020-04-17
Normalization reboot - Refactor normalization code

chunyong-lin:
to: @airbnb/streamalert-maintainers 
related to: #1237 #1230 
resolves:

## Background
This is 3rd PR to implement Normalization v2 feature. It is focus on refactor normalization code to use both Normalization v1 and v2 configuration and add additional information `function` to normalized values.

### New Support
The biggest change in this PR is the normalization configuration change in `conf/schemas/*.json`. In StreamAlert schemas, provide a new configuration option:
```
“cloudtrail:something_logs”: {
  “schema”: {
    “Field1”: “string”,
    “Field2”: “string”,
    “Field3”: “string”
  },
  “parser”: “json”,
  “configuration”: {
    “normalization”: {
      -- NEW CONFIGURATION HERE --
      “ip_address”: [
        {
          “path”: [“path”, "to", “srcIpAddr”], # standard configuration including path to the key and function
          “function”: “source”
        },
        {
          “path”: ["path", "to", “dstIpAddr”], 
          “function”: “destination”
        }
      ],
      “hostname”: [“path”, “to”, “hostname”] # simplified configuration including path to the key only
    }
  }
}
```

### Archtecture
normalization-arch

### Deprecate Normalization v1
We deprecate Normalization v1 and `conf/normalized_types.json` will have no effect. Make sure migration our normalized configuration to `conf/schemas/*.json` along with log schemas.

## Changes
* Fixed couple bugs in #1237 which were discovered during staging testing
* Refactor normalization code to support normalization v2 
* Update existing test cases
* Add normalization configuration to `conf/schemas/carbonblack.json`, `conf/schemas/cloudwatch` and `conf/schemas/osquery.json`. Also update rule `right_to_left_character` to use normalization v2.
* Add more test cases to improve the code coverage
* Add architecture diagram to doc

## Testing
Deploy the changes to staging environment and created a kinesis stream for testing. Sent two fake cloudwatch events to the kinesis.
* Those two cloudwatch events showed up in `cloudwatch:events` table.
cloudwatch_events

* Artifacts extracted from those two cloudwatch events showed up in `artifacts` table.
artifacts

## Next Step
Still have one more PR to complete this feature that is to add advanced features to apply filters in normalization.

PR #1245, opened 2020-05-05
Normalization reboot - Add condition support to normalizer

chunyong-lin:
related to: #1178, #1230, #1238, #1237, #1242
resolves:

## Background
This PR is to add the optional block `condition` to normalization configure. With this support, we will have more flexibility to normalize only a subset of data. The normalizer will be skipped if the `condition` block presented and the condition is not met.

See [the updated doc](https://github.com/airbnb/streamalert/blob/fca035ed278d39ed2a98d037b9326c568d376f1e/docs/source/normalization.rst#configuration) for the usage of the `condition` block.

## Changes

* Upgrade terraform aws provider to `2.48.0` to avoid the bug [Unable to remove aws_kinesis_firehose_delivery_stream S3 processors](https://github.com/terraform-providers/terraform-provider-aws/issues/11305)
* Add `condition` block support
* Update the docs
* Add unit test cases

## Testing
* Unit test cases are passed
* Tested in staging account

### DIFF
--- a/streamalert/shared/normalize.py
+++ b/streamalert/shared/normalize.py
@@ -246,7 +246,7 @@ class Normalizer:
                 }
         """
         for param in paths_to_normalize.parsed_params:
-            if param.get('condition') and not cls._match_condition(record, param['condition']):
+            if param.get(CONST_CONDITION) and not cls._match_condition(record, param['condition']):
                 # If optional 'condition' block is configured, it will only extract values if
                 # condition is matched.
                 continue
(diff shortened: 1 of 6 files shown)


streamalert/shared/normalize.py
```diff
@@ -229,14 +246,12 @@ def _extract_values(cls, record, paths_to_normalize):
...
         for param in paths_to_normalize.parsed_params:
-            found_value = False
-            value = record
-            for key in param.get(CONST_PATH):
-                value = value.get(key)
-                if not value:
-                    found_value = False
-                    break
-                found_value = True
+            if param.get('condition') and not cls._match_condition(record, param['condition']):
```
Ryxias, line 249: `condition` —> `CONST_CONDITION`?

PR #1252, opened 2020-06-10
updating cli description for supplying test files

ryandeivert:
to: @chunyong-lin / @blakemotl 

## Background

There is a confusing description for a command line flag related to test files. 

## Changes

* Updating description and removing metavar to make it more obvious that this should be full paths, not file names.

## Testing

n/a


ryandeivert:
@chunyong-lin pls stamp

PR #1253, opened 2020-06-10
support for packaging user specified conf directory

ryandeivert:

## Background

The current `conf` directory is hard coded into the packaging logic. This should not be the case.

## Changes

* Updating packaging logic to zip up the user-specified config directory (which has a default value of `conf`)

## Testing

Adding unit test to check for copying to custom destination.

### DIFF
--- a/streamalert_cli/manage_lambda/package.py
+++ b/streamalert_cli/manage_lambda/package.py
@@ -26,14 +26,11 @@ LOGGER = get_logger(__name__)
 
 class LambdaPackage:
     """Build the deployment package for StreamAlert Lambdas"""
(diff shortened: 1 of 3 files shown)


streamalert_cli/manage_lambda/package.py
```diff
@@ -26,14 +26,11 @@
 
 class LambdaPackage:
     """Build the deployment package for StreamAlert Lambdas"""
-    package_name = 'streamalert'       # The basename of the generated .zip file
+    # The name of the directory to pacakge and basename of the generated .zip file
```
Ryxias, line 29: `pacakge`

PR #1257, opened 2020-06-11
another attempt to fix coverage in forks..

ryandeivert:
related to: #1256 

FML


ryandeivert:
this is actually it now I'm sure of it :)

PR #1260, opened 2020-06-15
Add additional G-Suite Admin Audit types.

gavinelder:

## Background

When configuring the GSuite collector I noticed a number of additional audit endpoints had been added and upon investigation I noticed some further additions by Google not present in SA, The following PR adds these. 

## Changes

Addition of the following Audit endpoints. 

1. [GCP](https://developers.google.com/admin-sdk/reports/v1/appendix/activity/gcp)
2. [Meet](https://developers.google.com/admin-sdk/reports/v1/appendix/activity/meet)
3. [SAML](https://developers.google.com/admin-sdk/reports/v1/appendix/activity/saml)

## Notes 

The SAML audit app was removed in https://github.com/airbnb/streamalert/pull/1046 as deprecated however I cannot see anything to suggest that Google has done this (This does not mean they have not undocumented changes is a normal google move) however I think this was replaced with the `token` app which has a similar feature set but is not a complete replacement. 

## Testing

Deployed confirmed ingestion, no rules written around these yet.


gavinelder:
Should add some documentation for this but to setup the Gsuite collector. 

Create a GSuite Service account with [G-Suite delegation rights](https://developers.google.com/admin-sdk/directory/v1/guides/delegation)

In the above guide replace the scopes with 
```
https://www.googleapis.com/auth/admin.reports.audit.readonly
```

PR #1263, opened 2020-06-17
[config] Add Okta log schema

themullinator:
to:
related to:
resolves:

## Background

Adds a schema for classifying Okta System LogEvent records based on https://developer.okta.com/docs/reference/api/system-log/#logevent-object-annotated-example

## Changes

* Adds `okta.json` under conf/schemas with an `okta:logevents` subtype

## Testing
Unit tests don't break


themullinator:
Looks like the tests actually passed, but the failing CI is due to this error:
```
coveralls.exception.CoverallsException: Not on TravisCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var.
##[error]Process completed with exit code 1.
```

gavinelder:
@themullinator if you target release 3.3.0 vs master and rebase onto that CI should pass.

chunyong-lin:
Hi @themullinator thank for the contribution. As @gavinelder pointed out, CI should be happy after change your based to `release-3-3-0`, I believe you can do it from GUI as well.
Also, can you add test event for validate the schema? It will be a good reference for the future users to understand what's the structure of Okta log. 

You don't need to write a rule, just to have a `*.json` file containing Okta log examples similar to [fleet test events](https://github.com/airbnb/streamalert/blob/cd5c1e7036352e47883928e7f0e046eb9517f578/rules/community/fleet/fleet_bad_action.json#L1-L25).

themullinator:
We're probably gonna need a squash commit, but finally got there in the end :) Thanks for the assist!

PR #1267, opened 2020-07-02
CLI support for extra user supplied terraform files

ryandeivert:

## Background

Users may have additional resources they'd like to create in terraform and track in the streamalert deployment state. We should support this concept.

## Changes

- adding cli arg (--terraform-file) to supply additional terraform config files. the flag can be used multiple times to supply more than one file (similar to the `terraform --target=... --target=....` logic)
- removing old tf cleanup code since temp path will be used for terraform configuration
  - Since runs should now be idempotent, a new temp file will be used every time the CLI is invoked
- adding cliconfig support for temp tf directory
- updates to tf_runner and run_command for temp tf path
- packaging change for tf temp path
- adding logic for copying files to tf temp path
- adding `terraform_files` section to `general` in config
- removing init backend option for CLI since backend init will have to happen every time the CLI is invoked now
  - [ ] TODO: decide if there is a better way for this since it could get annoying/slow?

## Testing

Updates to various unit tests

### DIFF
--- a/conf/global.json
+++ b/conf/global.json
@@ -5,6 +5,7 @@
     "region": "us-east-1"
   },
   "general": {
+    "terraform_files": [],
     "matcher_locations": [
       "matchers"
     ],
--- a/docs/source/config-global.rst
+++ b/docs/source/config-global.rst
@@ -61,6 +61,9 @@ Configuration
 
   {
     "general": {
+      "terraform_files": [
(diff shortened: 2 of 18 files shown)


blakemotl:
Since the paths are supplied via command line arguments won't the user have to include them every time they deploy/build?

ryandeivert:
@blakemotl good call, added a config option that gets merged with ones provided as cli args

PR #1274, opened 2020-07-22
[core][apps] Increase aliyun timeout

chunyong-lin:
related to:
resolves:

## Background
Increase default connection and read timeout to aliyun server to prevent `aliyun_actiontrail` app getting timed out. Aliyun SDK team separates timeout to connection and read timeout and also lower the default connection timeout time from 10 to 5 seconds ([Aliyun Python SDK PR 216](https://github.com/aliyun/aliyun-openapi-python-sdk/pull/216)), so we can increase both connection and read timeout.

## Changes

* Increase both connection and read timeout to 15 seconds.
* Raise exception instead of silent the timeout error and next app run should pick up the previous logs.

## Testing
* Updated unit test


chunyong-lin:
The Action CI failure is related to [this comment](https://github.com/actions/setup-python/issues/114#issuecomment-661906768)

> Hello everyone, this issue is related to image, not the setup-python task. We are working on investigation and fix. Progress can be tracked in actions/virtual-environments#1202 (comment)

I checked, our `PyYAML` package is pinned to `5.1.2` which kinda conflict with Action CI image where expects to have `5.3.1`. So we either wait for Action CI image to be fixed or upgrade our `PyYAML` to `5.3.1`. 🤷‍♀️

Issue #1289, opened 2020-08-23
[terraform] upgrade to terraform v0.13.0


ryandeivert:
## Background

Terraform version 13 [went GA](https://www.hashicorp.com/blog/announcing-hashicorp-terraform-0-13/) recently.

## Desired Change

We should update to the latest version to take advantage of new features. Specifically, the improve usage of `module`: https://www.hashicorp.com/blog/terraform-0-13-brings-powerful-meta-arguments-to-modular-workflows/

We can likely DRY out a lot of code by leveraging some of these features.

ryandeivert:
finished with #1290

PR #1290, opened 2020-08-23
updating to terraform v0.13.0

ryandeivert:
resolves: #1289 

## Background

See #1289 

## Changes

* Updating terraform version and aws provider version

### DIFF
--- a/streamalert_cli/_infrastructure/_include.tf
+++ b/streamalert_cli/_infrastructure/_include.tf
@@ -3,7 +3,7 @@ terraform {
 
   required_providers {
     aws = {
-      source  = "aws"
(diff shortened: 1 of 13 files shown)


streamalert_cli/_infrastructure/_include.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = "~> 0.13.0"
+
+  required_providers {
+    aws = {
+      source  = "aws"
```
chunyong-lin, line 6: Probably we need this `source  = "hashicorp/aws"`

ryandeivert: yeah you're right - updated

PR #1295, opened 2020-10-05
adding fix for #1294 to only copy terraform files once during init

### LINKED ISSUE #1294: Cannot create new streamalert install in 3.4.0
### Description
Fresh install does not seem to work on 3.4.0

### Steps to Reproduce
Can reproduce on both macOS and Linux.

Start with a new clone of 3.4.0 and a new AWS account. 

Run `python manage.py init`.
When prompted to create initial TF resources, say yes. After completion, the following error occurs:

```
State path: terraform.tfstate
[INFO 2020-10-05 17:02:46,172 (streamalert_cli.terraform.handlers:87)]: Configuring Terraform Remote State
  File "manage.py", line 151, in <module>
    sys.exit(main())
  File "manage.py", line 147, in main
    return not cli_runner(options)
  File "/opt/streamalert/streamalert_cli/runner.py", line 69, in cli_runner
    result = cmds[args.command](args)
  File "/opt/streamalert/streamalert_cli/runner.py", line 124, in <lambda>
    command: lambda opts, cmd=cli_command: cmd.handler(opts, config)
  File "/opt/streamalert/streamalert_cli/terraform/handlers.py", line 88, in handler
    if not terraform_generate_handler(config=config, check_tf=False, check_creds=False):
  File "/opt/streamalert/streamalert_cli/terraform/generate.py", line 427, in terraform_generate_handler
    _copy_terraform_files(config)
  File "/opt/streamalert/streamalert_cli/terraform/generate.py", line 399, in _copy_terraform_files
    ignore=shutil.ignore_patterns('*.tf.json')  # TODO: remove this eventually
  File "/usr/lib64/python3.7/shutil.py", line 324, in copytree
    os.makedirs(dst)
  File "/usr/lib64/python3.7/os.py", line 221, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/tmp/streamalert_build-98hs47_x'
```

It seems to be attempting to write to the directory it has already created. This causes the created tf state to become orphanned.

ryandeivert:
cc @blakemotl 
resolves #1294

## Background

See #1294

## Changes

* Ensuring copy of terraform files happens only once during init.. instead of attempting multiple times throughout the process.

## Other Changes

* Defaulting athena settings to parquet and removing prior warning.

## Testing

Updates to unit tests


streamalert_cli/terraform/generate.py
```diff
@@ -634,23 +608,6 @@ def generate_global_lambda_settings(
         tf_tmp_file (str): filename of terraform file, generated by CLI.
         message (str): Message will be logged by LOGGER.
     """
-    if conf_name == 'athena_partitioner_config':
-        # Raise ConfigError when user doesn't explicitly set `file_format`
-        # in `athena_partitioner_config` in conf/lambda.json when upgrade to v3.1.0.
-        file_format = get_data_file_format(config)
-
-        if not file_format or file_format not in ('parquet', 'json'):
```
chunyong-lin, line 642: Can we keep this warning message here? There is chance that users may mis-config the file_format and we have warning message for the errors. But we need to update the message
```
            message = '[ERROR] Available values are "parquet" (preferred) and "json". You set the file_format to {} '.format(file_format)

```

chunyong-lin (approved):
LGTM with one comment unrelated to the issue you fix for terraform init.

streamalert_cli/terraform/generate.py
ryandeivert: @chunyong-lin if you'd like this check, I'd rather have this be centralized as part of the config validation stuff here: https://github.com/airbnb/streamalert/blob/b7971a0020c10049256020e20e1850776ba4b77e/streamalert/shared/config.py#L338

chunyong-lin: Fair!

chunyong-lin: Let's do this later. I will open an issue about this.

ryandeivert: cool thanks!

PR #1301, opened 2021-01-12
Update cloudtrail schema to 1.08

gavinelder:
to:
related to:
resolves:

## Background

AWS releases a new version of the cloud trail schema `1.08` as found in https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-event-reference-record-contents.html

The following adds relevant fields from `1.07` & `1.08` where it seemed appropriate. 

I have commented on each field added the relevant AWS documentation which was available. 

### Exclusions 

At this time I have excluded the field `edgeDeviceDetails` This is exclusive to outpost and I cannot find any log examples to reference and confirm the type. 

```
edgeDeviceDetails
Shows information about edge devices that are targets of a request. Currently, S3 Outposts device events include this field.

Since: 1.08

Optional: True
```

## Changes

Added fields 

- addendum
- eventCategory
- insightDetails
- sessionCredentialFromConsole
- tlsDetails

## Testing

No testing due to lack of logs to check against, I welcome examples submissions.


conf/schemas/cloudtrail.json
```diff
@@ -22,26 +22,31 @@
   "cloudtrail:events": {
     "schema": {
       "additionalEventData": {},
+      "addendum": {},
```
gavinelder, line 25:
```
addendum
If an event delivery was delayed, or additional information about an existing event becomes available after the event is logged, an addendum field shows information about why the event was delayed. If information was missing from an existing event, the addendum field includes the missing information and a reason for why it was missing. Contents include the following.

reason - The reason that the event or some of its contents were missing. Values can be any of the following.

DELIVERY_DELAY – There was a delay delivering events. This could be caused by high network traffic, connectivity issues, or a CloudTrail service issue.

UPDATED_DATA – A field in the event record was missing or had an incorrect value.

SERVICE_OUTAGE – A service that logs events to CloudTrail had an outage, and couldn’t log events to CloudTrail. This is exceptionally rare.

updatedFields - The event record fields that are updated by the addendum. This is only provided if the reason is UPDATED_DATA.

originalRequestID - The original unique ID of the request. This is only provided if the reason is UPDATED_DATA.

originalEventID - The original event ID. This is only provided if the reason is UPDATED_DATA.
```

conf/schemas/cloudtrail.json
```diff
@@ -22,26 +22,31 @@
...
       "apiVersion": "string",
       "awsRegion": "string",
       "errorCode": "string",
       "errorMessage": "string",
+      "eventCategory": "string",
```
gavinelder, line 30:
```
      eventCategory
Shows the event category that is used in LookupEvents calls.

For management events, the value is management.

For data events, the value is data.

For Insights events, the value is insight.

Since: 1.07

Optional: False
```

conf/schemas/cloudtrail.json
```diff
@@ -22,26 +22,31 @@
...
       "eventID": "string",
       "eventName": "string",
       "eventSource": "string",
       "eventTime": "string",
       "eventType": "string",
       "eventVersion": "string",
+      "insightDetails": {},
```
gavinelder, line 37:
```
insightDetails
Insights events only. Shows information about the underlying triggers of an Insights event, such as event source, user agent, statistics, API name, and whether the event is the start or end of the Insights event. For more information about the contents of the insightDetails block, see CloudTrail Insights insightDetails element.

Since: 1.07

Optional: False
```

conf/schemas/cloudtrail.json
```diff
@@ -22,26 +22,31 @@
...
       "managementEvent": "boolean",
       "readOnly": "boolean",
       "recipientAccountId": "string",
       "requestID": "string",
       "requestParameters": {},
       "resources": [],
       "responseElements": {},
+      "sessionCredentialFromConsole": "boolean",
```
gavinelder, line 45:
```
sessionCredentialFromConsole
Shows whether or not an event originated from a AWS Management Console session. This field is not shown unless the value is true, meaning that the client that was used to make the API call was either a proxy or an external client. If a proxy client was used, the tlsDetails event field is not shown.

Since: 1.08

Optional: True
```

conf/schemas/cloudtrail.json
```diff
@@ -22,26 +22,31 @@
...
       "serviceEventDetails": {},
       "sharedEventID": "string",
       "sourceIPAddress": "string",
+      "tlsDetails": {},
```
gavinelder, line 49:
```
tlsDetails
Shows information about the Transport Layer Security (TLS) version, cipher suites, and the FQDN of the client-provided host name of a service API call. Contents include the following. CloudTrail still logs partial TLS details if expected information is missing or empty. For example, if the TLS version and cipher suite are present, but the HOST header is empty, available TLS details are still logged in the CloudTrail event.

If sessionCredentialFromConsole is present with a value of true, tlsDetails is present in an event record only if an external client was used to make the API call.

tlsVersion - The TLS version of a request.

cipherSuite - The cipher suite (combination of security algorithms used) of a request.

clientProvidedHostHeader - The FQDN of the client that made the request.

Since: 1.08

Optional: True

```

Ryxias (approved):

Ryxias:
Rebased onto `release-3-5-0`

jlundstedt:
@gavinelder FYI - `edgeDeviceDetails` is documented to be a JSON object by AWS [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/cloudtrail-logging-understanding-s3-entries.html)

PR #1302, opened 2021-01-12
[OUTPUT] Jira Cloud SASS (jira-v2) 

gavinelder:
to:
related to:
resolves:

## Background

Jira with their SASS offering and later product versions have deprecated the use of basic username + password authentication and instead, they require an API key to be used. 

The following takes the existing `jira` output and adds does the relevant b64 encoding + removes the cookie path. 

## Changes

* Summary of changes
* ...

## Testing

Steps for how this change was tested and verified.

- Added Output to StreamAlert instance and carried out test_rules / live to see a jira issue + comment created.


blakemotl:
Thanks for taking care of this! More or less the same as I have internally and it works great.

Ryxias (approved):

Ryxias:
Rebased onto `release-3-5-0`

PR #1304, opened 2021-01-13
Add victorops as an output in streamalert.

zachzeid:
to:
related to:
resolves:

## Background

Increasing coverage of outputs to include VictorOps

We use VictorOps internally, and we wanted specific alerts so that Security Operations can be paged and begin triaging.

## Changes

Added `victorops.py` as an output in `streamalert/outputs`

## Testing

Tested and deployed in local instance of Streamalert


streamalert/alert_processor/outputs/victorops.py
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        data = {
+            "message_type": "CRITICAL",
+            "entity_id": "streamalert/alert",
+            "entity_display_name": alert.rule_name,
+            "record": publication['record']
+        }
+
+        LOGGER.critical('Sending alert to VictorOps')
```
Ryxias, line 106: Maybe change to `LOGGER.info`

streamalert/alert_processor/outputs/victorops.py
```diff
@@ -0,0 +1,115 @@
...
+        url = creds['url'] + '/' + creds['routing_key']
+        resp = self._post_request(
```
Ryxias, line 108: Any value to use [`_post_request_retry`](https://github.com/airbnb/streamalert/blob/27b86691c5cfa8d86e98b18b91a07a36786385a9/streamalert/alert_processor/outputs/output_base.py#L290) instead?

Ryxias (approved):

Ryxias:
Changed base onto `release-3-5-0`

PR #1311, opened 2021-04-15
Enhancements to JIRA v1 output@z

Ryxias:
cc: @gavinelder 

## Changes

* Adds option to perform clientside SSL cert verification
* Adds aggregation additional filter
* Adds ability to add additional fields through publisher

## ClientSide SSL Verification
To enable this, simply put `yes` in the new `ssl_verify` output parameter.

## Aggregation Additional Filter
```
# When aggregation is enabled, it will fuzzy-search any JIRA ticket that best-matches
# the "summary ~ ..." statement, within the project key. For each matching rule,
# instead of creating new JIRA tasks over and over, it will instead opt to append a
# comment to a similar(ish) JIRA task.
#
# However, this can result in  very long-lived JIRA tickets getting tons of comments
# appended on. This optional parameter allows users to specify an additional JQL clause
# to filter out these older tickets, encouraging new JIRA tasks to be created from
# time to time. It can also be used to increase the accuracy of finding the parent
# task (maybe filtering on a component) in case you find the StreamAlert integration
# is appending comments to unrelated issues.
#
# Example: A highly effective JQL suffix is "created > startOfWeek(-1w)"
```

## Additional Fields
```
# For example, if your JIRA project requires a custom field called "custom_field_1",
# you can set the following json-encoded string in this:
# {"custom_field_1": {"value": "FooBar"}}
#
# These fields are DEFAULT values. You can still override them using the
# @jira.additional_fields publisher parameter.
```
 
## Testing

Deployed internally


gavinelder:
I really like the changes here to the aggregation filter and also the additional fields.

To confirm was the intent in the CC to see if they should also be added to the Jira_v2 output. 

Also in hindsight I wonder if supporting both auth methods in the same output behind a boolean may be better than two distinct outputs.

Ryxias:
Yeah it was kind of an "FYI" since I knew you wrote a lot of the JIRA v2 integration. Admittedly we didn't sync the v2 internally yet so I actually forgot it existed on the OSS repo; only **_just_** noticed it when I synced these JIRA changes out. 

And yeah, I took a cursory glance over the JIRA integration and I see the main difference is the auth; v1 uses username/password login whereas v2 uses api key in basic auth header. Since they both use REST API v1/v2 I think would be worth DRYING it out. I'll think about doing that if I have time

Issue #1261, opened 2020-06-15
[BUG] App Creation stores path to file vs file contents. 


gavinelder:
## Background

Currently in `streamalert_cli/helpers.py` there is a function called `user_input` historically `user_input` used python2 `raw_input` and as part of calling `response_is_valid` called a validation function which returned a string.

This string would then replace the contents of `raw_input()` overriding the user inputed string.

In the later releases of StreamAlert and migration to python3 it appears that `input()` is no longer having the value manipulated in the same way.  

### Steps to Reproduce

On a StreamAlert 3.0+ release create a new application such as `gsuite` or `box` which utilise this value returned by the validation function, once the application is created go to parameter store and look at the contents you should see the String to the path on disk vs the file contents. 

## Desired Change

Contents of the file on disk to be stored in parameter store.

chunyong-lin:
As @gavinelder  pointed out in py3 it uses `input()`  while py2 uses `raw_input()`, but they should behave the same to read the user input as a string which is the `keyfile path` in our case. Later, we should read the content of the `keyfile` using `_required_auth_info`  classmethod https://github.com/airbnb/streamalert/blob/a0a284c6bfdb9de40e76fc7581627b5863445cc0/streamalert/apps/_apps/gsuite.py#L200-L227

Definitely there is something wrong when calls `_required_auth_info` or it may not be called. Will investigate more.

To add more context about @gavinelder 's issue, the value of `gsuite_token_gtoken_app_auth` should be a json blob containing the context of private key. However, instead of storing the context of private key, it stores the path to the private key file which is wrong.
![gsuite_gtoken_app_auth](https://user-images.githubusercontent.com/21151436/84693668-b501b100-aefc-11ea-8064-8e27638a8a34.png)

gavinelder:
Opened a PR to address this, was forced to spend more time than a quick glance as part of building a new app.

PR #1316, opened 2021-08-24
Return response from response_is_valid in streamalert helpers

gavinelder:
related to: #1261 
resolves: #1261 

## Background

Historically SA used to do validation of the response within the `user_input` function and if the input was callable overwrite the initial response with the return value from the function. 

When this was refactored this response by the function was no longer returned and as such the user input was unmodified. 

The following change returns the response.    

## Changes

* Return the response from `response_is_valid` within `streamalert_cli/helpers.py`

## Testing

manual testing.

### DIFF
--- a/streamalert_cli/helpers.py
+++ b/streamalert_cli/helpers.py
@@ -142,7 +142,7 @@ def user_input(requested_info, mask, input_restrictions):
         while not response:
             response = input(prompt)  # nosec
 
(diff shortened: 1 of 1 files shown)


chunyong-lin:
Hi @gavinelder thanks for the contribution. Can you change the base to [release-4-0-0 branch](https://github.com/airbnb/streamalert/tree/release-4-0-0)?

Issue #1318, opened 2021-09-24
Deployment Fails with "ResourceConflictException" in Lambda 


gavinelder:
## Background

AWS has added some additional state checking into the lambda deployment lifecycle, however, the current TF provider version doesn't know how to handle this. 

https://aws.amazon.com/blogs/compute/coming-soon-expansion-of-aws-lambda-states-to-all-functions/

Users can do a temporary state opt-out via adding `aws:states:opt-out` in the description until the 1st of October...

PR #1319, opened 2021-09-24
[ISSUE-1318] Deployment Fails with "ResourceConflictException" in Lambda

gavinelder:
to: @ryandeivert @Ryxias @chunyong-lin 
cc:
related to: https://github.com/airbnb/streamalert/issues/1318
resolves: https://github.com/airbnb/streamalert/issues/1318
closes: https://github.com/airbnb/streamalert/issues/1318

## Background

AWS has added some additional state checking into the lambda deployment lifecycle, however, the current TF provider version doesn't know how to handle this.

https://aws.amazon.com/blogs/compute/coming-soon-expansion-of-aws-lambda-states-to-all-functions/

Users can do a temporary state opt-out via adding `aws:states:opt-out` in the description until the 1st of October...

The smallest change is to update the TF AWS provider to use a new version of the AWS SDK that can handle this. 

## Changes

* Summary of changes
* ...

## Testing

### DIFF
--- a/streamalert_cli/_infrastructure/_include.tf
+++ b/streamalert_cli/_infrastructure/_include.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 3.60.0"
+      version = ">= 3.60.0, < 4.0.0"
     }
   }
 }
--- a/streamalert_cli/_infrastructure/_include.tf
+++ b/streamalert_cli/_infrastructure/_include.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 3.3.0"
+      version = "~> 3.60.0"
     }
   }
 }


streamalert_cli/_infrastructure/_include.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 3.3.0"
+      version = "~> 3.60.0"
```
ryandeivert, line 7: this terraform syntax should be avoided for the aws provider. the `~>` constraint only allows for the RIGHT MOST member of the tuple to increment (eg: in `3.N.n` only `n` can change) Effectively, this is actually `== 3.60.0` since this provider never publishes patch semvers.

we should update this to a constraint that more loosely allows incrementing, like: `>= 3.60.0`. possibly also restrict to `< 4.0.0`

also see the note here: https://www.terraform.io/docs/language/modules/develop/providers.html#provider-aliases-within-modules

>If you are writing a shared Terraform module, constrain only the minimum required provider version using a >= constraint.

gavinelder: I have opted for the latter approach of constraining between two versions vs pinning to a specific version to allow the most flexibility.

ryandeivert (approved):

PR #1333, opened 2022-07-19
Dry Out Jira-V2 Output

gavinelder:
cc : @blakemotl
related to: https://github.com/airbnb/streamalert/pull/1311

## Background

The following change was discussed on https://github.com/airbnb/streamalert/pull/1311 where the original Jira-v2 was just a heavy cut/paste job with some slight modifications, which is not ideal.

https://streamalert.slack.com/archives/C3BHE2Z0S/p1610399094191300

This is cleaning up all of that tech debt.

## Changes

* Dry Out the config.
* `_api_key` is now `password` and we use the existing credentials helper.

## Contentious points

* This will break existing Jira-V2 users as the credential for `api_key` has changed to `password` , I can add a small helper to pull from this field if `password` is none etc but curious thoughts on that one.
* This overloads `self._cookie` which is now holding the base64 encoded authorization bearer token.

## Testing

This was tested in conjunction with https://github.com/airbnb/streamalert/pull/1311 against a live environment.

### DIFF
--- a/streamalert/alert_processor/outputs/jira_v2.py
+++ b/streamalert/alert_processor/outputs/jira_v2.py
@@ -31,7 +31,7 @@ LOGGER = get_logger(__name__)
 @StreamAlertOutput
 class JiraSaaSOutput(JiraOutput):
     """
-    JiraSaaSOutput handles all alert dispatching for Jira SaaS 
+    JiraSaaSOutput handles all alert dispatching for Jira SaaS.
     and is a subclass of the "JiraOutput" dispatcher which is typically used for on-prem Jira."""
     __service__ = 'jira-v2'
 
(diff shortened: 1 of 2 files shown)


chunyong-lin (approved):
It is lots of copy and paste 😄 Thanks for the contribution.

Side note, we probably can write a new output based on JIRA's official python package [jira](https://jira.readthedocs.io/installation.html), maybe name it `jira-cloud` or `jira-saas`. I actually use it for other projects and works well.

gavinelder:
@chunyong-lin `jira-cloud` SGTM I would like to see what #1311 and v4 look like before a major overhaul, however building and introducing before a major breaking change and removing old Outputs like PD v1, Jira, etc may make sense.

